PR #3, opened 2014-08-01
Can I build PEX with system wide site-packages?

kamilchm:
I build an app for Linux that uses Python `gobject` bindings provided by my system (Debian in my case). It's in `/usr/lib/python2.7/dist-packages`.
When I start pex with `--inherit-path` it prints:

```
Scrubbing from site-packages: /usr/lib/python2.7/dist-packages
```

When I try to import `gobject` i get:

```
>>> import gobject
ImportError: No module named gobject
```

I want to distribute my app as a `deb` and resolve system wide dependencies when installing it by `dpkg`.
Is there any way I can use system provided lib in my PEX?

### DIFF
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -116,8 +116,21 @@ class PEX(object):
 
     return new_modules
 
+  def minimum_sys_path(self, site_libs):
+    sys_path = OrderedSet(sys.path)
+
+    if not self._pex_info.inherit_path:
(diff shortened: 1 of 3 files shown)


jsirois:
Thanks for attempting a fix. Please accept our apologies for the epically long delay in responding! This has since been fixed a few times, the latest in #449.

PR #20, opened 2014-10-18
Helper method to access files within a zipped pex

Yasumoto:


pex/common.py
```diff
@@ -276,7 +276,10 @@ def write(self, data, dst, label=None, mode='wb'):
     self._tag(dst, label)
     self._ensure_parent(dst)
     with open(os.path.join(self.chroot, dst), mode) as wp:
-      wp.write(data)
+      try:
+        wp.write(data)
+      except TypeError:
+        wp.write(bytes(data, 'UTF-8'))
```
wickman: this doesn't seem right.  it should not be the responsibility of write() to do the encoding/decoding.  it should be the responsibility of the caller.

pex/util.py
```diff
@@ -41,6 +41,32 @@ def zipsafe(dist):
       return False
 
   @classmethod
+  def access_zipped_assets(cls, static_module_name, static_path, asset_path, dir_location='.'):
+    """
+    Create a copy of static resource files as we can't serve
+    them from within the pex file.
+
+    :param static_module_name: Module name containing module to cache in a tempdir
+    :type static_module_name: string in the form of 'twitter.common.zookeeper'
```
wickman: you really mean "string, for example 'twitter.common.zookeeper'" or similar

pex/util.py
```diff
@@ -41,6 +41,32 @@ def zipsafe(dist):
...
+    :param static_path: Module name of the form 'serverset'
+    :param asset_path: Initially a module name that's the same as the static_path, but will be
+        changed to walk the directory tree
+    :param dir_location: directory to create a new temporary directory in
+    """
+    temp_dir = safe_mkdtemp(dir=dir_location)
+    for asset in resource_listdir(static_module_name, asset_path):
+      asset_target = os.path.join(os.path.relpath(asset_path, static_path), asset)[2:]
```
wickman: what is going on with [2:] ?

tests/test_util.py
```diff
@@ -74,3 +86,68 @@ def test_zipsafe():
...
+  try:
+    import __builtin__
+    builtin_path = __builtin__.__name__
+  except ImportError:
+    # Python3
+    import builtins
+    builtin_path = builtins.__name__
+
+  mock_open = mock.mock_open()
+  with mock.patch('%s.open' % builtin_path, mock_open, create=True):
```
wickman: not sure if this is available in our version but we should use it:
https://docs.python.org/3/library/unittest.mock-helpers.html#mock-open

if that's not available, '__builtins__' is consistent in python 2.x and python 3.x

Yasumoto: We're already using mock_open, but need to patch it in still.

Hm, I don't think so:

```
[tw-mbp-jsmith pex (jsmith/access_zipped_assets)]$ python
Python 2.7.9 (default, Mar  4 2015, 17:22:02) 
[GCC 4.2.1 Compatible Apple LLVM 6.1.0 (clang-602.0.42)] on darwin
>>> import builtins
  File "<stdin>", line 1, in <module>
ImportError: No module named builtins
>>> 
```

pex/util.py
Yasumoto: It's gotta remote a `./` which prepends the file name. It looks like this before/after the `[2:]`
```
./mod.py
mod.py
./mod.pyc
mod.pyc
./__init__.py
__init__.py
```

I've swapped this to use `os.path.normpath` to get what I'm actually looking for here.

Happy for other suggestions though.

pex/util.py
```diff
@@ -41,6 +42,33 @@ def zipsafe(dist):
...
+      asset_target = os.path.normpath(os.path.join(os.path.relpath(asset_path, static_path), asset))
+      if resource_isdir(static_module_name, os.path.join(asset_path, asset)):
+        safe_mkdir(os.path.join(temp_dir, asset_target))
+        cls.access_zipped_assets(static_module_name, static_path, os.path.join(asset_path, asset))
+      else:
+        with open(os.path.join(temp_dir, asset_target), 'wb') as fp:
+          path = os.path.join(static_path, asset_target)
+          file_data = resource_string(static_module_name, path)
+          fp.write(file_data)
+    atexit.register(lambda: safe_rmtree(temp_dir))
```
wickman: safe_mkdtemp already does atexit.register so this is unnecessary

tests/test_util.py
```diff
@@ -3,15 +3,27 @@
...
+from pex.compatibility import nested
 from pex.installer import EggInstaller, WheelInstaller
+from pex.pex_builder import PEXBuilder
 from pex.testing import make_bdist, temporary_content, write_zipfile
 from pex.util import CacheHelper, DistributionHelper
 
+try:
+  import mock
+except ImportError:
+  from unittest import mock
```
wickman: should be the other way around due to http://code.google.com/p/mock/issues/detail?id=249

PR #32, opened 2015-01-22
More specific information in building pex documentation

Moult:
- Mention that pex needs wheel as a dependency
- Explain how to use the -s option
- Fix a typo

### DIFF
--- a/docs/buildingpex.rst
+++ b/docs/buildingpex.rst
@@ -38,6 +38,7 @@ absolute path of a Python binary or the name of a Python interpreter within the 
                                ^
     SyntaxError: invalid syntax
 
+Note that ``pex`` requires ``wheel`` to be installed.
 
 
 Specifying requirements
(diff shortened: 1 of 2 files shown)


docs/buildingpex.rst
```diff
@@ -63,6 +64,22 @@ You can then import and manipulate modules like you would otherwise::
...
+Specifying source directories
+-----------------------------
+
+Source directories are source distributions typically built with ``python setup.py sdist``.
+The ``-s`` option specifies the path to a source directory.::
+
+    $ python setup.py sdist
+    $ tar -xzvf dist/foo-x.y
+    $ pex -s foo-x.y
+    Python 2.6.9 (unknown, Jan  2 2014, 14:52:48)
```
Yasumoto: s/2.6/2.7 if you don't mind?

Yasumoto:
Any chance you could take a look at the style error at https://travis-ci.org/pantsbuild/pex/jobs/47876431 and fixup + re-post?

Thanks!

Yasumoto:
Ah, actually, looks like this is combined with the diff in #33. If you can split that out of this one, we can get it shipped.

Moult:
Aha, didn't realise pull requests were branch specific. Changed the python version as you suggested, and reverted the ABI change commit.

Yasumoto:
LGTM, I'll leave it to @wickman to give the final +1 and merge.

docs/buildingpex.rst
```diff
@@ -63,6 +64,22 @@ You can then import and manipulate modules like you would otherwise::
...
 
+Specifying source directories
+-----------------------------
+
+Source directories are source distributions typically built with ``python setup.py sdist``.
+The ``-s`` option specifies the path to a source directory.::
+
+    $ python setup.py sdist
+    $ tar -xzvf dist/foo-x.y
+    $ pex -s foo-x.y
```
wickman: This is equivalent to 'pex -s .'

I think it makes sense to skip the sdist part because it might confuse readers into thinking they need to first create an sdist, when pex does exactly this under the covers.  An alternative would be to describe either git cloning a repository with a setup.py or fetching a source distribution and then unpacking and running 'pex -s' on the target.

docs/buildingpex.rst
```diff
@@ -63,6 +64,22 @@ You can then import and manipulate modules like you would otherwise::
     >>> psutil.__path__
     ['/private/var/folders/4d/9tz0cd5n2n7947xs21gspsxc0000gp/T/tmpYuGpFW/.deps/psutil-2.0.0-py2.6-macosx-10.4-x86_64.egg/psutil']
 
+Specifying source directories
+-----------------------------
+
+Source directories are source distributions typically built with ``python setup.py sdist``.
```
wickman: This wording is slightly backward.  Instead it should be something like

"Python source distributions are source directories with a properly defined setup.py."

PR #35, opened 2015-01-25
Fix bug that caused pex not to cache the results of translation.

ugodiggi:
I'm under the impression that this is a simple "typo" error, and the intention was also to pass through the parameter to the chained translators.

Of course it's possible that this was intentional and I'm really missing the point :-)

See https://rbcommons.com/s/twitter/r/1666/


wickman:
nice catch @ugodiggi!

PR #37, opened 2015-01-29
Speed up the best-case scenario of dependency resolution.

ugodiggi:
A few small touchups that improve the speed of dependency resolution when
many packages are already in the local cache.

- when dispatching multiple crawling threads:
  - make each thread a bit more reactive by reducing the polling timeout from
    100ms to 10ms
  - do not wait for each of the workers threads to complet, just wait for them to complete their
    workload.
  Without this change, dependency resolution is guaranteed to take longer than 100ms per
  dependency, which is a large amount of time for just checking a local zipfile's content.

- cache the result of a couple calls that are repeated many times:
  - Link.from_filename
  - Package.from_href
  Each of this call is performed for each file in the cache, for each dependency that is resolved.
  While both these calls are not especially expensive, when we repeat them n^2 times in a largish
  local cache * set of dependencies they do add up.

Somewhat unscientific benchmarking on my system show that the average time for resolving a single
dependency (namely 'pytz==2013b') goes down from 150ms to 30ms.

Running the modified code on the urbancompass codebase produced similarly desireable timings.

### DIFF
--- a/pex/crawler.py
+++ b/pex/crawler.py
@@ -101,7 +101,7 @@ class Crawler(object):
     def execute():
       while not converged.is_set():
         try:
-          link = queue.get(timeout=0.1)
+          link = queue.get(timeout=0.01)
         except Empty:
           continue
         if link not in seen:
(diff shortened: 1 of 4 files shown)


wickman:
Only minor nit is that the package memoizer should be flushed each time we call .register, but this is purely pedantic.  Will merge as-is.

Yasumoto:
This is awesome! Thanks @ugodiggi !!

PR #39, opened 2015-01-30
Change to uuid4().hex for Python3

Moult:


wickman:
Thanks, this is a duplicate of #8 but it looks like the amended PR never made it.

PR #40, opened 2015-01-30
Support Python 2.6

sholsapp:
This string formatting nicety was introduced in Python 2.7 and breaks
Python 2.6 users like myself. This change is forwards and backwards
compatible between all Pythons.

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -54,7 +54,7 @@ def configure_clp():
       '%prog builds a PEX (Python Executable) file based on the given specifications: '
       'sources, requirements, their dependencies and other options')
 
-  parser = OptionParser(usage=usage, version='%prog {}'.format(__version__))
+  parser = OptionParser(usage=usage, version='%prog {0}'.format(__version__))
 
   parser.add_option(
       '--pypi', '--no-pypi',


wickman:
Thanks for the catch!  Was unaware that this formatting quirk failed in 2.6.

This points out the need for better integration/coverage testing of the pex tool itself since 2.6 is supported and definitely covered by the rest of the tests.

sholsapp:
You're welcome!

Awesome tool, btw, we're playing with pex now to see if we can use it to manage a somewhat large codebase. So far it is looking great.

wickman:
Cool.  If you have any questions or need any help, don't hesitate to reach out.  We don't have a pex-specific mailing list but the pants mailing list can be pretty helpful, and there's usually at least one of us in #pypa or #pypa-dev.

sholsapp:
Hi @wickman, I know the change was trivial, but do you have any idea when a new version of `pex` might be available on pypi? We're largely a Python 2.6 shop and I'm trying to work `pex` into our artifact generating machinery... I can't do that until we get this fix released.

wickman:
unless i'm missing something, '{}'.format(foo) is not a syntax error in Python 2.6 so the toolchain should work just fine (just that the 'pex --version' command is broken.)  we should still roll a patch release but i just wanted to be clear that you shouldn't be blocked.

sholsapp:
You're right, it's not a syntax error, but every command I try to run on Python 2.6 yields an error like this.

```
  File "/home/sholsapp/.virtualenvs/py509/bin/pex", line 11, in <module>
    sys.exit(main())
  File "/home/sholsapp/.virtualenvs/py509/lib/python2.6/site-packages/pex/bin/pex.py", line 320, in main
    parser = configure_clp()
  File "/home/sholsapp/.virtualenvs/py509/lib/python2.6/site-packages/pex/bin/pex.py", line 57, in configure_clp
    parser = OptionParser(usage=usage, version='%prog {}'.format(__version__))
ValueError: zero length field name in format
```

wickman:
Ah, ok.  I see.  Thanks for the context.  I'll get this published asap.

wickman:
Published.  Thanks for the prod.

sholsapp:
Right on. Thank you for the quick response!

Issue #44, opened 2015-02-09
no errors keyword for decode in python 2.6


blalor:
Attempting to run `pex` on Python 2.6.8 fails.  I had to add `-v` to see the error, otherwise it just says it can't resolve the requirement.

```
[root@08b6ba92511a ~]# pex --python=/usr/bin/python2.6 -r fabric -v
pex: Unknown exception encountered: decode() takes no keyword arguments
pex: Traceback (most recent call last):
pex:   File "/usr/lib/python2.6/site-packages/pex/crawler.py", line 110, in execute
pex:     roots, rels = self.crawl_link(self.context, link)
pex:   File "/usr/lib/python2.6/site-packages/pex/crawler.py", line 87, in crawl_link
pex:     return cls.crawl_remote(context, link)
pex:   File "/usr/lib/python2.6/site-packages/pex/crawler.py", line 74, in crawl_remote
pex:     content = context.content(link)
pex:   File "/usr/lib/python2.6/site-packages/pex/http.py", line 122, in content
pex:     return fp.read().decode(encoding, errors='replace')
pex: TypeError: decode() takes no keyword arguments
Traceback (most recent call last):
  File "/usr/bin/pex", line 11, in <module>
    sys.exit(main())
  File "/usr/lib/python2.6/site-packages/pex/bin/pex.py", line 325, in main
    pex_builder = build_pex(args, options)
  File "/usr/lib/python2.6/site-packages/pex/bin/pex.py", line 303, in build_pex
    cache_ttl=options.cache_ttl)
  File "/usr/lib/python2.6/site-packages/pex/resolver.py", line 220, in resolve
    raise Unsatisfiable('Cannot satisfy requirements: %s' % requirement_set[requirement.key])
pex.resolver.Unsatisfiable: Cannot satisfy requirements: [Requirement.parse('fabric')]
```

wickman:
Fixed by #46

PR #45, opened 2015-02-09
Fix for decode() on Python 2.6

blalor:
Addresses #44, but I don't know if it breaks compatibility with Python 3.


blalor:
Hey, what do you know, all tests passed! 👍 ;-)

jsirois:
@blalor but they passed before your change and the tests include a python2.6 environment.  I think there is a missing test that could exercise what you fixed here and pre-fix make the python 2.6 CI shard go red.

wickman:
Submitted #46 which is basically #45 + test coverage.

PR #73, opened 2015-04-14
correctly handle index urls which do not end in trailing slash

### LINKED ISSUE #70: index urls without trailing / are improperly handled
given `https://user:pass@my.python-index.com/simple` pex 1.0.0dev0 fails to construct the appropriate url for packages in the index.

Expected:

```
--extra-index-url https://user:pass@python-pkgs.ernest.ly/simple
my-private-dep==1.0.0
```

should search `https://user:pass@python-pkgs.ernest.ly/simple/my-private-dep` for version matching 1.0.0

instead, it searches `https://user:pass@python-pkgs.ernest.ly/my-private-dep`

ewdurbin:
equivalent index url handling to pip. fixes #70


Yasumoto:
Mind an s/test_fetcher`s`.py/`test_fetcher.py` to match the name of the module it's testing? Otherwise :+1:

PR #86, opened 2015-04-16
Unify coverage and profiling handling and semantics. Fixes #81.

### LINKED ISSUE #81: PEX_COVERAGE and PEX_PROFILE don't work if SystemExit is raised
they should be in a try/finally loop.  right now they only work if the application does not SystemExit.

wickman:


wickman:
pantsbuild travis queue way behind.  this merges cleanly / passes at https://travis-ci.org/wickman/pex/builds/58804067

PR #106, opened 2015-05-01
Fix --inherit-path

mbachry:
`--inherit-path` doesn't work with current master. Normally pex bootstrapper strips down `sys.path` to bare minimum, but I believe it should leave `sys.path` alone if `--inherit-path` was passed. I'm not sure about other stuff patched in `patch_sys` function (like `sys.modules`), but I think it's safer not to touch these either in this case.

I bumped on this bug trying to use `lxc` module. It's provided by `python-lxc` package on Ubuntu, but unfortunately it's not uploaded to pypi.

### DIFF
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -222,30 +222,18 @@ class PEX(object):  # noqa: T000
   # Thar be dragons -- when this contextmanager exits, the interpreter is
   # potentially in a wonky state since the patches here (minimum_sys_modules
   # for example) actually mutate global state.  This should not be
-  # considered a reversible operation despite being a contextmanager.
+  # considered a reversible operation.
   @classmethod
-  @contextmanager
   def patch_sys(cls):
     """Patch sys with all site scrubbed."""
     def patch_dict(old_value, new_value):
       old_value.clear()
       old_value.update(new_value)
 
-    def patch_all(path, path_importer_cache, modules):
-      sys.path[:] = path
-      patch_dict(sys.path_importer_cache, path_importer_cache)
-      patch_dict(sys.modules, modules)
-
-    old_sys_path, old_sys_path_importer_cache, old_sys_modules = (
-        sys.path[:], sys.path_importer_cache.copy(), sys.modules.copy())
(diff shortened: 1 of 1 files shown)


rouge8:
:+1: I just spent the past hour or so banging my head against this...

brbsix:
I was trying to fix this earlier today. I was packaging up a pip helper utility, so access to user site distributions was essential. You can see what I did at https://github.com/brbsix/pip-utils/commit/b27f9ddb784eb49bbb7a8894d0f13067fb42b4e4. It's a really dirty `sed` one-liner, this gist of which is:

    git clone --depth=1 https://github.com/pantsbuild/pex
    sed -i 's/^\(\s*scrub_paths\) = user_site_distributions$$/\1 = OrderedSet()/' pex/pex/pex.py

This obviously doesn't fix log messages or anything like that, but it gets the job done. It would be nice to see this properly addressed so that `.pex` can be more akin to ZIP applications if need be (but with decent tooling and dependency support).

kwlzn:
as is, this PR is over a year old and has merge conflicts - but I'd be happy to help land a proper fix for this if someone wanted to dust things off and revise for current state.

it'd be ideal to get a test case in place for this as well as part of that - ideally one that breaks without the patch and succeeds with it in place so we can prevent accidental regression of this feature.

it also looks like there were some specific issues called out on #138, so it'd be good for the author to at least address those concerns at a discussion level in the ticket in conjunction with a code change.

brbsix:
I really like @wickman's suggestion at https://github.com/pantsbuild/pex/issues/138#issuecomment-126116681 for whitelisted packages. Perhaps the ability to whitelist system site, user site, individual packages, or all of the above.

Note I also opened a related PR, https://github.com/pantsbuild/pex/pull/284, to include the user site with `--inherit-path`. I'd like to contribute further on this issue, I just need to become familiarized with the project/organization standards and come to a consensus on the design requirements for this feature.

jsirois:
Thanks for attempting a fix here. This has since been fixed as demonstrated in https://github.com/pantsbuild/pex/issues/138#issuecomment-529264375

PR #116, opened 2015-06-01
sys.exit instead of returning pex exit code.

### LINKED ISSUE #115: the pex cli doesn't return the correct exit code
exit.py:
```
    import sys
    sys.exit(1)
```

pex:
```
    $ python exit.py
    $ echo $?
    1

    $ pex -- exit.py
    $ echo $?
    0
```

wickman:
Fix #115.


wickman:
Passed at https://travis-ci.org/wickman/pex/builds/64953453

PR #123, opened 2015-06-11
[CI] Two enhancements to PEXBuilder:

benjyw:
1. Allow it to optionally copy instead of hard-linking files into the
   environment.  This is useful for creating long-running environments
   without corrupting them by modifying the underlying files from under
   them.

2. Make precompiling .py sources optional, as mentioned in a TODO.
   Again, in long-running environments it may make more sense to save
   this time when building the environment and pay it when actually
   using the environment.

### DIFF
--- a/pex/pex_builder.py
+++ b/pex/pex_builder.py
@@ -55,7 +55,8 @@ class PEXBuilder(object):
 
   BOOTSTRAP_DIR = ".bootstrap"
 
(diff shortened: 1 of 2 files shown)


fr-anthony:
+1 for this functionality, as `vagrant` shared folders do not support hard or soft links, rendering pex unable to build in such an environment.

jordal:
+1 would like to see this as well

wickman:
This seems to have been merged @ 27fa349a8e

PR #126, opened 2015-06-12
Bytecode compile using the correct interpreter.

jsirois:
Previously a PEXBuilder with a custom interpeter would not use it to
bytecode compile sources when building a pex.  This could lead to errors
when the current interpreter and the target interpreter spanned breaking
language changes.

Kill the CodeMarshaller which has only its marshalling half used in
favor of using py_compile from the stdlib in the new
interpreter-sensitive Compiler class.

Additionally, actually compile all PEXBuilder source including any
generated `__init__.py`s, the generated `__main__.py` and everything in the
`_pex/` bootstrap.

https://rbcommons.com/s/twitter/r/2350/


jsirois:
This is part of addressing https://github.com/pantsbuild/pants/issues/1658

jsirois:
This addresses https://github.com/pantsbuild/pex/issues/127

jsirois:
Submitted @ https://github.com/pantsbuild/pex/commit/ac3c7f17e7ab5ac13eba78845b24936eab78b9fa

Issue #135, opened 2015-07-22
Variables values with non-None defaults trample PexInfo settings.


jsirois:
For example, PexInfo's zip_safe is trampled by PEX_FORCE_LOCAL.  Any non-None Variables value default currently is treated as-if the env var was actually set by a user:
  https://github.com/pantsbuild/pex/blob/master/pex/pex_info.py#L100-L101

jsirois:
This issue was spotted over in pants after the upgrade to pex 1.0.x: https://github.com/pantsbuild/pants/issues/1852

In the pants case, a python_binary marked zip_safe=False correctly generates a pex with a zip_safe=false PEX-INFO manifest, but when the binary run it acts as if PEX_FORCE_LOCAL has been explicity set to false, thus ignoring the PEX-INFO setting systemically.

jsirois:
Fix is in @ https://github.com/pantsbuild/pex/commit/7654a4c762247a94760b1d593e84b1ca710a3c34

PR #136, opened 2015-07-23
Fix from_env to capture only explicitly set values.

jsirois:
Previously from_env also captured Variables defaults which had the
effect of defaulted Variables always over-riding user supplied PexInfo
values.

https://rbcommons.com/s/twitter/r/2517/

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,14 @@
 CHANGES
 =======
 
+----------
+1.0.2.dev0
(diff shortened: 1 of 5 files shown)


jsirois:
This addresses #135

jsirois:
Submitted @ https://github.com/pantsbuild/pex/commit/7654a4c762247a94760b1d593e84b1ca710a3c34

PR #139, opened 2015-07-29
Support PathResource that comes from egg

palexander:
I was seeing failures when creating an entry point to a script that was included in an egg. I'm not super familiar with the pex internals, but it seemed to only happen when the pex was actually executing (IE it found the script during the build process but not during execution).

My guess is that the eggs get converted to a PathResource at some point and since the PathResource is treated the same as a WheelResource it fails on paths that came from eggs because the scripts directory is in a different location.

My fix was to try both finding scripts from an egg location and from a wheel location when a PathResource is found. I'm happy to add/update tests if you can point me to the relevant ones.

### DIFF
--- a/pex/finders.py
+++ b/pex/finders.py
@@ -268,7 +268,7 @@ def get_script_from_distribution(name, dist):
     return get_script_from_whl(name, dist)
   elif isinstance(dist._provider, (pkg_resources.PathMetadata)):
     script_path, script = get_script_from_egg(name, dist)
-    if script_path == None:
+    if script_path is None:
       script_path, script = get_script_from_whl(name, dist)
     return script_path, script
   return None, None
--- a/pex/finders.py
+++ b/pex/finders.py
@@ -238,7 +238,7 @@ def unregister_finders():
 
 def get_script_from_egg(name, dist):
   """Returns location, content of script in distribution or (None, None) if not there."""
-  if name in dist.metadata_listdir('scripts'):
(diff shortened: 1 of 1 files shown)


wickman:
Thanks for the report; I am able to repro simply with boto.

With wheel:
```bash
mba=pex=; pex boto -o /tmp/boto.pex
mba=pex=; PEX_SCRIPT=glacier /tmp/boto.pex | head -5
glacier <command> [args]

    Commands
        vaults    - Operations with vaults
```

With egg:
```bash
mba=pex=; pex boto --no-wheel -o /tmp/boto.pex
mba=pex=; PEX_SCRIPT=glacier /tmp/boto.pex
  File ".bootstrap/_pex/pex.py", line 319, in execute
  File ".bootstrap/_pex/pex.py", line 254, in _wrap_coverage
  File ".bootstrap/_pex/pex.py", line 286, in _wrap_profiling
  File ".bootstrap/_pex/pex.py", line 356, in _execute
  File ".bootstrap/_pex/pex.py", line 389, in execute_script
_pex.pex.NotFound: Could not find script glacier in pex!
```

Let me dig in to see what precisely is going on (since I believe we have test coverage for this) and then I will address your PR.

wickman:
Confirmed -- we were only testing zip-safe eggs/wheels in ``test_pex_script`` in tests/test_pex.py.  When parameterizing on zip_safe=True/False, the test case failed for 2 of 8 combinations.

PR #143, opened 2015-08-06
Remove stderr print on SystemExit().code == None

kwlzn:
The existing SystemExit handler causes an extraneous 'None' to be printed at exit when pex is used with twitter.common.app (and presumably sys.exit() is called with no arguments, which leads to SystemExit().code == None). This adds a None check on se.code.

### DIFF
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -321,7 +321,7 @@ class PEX(object):  # noqa: T000
     except SystemExit as se:
       # Print a SystemExit error message, avoiding a traceback in python3.
       # This must happen here, as sys.stderr is about to be torn down
-      if not isinstance(se.code, int):
+      if not isinstance(se.code, int) and se.code is not None:
         print(se.code, file=sys.stderr)
       raise
     finally:
--- a/tests/test_pex.py
+++ b/tests/test_pex.py
@@ -60,6 +60,14 @@ def test_pex_sys_exit_prints_non_numeric_value_no_traceback():
   _test_sys_exit(sys_exit_arg, expected_output, 1)
 
 
+def test_pex_sys_exit_doesnt_print_none():
+  _test_sys_exit('', to_bytes(''), 0)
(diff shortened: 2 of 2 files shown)


pex/pex.py
```diff
@@ -321,7 +321,7 @@ def execute(self):
     except SystemExit as se:
       # Print a SystemExit error message, avoiding a traceback in python3.
       # This must happen here, as sys.stderr is about to be torn down
-      if not isinstance(se.code, int):
+      if not isinstance(se.code, int) and se.code is not None:
         print(se.code, file=sys.stderr)
```
orsenthil: This change LGTM.

However, w.r.t to previous change, we never want any exit code / message to be explicitly printed on stderr. Isn't it?  I couldn't realize the need of it even after reading comment re: avoiding traceback in python3.

kwlzn: per the sys.exit docs:

'The optional argument arg can be an integer giving the exit status (defaulting to zero), or another type of object. If it is an integer, zero is considered “successful termination” and any nonzero value is considered “abnormal termination” by shells and the like. Most systems require it to be in the range 0-127, and produce undefined results otherwise. Some systems have a convention for assigning specific meanings to specific exit codes, but these are generally underdeveloped; Unix programs generally use 2 for command line syntax errors and 1 for all other kind of errors. If another type of object is passed, None is equivalent to passing zero, and any other object is printed to stderr and results in an exit code of 1. In particular, sys.exit("some error message") is a quick way to exit a program when an error occurs.'

jamesbroadhead:
lgtm.   Thanks, sorry.

wickman:
@kwlzn could you rebase this PR?  I can get it merged in and a fix pushed to unbreak pants

kwlzn:
@wickman ah my bad, should be clean now (and apologies for the delay, just saw this response).

PR #145, opened 2015-08-10
Accommodate OSX `Python` python binaries.

jsirois:
https://rbcommons.com/s/twitter/r/2586/


jamesbroadhead:
lgtm, no tests. Would be nice if the version bump was PR'd separately

PR #148, opened 2015-08-11
Normalize all names in ResolvableSet.  Fixes #147.

### LINKED ISSUE #147: It's possible to build a pex with multiple versions of the same package
```bash

$ git clone https://github.com/sphinx-doc/sphinx && cd sphinx
$ pex . -v
  snowballstemmer 1.2.0
  MarkupSafe 0.23
  Jinja2 2.8
  six 1.9.0
  Babel 2.0
  Sphinx 1.3.1
  pytz 2015.4
  alabaster 0.7.6
  Pygments 2.0.2
  sphinx-rtd-theme 0.1.8
  Sphinx 1.4a0.dev20150811
  docutils 0.12
```

Note the two copies of sphinx pulled into the environment.  This messes with things like PEX_SCRIPT which expects to see just one version of an entry point.

wickman:


pex/resolver.py
```diff
@@ -56,6 +58,10 @@ def merge(self, other):
 
 
 class _ResolvableSet(object):
+  @classmethod
```
kwlzn: why @classmethod here vs @staticmethod?

kwlzn:
LGTM - one comment.

pex/resolver.py
wickman: I just generally always use classmethods instead of staticmethods to allow for future subclassing/testing.  For example, we could write an implementation like so:

```python
class BuggyResolvableSet(_ResolvableSet):
  @classmethod
  def normalize(cls, name):
    return name
```

and then assert that it fails when running the same test suite as the regular implementation.

As it stands, we're subclassing it nowhere and it's an internal class so it doesn't really give us any leverage over a staticmethod but why be inconsistent.

jsirois:
LGTM

PR #149, opened 2015-08-12
Implement bdist_pex.

wickman:
This is my cut at bdist_pex with @msabramo's fixes rebased on top.  The main question is if the implemented/documented behavior is sane.  It seems reasonable to me but I'm also unsure about what potential customers think.

Addresses feature request #99.

### DIFF
--- a/pex/commands/bdist_pex.py
+++ b/pex/commands/bdist_pex.py
@@ -61,7 +61,7 @@ class bdist_pex(Command):
         log.info('Writing %s to %s.pex' % (script_name, script_name))
         self._write(pex_builder, script_name, script=script_name)
     else:
-      if len(self.distribution.entry_points['console_scripts']) == 1:
+      if len(self.distribution.entry_points.get('console_scripts', [])) == 1:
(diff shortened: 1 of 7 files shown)


pex/commands/bdist_pex.py
```diff
@@ -0,0 +1,87 @@
...
+        log.info('Writing %s to %s' % (script_name, target))
+        self._write(pex_builder, target, script=script_name)
+    elif name in console_scripts:
+      # The package has a namesake entry point, so use it.
+      target = os.path.join(self.bdist_dir, name + '-' + version + '.pex')
+      log.info('Writing %s to %s' % (name, target))
+      self._write(pex_builder, target, script=name)
+    else:
+      # The package has no namesake entry point, so build an environment pex.
+      # TODO(wickman) Is this a reasonable default behavior?  Should we fall back
```
jsirois: I think its a reasonable default.

jsirois:
This is very very cool.  Thanks @msabramo!  And thank you @wickman as always.

PR #153, opened 2015-08-16
Use Travis's container-based infrastructure

rouge8:
Jobs on the container-based infrastructure should start faster.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: python
 python: 2.7
 env:


rouge8:
Replaced by #168

jsirois:
Aha - did not see this.  I just started helping out on pex maintenance.

Thanks for having done this despite being trampled.

piotr-dobrogost:
Actually for pip they are slower – https://github.com/pypa/pip/pull/3095

jsirois:
We shall see as CI runs accumulate, there is more at play than pip here.  Pex has many shards to schedule, and that may give an overall win.  We've had 1 container based build land and the variances are high, but it is ~10% faster overall: https://travis-ci.org/pantsbuild/pex/builds.

For pants - which also has many shards - the win was noticeable.  We switched back on 2014-12-19 and have been happy customers.

PR #156, opened 2015-09-07
crawler: Decode "&amp;" to "&" in urls

jdanbrown:
- Encountered when trying to fetch from a private pypicloud that redirects to s3 uris with query string params
- Added failing unit test that now passes
- Left a TODO for replacing all html entities, not just `&amp;`, since that requires some nontrivial python 2/3 compatibility work that wasn't obvious after a few googles


jsirois:
Pex has since switched to using vendored Pip to do crawling and resolving in general so I'm going to close this out. Thanks for the attempt @jdanbrown back in the day.

PR #61, opened 2015-04-06
RFC: Add the ability to include extensions & shared libraries in pexs.

mikekap:
Extensions already (sort-of) worked via add_source - this makes sure that when adding a .(so|dll|pyd), we force zip_safe to False.
In addition this adds the ability to include other shared libraries with the pex. This is intended to be used for dependencies of the python extensions (e.g. a cityhash extension will depend on a cityhash library).

The way this works is a bit hacky and only works well on unix-ish platforms. It works by changing LD_LIBRARY_PATH and re-execing the interpreter (so that the loader picks up the path change). If this should ever need to support windows, using something like AddDllDirectory will likely be cleaner.

PR #166, opened 2015-10-07
Add the ability to include extensions & shared libraries in pexs.

rouge8:
Extensions already (sort-of) worked via add_source - this makes sure that when adding a .(so|dll|pyd), we force zip_safe to False.

In addition this adds the ability to include other shared libraries with the pex. This is intended to be used for dependencies of the python extensions (e.g. a cityhash extension will depend on a cityhash library).

This is an updated version of #61.

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -330,6 +330,14 @@ def configure_clp():
            'times.')
 
   parser.add_option(
+      '--native-library',
+      dest='native_libraries',
+      metavar='FILE',
+      default=[],
+      action='append',
+      help='Native library to include in the archive & loader path.')
+
+  parser.add_option(
       '-v',
       dest='verbosity',
       default=0,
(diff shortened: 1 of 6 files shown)


wickman:
pex adheres to what's in the standard lib and in PEPs.  My reluctance with this feature is that we're going out of the bounds of the python spec and more into the realm of cx_Freeze and p2app, which is a slippery slope of elaborate hacks to make stuff work when we already have ubiquitous solutions for managing dependencies holistically: docker, appc, etc.

It feels like this feature could be a wrapper around pex rather than baked into pex itself, where the wrapper is either heavyweight (docker) or lightweight (walk the pex, introspect the necessary .dylib/.so via something like pyelftools, and construct a shar or zip that sets up the path and execve's the pex.)

rouge8:
The PyPy tests pass for me locally with PyPy 2.6.1, which makes me suspect they're failing on Travis from the older PyPy...

mikekap:
FYI one way to fix this (as is done at https://github.com/facebook/buck/commit/f485090c08cfee550a7fd7f989b6556073c1c01b) - add an `-rpath @loader_path/` (or `-rpath $ORIGIN/` on linux) to your .so when linking. Then set zip_safe=False and everything should work :).

PR #169, opened 2015-10-09
Fixup `test_variables.test_rc_ignore`.

jsirois:
Previously the test wasn't testing what it advertised and as a result
was papering over some test state setup issues.


jsirois:
This is a fix for the test issue in #128

PR #182, opened 2015-11-19
fix infinite recursion when PEX_PYTHON points at a symlink

landism:
When PEX_PYTHON is in ~/.pexrc and set to a symlink, pex_bootstrapper repeatedly compares the symlink to the real path, determines they're not the same, and continues recursing.  This goes on for hundreds of iterations until (it seems) something gets shoved out of the max env var length and the whole thing blows up because it can't find "locale".


kwlzn:
merged @ a3b0a6a2b02eee483946368e853ebb2535820492 using the rbt workflow. closing.

PR #183, opened 2015-11-19
support /etc/pexrc

landism:
This will allow system administrators to make it more likely that pexes are executed using a "clean" python instead of whatever users have installed on their path.  /etc/pexrc is the first rc read so that it has the lowest precedence and determined users can still override whatever they like.


kwlzn:
merged @ 62ffd3b17f3bcca16e52d2fcb4879c345c383b9a via the rbt workflow. closing.

PR #185, opened 2015-12-01
Add the ability to specify pex_root on command line

digwanderlust:
* add pex_root option
* fix some places where ~/.pex was hard coded

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -159,7 +159,7 @@ def configure_clp_pex_resolution(parser, builder):
   group.add_option(
       '--cache-dir',
       dest='cache_dir',
-      default=os.path.expanduser('~/.pex/build'),
+      default=os.path.expanduser('{pex_root}/build'),
       help='The local cache directory to use for speeding up requirement '
            'lookups. [Default: %default]')
 
@@ -266,7 +266,7 @@ def configure_clp_pex_environment(parser):
   group.add_option(
       '--interpreter-cache-dir',
       dest='interpreter_cache_dir',
-      default=os.path.expanduser('~/.pex/interpreters'),
+      default=os.path.expanduser('{pex_root}/interpreters'),
       help='The interpreter cache to use for keeping track of interpreter dependencies '
            'for the pex tool. [Default: %default]')
 
@@ -336,6 +336,13 @@ def configure_clp():
       action='callback',
       callback=increment_verbosity,
       help='Turn on logging verbosity, may be specified multiple times.')
+
+  parser.add_option(
+    '--pex-root',
+    dest='pex_root',
+    default=None,
+    help='Specify the pex root used in this invocation of pex'
+  )
 
   parser.add_option(
       '--help-variables',
@@ -492,6 +499,10 @@ def build_pex(args, options, resolver_option_builder):
 
 
 def main():
+  def make_relative_to_root(path):
+    """Update options so that defaults are user relative to specified pex_root"""
+    return os.path.normpath(path.format(pex_root=ENV.PEX_ROOT))
+
   parser, resolver_options_builder = configure_clp()
 
   # split arguments early because optparse is dumb
@@ -503,8 +514,12 @@ def main():
     args, cmdline = args, []
 
   options, reqs = parser.parse_args(args=args)
+  if options.pex_root:
+    ENV.set('PEX_ROOT', options.pex_root)
+  options.cache_dir = make_relative_to_root(options.cache_dir)
+  options.interpreter_cache_dir = make_relative_to_root(options.interpreter_cache_dir)
 
-  with ENV.patch(PEX_VERBOSE=str(options.verbosity)):
+  with ENV.patch(PEX_VERBOSE=str(options.verbosity), PEX_ROOT=options.pex_root):
     with TRACER.timed('Building pex'):
       pex_builder = build_pex(reqs, options, resolver_options_builder)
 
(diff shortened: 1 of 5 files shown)


pex/pex_info.py
```diff
@@ -30,7 +30,7 @@ class PexInfo(object):
   requirements: list                # list of requirements for this environment
 
   # Environment options
-  pex_root: ~/.pex                   # root of all pex-related files
+  pex_root: eg: ~/.pex               # root of all pex-related files
```
kwlzn: shouldn't this actually include the type information per the format of the other lines, i.e. 'string'? maybe something like:

```pex_root: string, default: ~/.pex                              # root of all pex-related files```

File: pex/bin/pex.py [hunk 2]
```diff
@@ -159,7 +159,7 @@ def configure_clp_pex_resolution(parser, builder):
   group.add_option(
       '--cache-dir',
       dest='cache_dir',
-      default=os.path.expanduser('~/.pex/build'),
+      default=os.path.expanduser('{pex_root}/build'),
```
kwlzn: should this lose the `expanduser` treatment now that we don't specify a string literal with ~ in it?

pex/bin/pex.py
```diff
@@ -266,7 +266,7 @@ def configure_clp_pex_environment(parser):
   group.add_option(
       '--interpreter-cache-dir',
       dest='interpreter_cache_dir',
-      default=os.path.expanduser('~/.pex/interpreters'),
+      default=os.path.expanduser('{pex_root}/interpreters'),
```
kwlzn: should this also lose the `expanduser` treatment now that we don't specify a string literal with ~ in it?

pex/bin/pex.py
```diff
@@ -338,6 +338,13 @@ def configure_clp():
       help='Turn on logging verbosity, may be specified multiple times.')
 
   parser.add_option(
+    '--pex-root',
```
kwlzn: nit: this doesn't match the existing indentation style of 4 spaces.

pex/bin/pex.py
```diff
@@ -492,6 +499,10 @@ def build_pex(args, options, resolver_option_builder):
 
 
 def main():
+  def make_relative_to_root(path):
+    """Update options so that defaults are user relative to specified pex_root"""
```
kwlzn: nit: punctuation

pex/bin/pex.py
```diff
@@ -492,6 +499,10 @@ def build_pex(args, options, resolver_option_builder):
 
 
 def main():
+  def make_relative_to_root(path):
```
kwlzn: could this get de-nested outside of `main()` to declutter things?

pex/variables.py
```diff
@@ -238,7 +238,7 @@ def PEX_ROOT(self):
     The directory location for PEX to cache any dependencies and code.  PEX must write
     not-zip-safe eggs and all wheels to disk in order to activate them.  Default: ~/.pex
     """
-    return self._get_path('PEX_ROOT', default=None)
+    return self._get_path('PEX_ROOT', default='~/.pex')
```
kwlzn: should we `os.path.expanduser` this string here?

pex/bin/pex.py
```diff
@@ -503,8 +514,12 @@ def main():
     args, cmdline = args, []
 
   options, reqs = parser.parse_args(args=args)
+  if options.pex_root:
+    ENV.set('PEX_ROOT', options.pex_root)
+  options.cache_dir = make_relative_to_root(options.cache_dir)
+  options.interpreter_cache_dir = make_relative_to_root(options.interpreter_cache_dir)
 
-  with ENV.patch(PEX_VERBOSE=str(options.verbosity)):
+  with ENV.patch(PEX_VERBOSE=str(options.verbosity), PEX_ROOT=options.pex_root):
```
kwlzn: what happens if `options.pex_root` is `None` here? does that clobber the default `PEX_ROOT`?

pex/commands/bdist_pex.py
```diff
@@ -55,7 +55,7 @@ def run(self):
 
     reqs = [package_dir] + reqs
 
-    with ENV.patch(PEX_VERBOSE=str(options.verbosity)):
+    with ENV.patch(PEX_VERBOSE=str(options.verbosity), PEX_ROOT=options.pex_root):
```
kwlzn: what happens if `options.pex_root` is `None` here? does that clobber the default `PEX_ROOT`?

pex/bin/pex.py
```diff
@@ -503,8 +514,12 @@ def main():
     args, cmdline = args, []
 
   options, reqs = parser.parse_args(args=args)
+  if options.pex_root:
```
kwlzn: should we be doing the inverse here (e.g. set options.pex_root from ENV default) in the case that `options.pex_root` is `None`?

kwlzn:
apologies for the delay in getting to this - have a handful of comments/questions on the PR.

would also like to see some basic test coverage to avoid future regressions.

pex/bin/pex.py
digwanderlust: It is possible for pex root to have ~ in it.

pex/bin/pex.py
digwanderlust: its possible for pex_root to have ~ in it.

kwlzn:
this was abandoned in favor of #206, no?

PR #191, opened 2016-01-01
Unescapes html in PageParser.href_match_to_url

daveFNbuck:
PageParser breaks if the links contain any escaped html characters. This fixes that bug.

### DIFF
--- a/pex/crawler.py
+++ b/pex/crawler.py
@@ -22,6 +22,15 @@ else:
   from urlparse import urlparse
 
 
+def unescape(s):
+  """Unescapes html. Taken from https://wiki.python.org/moin/EscapingHtml"""
+  s = s.replace("&lt;", "<")
+  s = s.replace("&gt;", ">")
+  # this has to be last:
+  s = s.replace("&amp;", "&")
+  return s
+
+
 class PageParser(object):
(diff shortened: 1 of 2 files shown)


kwlzn:
would you mind also adding a quick test for this?

one mild concern to call out: this introduces a dependency on the `xml.*` class of stdlib modules which, depending on the environment python was compiled in (e.g. lacking supporting xml devel packages), may or may not be present in the stdlib on some systems/interpreters. I suppose that's sort of outside of our scope here tho - and I don't see a great portable alternative to `xml.sax.saxutils.unescape` short of writing our own.

daveFNbuck:
Sure, I can add a quick test. https://wiki.python.org/moin/EscapingHtml
suggests a short function we can use if the xml lib not existing is a
problem for some people. All I really need is the s.replace('&amp;', '&')
line, as the issue I'm having is with parameters being passed in the url on
my custom pypi server.

daveFNbuck:
ping

kwlzn:
lgtm! :shipit:

kwlzn:
thanks @daveFNbuck - this should be going out in the 1.1.2 release later today/tomorrow.

daveFNbuck:
Awesome, thanks!

Issue #140, opened 2015-07-30
"too many values to unpack" when parsing interpreter versions


kwlzn:
in pex as used by pants, we've observed about a half dozen users over the months reporting occasional issues with pants/pex throwing "too many values to unpack". usually clearing the interpreter cache fixes this - but it happens frequently enough that I think it should be repaired.

the traceback is:

```
Exception caught:
  File "/Users/easebiomo/.pants.d/bin/pants.pex/pants-1.2.3rc7-testing-py27.pex/.bootstrap/_pex/pex.py", line 271, in execute
    self.execute_entry(entry_point, args)
  File "/Users/easebiomo/.pants.d/bin/pants.pex/pants-1.2.3rc7-testing-py27.pex/.bootstrap/_pex/pex.py", line 320, in execute_entry
    runner(entry_point)
  File "/Users/easebiomo/.pants.d/bin/pants.pex/pants-1.2.3rc7-testing-py27.pex/.bootstrap/_pex/pex.py", line 343, in execute_pkg_resources
    runner()
  File "/Users/easebiomo/workspace/source/.pex/install/pantsbuild.pants-0.0.33-py2-none-any.whl.cd9bdb0408ea39e69875c17237da93bc5427dcdf/pantsbuild.pants-0.0.33-py2-none-any.whl/pants/bin/pants_exe.py", line 81, in main
    _run(exiter)
  File "/Users/easebiomo/workspace/source/.pex/install/pantsbuild.pants-0.0.33-py2-none-any.whl.cd9bdb0408ea39e69875c17237da93bc5427dcdf/pantsbuild.pants-0.0.33-py2-none-any.whl/pants/bin/pants_exe.py", line 74, in _run
    result = goal_runner.run()
  File "/Users/easebiomo/workspace/source/.pex/install/pantsbuild.pants-0.0.33-py2-none-any.whl.cd9bdb0408ea39e69875c17237da93bc5427dcdf/pantsbuild.pants-0.0.33-py2-none-any.whl/pants/bin/goal_runner.py", line 218, in run
    result = self._do_run()
  File "/Users/easebiomo/workspace/source/.pex/install/pantsbuild.pants-0.0.33-py2-none-any.whl.cd9bdb0408ea39e69875c17237da93bc5427dcdf/pantsbuild.pants-0.0.33-py2-none-any.whl/pants/bin/goal_runner.py", line 275, in _do_run
    return engine.execute(context, self.goals)
[...]

PR #192, opened 2016-01-01
Updates interpreter path parsing to handle Jenkins

daveFNbuck:
When building a pex on Jenkins inside an auto-generated virtualenv, I keep ending up with interpreter names like `CPython-2.7.3.tmp.ba3a2c39daab46ef978ba07fe335079d`. If we simply split on dots to get version number, we'll end up with an invalid version number. To avoid this, we now just take the first 3 items when splitting on a dot. This allows pex building for me on Jenkins.

### DIFF
--- a/pex/interpreter.py
+++ b/pex/interpreter.py
@@ -95,7 +95,7 @@ class PythonIdentity(object):
   @classmethod
   def from_path(cls, dirname):
     interp, version = dirname.split('-')
-    major, minor, patch = version.split('.')
+    major, minor, patch = version.split('.')[:3]
     return cls(str(interp), int(major), int(minor), int(patch))
 
   def __init__(self, interpreter, major, minor, patch):


pex/interpreter.py
```diff
@@ -95,7 +95,7 @@ def from_id_string(cls, id_string):
   @classmethod
   def from_path(cls, dirname):
     interp, version = dirname.split('-')
-    major, minor, patch = version.split('.')
+    major, minor, patch = version.split('.')[:3]
```
kwlzn: I think the larger question here is why the interpreter discovery mechanism is finding what it deems valid interpreters that look like 'CPython-2.7.3.tmp.ba3a2c39daab46ef978ba07fe335079d' or where they're coming from to begin with. 

any clues from your analysis? here's what these should normally look like in cache:

```
$ find ~/.pex/interpreters
/Users/kw/.pex/interpreters
/Users/kw/.pex/interpreters/CPython-2.7.10
/Users/kw/.pex/interpreters/CPython-2.7.10/wheel
/Users/kw/.pex/interpreters/CPython-2.7.10/wheel-0.24.0-py2.7.egg
/Users/kw/.pex/interpreters/CPython-2.7.6
/Users/kw/.pex/interpreters/CPython-2.7.6/wheel
/Users/kw/.pex/interpreters/CPython-2.7.6/wheel-0.24.0-py2.7.egg
```

but in your interpreter cache, you'd probably find the `*.tmp.*` variety which is likely where this code is choking.

related:
- https://github.com/pantsbuild/pex/issues/140
- https://github.com/pantsbuild/pants/issues/2038

daveFNbuck: It seems that this happened because of the failure fixed in #191. I can't
replicate it on my laptop, so it's probably something weird with how my CI
generates temporary workspaces. With #191 applied I don't have this issue,
so I'm ok to close.

PR #198, opened 2016-01-12
Fixes to get pex to work on windows.

mikekap:
The fixes are:
 - os.link doesn't exist on windows. Always copy instead.
 - NamedTemporaryFile doesn't work correctly on windows (see https://bugs.python.org/issue14243)
 - sys.prefix is part of site.getsitepackages() on windows. Don't remove it.

### DIFF
--- a/pex/common.py
+++ b/pex/common.py
@@ -28,18 +28,26 @@ def safe_copy(source, dest, overwrite=False):
     shutil.copyfile(source, temp_dest)
     os.rename(temp_dest, dest)
 
-  try:
-    os.link(source, dest)
-  except OSError as e:
-    if e.errno == errno.EEXIST:
-      # File already exists.  If overwrite=True, write otherwise skip.
-      if overwrite:
+  # If the platform supports hard-linking, use that and fall back to copying.
+  # Windows does not support hard-linking.
+  if hasattr(os, 'link'):
+    try:
+      os.link(source, dest)
+    except OSError as e:
+      if e.errno == errno.EEXIST:
+        # File already exists.  If overwrite=True, write otherwise skip.
+        if overwrite:
+          do_copy()
+      elif e.errno == errno.EXDEV:
+        # Hard link across devices, fall back on copying
         do_copy()
-    elif e.errno == errno.EXDEV:
-      # Hard link across devices, fall back on copying
+      else:
+        raise
+  elif os.path.exists(dest):
+    if overwrite:
       do_copy()
-    else:
-      raise
+  else:
+    do_copy()
 
 
 # See http://stackoverflow.com/questions/2572172/referencing-other-modules-in-atexit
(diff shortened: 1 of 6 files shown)


pex/common.py
```diff
@@ -28,18 +28,24 @@ def do_copy():
     shutil.copyfile(source, temp_dest)
     os.rename(temp_dest, dest)
 
-  try:
-    os.link(source, dest)
-  except OSError as e:
-    if e.errno == errno.EEXIST:
-      # File already exists.  If overwrite=True, write otherwise skip.
-      if overwrite:
+  if hasattr(os, 'link'):
```
kwlzn: how about a quick comment here to indicate the reason for doing this (i.e. windows) so that a casual observer doesn't get confused by the otherwise odd `hasattr` check here.

pex/common.py
```diff
@@ -264,17 +270,8 @@ def link(self, src, dst, label=None):
...
-      if e.errno == errno.EEXIST:
-        # File already exists, skip XXX -- ensure target and dest are same?
-        pass
-      elif e.errno == errno.EXDEV:
-        # Hard link across devices, fall back on copying
-        shutil.copyfile(abs_src, abs_dst)
-      else:
-        raise
+    safe_copy(abs_src, abs_dst, overwrite=False)
+    # If the file already exists, skip XXX -- ensure target and dest are same?
```
kwlzn: mind making this a TODO and rewording it a bit to make sense in the modified context?

pex/util.py
```diff
@@ -200,3 +201,17 @@ def get(self, key, default=None):
   def store(self, key, value):
     with self._lock:
       self._data[key] = value
+
+
+@contextlib.contextmanager
+def named_temporary_file():
```
kwlzn: should this plumb `*args, **kwargs` through to `NamedTemporaryFile` in the event someone wants to pass params to this helper in the future?

kwlzn:
lgtm overall mod a few comments. also, test coverage simulating windows environmental conditions (+ a quick unit test for `named_temporary_file`?) would be ideal.

mikekap:
Might it be better to set up windows CI (e.g. Appveyor) to make sure this works? Unfortunately I don't think there are currently tests for _site_libs other than integration tests.

kwlzn:
generally speaking, pex doesn't claim or strongly desire windows support - my understanding is that it's essentially "best effort" via submission of PRs from interested contributors. with this said, I'm not sure it makes sense for us to add Appveyor at the moment without a stronger sense of ongoing Windows support ownership - hence the suggestion to simply mock out some of the windows conditions in test.

mikekap:
Added a couple of tests for site_libs. Let me know if that's what you had in mind.

I also managed to finally get all the tests to pass on windows (you can see the commits @ https://github.com/pantsbuild/pex/compare/master...mikekap:windows?expand=1 & the build @ https://ci.appveyor.com/project/mikekap/pex/build/1.0.33). I'll open another PR once this one goes in since the changes there are a bit larger.

mikekap:
ping :)

sdwilsh:
The Buck project would love to see these upstreamed so we can stop having a variety of changes to PEX in our codebase :)

kwlzn:
lgtm, thanks @mikekap!

will you guys need a pypi release to consume this or are you vendoring?

Issue #68, opened 2015-04-14
Make --platform/--interpreter multi-options


wickman:
This is to make it easier to build multi-platform pex files a la pants.

bluepeppers:
Is pants the only way to do this at the moment? I can't seem to find any documentation on it

wickman:
sadly yes but this should be a pretty straightforward PR

bluepeppers:
If I understand it correctly, it would be a case of iterating through the combinations and creating resolvers for each, installing packages from those resolvers and then only after that, creating/freezing the pex file? I can't see where Pants is handling its multiplatform stuff

jjjake:
@wickman is anybody working on this? I'm not very familiar with the pex codebase, but I would like to take a try at implementing this if nobody else is.

We're using pex at the Internet Archive to generate binaries for our `ia` command-line tool (https://github.com/jjjake/internetarchive). I would like to add some dependencies that would require platform specific distributions, and it would be great if we could package everything into a single binary that could be used on multiple platforms.

Thanks for all the work on pex, it's been a big win for us at the Archive being able to easily package the `ia` command-line tool as a single binary that works on many platforms. : )

kwlzn:
@jjjake I've not heard of anyone planning to tackle this in `pex` anytime soon, so I'd say feel free to take a stab at it.

btw, per the earlier conversation on this thread here's the relevant code from `pants`: https://github.com/pantsbuild/pants/blob/97c0bdba3eb39ee8dd4836a6547bed7e6cf937d6/src/python/pants/backend/python/python_chroot.py#L241

it'd be ideal if we could subsume this functionality directly into `pex` at the library level (e.g. a pluralized `platforms=` kwarg to `pex.resolver.resolve` etc), such that `pants` could utilize baked-in multi-platform support via the API and drop its own multi-platform logic.

jjjake:
Awesome. Thank you for the details @kwlzn, those are helpful. I'll go ahead and take a stab at it.

jjjake:
@kwlzn here's my first pass at making --platform/--interpreter multi-options: https://github.com/jjjake/pex/commit/d948fa9c5de366b651ae891aa59c016c4287d4b5

Please let me know if I'm on the right track, and I'll work on cleaning things up, docstrings and tests for this change. Thanks!

kwlzn:
[...]

PR #201, opened 2016-01-20
Make --platform/--interpreter multi-options

jjjake:


kwlzn:
looks like this PR failed CI, so no one took a look - but it'd be great to have this feature in master.

any thoughts on getting this finalized? if time doesn't permit, you could also offer this up to another contributor to run with as it seems to be mostly there.

jjjake:
Thanks for following up @kwlzn, 'preciate it.

I do not have the time to work on this at the moment, and would be happy to hand it off to another contributor.

FWIW, I am currently using my resolve_multi branch to generate multi-platform pex files, so hopefully there won't be too much more work. I'm not very familiar with the pex source code, though, so I may be underestimating some of the corner cases.

mzdaniel:
Hey @jjjake . Thank you for your implementing in this feature. So far, we have now all test passing. Not sure though how the CLI should work. As per @kwkn request, I am following this discussion at #68

kwlzn:
this has been superseded by #394 which is now in master, so closing this out. thanks @jjjake for getting the ball rolling.

PR #202, opened 2016-01-20
More fixes for windows

mikekap:
Now all the tests pass.

One thing I'm not sure about is in the first commit - perhaps we should do away with Link.local_path & make Link.path do the proper thing? I don't think there's a single user of Link.path that uses it for real urls. Let me know what you think.

### DIFF
--- a/pex/compatibility.py
+++ b/pex/compatibility.py
@@ -4,6 +4,7 @@
 # This file contains several 2.x/3.x compatibility checkstyle violations for a reason
 # checkstyle: noqa
 
+import os
 from abc import ABCMeta
 from numbers import Integral, Real
 from sys import version_info as sys_version_info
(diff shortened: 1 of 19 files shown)


mikekap:
Ping :)

mikekap:
Ping.

mikekap:
Would this be easier to review if I made separate PRs for each commit? That would be fairly easy to do since they're not really dependent on each other.

mikekap:
Fixed the comments; PTAL @ the link commit, since it involved some urllib.unquote additions (which I think make Link.py even more correct).

Almost feels like link.py should be open-sourced on it's own :P

kwlzn:
lgtm! :+1:

kwlzn:
thanks for the PR @mikekap - I'll get a release out soon so folks can consume this via pypi.

PR #204, opened 2016-01-25
Don't normalize version numbers as names.

mikekap:
Version numbers can have + in them as a local version - `0.1-pex.1` & `0.1+pex.1` take completely different paths in pkg_resources.


mikekap:
Ping

kwlzn:
lgtm!

PR #211, opened 2016-02-16
Specify alter_sys=True when invoking runpy.run_module

### LINKED ISSUE #210: unittest.main() doesn't work
Background:
A typical python unit test code file looks like

```
import unittest

class MyTest(unittest.TestCase):
  def test_xyz(self):
    ...

if __name__ == '__main__':
  unittest.main()
```

`unittest.main` by default runs all the test cases in module `__main__`.

`pex` executes the entry point via `runpy.run_module(module_name, run_name='__main__')`: https://github.com/pantsbuild/pex/blob/a253efd4a4d9f33f5ac7f6f397902f62fb1e517f/pex/pex.py#L421

This sets `__name__` to be `__main__` in the entry point's execution context, and therefore the `if` check in the example code above passes. However, when `unittest.main()` tries to load the `__main__` module, what it loads is the `__main__.py` file generated by pex, not the entry point.

A simple fix is to specify `alter_sys=True` when invoking `runpy.run_module`, which will set `sys.modules[__name__]` to the entry point code.

guoshimin:
Specify `alter_sys=True` when invoking `runpy.run_module` so that `__main__` points to the entry point module.

Fixes #210

### DIFF
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -427,7 +427,7 @@ class PEX(object):  # noqa: T000
   @staticmethod
   def execute_module(module_name):
     import runpy
-    runpy.run_module(module_name, run_name='__main__')
+    runpy.run_module(module_name, run_name='__main__', alter_sys=True)
 
   @staticmethod
   def execute_pkg_resources(spec):


mikekap:
I think there's quite a few scripts out there that expect `sys.argv[0]` to be the path to the pex package itself. I think it might be safer to just patch `sys.modules` by hand to avoid changing `sys.argv[0]`.

guoshimin:
Fair point about code expecting the existing behavior of `sys.argv[0]`. Alas the `runpy.run_module` function doesn't provide a way to override only `sys.modules` but not `sys.argv[0]`. So patching `sys.modules` by hand would amount to reinventing much of the logic in `run_module`.

guoshimin:
It will essentially be inlining the `run_module` method. Something like

```
def run_module_with_sys_modules_override(mod_name):
  run_name = '__main__'
  _, loader, code, fname = runpy._get_module_details(mod_name)
  pkg_name = mod_name.rpartition('.')[0]
  with runpy._TempModule(run_name) as temp_module:
      mod_globals = temp_module.module.__dict__
      runpy._run_code(code, mod_globals, None,
                      run_name, fname, loader, pkg_name)
```
Calling the internal methods of runpy isn't ideal.

mikekap:
Definitely agree that depending on runpy internals isn't great. Before we go down too much into thinking how to implement this, I'd like to make a case that leaving `sys.argv[0]` alone is probably the more correct behavior in general. `sys.argv[0]` is how the program was started. You should be able to `os.exec([sys.executable] + sys.argv)` for essentially restarting the program. Does that seem reasonable?

From reading https://www.python.org/dev/peps/pep-0338/ , it looks like the argument for changing `sys.argv[0]` is that some scripts expect it. The example given is `regrtest.py`, and indeed, the script would not work due to an assert: https://github.com/python/cpython/blob/2.7/Lib/test/regrtest.py#L1588 . However, I can't figure out why that assert is there at all - nothing else seems to read `sys.argv[0]`. It seems to be a remnant of another time where the script used `sys.argv[0]` instead of `__file__` and slowly transitioned to `__file__` (which matches the pattern in other parts of the python world).

If the above argument makes sense, and others agree with it, the right thing to do here would probably be to propose changing runpy for python 3.6+. Assuming that's accepted, we can backport the functionality with the "hack" you posted for < 3.6. I'll send an mail to python-dev to see how folks feel about this.

guoshimin:
Yeah totally agree `sys.argv[0]` is best left alone. Looks like I'll need to patch pex for my organization for the time being.

ncoghlan:
The rationale for changing sys.argv[0] is so tools like debuggers and coverage analysers can use run_module to invoke the script to be run, while still making the invoked script believe its running normally.

However, there are currently still some significant challenges in doing that in practice: https://bugs.python.org/issue9325

PR #212, opened 2016-02-16
Fix creating an environment pex via bdist_pex

mikekap:
target wasn't defined in a branch.

I couldn't find any tests for bdist_pex though :(


kwlzn:
thx!

Issue #213, opened 2016-03-04
Support abi types for python 2.x


guoshimin:
Currently pex only supports abi types like 'dmu' for python 3: https://github.com/pantsbuild/pex/blob/f5721a98f21dc680a0f541cede1afa724d2a4159/pex/pep425.py#L102

I built a numpy wheel on linux, and it has filename numpy-1.10.4-cp27-cp27mu-linux_x86_64.whl. Because there is no support for abi type mu for python 2.7, pex doesn't think it's a valid package for numpy on linux_x86_64.

kwlzn:
confirming this is indeed an issue. repro:

```
[illuminati zzz]$ find deps -type f
deps/cryptography-1.2.3-cp27-cp27m-macosx_10_10_intel.whl
[illuminati zzz]$ pex --disable-cache --no-pypi -f ./deps cryptography==1.2.3 -v -v -v
pex: crawling link i=0 link=Link('file:///private/tmp/zzz/deps') follow_links=False                                    
Could not satisfy all requirements for cryptography==1.2.3:
    cryptography==1.2.3
[illuminati zzz]$ cp deps/cryptography-1.2.3-cp27-{cp27m,none}-macosx_10_10_intel.whl 
[illuminati zzz]$ pex --disable-cache --no-pypi -f ./deps cryptography==1.2.3 -v -v -v
pex: crawling link i=0 link=Link('file:///private/tmp/zzz/deps') follow_links=False                                    
pex: Building pex :: Resolving distributions :: Fetching file:///private/tmp/zzz/deps/cryptography-1.2.3-cp27-none-macospex: Building pex :: Resolving distributions :: Fetching file:///private/tmp/zzz/deps/cryptography-1.2.3-cp27-none-macospex: Building pex :: Resolving distributions :: Translating /private/var/folders/3t/xkwqrkld4xxgklk2s4n41jb80000gn/T/tmpCould not satisfy all requirements for enum34:10_intel.whl into distribution                                  
    enum34(from: cryptography==1.2.3)
```

ewang:
I added a PR for this in #214. But not quite sure if the implementation is 100% accurate since I couldn't find much documentation online regarding py2.x ABI tags.

kwlzn:
this should now be fixed in master (thanks @ewang!) - I'll get a pypi release cut ~tomorrow for general consumption.

can follow #215 for the release.

guoshimin:
Awesome! Thanks for the quick fix!

PR #214, opened 2016-03-11
Add python 2.x abi tag support

ewang:
Not quite sure how to test this short of running it across a whole bunch of libraries. I updated the test case accordingly.

Left the existing code pertaining to the cp3 abi tags as is in case we need to change the supported tags for cp2 and cp3 independently.

But it has been tested with the cryptography==1.2.3 issue mentioned in #213.

### DIFF
--- a/pex/pep425.py
+++ b/pex/pep425.py
@@ -99,14 +99,18 @@ class PEP425(object):  # noqa
     """
     # Predict soabi for reasonable interpreters.  This is technically wrong but essentially right.
     abis = []
-    if impl == 'cp' and version.startswith('3'):
+    if impl == 'cp' and (version.startswith('2') or version.startswith('3')):
       abis.extend([
         'cp%s' % version,
         'cp%sdmu' % version, 'cp%sdm' % version, 'cp%sdu' % version, 'cp%sd' % version,
         'cp%smu' % version, 'cp%sm' % version,
-        'cp%su' % version,
-        'abi3'
+        'cp%su' % version
       ])
+
+      if version.startswith('3'):
+        abis.extend([
+          'abi3'
+        ])
 
     major_version = int(version[0])
     minor_versions = []
--- a/tests/test_pep425.py
+++ b/tests/test_pep425.py
@@ -46,6 +46,18 @@ def test_iter_supported_tags():
     for interp in ('cp', 'py'):
       for interp_suffix in ('2', '20', '21', '22', '23', '24', '25', '26'):
         for platform in ('linux_x86_64', 'any'):
-          yield (interp + interp_suffix, 'none', platform)
+          abis = ['none']
+
+          if interp == 'cp' and interp_suffix == '26' and platform == 'linux_x86_64':
+            abis.extend([
+              'cp%s' % interp_suffix,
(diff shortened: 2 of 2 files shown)


pex/pep425.py
```diff
@@ -99,6 +99,14 @@ def _iter_supported_tags(cls, impl, version, platform):
     """
     # Predict soabi for reasonable interpreters.  This is technically wrong but essentially right.
     abis = []
+    if impl == 'cp' and version.startswith('2'):
```
kwlzn: how about just checking for either '2' or '3' and reusing the abis.extend call + appending the 'abi3' bit to eliminate duplication?

something like:

```
if impl == 'cp' and (version.startswith('2') or version.startswith('3')):
  abis.extend(...)
  if version.startswith('3'):
    abis.append('abi3')
```

kwlzn:
one comment, otherwise lgtm.

btw, here's the relevant bit from pip for py2.x and pypy if you wanted to go all the way and match that behavior (i.e. `if impl in ('cp', 'pp') and (version...):`)

https://github.com/natefoo/pip/commit/0b9afeaca433d98eda363d564d40bb9484be43a9#diff-41909e50a87e91a77f6c81f9521a9329R59

pex/pep425.py
ewang: I did that initially, but wasn't sure if all the abi tags from cp3 also applied to cp2; the docs on seems to be lacking on this topic. The 2 separate conditionals allows for keeping the 2 lists independent.

I can extract out the common ones if that's preferred. Let me know.

kwlzn: I think it's safe to assume based on pip's implementation that this initial set will apply to both cp2 and cp3. if other python3 specific items beyond 'abi3' are needed, presumably they'd be additive and could be handled in the `if version.startswith('3')` branch.

kwlzn: err, I spoke too soon - looks like the current implementation is:

https://github.com/pypa/pip/blob/develop/pip/pep425tags.py#L85

which seems to have special case for 'u' requiring python < 3.3

ewang:
Fixed the PR and squashed the commits. I think this one should be good to go for now. Any idea if we could get this included in the next weekly pants release?

I will look into the pip implementation, but that implementation will require rewriting the test cases to simulate the various interpreter flags. That said, the pip implementation only returns a single tag for the current interpreter, whereas the pex implementation is to return the entire list of compatible tags, so they aren't doing quite the same thing.

kwlzn:
sounds good - feel free to also just log an issue if you don't feel like dealing with it and then when a consumer comes along that needs it they can implement it themselves.

anyway, thanks for the PR! I'll try to get a pex release out tomorrow - but it's probably a good idea to let a version bump bake in pants master for at least a couple days, so tomorrows pants release may be a little early. would next friday's pants release work?

ewang:
Yea, next week's pants release should be fine.

Also, just wondering if the pex version used by pants could be overridden the same way the bootstrapped jar tools could be overridden via the BUILD.tools file as a python_requirement_library?

kwlzn:
@ewang depending on how you build/run pants, you could probably just cherrypick the requirements.txt version bump before building to use this (or just build from master after the version bump lands).

we actually build from master vs release tags and use cherry picks quite often in our pants deployments.

PR #224, opened 2016-04-04
Add a test which enables PEX_VERBOSE and checks whether enabling inhe…

KatieLucas-Grapeshot:
Grapeshot uses Pants to build PEX objects. We install MySQL interface code in site-packages and hence would like PEXes to collect this path on start-up. We have discovered the PANTS option to do this doesn't work. Fault traced to the environment scrubbing not honouring the setting or the equiv env variable.

PR adds a test which build such a PEX, enables PEX_VERBOSE, launches PEX and checks whether the enabling of inherit_path results in site-packages being not scrubbed. Since this fails, add processing of inherit_path to the pex environment booting. Test now passes.

### DIFF
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -171,7 +171,6 @@ class PEX(object):  # noqa: T000
 
     user_site_distributions.update(all_distribution_paths(USER_SITE))
 
-
     for path in user_site_distributions:
       TRACER.log('Scrubbing from user site: %s' % path)
 
@@ -318,7 +317,8 @@ class PEX(object):  # noqa: T000
     """
     teardown_verbosity = self._vars.PEX_TEARDOWN_VERBOSE
     try:
-      pex_inherit_path = True if self._vars.PEX_INHERIT_PATH or self._pex_info.inherit_path else False
+      pex_inherit_path = True if (
+        self._vars.PEX_INHERIT_PATH or self._pex_info.inherit_path) else False
       with self.patch_sys(pex_inherit_path):
         working_set = self._activate()
         TRACER.log('PYTHONPATH contains:')
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -154,7 +154,7 @@ class PEX(object):  # noqa: T000
     return new_modules
 
   @classmethod
-  def minimum_sys_path(cls, site_libs):
+  def minimum_sys_path(cls, site_libs, inherit_path):
     site_distributions = OrderedSet()
     user_site_distributions = OrderedSet()
 
@@ -171,13 +171,17 @@ class PEX(object):  # noqa: T000
 
     user_site_distributions.update(all_distribution_paths(USER_SITE))
 
-    for path in site_distributions:
-      TRACER.log('Scrubbing from site-packages: %s' % path)
 
     for path in user_site_distributions:
       TRACER.log('Scrubbing from user site: %s' % path)
 
-    scrub_paths = site_distributions | user_site_distributions
+    if inherit_path:
+      scrub_paths = user_site_distributions
+    else:
+      scrub_paths = site_distributions | user_site_distributions
+      for path in site_distributions:
+        TRACER.log('Scrubbing from site-packages: %s' % path)
+
     scrubbed_sys_path = list(OrderedSet(sys.path) - scrub_paths)
     scrub_from_importer_cache = filter(
       lambda key: any(key.startswith(path) for path in scrub_paths),
@@ -191,7 +195,7 @@ class PEX(object):  # noqa: T000
     return scrubbed_sys_path, scrubbed_importer_cache
 
   @classmethod
-  def minimum_sys(cls):
+  def minimum_sys(cls, inherit_path):
     """Return the minimum sys necessary to run this interpreter, a la python -S.
 
     :returns: (sys.path, sys.path_importer_cache, sys.modules) tuple of a
@@ -205,7 +209,7 @@ class PEX(object):  # noqa: T000
       site_libs.add(extras_path)
     site_libs = set(os.path.normpath(path) for path in site_libs)
 
-    sys_path, sys_path_importer_cache = cls.minimum_sys_path(site_libs)
+    sys_path, sys_path_importer_cache = cls.minimum_sys_path(site_libs, inherit_path)
     sys_modules = cls.minimum_sys_modules(site_libs)
 
     return sys_path, sys_path_importer_cache, sys_modules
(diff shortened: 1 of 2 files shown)


tests/test_inherits_path_option.py
```diff
@@ -0,0 +1,47 @@
...
+  :type exe_contents: string
+  :param dists: distributions to include, typically sdists or bdists
+  :param coverage: include coverage header
+  """
+  dists = dists or []
+
+  with open(os.path.join(td, 'exe.py'), 'w') as fp:
+    fp.write(exe_contents)
+
+  pb = PEXBuilder(path=td, preamble=COVERAGE_PREAMBLE if coverage else None)
```
jsirois: This line doesn't fail only because coverage is never passed an defaults to `False`, otherwise boom on `COVERAGE_PREAMBLE` which is not a symbol in-scope.  Killing the unused `dists` and `coverage` params would be good.

tests/test_inherits_path_option.py
```diff
@@ -0,0 +1,47 @@
...
+  pb.info.inherit_path = True
+
+  for dist in dists:
+    pb.add_egg(dist.location)
+
+  pb.set_executable(os.path.join(td, 'exe.py'))
+  pb.freeze()
+
+  return pb
+
```
jsirois: 1 more blank line to make 2 between top-level functions (http://legacy.python.org/dev/peps/pep-0008/#blank-lines).

tests/test_inherits_path_option.py
```diff
@@ -0,0 +1,47 @@
...
+def test_inherits_path_option():
+  with temporary_dir() as td:
+    pb = write_inheriting_pex(td, 'import sys\nimport os\nprint(os.environ)\nprint(sys.path)')
+    pex_path = os.path.join(td, 'show_path.pex')
+    pb.build(pex_path)
+
+    env = os.environ.copy()
```
jsirois: The `twitter.common.contextutil` module also has the `environment_as` context manager (use here: https://github.com/pantsbuild/pex/blob/master/tests/test_integration.py#L42) - prefer using that here to polluting the test runtime env.

jsirois:
A few things to fix above, but also Travis-CI should be green as well.
If you can get these issues and Travis all fixed up I can dive in for a bit of a closer look.

tests/test_inherits_path_option.py
KatieLucas-Grapeshot: Done!

(Also, might look at using that code in our stuff here; there's a lot of environment juggling)

tests/test_inherits_path_option.py
KatieLucas-Grapeshot: Done.

tests/test_inherits_path_option.py
KatieLucas-Grapeshot: Vars killed.

KatieLucas-Grapeshot:
Am looking at Travis fails.

pex/pex.py
```diff
@@ -314,7 +317,9 @@ def execute(self):
     """
     teardown_verbosity = self._vars.PEX_TEARDOWN_VERBOSE
     try:
-      with self.patch_sys():
+      pex_inherit_path = True if (
+        self._vars.PEX_INHERIT_PATH or self._pex_info.inherit_path) else False
```
jsirois: Just the bool expression itself would be more idiomatic and I think clearer:
```python
pex_inherit_path = self._vars.PEX_INHERIT_PATH or self._pex_info.inherit_path
```

tests/test_inherits_path_option.py
```diff
@@ -0,0 +1,51 @@
...
+  pb.info.inherit_path = True
+  pb.set_executable(os.path.join(td, 'exe.py'))
+  pb.freeze()
+
+  return pb
+
+
+def test_inherits_path_option():
+  with temporary_dir() as td:
+    pb = write_inheriting_pex(td, 'import sys\nimport os\nprint(os.environ)\nprint(sys.path)')
```
jsirois: This is a bit deceiving (here and below in `test_does_not_inherit_path_option`) since the environ and sys.path printing goes unused.  You might just use an empty script to emphasize the point of the test is checking the verbose output.

tests/test_inherits_path_option.py
```diff
@@ -0,0 +1,51 @@
...
+    pb.build(pex_path)
+
+    with environment_as(PEX_VERBOSE='1'):
+      so, rc = run_simple_pex(pex_path)
+      assert 'Scrubbing from site-packages' not in str(so), 'Site packages should not be scrubbed.'
+
+
+def test_does_not_inherit_path_option():
+  with temporary_dir() as td:
+    pb = write_simple_pex(td, 'import sys\nimport os\nprint(os.environ)\nprint(sys.path)')
```
jsirois: You might also `s/write_inheriting_pex(...)/write_pex(inherit_path)/`, drop use of  `write_simple_pex` and DRY most this up.  The final tests minus the helper function(s) could just be:
```python
def test_inherits_path_option():
  with write_pex(inherit_path=True) as so:
    assert 'Scrubbing from site-packages' not in str(so), 'Site packages should not be scrubbed.'

def test_does_not_inherit_path_option():
  with write_pex(inherit_path=False) as so:
    assert 'Scrubbing from site-packages' in str(so), 'Site packages should be scrubbed.'
```

jsirois:
The last 3 comments are all debateable nits only, this change LGTM.  Let me know on each, your call.  I'll then merge this in.

Thanks for this fix!

jsirois:
I glossed over this when reading the PR description, but it sounds like this fix will be consumed via pants so I'm linking this to the 1.1.5 release ticket #237.  Would you all be insterested in contributing Grapeshot to the powered-by page?: http://pantsbuild.github.io/powered_by.html

KatieLucas-Grapeshot:
Oh, yes, it's Pantsbuild we're actually using to compile things. (I'm
trying to shove it into more of the work after reading what people have
done to GNU Make in the interests of expediency).

I'm pretty sure the CEO will do that big happy grin of his to see us listed
up there :-)

I'll send an email.

KatieLucas-Grapeshot:
Oh, we should do this properly. I haven't done any OS contribs in *yonks*
(and this'll be my first Python one). Might as well get it right :-)

I hadn't realised the pex writer took options, which does sort of collapse
the testcase right down...

jsirois:
Thanks @KatieLucas-Grapeshot!  This has been patched into master as 1 squashed commit keeping the initial description as the commit message here: https://github.com/pantsbuild/pex/commit/adefd7e9f862468add47bcb279684a4448d1036a

PR #227, opened 2016-04-07
Fix old pkg_resources egg version normalization.

### LINKED ISSUE #226: Handle eggs for libraries with versions containing _ in the filename
In 1.1.3, there was a change that caused the versions of packages to not be normalized anymore. This is a problem because the egg filename spec normalizes versions using to_filename, which turns '-'s into '_'s. So, any egg with a version containing an '-' or any character normalized to a '-' can't be resolved with pex's resolver.

mikekap:
Before recent releases of `setuptools`, version numbers in egg filenames were mangled. `safe_version` handles this correctly in modern `setuptools` and should work on older setuptools as well.

Closes #226.

### DIFF
--- a/pex/package.py
+++ b/pex/package.py
@@ -3,7 +3,7 @@
 
 import os
 
-from pkg_resources import EGG_NAME, parse_version, safe_name
+from pkg_resources import EGG_NAME, parse_version, safe_name, safe_version
 
 from .archiver import Archiver
 from .base import maybe_requirement
@@ -133,7 +133,7 @@ class SourcePackage(Package):
 
   @property
   def raw_version(self):
-    return self._raw_version
+    return safe_version(self._raw_version)
 
   # SourcePackages are always compatible as they can be translated to a distribution.
   def compatible(self, identity, platform=Platform.current()):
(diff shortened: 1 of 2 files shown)


jsirois:
The function at least goes back ~forever So LGTM:
```
jsirois@gill ~/dev/pypa/setuptools (master) $ git log -S"def safe_version"
commit 0c9e886999bb42cf825054778b1664e41f8164b4
Author: PJ Eby <distutils-sig@python.org>
Date:   Sun Jun 5 18:59:37 2005 +0000

    Add "safe_name" and "safe_version" functions to allow sanitizing of
    distribution names and versions in arbitrary packages that might be built
    using EasyInstall.
    
    --HG--
    branch : setuptools
    extra : convert_revision : svn%3A6015fed2-1504-0410-9fe1-9d1591cc4771/sandbox/trunk/setuptools%4041040
```

jsirois:
Thanks @mikekap and sorry for the delay reviewing this.  Patched into master @ https://github.com/pantsbuild/pex/commit/4966afaee400b7d68074695873ce8155f6a759cb

PR #231, opened 2016-04-13
open with utf-8 support

acaire:
Without this fix, pex will raise pex.resolver.Untranslateable due to
non utf-8 characters in setup.py or \__init\__.py:

$ pex -r requirements.txt -c foobar -o foobar.pex
**** Failed to install pygments-markdown-lexer-0.1.0.dev39. stdout:

**** Failed to install pygments-markdown-lexer-0.1.0.dev39. stderr:
  File "<stdin>", line 7, in <module>
  File "/tmp/virtualenv/lib/python3.4/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 158: ordinal not in range(128)

  File "/tmp/virtualenv/bin/pex", line 11, in <module>
    sys.exit(main())
  File "/tmp/virtualenv/lib/python3.4/site-packages/pex/bin/pex.py", line 509, in main
  File "/tmp/virtualenv/lib/python3.4/site-packages/pex/bin/pex.py", line 471, in build_pex
    resolveds = resolver.resolve(resolvables)
  File "/tmp/virtualenv/lib/python3.4/site-packages/pex/resolver.py", line 200, in resolve
  File "/tmp/virtualenv/lib/python3.4/site-packages/pex/resolver.py", line 257, in build
    dist = super(CachingResolver, self).build(package, options)
  File "/tmp/virtualenv/lib/python3.4/site-packages/pex/resolver.py", line 168, in build
pex.resolver.Untranslateable: Package SourcePackage('file:///root/.pex/build/pygments-markdown-lexer-0.1.0.dev39.zip') is not translateable by ChainedTranslator(WheelTranslator, EggTranslator, SourceTranslator)

### DIFF
--- a/pex/installer.py
+++ b/pex/installer.py
@@ -32,12 +32,12 @@ def after_installation(function):
 
 
 class InstallerBase(object):
-  SETUP_BOOTSTRAP_HEADER = "import sys"
+  SETUP_BOOTSTRAP_HEADER = "import io, sys"
   SETUP_BOOTSTRAP_MODULE = "sys.path.insert(0, %(path)r); import %(module)s"
   SETUP_BOOTSTRAP_FOOTER = """
 __file__ = 'setup.py'
 sys.argv[0] = 'setup.py'
-exec(compile(open(__file__).read().replace('\\r\\n', '\\n'), __file__, 'exec'))
+exec(compile(io.open(__file__, encoding='utf-8').read().replace('\\r\\n', '\\n'), __file__, 'exec'))
 """
 
   class Error(Exception): pass


pex/installer.py
```diff
@@ -32,12 +32,12 @@ def function_wrapper(self, *args, **kw):
 
 
 class InstallerBase(object):
-  SETUP_BOOTSTRAP_HEADER = "import sys"
+  SETUP_BOOTSTRAP_HEADER = "import sys; from codecs import open"
```
kwlzn: any reason to not just use `io.open` here instead? it should be in python 2.6 and up and also supports an `encoding=` kwarg - and is the default `open` in python3.

I think you could also avoid the semicolon by using the multiple import syntax at the module level:

```
...
SETUP_BOOTSTRAP_HEADER = "import io, sys"
...
exec(compile(io.open(..., encoding='utf-8')...
```

acaire: I just took this [SO question](http://stackoverflow.com/questions/10971033/backporting-python-3-openencoding-utf-8-to-python-2) as gospel in terms of speed improvements but admittedly didn't do any benchmarks - I'll push again with `io.open`

acaire:
@kwlzn cheers for the feedback - Updated.

kwlzn:
@acaire merged - thanks for the PR!

Issue #243, opened 2016-04-26
Incorrect capitalisation of "ConfigParser"


tjwalton:
Having upgraded pex from 1.1.4 to 1.1.5, my build now fails with:

```
pex: Building pex :: Resolving distributions :: Translating /var/lib/jenkins/home/jobs/deploy_dashboard/workspace/build/platforms/dashboard/tmp/tmptJvrim/python-jenkins-0.4.12.tar.gz into distribution
pex: Building pex :: Resolving distributions :: Translating /var/lib/jenkins/home/jobs/deploy_dashboard/workspace/build/platforms/dashboard/tmp/tmptJvrim/python-jenkins-0.4.12.tar.gz into distribution :: Packaging python-jenkins
pex: Building pex :: Resolving distributions :: Translating /var/lib/jenkins/home/jobs/deploy_dashboard/workspace/build/platforms/dashboard/tmp/tmptJvrim/python-jenkins-0.4.12.tar.gz into distribution :: Packaging python-jenkins :: Installing /var/lib/jenkins/home/jobs/deploy_dashboard/workspace/build/platforms/dashboard/tmp/tmpbRO1kr
**** Failed to install python-jenkins-0.4.12. stdout:

**** Failed to install python-jenkins-0.4.12. stderr:
ERROR:root:Error parsing
Traceback (most recent call last):
  File "/var/lib/jenkins/home/jobs/deploy_dashboard/workspace/build/platforms/dashboard/venv2/lib/python2.7/site-packages/pbr/core.py", line 111, in pbr
    attrs = util.cfg_to_args(path, dist.script_args)
  File "/var/lib/jenkins/home/jobs/deploy_dashboard/workspace/build/platforms/dashboard/venv2/lib/python2.7/site-packages/pbr/util.py", line 264, in cfg_to_args
    wrap_commands(kwargs)
  File "/var/lib/jenkins/home/jobs/deploy_dashboard/workspace/build/platforms/dashboard/venv2/lib/python2.7/site-packages/pbr/util.py", line 566, in wrap_commands
    cmdclass = ep.resolve()
  File "/var/lib/jenkins/home/jobs/deploy_dashboard/workspace/build/platforms/dashboard/venv2/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2386, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
[...]

PR #246, opened 2016-04-26
Fix bdist_pex to work with python2.

jsirois:
Previously `entry_points` handling for ini-file style only worked under
python3.  This adds tests for this narrow use case, more are needed for
other existing functionality.

This addresses #243.

https://rbcommons.com/s/twitter/r/3765/

### DIFF
--- a/pex/commands/bdist_pex.py
+++ b/pex/commands/bdist_pex.py
@@ -1,12 +1,11 @@
 import os
 from distutils import log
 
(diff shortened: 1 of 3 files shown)


jsirois:
Submitted @ https://github.com/pantsbuild/pex/commit/a255b3ad7f0c72e6588be1059f70d14cdc94c7ce

PR #258, opened 2016-05-17
Add README and supported python versions to PyPI description

### LINKED ISSUE #256: Update PyPI page
It would be nice if the `README.rst` were included in the `setup.py` `long_description` in addition to the `CHANGES.rst` so that users browsing PyPI could read the README without having to travel to GitHub.

Would also be nice if the trove classifiers in `setup.py` reflected which versions of Python were officially supported (e.g. `'Programming Language :: Python :: 3.5'`).

johnthagen:
Fix #256.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,7 +35,9 @@ setup(
   classifiers = [
     'Intended Audience :: Developers',
     'License :: OSI Approved :: Apache Software License',
-    'Operating System :: OS Independent',
+    'Operating System :: Unix',
+    'Operating System :: POSIX :: Linux',
+    'Operating System :: MacOS :: MacOS X',
     'Programming Language :: Python',
     'Programming Language :: Python :: 2',
     'Programming Language :: Python :: 2.6',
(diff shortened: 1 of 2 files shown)


johnthagen:
Per #259, updated operating system classifiers so users are aware Windows is not currently supported.

setup.py
```diff
@@ -5,9 +5,11 @@
 
 from setuptools import setup
 
+with open(os.path.join(os.path.dirname(__file__), 'README.rst')) as fp:
+  LONG_DESCRIPTION = fp.read() + '\n'
```
kwlzn: here's a preview of what this concatenation would look like: https://gist.github.com/kwlzn/97f740834edcc03b6e729a3b18d641dc

note the disjoint styling between the concatenation. seems to me that we'd probably want to change the style of the CHANGES.rst header, which is currently:

```
=======
CHANGES
=======
```

to something like:

```
Release Notes
===========
```

to make this line up nicely.

johnthagen:
@kwlzn Good catch, updated PR to appropriately.

kwlzn:
LGTM!

kwlzn:
merged! thanks @johnthagen!

Issue #260, opened 2016-05-23
1.1.17 throws AttributeError with `--disable-cache`


wayne-luminal:
Our build system runs pex and throws the following stack trace. I've confirmed I don't see this with 1.1.16.

```
Traceback (most recent call last):
  File ".tox/package/bin/pex", line 11, in <module>
    sys.exit(main())
  File ".../lib/python3.4/site-packages/pex/bin/pex.py", line 533, in main
    options.cache_dir = make_relative_to_root(options.cache_dir)
  File ".../.tox/package/lib/python3.4/site-packages/pex/bin/pex.py", line 514, in make_relative_to_root
    return os.path.normpath(path.format(pex_root=ENV.PEX_ROOT))
AttributeError: 'list' object has no attribute 'format'
```

With the additional message

```
ERROR: InvocationError: 'pex . -o dist/component.pex -c component --disable-cache -v -i <url>'
```

kwlzn:
@wayne-luminal thanks for the bug report and apologies for the breakage. we'll get this fixed up asap.

@digwanderlust looks like this was related to your change that just went out in 1.1.7. mind taking a look?

kwlzn:
fwiw, it seems the use of `--disable-cache` is the bit that causes this.

digwanderlust:
For some readon disable cache sets the cache dir to [], I'll add a test and fix this.

digwanderlust:
Fix is up at: https://github.com/pantsbuild/pex/pull/261

kwlzn:
this should be fixed in 1.1.8 - now available on pypi: https://pypi.python.org/pypi/pex/1.1.8

verified I can no longer reproduce the breakage using this release. closing.

wayne-luminal:
👍  Thank you!

PR #261, opened 2016-05-23
Fix disable cache (#260)

digwanderlust:
Fix a bug introduced with --pex-root that raised an exception on
when --disable-cache was used.


kwlzn:
lgtm! thanks for the quick turnaround @digwanderlust

PR #271, opened 2016-05-26
Improve failure modes for os.rename() as used in distribution caching.

### LINKED ISSUE #265: Failure to extract dependencies to ~/.pex/install if they already exist with some Python version
Specifically on a host running Python 2.7.5 we ran into the following stack trace:

```
  File ".bootstrap/_pex/pex.py", line 309, in execute
  File ".bootstrap/_pex/pex.py", line 78, in _activate
  File ".bootstrap/_pex/environment.py", line 132, in activate
  File ".bootstrap/_pex/environment.py", line 176, in _activate
  File ".bootstrap/_pex/environment.py", line 121, in update_candidate_distributions
  File ".bootstrap/_pex/environment.py", line 107, in load_internal_cache
  File ".bootstrap/_pex/environment.py", line 95, in write_zipped_internal_cache
  File ".bootstrap/_pex/util.py", line 177, in cache_distribution
OSError: [Errno 17] File exists
```

In this case we're on a slightly older version of pex, line 177 in util.py is: https://github.com/pantsbuild/pex/blob/ba58a854d561dcf6963c0097cad544bfe2861c5a/pex/util.py#L177

So, the rename fails because the directory exists. In this case it's raising EEXIST, whereas pex only expects ENOTEMPTY to be a continuable error code.

Pex should handle EEXIST with the same behavior as ENOTEMPTY in this case.

kwlzn:
Fixes #265.

### DIFF
--- a/pex/common.py
+++ b/pex/common.py
@@ -140,6 +140,21 @@ def safe_rmtree(directory):
     shutil.rmtree(directory, True)
 
 
+def rename_if_empty(src, dest, allowable_errors=(errno.EEXIST, errno.ENOTEMPTY)):
+  """Rename `src` to `dest` using `os.rename()`.
+
+  If an `OSError` with errno in `allowable_errors` is encountered during the rename, the `dest`
+  dir is left unchanged and the `src` directory will simply be removed.
+  """
+  try:
+    os.rename(src, dest)
+  except OSError as e:
+    if e.errno in allowable_errors:
+      safe_rmtree(src)
+    else:
+      raise
+
+
 def chmod_plus_x(path):
   """Equivalent of unix `chmod a+x path`"""
   path_mode = os.stat(path).st_mode
--- a/pex/util.py
+++ b/pex/util.py
@@ -4,7 +4,6 @@
 from __future__ import absolute_import
 
 import contextlib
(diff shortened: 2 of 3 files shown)


pex/common.py
```diff
@@ -140,6 +140,21 @@ def safe_rmtree(directory):
...
+def rename_if_empty(src, dest, allowable_errors=(errno.EEXIST, errno.ENOTEMPTY)):
+  """Rename `src` to `dest` using `os.rename()`.
+
+  If an `OSError` with errno in `allowable_errors` is encountered during the rename, this is
+  a functional no-op and the src directory will be removed.
+  """
+  try:
+    os.rename(src, dest)
+  except OSError as e:
+    if e.errno in allowable_errors:
```
peiyuwang: in the case of `ENOTEMPTY`, it's no-op only if `dest == src`, otherwise we remove `src` but `dest != src`

kwlzn: correct - it's not pretty, but this is the current behavior.

kwlzn: eliminated the `no-op` terminology.

peiyuwang: :shipit:

Issue #274, opened 2016-05-30
bdist_pex.run() doesn't resolve cache-dir's "{pex_root"}


weigon:
bdist_pex.run() doesn't rewrite the {pex_root} token as the "pex" cmdline frontend does which leads
to a directory named "{pex_root}/" in the root-folder of the project.

Adding the same rewrite to pex/commands/bdist_pex.run() fixes it:

    if options.cache_dir:
      options.cache_dir = make_relative_to_root(options.cache_dir)

Introduced by #206

kwlzn:
@weigon thanks for the bug report.

@digwanderlust this is related to your `--pex-root` change - can you take a look?

PR #278, opened 2016-06-08
Update cache dir when bdist_pex.run called directly (#274)

digwanderlust:
* Add unti test to make sure {pex_root} directory is not created
* Update the cache_dir in bdist_pex


kwlzn:
LGTM - thanks @digwanderlust

PR #280, opened 2016-07-06
Don't ignore exit codes when using setuptools entry points

### LINKED ISSUE #137: pex -c incorrectly alters exit codes
Running pytest from a pex always exits 0 if there are test failures, but not if there are command line argument processing errors.

For example, given this `test_fail.py`:

```python
assert False
```

A pytest pex will exit zero:

```console
$ pex --version && pex pytest -c py.test -o pytest.pex && ./pytest.pex test_fail.py ; echo $?
pex 1.0.1
=========================================================================== test session starts ============================================================================
platform darwin -- Python 2.7.10 -- py-1.4.30 -- pytest-2.7.2
rootdir: /Users/rouge8/forks/pex, inifile:
collected 0 items / 1 errors

================================================================================== ERRORS ==================================================================================
______________________________________________________________________ ERROR collecting test_fail.py _______________________________________________________________________
test_fail.py:1: in <module>
    assert False
E   assert False
========================================================================= 1 error in 0.00 seconds ==========================================================================
0
```

But running `py.test` in a virtualenv or similar will exit 1.

On the other hand, giving a bad option to `py.test` will exit non-zero, as expected:

```console
$ pex --version && pex pytest -c py.test -o pytest.pex && ./pytest.pex --bad-option test_fail.py ; echo $?
pex 1.0.1
usage: pytest.pex [options] [file_or_dir] [file_or_dir] [...]
pytest.pex: error: unrecognized arguments: --bad-option
2
```

rouge8:
Setuptools does not require `console_scripts` entry points to call `sys.exit()` (or similar) to exit non-zero:

> The functions you specify are called with no arguments, and their return value is passed to `sys.exit()`, so you can return an errorlevel or message to print to stderr

This patch updates pex to follow the same behavior.

Fixes #137.

### DIFF
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -391,7 +391,7 @@ class PEX(object):  # noqa: T000
 
     entry_point = get_entry_point_from_console_script(script_name, dists)
     if entry_point:
-      return self.execute_entry(entry_point)
+      sys.exit(self.execute_entry(entry_point))
 
     dist, script_path, script_content = get_script_from_distributions(script_name, dists)
     if not dist:
@@ -426,7 +426,7 @@ class PEX(object):  # noqa: T000
   @classmethod
   def execute_entry(cls, entry_point):
     runner = cls.execute_pkg_resources if ':' in entry_point else cls.execute_module
-    runner(entry_point)
+    return runner(entry_point)
 
   @staticmethod
   def execute_module(module_name):
@@ -444,7 +444,7 @@ class PEX(object):  # noqa: T000
     else:
       # setuptools < 11.3
       runner = entry.load(require=False)
-    sys.exit(runner())
+    return runner()
 
   def cmdline(self, args=()):
     """The commandline to run this environment.
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -444,7 +444,7 @@ class PEX(object):  # noqa: T000
     else:
       # setuptools < 11.3
       runner = entry.load(require=False)
-    runner()
+    sys.exit(runner())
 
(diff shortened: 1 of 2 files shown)


pex/pex.py
```diff
@@ -444,7 +444,7 @@ def execute_pkg_resources(spec):
     else:
       # setuptools < 11.3
       runner = entry.load(require=False)
-    runner()
+    sys.exit(runner())
```
kwlzn: this seems like a potentially awkward place to `sys.exit` - esp as potentially called by a library consumer. is there no better path to plumbing the return value/execution context and `sys.exit`'ing at the end of `PEX.execute()`?

at a minimum, moving this to either `PEX.execute_script()` or its calling branch in `PEX._execute()` would make slightly more sense to me given the context.

rouge8: I thought about that, but most packages will already have done a `sys.exit()` or `raise SystemExit` here anyways, [including pex](https://github.com/pantsbuild/pex/blob/v1.1.10/pex/bin/pex.py#L557). I can still move it if you'd like

rouge8: I pushed a commit moving it so this should be ready to merge either way

kwlzn: there is the generic case of non-console script entry point execution to consider, esp as used as a library - moving it to `execute_script()` works for me tho.

kwlzn:
thanks for the PR @rouge8 !

pex/pex.py
rouge8: Oh, definitely did not consider that

PR #284, opened 2016-07-13
Inherit user site with --inherit-path

brbsix:
How do people feel about also inheriting the user site when `--inherit-path` is used?


kwlzn:
seems reasonable to me and the change itself looks good - how would you feel about adding a basic test to exercise and enforce the behavior going forward?

brbsix:
Just extended the (minimal) existing tests covering `--inherit-path`.

kwlzn:
thanks @brbsix - this one is also in.

PR #285, opened 2016-07-13
Fix bdist_pex --pex-args

brbsix:
Splitting on spaces will break on commands like:

    python setup.py bdist_pex \
        --pex-args="--python-shebang='#!/usr/bin/env python'"

This uses `shlex.split()` to do it safely.

### DIFF
--- a/pex/commands/bdist_pex.py
+++ b/pex/commands/bdist_pex.py
@@ -1,4 +1,5 @@
 import os
+import shlex
 from distutils import log
 
 from setuptools import Command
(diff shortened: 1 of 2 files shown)


kwlzn:
lgtm - how would you feel about a quick test or modification to the existing test that repros the pre-fix breakage for the shebang+space case?

brbsix:
Just added tests for bdist_pex `--pex-args` shebangs both with and without spaces. `#!/usr/bin/env python` and `#!/usr/bin/python`, respectively.

kwlzn:
thanks @brbsix - this is in master.

PR #286, opened 2016-07-13
Fix bdist_pex interpreter cache directory

brbsix:
Running bdist_pex results in the creation of a directory literally
named *{pex_root}*. Within it is a subdirectory named *interpreter*.

This fixes it by following behavior similar to that found at https://github.com/pantsbuild/pex/blob/master/pex/bin/pex.py#L536.


kwlzn:
thanks for the PR @brbsix - I'm planning to cut the 1.1.12 release this week which this should go out in.

@digwanderlust - FYI.

PR #310, opened 2016-10-28
Fix #309 by deduplicating output of the distribution finder.

### LINKED ISSUE #309: later versions of setuptools add new wheel finders
yay!  this may obviate the need for pex/finders.py, except now the following breaks:
https://github.com/pantsbuild/pex/blob/85ef56ad/pex/util.py#L93

the reason: now both the monkeypatched finder provided by pex and the new setuptools finder find a distribution in the pex cache, thus len(distributions) == 2 even though the distributions are identical.  for whatever reason, pex throws up its hands and returns None instead, causing PEXEnvironment.update_candidate_distributions to barf.

i have no idea why the code was written the way it was (senility has crept in), but if you replace list() with set() in order to dedupe and return distributions.pop(), things work regardless of setuptools version.

wickman:


jsirois:
LGTM

Issue #306, opened 2016-10-06
Update minimum setuptools requirement to supported version


baroquebobcat:
The minimum, currently 2.2 is too low. 2.2 doesn't support local version segments in requirements. 5.4.1 also doesn't. We should bump it to the first working version that does.

kwlzn:
fixed in #352

PR #311, opened 2016-10-30
URL fix to Tox Documentation in README, setuptools requirements update

eliotwinchell:
Link is currently broken in documentation. 

Also updating Setuptools requirements as per [Issue #306](https://github.com/pantsbuild/pex/issues/306)

According to the Setuptools changelog, [Version 20.2](https://setuptools.readthedocs.io/en/latest/history.html#id138) is when local version segments were added (PEP-440)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -113,7 +113,7 @@ is available at https://pex.readthedocs.io.
 Development
 ===========
 
-pex uses `tox <https://testrun.org/tox/latest/>`_ for test and development automation.  To run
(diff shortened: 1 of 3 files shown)


eliotwinchell:
Realized that there are issues with Pex and the newer setuptools version, I will have to play around with that

kwlzn:
@eliotwinchell out of curiosity, what were the issues you saw with the newer setuptools versions and pex?

Issue #302, opened 2016-09-28
import path leak for `six` with OSX default python


kwlzn:
when running a pex that bundles `six` with the OSX system python (`/usr/bin/python`), the import of `six` references the system path (an ancient version of `six`) rather than the one provided with the pex. this seems to relate to OSX's bespoke usage of the "Extras" path in their installation, which pex does not classify as user-site or site-packages during the sys.path scrubbing.

repro:

```
[illuminati source (master)]$ PATH=/usr/bin ~/bin/pex six -o /tmp/throwaway.pex
[illuminati source (master)]$ PATH=/usr/bin /tmp/throwaway.pex
Python 2.7.10 (default, Jul 30 2016, 18:31:42)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.34)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
(InteractiveConsole)
>>> import sys, six
>>> print(sys.executable, six.__file__)
('/System/Library/Frameworks/Python.framework/Versions/2.7/Resources/Python.app/Contents/MacOS/Python', '/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/six.pyc')
>>> from pprint import pprint
>>> pprint(sys.path)
['/tmp/throwaway.pex/.bootstrap',
 '/tmp/throwaway.pex',
 '/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python27.zip',
 '/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7',
 '/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/plat-darwin',
 '/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/plat-mac',
 '/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/plat-mac/lib-scriptpackages',
 '/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/lib-tk',
 '/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/lib-old',
 '/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/lib-dynload',
 '/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python',
 '/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/PyObjC',
[...]

PR #312, opened 2016-11-02
scrub Extras from sys.path on macOS Sierra

JustinVenus:
### Pre ReviewBoard Travis/CI run

This change removes two paths from sys.path on macOS/OSX.  This prevents outdated/undesirable packages such as `six==1.4` from being loaded before the desired versions specified in PEX-INFO.

### DIFF
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -94,6 +94,18 @@ class PEX(object):  # noqa: T000
     extras_paths = filter(None, makefile.get('EXTRASPATH', '').split(':'))
     for path in extras_paths:
       yield os.path.join(standard_lib, path)
+    # macOS Sierra support
+    framework = sysconfig.get_config_var("PYTHONFRAMEWORK")
+    if not framework:
+      raise StopIteration()
+    third_party = os.path.join("/Library", framework, sys.version[:3], "site-packages")
+    yield third_party
+    extras_pth = os.path.join(third_party, 'Extras.pth')
+    if not os.path.exists(extras_pth):
+      raise StopIteration()
+    with open(extras_pth, 'r') as fd:
+      for line in fd.readlines():
+        yield line.strip()
 
   @classmethod
   def _get_site_packages(cls):


JustinVenus:
This should address https://github.com/pantsbuild/pex/issues/302.

JustinVenus:
RB posted https://rbcommons.com/s/twitter/r/4347/

pex/pex.py
```diff
@@ -155,9 +155,14 @@ def minimum_sys_modules(cls, site_libs, modules=None):
 
   @classmethod
   def minimum_sys_path(cls, site_libs, inherit_path):
+    ver = sys.version[:3]
     scrub_paths = OrderedSet()
     site_distributions = OrderedSet()
     user_site_distributions = OrderedSet()
+    undesirable_paths = OrderedSet([
+      '/System/Library/Frameworks/Python.framework/Versions/%s/Extras/lib/python' % ver,
```
wickman: we should never ever hardcode paths like this into pex.  it sounds like PEX._extras_paths is not behaving the way as intended and should be fixed.  these paths are available programmatically:

```
=; PEX_INTERPRETER=1 PEX_IGNORE_RCFILES=1 PEX_PYTHON=/usr/bin/python2.7 pex
Python 2.7.5 (default, Mar  9 2014, 22:15:05) 
[GCC 4.2.1 Compatible Apple LLVM 5.0 (clang-500.0.68)] on darwin
(InteractiveConsole)
>>> from pex.pex import PEX
>>> for path in PEX._extras_paths():
...   print(path)
... 
/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/../../Extras/lib/python
```

JustinVenus: @wickman thanks for the feedback I'll see if I can track it down.

JustinVenus: @wickman it seems 'EXTRASPATH' isn't defined in Sierra I'll dig around some more.

as seen in https://github.com/pantsbuild/pex/blob/master/pex/pex.py#L94

```sh
jvenus-signal-mbp13:clowncar jvenus$ /usr/bin/python
Python 2.7.10 (default, Jul 30 2016, 18:31:42)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.34)] on darwin
>>> from distutils import sysconfig
>>> makefile = sysconfig.parse_makefile(sysconfig.get_makefile_filename())
>>> for l in makefile.keys():
...   if 'EXT' in l:
...     print l
...
EXTRAPLATDIR
EXTRATESTOPTS
EXTRA_CFLAGS
EXTRAMACHDEPPATH
```

JustinVenus: PEX.site_libs() first path `/Library/Python/2.7/site-package` contains `Extras.pth` and `six.py` which seems to be causing most of the headache.
```sh
jvenus-signal-mbp13:pex jvenus$ env PEX_INTERPRETER=1 PEX_IGNORE_RCFILES=1 PEX_PYTHON=/usr/bin/python2.7 pex --python=/usr/bin/python2.7 pex
Python 2.7.10 (default, Jul 30 2016, 18:31:42)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.34)] on darwin
(InteractiveConsole)
>>> from pex.pex import PEX
>>> PEX.site_libs()
set(['/Library/Python/2.7/site-packages', '/System/Library/Frameworks/Python.framework/Versions/2.7/lib/site-python', '/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages'])
```

```sh
jvenus-signal-mbp13:pex jvenus$ ls /Library/Python/2.7/site-packages/
Extras.pth                       README                           six-1.9.0-py2.7.egg-info         six.pyc                          vboxapi-1.0-py2.7.egg-info
RBTools-0.7.2.dev-py2.7.egg-info rbtools                          six.py                           vboxapi
jvenus-signal-mbp13:pex jvenus$ cat /Library/Python/2.7/site-packages/Extras.pth
/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python
/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/PyObjC
```

kwlzn: yeh, the Extras.pth in site-packages seems to be the primary linkage here.

I alluded to this in #312, but I think a semi-sane workaround approach here to identifying these paths (which could merge directly into `_extras_paths`) could be:

```
>>> def yield_extras_paths():
...   extras_prefix = os.path.join(sys.prefix, 'Extras')
...   for path in sys.path:
...     if path.startswith(extras_prefix):
...       yield path
... 
>>> extras_paths = list(yield_extras_paths())
>>> pprint(extras_paths)
['/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python',
 '/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/PyObjC']
```

kwlzn:
+1 to wickmans comment. also, it'd be great to add some quick test coverage for this case before landing.

JustinVenus:
This should now work.
```sh
jvenus-signal-mbp13:pex jvenus$ env PEX_INTERPRETER=1 PEX_IGNORE_RCFILES=1 PEX_PYTHON=/usr/bin/python2.7 pex --python=/usr/bin/python2.7 .
Python 2.7.10 (default, Jul 30 2016, 18:31:42)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.34)] on darwin
(InteractiveConsole)
>>> import sys
>>> sys.path.insert(0, '')
>>> from pex.pex import PEX
>>> for path in PEX._extras_paths():
...   print(path)
...
/Library/Python/2.7/site-packages
/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python
/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/PyObjC
```

JustinVenus:
@kwlzn how'd you like this tested?

pex/pex.py
```diff
@@ -94,6 +94,18 @@ def _extras_paths(cls):
     extras_paths = filter(None, makefile.get('EXTRASPATH', '').split(':'))
     for path in extras_paths:
       yield os.path.join(standard_lib, path)
+    # macOS Sierra support
+    framework = sysconfig.get_config_var("PYTHONFRAMEWORK")
+    if not framework:
+      raise StopIteration()
```
kwlzn: better to simply `return` in these cases than to manually raise `StopIteration`:

```
if not framework:
  return
```

pex/pex.py
```diff
@@ -94,6 +94,18 @@ def _extras_paths(cls):
...
+    third_party = os.path.join("/Library", framework, sys.version[:3], "site-packages")
+    yield third_party
+    extras_pth = os.path.join(third_party, 'Extras.pth')
+    if not os.path.exists(extras_pth):
+      raise StopIteration()
+    with open(extras_pth, 'r') as fd:
+      for line in fd.readlines():
+        yield line.strip()
```
kwlzn: I see a few problems with this approach for backward and forward compatibility:

1) .pth files can contain comments, which aren't handled here
2) .pth files can contain relative paths, which wouldn't make sense to directly yield
3) .pth files can contain paths that don't exist, which isn't checked here
4) technically, .pth files that begin with `import` can contain executable code, which isn't handled here (and probably should not be - but at least should detect and avoid emitting raw code)

so, if we must rely on our own .pth file parsing it'd probably be good to extract a separate, general purpose .pth file parser function into `pex.util` and test that separately.

you can probably lift much of the required logic directly from the `site` module's .pth handling implementation here: https://github.com/kwlzn/python-sources/blob/master/Python-2.7.12/Lib/site.py#L131

pex/pex.py
```diff
@@ -94,6 +94,18 @@ def _extras_paths(cls):
     extras_paths = filter(None, makefile.get('EXTRASPATH', '').split(':'))
     for path in extras_paths:
       yield os.path.join(standard_lib, path)
+    # macOS Sierra support
+    framework = sysconfig.get_config_var("PYTHONFRAMEWORK")
+    if not framework:
+      raise StopIteration()
+    third_party = os.path.join("/Library", framework, sys.version[:3], "site-packages")
```
kwlzn: this seems like a poor way of determining the `site-packages` dir - it hardcodes `/Library` and extracts version information from the `sys.version` string (rather than say, `sys.version_info`).

instead, prefer iterating over entries provided by the `site` module's `getsitepackage()`:

```
>>> import site; site.getsitepackages()
['/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages', '/System/Library/Frameworks/Python.framework/Versions/2.7/lib/site-python', '/Library/Python/2.7/site-packages']
```

which already has an existing helper class method in `PEX._get_site_packages()`.

this probably also obviates the need to fetch the 'PYTHONFRAMEWORK' config var?

pex/pex.py
```diff
@@ -94,6 +94,18 @@ def _extras_paths(cls):
...
+    yield third_party
```
kwlzn: doesn't seem like this should be yielded here since it's not actually an extras path?

pex/pex.py
```diff
@@ -94,6 +94,18 @@ def _extras_paths(cls):
...
+    extras_pth = os.path.join(third_party, 'Extras.pth')
```
kwlzn: I wonder if it would make sense to classify all .pth-based path additions as "extras" in this context? i.e. to actually look for any files named `*.pth` and parse/emit their fully qualified paths akin to the .pth file handling here: https://github.com/kwlzn/python-sources/blob/master/Python-2.7.12/Lib/site.py#L183-L188 ?

this would more generally handle the case of .pth files injecting new paths that wouldn't be otherwise noticed by pex, with the `Extras.pth`case included.

kwlzn:
@JustinVenus would take a look at the existing test cases in `tests/test_pex.py` and copy/paste/port one of those over to get some minimal coverage for the `Extras` path handling via mocks etc.

pex/pex.py
JustinVenus: agreed

pex/pex.py
JustinVenus: This seems like a good approach.  I'll take a look at doing this

pex/pex.py
JustinVenus: It's pretty explicit in the readme about being a thirdparty/extra location.
```sh
jvenus-signal-mbp13:pex jvenus$ cat /Library/Python/2.7/site-packages/README
This directory exists so that 3rd party packages can be installed
here.  Read the source for site.py for more details.
```

pex/pex.py
JustinVenus: No particular disagreement, however this is the way `framework` detection is done in [site.py](https://github.com/kwlzn/python-sources/blob/master/Python-2.7.12/Lib/site.py#L302).

pex/pex.py
JustinVenus: sure.

JustinVenus:
@wickman @kwlzn how's this for a solution?

1) search for` *.pth` files whenever a tainted path is detected in 'pex/pex.py'
2) use `path_from_pth` in `pex/util.py` to provide addtional paths to scrub

#### roughly how I'd implement `path_fom_pth` *it clearly doesn't meet guidelines for inclusion yet*
```python
import os, sys, traceback
from site import makepath

#based on site.addpackage
def path_from_pth(sitedir, name, known_paths):
  fullname = os.path.join(sitedir, name)
  namespace = {}
  local_vars = {
      'sitedir': sitedir,
      'name': name,
  }
  exec "import site" in namespace
  try:
    f = open(fullname, "rU")
  except IOError:
    return

  with f:
    for n, line in enumerate(f):
      if line.startswith("#"):
        continue
      try:
        if line.startswith(("import ", "import\t")):
          exec line in namespace, local_vars
          for mod in local_vars['sys'].modules.values():
            if hasattr(mod, '__path__'):
              dir, dircase = makepath(sitedir, *mod.__path__)
              if not dircase in known_paths and os.path.exists(dir):
                known_paths.add(dircase)
          continue
        line = line.rstrip()
        dir, dircase = makepath(sitedir, line)
        if not dircase in known_paths and os.path.exists(dir):
          known_paths.add(dircase)
      except Exception as err:
        print >>sys.stderr, "Error processing line {:d} of {}:\n".format(n + 1, fullname)
        for record in traceback.format_exception(*sys.exc_info()):
          for line in record.splitlines():
            print >>sys.stderr, '  '+line
        print >>sys.stderr, "\nRemainder of file ignored"
        break
  return known_paths

# example usage
if __name__ == '__main__':
  path = '/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python'
  print addpackage(path, 'matplotlib-1.3.1-py2.7-nspkg.pth', set())

  path = '/Library/Python/2.7/site-packages'
  print addpackage(path, 'Extras.pth', set())
```

```sh
jvenus-signal-mbp13:~ jvenus$ /usr/bin/python test.py
set(['/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/mpl_toolkits', '/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/encodings'])
set(['/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python', '/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/PyObjC'])
```

kwlzn:
@JustinVenus seems reasonable to me, tho the utility function would obviously want to adopt the repo style wherever it makes sense (e.g. s/`print >>sys.stderr`/`sys.stderr.write()` etc).

JustinVenus:
@kwlzn I'll circle back on this sometime early next week.  Thanks for the feedback.

mikesown:
@JustinVenus Is this still on the near-term horizon? It's causing a nontrivial amount of pain for us, so please let me know if I can do anything to help push it along!

JustinVenus:
@mikesown yeah I'd like to get to this soonish ... I've just been tied up w/ other projects and holiday stuff.

benjyw:
Pinging this - is any progress pending?

kwlzn:
picking this one up over in #370 based on the approach outlined in the discussion here.

PR #316, opened 2016-11-17
Add manylinux wheel support and fix a few bugs along the way

### LINKED ISSUE #281: Support manylinux1 wheels
Pex doesn't seem to support manylinux1 wheels:

``` console
$ pex --version && pex --no-build -vvv numpy
pex 1.1.11
pex: crawling link i=0 link=Link('file:///home/andy/.pex/build') follow_links=False
pex: crawling link i=0 link=Link('https://pypi.python.org/simple/numpy/') follow_links=False
Could not satisfy all requirements for numpy:
    numpy
$ pip install --only-binary :all: numpy
Collecting numpy
  Downloading numpy-1.11.1-cp27-cp27mu-manylinux1_x86_64.whl (15.3MB)
    100% |████████████████████████████████| 15.3MB 102kB/s
Installing collected packages: numpy
Successfully installed numpy-1.11.1
```

This was mentioned briefly in #250, but that was closed.

dan-blanchard:
- Adds support for `manylinux1_x86_64` and `manylinux1_i686` platforms. I've treated them as platforms, because that's how [PEP-513](https://www.python.org/dev/peps/pep-0513/) describes them. This fixes #281.
- Fixes a number of bugs related to the handling of wheels that have hyphens in the package names (e.g., `cassandra-driver` and `msgpack-python`).
- Fixes a couple places where `str.replace` was being called without storing the result (thereby accomplishing nothing since `str` is immutable)


dan-blanchard:
Will update test data so they pass.

dan-blanchard:
Tests now pass

kwlzn:
@dan-blanchard awesome. let me know if/when this is ready for review and I'd be happy to take a look.

dan-blanchard:
This is all ready for review

rouge8:
One thing that this doesn't handle that I was worried about is detecting whether a platform actually supports manylinux: https://github.com/pantsbuild/pex/issues/281#issuecomment-231124725

I didn't have a chance to think about where to add runtime compatibility checking in PEX though.

.gitignore
```diff
@@ -9,3 +9,5 @@
 /.idea
 /.coverage*
 /htmlcov
+.cache/*
```
dan-blanchard: I was running the tests locally and getting annoyed that this stuff kept getting making my repo dirty. Not directly related, but should have been there before.

dan-blanchard:
> One thing that this doesn't handle that I was worried about is detecting whether a platform actually supports manylinux: #281 (comment)

Would there be any objections to adding `pip` as an explicit dependency for `pex`? It seems silly to reimplement a lot of the work they've done on checking for manylinux compatibility when we could just call `pip.pep425tags.is_manylinux1_compatible()`.

dan-blanchard:
One thing I've noticed with this PR that the wheels it picks aren't don't always match the ABI tags they should. At first I thought this was just when using the `--platform` tag on OS X, but even without `--platform` I've found that on Linux I get some `cp27m` wheels and some `cp27mu` wheels in the same archive. I think maybe we'll need to add the ability to specify an ABI tag when using `--platform` and we should use `pip.pep425tags.get_abi_tag()` to determine the correct ABI tag without it.

dan-blanchard:
Also, more about the depending on `pip` thing... pypa/pip@bbe99ce62d626f6c01552c3fcab953b014674c53 adds support for downloading packages for different interpretters and ABIs to pip. It feels like all the fetching/downloading stuff `pex` is doing could just piggyback on `pip` and `pex` could just concentrate on building the zip with `__main__.py`.

kwlzn:
@dan-blanchard 

> Would there be any objections to adding pip as an explicit dependency for pex?

I think it'd be great if we could avoid that for now just due to precedent/historical reasons, but wouldn't be opposed in the longer term if a proposal materialized that thoughtfully motivated the case.

> One thing I've noticed with this PR is that the wheels it picks don't always match the ABI tags they should.

yeah, this is a semi-known wart with current pex that is particularly relevant for shops without isolated indexes. pex differs from (pre-8.0) pip et al because fundamentally you can build a pex on one host and run it on another which may or may not share the same abi compatibility. I think to solve this, we probably would want to plumb an explicit argument or other config for constraining the resolve to only allowable flags with some sensible defaults. the runtime resolve checking should probably also detect compatible abis and error early if they fail to resolve - which could additionally pave the way for multi-abi pexes (e.g. one pex with both `m` and `mu` wheels that could exec on ucs2 *and* ucs4 systems and use the right deps accordingly). feel free to file an issue for this one if you have time.

> It feels like all the fetching/downloading stuff pex is doing could just piggyback on pip ...

yes, totally. we've effectively been waiting (roughly >a year now) on the change/set of changes you mentioned (patricklaw is also a pants/pex contributor) to land in `pip` to create a path for us to move to a pluggable resolver model that would let us slot pip directly into pex for resolution. this would allow us to deprecate the bespoke pex resolver in favor of pip's while also providing a backwards-compatible upgrade path. it's something we want to do, but is relatively invasive and at odds with other priorities - so if you had any interest in picking that work up, it'd be most welcome.

kwlzn:
@dan-blanchard what were you thinking here in terms of best way to proceed? perhaps a quick `--manylinux` flag in this PR defaulting to off to explicitly turn on this behavior - and then circling back with improvements around the abi/capability detection before defaulting that to on?

dan-blanchard:
> what were you thinking here in terms of best way to proceed? perhaps a quick --manylinux flag in this PR defaulting to off to explicitly turn on this behavior - and then circling back with improvements around the abi/capability detection before defaulting that to on?

Yeah, that seems reasonable to me. I'll try to set that up soon.

dan-blanchard:
I just pushed a commit that I haven't updated the tests for yet, so it will likely fail, but it adds the CLI options necessary for finding the correct wheels for different platforms. I was unable to do this change in a backward-compatible way, so this will need a major version bump if people are using the internal PEX API. The biggest change is that I copied out pip's `pep425tags` module and started using that all over the place. This will make sure that we won't default to using manylinux wheels on Linux interpreters that aren't manylinux compatible.

There's some cleanup that's still left to do, but I wanted to show y'all what I had so far.

dan-blanchard:
I've built actual manylinux PEXs with this thing branch and they work fine. I cannot for the life of me figure out why the tests are failing right now. Would love another set of eyes.

tests/test_package.py
```diff
@@ -82,7 +82,7 @@ def test_wheel_package():
 
 
 @pytest.mark.parametrize("package_name", [
-  "transmute_core-0.4.5-py2.py3-none-any-gobbledy.whl"
+  "transmute_core-0.4.5-py2.py3-none.whl"
```
dan-blanchard: Had to change this test because we're now using `rsplit`, so this invalid name would now sneak by as valid. I think this types of screw ups are incredibly unlikely to occur as the wheel names aren't manually created.

dan-blanchard:
Tests are now passing for everything but pypy. I'll dig into that, but this could use another look now that it's gotten a bit bigger (because I fixed the issue where PEX would download wheels for the wrong ABI tags).

dan-blanchard:
The PyPy tests were failing due to a strange special case in pip's PEP425 handling for PyPy that we're now using. I've fixed that and now ALL TESTS PASS.

@kwlzn, I know you already approved this, but it's grown a little bit in scope since you did that, so you might want to take another look.

Also, I know giant PRs are a total pain to review, so if this is too much to get through, I could split off the last two commits (which removed the old `pep425` and `platforms` modules and replace them with pip's `glibc` and `pep425tags` modules) into their own PR. That would leave this PR basically like it was when @kwlzn approved it, but no one would actually be able to use the manylinux wheel support until the second PR was merged, because without the ABI compatibility fix, you end up with broken PEX files more often than not.  

I did not add a `--manylinux` flag in the end because by switching to pip's `pep425tags` implementation we now check if a platform is manylinux compatible before trying to use those wheels.

ntravis:
Pinging for feedback as I believe the PR would address workarounds we have in place in our environment. Cheers!

kwlzn:
thanks for updating the PR - I'm a bit swamped atm, but I'll try to carve out time to review sometime this week. apologies for the delay!

dan-blanchard:
@kwlzn Thanks, no need to apologize. I maintain more projects than I should, so I know what it's like.

kwlzn:
I was unfortunately not able to tackle this as I had hoped this week, but it's now at the top of my TODO list and will be diving in first thing on Monday as part of a pex 1.2.7 release (#379).

dan-blanchard:
Just saw there were conflicts after the pre-release related changes, so I rebased.

pex/pep425tags.py
```diff
@@ -0,0 +1,325 @@
...
+    supported.append(('py%s' % (version,), 'none', 'any'))
+    if i == 0:
+      supported.append(('py%s' % (version[0]), 'none', 'any'))
+
+  return supported
+
+supported_tags = get_supported()
+supported_tags_noarch = get_supported(noarch=True)
+
+implementation_tag = get_impl_tag()
```
kwlzn: might be nice to kill these import-time execs/global state.

pex/resolver.py
```diff
@@ -142,31 +142,48 @@ def map_packages(resolved_packages):
...
-  def __init__(self, allow_prereleases=None, interpreter=None, platform=None):
+  def __init__(self, allow_prereleases=None, interpreter=None, versions=None,
+               platform=None, impl=None, abi=None):
     self._interpreter = interpreter or PythonInterpreter.get()
-    self._platform = platform or Platform.current()
+    self._platform = platform
     self._allow_prereleases = allow_prereleases
+    self._impl = impl
+    self._abi = abi
+    self._versions = versions
```
dan-blanchard: It's just because of `get_supported`. I believe there was a point where I had myself convinced we would need to pass multiple versions to get prior minor versions in the list, but I don't think we actually need it. I'll get rid of the plural everywhere but `get_supported` (since that's copied pip code).

pex/bin/pex.py
```diff
@@ -290,8 +290,43 @@ def configure_clp_pex_environment(parser):
...
+
+  group.add_option(
+      '--implementation',
+      dest='implementation',
+      metavar='implementation',
+      default=None,
+      help=("Python implementation for which to build the PEX (e.g. 'pp', 'jy',"
+            " 'cp',  or 'ip'). If not specified, then the current interpreter "
+            "implementation is used.  Use 'py' to force implementation-"
+            "agnostic wheels."),
```
dan-blanchard: I'm less than thrilled with the idea of muddying up the already murky terminology around package compatibility tags, but I can see why passing a single arg could be more user friendly in certain cases. How about instead of overloading `--platform` we make `--tags` a new arg that supports passing in all the PEP 425 compatibility tags joined by hyphens?

In the future when you want to support specifying multiple supported tag sets at a time, you could just only allow multiple `--tags` args to be specified, or you could do combinations of passed in `--platform` and `--abi` args like you mentioned. Either seems reasonable to me.

dan-blanchard: I went ahead and added `--tags`, and you can let me know if that solution seems good enough or not.

dan-blanchard:
> I think it'd also be great if possible to at least pin the manylinux stuff down before this lands with a quick integration test. if you can find smaller manylinux wheels (<=300KB) for testing, feel free to commit them directly to the repo vs depending on pypi.

I'll try to dig some of those up tonight or tomorrow.

pex/bin/pex.py
kwlzn: > or you could do combinations of passed in --platform and --abi args like you mentioned. Either seems reasonable to me.

I think the problem with that is that there are no natural combinations here at the ABI level in one of the most common native cases we have/see: a OSX + Linux multi-platform resolve (aka how we build almost every pex at Twitter). OSX interpreters are traditionally always `m` while Linux is traditionally always `mu`. so in that common case, the combinatorial approach would then end up with two completely extraneous platforms (osx->mu, linux->m) which would need to be 1) pre-built/available and 2) resolved - for no apparent benefit. I also think relying on users to pass in position-dependent sets of all 4 facets is likely to be error prone.

so fundamentally to me, it seems like any form of split args for all 4 facets here - which are now necessary to pass all or none - will lead to an awkward, complicated interface for the CLI. having one clear way to pass in the platform + tags triple in all cases would be IMHO cleanest/most ideal - vs having one set of args you pass in for single-platform resolves (e.g. `--platform X --python-version X --implementation X --abi X`) and then an entirely different passing style when you want to pass in a second platform (e.g. `--tags X --tags X`). so I think that means either outright replacing or extending `--platform`.

fwiw, my implicit thinking with the idea of extending `--platform` is largely around backward compatibility. I had alluded briefly to the idea of implicit `any` wildcards for the latter 3 fields, which I think could be a good mechanism for that.

the idea there essentially is that:

```
./pants --platform=linux-x86_64 ...
```

could implicitly map to e.g.:

```
./pants --platform=linux-x86_64-any-any-any ...
```

so that existing users could continue to pass in `--platform` as they do today and get the ~current behavior. then, when the extended form is needed that could be passed in as well just as easily. I was thinking about tackling the backwards compatibility aspect as a fast-follow on top of your work here, prior to cutting a release.

it's also worth noting that this idea of overloading platform and remaining backward compatible would also readily extend to pants as well, where platforms are specified as an iterable of platform strings that could also work with either form to ease the need for migration on users (..tho pants is not exactly affected by the CLI API discussed here, the conceptual mapping and underlying functionality in pex to support is relevant). additionally, `tags` also has an existing conceptual meaning in pants that applies to all targets.

but I definitely hear you re: murky terminology - and this is certainly overloading the term `platform` (and `tags` too, unfortunately).. so I'm open to alternative tacks.

either way tho, it'd probably be a good idea to get some other perspectives on this as well. maybe @jsirois might be able to review and share his take?

sveitser:
If it helps, I looked through pypi for manylinux wheels smaller than 300kB and found these

```
# size pkg url 
296099 RelStorage-2.1a1-cp27-cp27m-manylinux1_x86_64.whl https://pypi.python.org/packages/15/51/6b158153470dd900d23f881da46206eec13fb2e38ab2e4d139bf5bb6b778/RelStorage-2.1a1-cp27-cp27m-manylinux1_x86_64.whl
289196 dumb_init-1.2.0-py2.py3-none-manylinux1_x86_64.whl https://pypi.python.org/packages/43/67/5697dd284b57c1839a1e4c5b9e8a50e2ccb6cffb00c53dbf03266204c49b/dumb_init-1.2.0-py2.py3-none-manylinux1_x86_64.whl
241958 msgpack_python-0.4.8-cp27-cp27m-manylinux1_i686.whl https://pypi.python.org/packages/ba/f1/3e53ae3cede8715e7e06503660624d22372f84e7bc37f6ff96ed77b15863/msgpack_python-0.4.8-cp27-cp27m-manylinux1_i686.whl
236545 pyrilla-0.0.2-cp27-cp27m-manylinux1_i686.whl https://pypi.python.org/packages/39/b5/2b97a67b4755de81889b7d7e80bfe00da71f59f0df0b593bea4badf7c3f3/pyrilla-0.0.2-cp27-cp27m-manylinux1_i686.whl
234604 python_bsonjs-0.1.1-cp26-cp26m-manylinux1_i686.whl https://pypi.python.org/packages/8d/bd/ce91ff40e990700cade754e887760a11b5aeda9957035f47811c692df639/python_bsonjs-0.1.1-cp26-cp26m-manylinux1_i686.whl
226251 japronto-0.1.1-cp35-cp35m-manylinux1_x86_64.whl https://pypi.python.org/packages/f2/7b/b711fbbeeeba2d747417d8903ad190a7ee977436da37432e44b204fce110/japronto-0.1.1-cp35-cp35m-manylinux1_x86_64.whl
218243 meinheld-0.6.1-cp36-cp36m-manylinux1_x86_64.whl https://pypi.python.org/packages/6f/66/442eba40331114798a4620373032bbb8f17be60609e7b12279eac644911c/meinheld-0.6.1-cp36-cp36m-manylinux1_x86_64.whl
196976 libMHCUDA-1.1.1-cp35-cp35m-manylinux1_x86_64.whl https://pypi.python.org/packages/87/25/b1814c8204258866fdd9308e5628f486c58f30cb4328db5c52a9f137b56f/libMHCUDA-1.1.1-cp35-cp35m-manylinux1_x86_64.whl
194387 opbeat-3.5.2-cp27-cp27m-manylinux1_i686.whl https://pypi.python.org/packages/6b/25/fe40368bb3ac0e8dab31e46974c29666893131627db758f6e31972f1c549/opbeat-3.5.2-cp27-cp27m-manylinux1_i686.whl
191413 coverage-4.4b1-cp26-cp26m-manylinux1_i686.whl https://pypi.python.org/packages/71/15/5b03021f4da2d220d2e799b78e3bd477ab325baebe9a6a3eff99d6357a62/coverage-4.4b1-cp26-cp26m-manylinux1_i686.whl
188967 cchardet-2.0.0-cp27-cp27m-manylinux1_i686.whl https://pypi.python.org/packages/0f/be/929162ca9fee3399eda89e59629136c16e1a88a058202a1b487dff8c7d05/cchardet-2.0.0-cp27-cp27m-manylinux1_i686.whl
161509 yarl-0.10.0-cp34-cp34m-manylinux1_i686.whl https://pypi.python.org/packages/eb/1d/2615e0dac685d0b0b9e3aa995b7293c75e4d4e4d6ab9e965635ee90fdb25/yarl-0.10.0-cp34-cp34m-manylinux1_i686.whl
158148 pytrip98-1.5.1-cp27-cp27m-manylinux1_i686.whl https://pypi.python.org/packages/d2/07/503adf102b79e6bacbd03d45845f28a5345e98739b9d2ab19ffbffa56eea/pytrip98-1.5.1-cp27-cp27m-manylinux1_i686.whl
155489 bencoder.pyx-1.2.0-cp26-cp26m-manylinux1_i686.whl https://pypi.python.org/packages/24/a4/73ab3992a7ef5700fae247c2d81253ac5f68d4661af912ce4b226332e312/bencoder.pyx-1.2.0-cp26-cp26m-manylinux1_i686.whl
145494 secp256k1-0.13.2-cp27-cp27m-manylinux1_x86_64.whl https://pypi.python.org/packages/c8/22/95038ee6e1362f8177a1ad7af0912ae2de290710a8b292c04edc8e125420/secp256k1-0.13.2-cp27-cp27m-manylinux1_x86_64.whl
139323 PyLBFGS-0.2.0.3-cp27-cp27m-manylinux1_i686.whl https://pypi.python.org/packages/e9/e0/ea444918b65bb8a1d80e73add89c840e4f8e689dd6b5dd201fac4f819a94/PyLBFGS-0.2.0.3-cp27-cp27m-manylinux1_i686.whl
135315 twisted_platform_support-16.12.3-cp27-cp27m-manylinux1_i686.whl https://pypi.python.org/packages/f8/8d/1fd1a6ff65295a7fea71d030429a9e3eadcbae25fd12995302b8c4bf8370/twisted_platform_support-16.12.3-cp27-cp27m-manylinux1_i686.whl
130658 svmloader-0.4-cp27-cp27m-manylinux1_i686.whl https://pypi.python.org/packages/e5/cd/172bf806bcc0e2759716984a5e22101b3d67f73c78555fe42341c9b1a2f4/svmloader-0.4-cp27-cp27m-manylinux1_i686.whl
123815 pendulum-1.2.0-cp27-cp27m-manylinux1_i686.whl https://pypi.python.org/packages/4a/cc/e897d0a073e2b30d3fa16f65d972e4f9efeac49783b07f7262cb081e8f5d/pendulum-1.2.0-cp27-cp27m-manylinux1_i686.whl
103538 greenify-0.3.2-cp27-cp27m-manylinux1_x86_64.whl https://pypi.python.org/packages/94/96/60b1904ed211e1b37b1b58bf4e22cc6e709609c4374aec15095e9bf1836c/greenify-0.3.2-cp27-cp27m-manylinux1_x86_64.whl
98205 zopfli-0.1.1-cp27-cp27m-manylinux1_i686.whl https://pypi.python.org/packages/b5/66/33b968013d54a84846b123e9cbdec627de80e7a800401ec52d9ef978b233/zopfli-0.1.1-cp27-cp27m-manylinux1_i686.whl
96341 numpy_sugar-1.0.35-cp27-cp27m-manylinux1_x86_64.whl https://pypi.python.org/packages/80/cc/fccc12a2311f4c7f7fbdb4c8208c4a67cc97ce0b3c9b8eacbc26460f6b91/numpy_sugar-1.0.35-cp27-cp27m-manylinux1_x86_64.whl
91474 ruamel.ordereddict-0.4.9-cp26-cp26m-manylinux1_x86_64.whl https://pypi.python.org/packages/7c/2d/552e914797da6f5a475452008f8450d4e3142ac2bcd59200356eed74e082/ruamel.ordereddict-0.4.9-cp26-cp26m-manylinux1_x86_64.whl
89084 ninja-1.7.2-cp27-cp27m-manylinux1_i686.whl https://pypi.python.org/packages/3f/ca/8573f4b0a561c340776562e2465f55b2b1e352b61e2b929c627601e3d869/ninja-1.7.2-cp27-cp27m-manylinux1_i686.whl
87933 argon2_cffi-16.3.0-cp27-cp27m-manylinux1_i686.whl https://pypi.python.org/packages/28/fe/a35f298c9a8887223ee29e30641205f23a97af91cf99a1fc9172bf4eede8/argon2_cffi-16.3.0-cp27-cp27m-manylinux1_i686.whl
79601 FXrays-1.3.3-cp27-cp27mu-manylinux1_x86_64.whl https://pypi.python.org/packages/11/bb/20649dfa9c82669ac7890b0ce04cb219760e48796f079a43b4f046875831/FXrays-1.3.3-cp27-cp27mu-manylinux1_x86_64.whl
74667 tonnikala-1.0.0b3-cp27-cp27m-manylinux1_x86_64.whl https://pypi.python.org/packages/65/fd/7ee2c5de0b057facefb493ced73a0794036fcee702022d6de7807801a2d0/tonnikala-1.0.0b3-cp27-cp27m-manylinux1_x86_64.whl
74061 cyrandom-0.1.3-cp35-cp35m-manylinux1_i686.whl https://pypi.python.org/packages/ae/61/610ab5d04df76002aad8e457338748b7564a602e4773b7adc9df529d9eb4/cyrandom-0.1.3-cp35-cp35m-manylinux1_i686.whl
72913 DoubleMetaphone-0.1-cp27-cp27m-manylinux1_i686.whl https://pypi.python.org/packages/20/dc/a9fec6d53b0629f017e6ff9ee73cb50bc33b7aa7a4784187d7cedfde9b83/DoubleMetaphone-0.1-cp27-cp27m-manylinux1_i686.whl
71877 porridge-0.1.4-cp27-none-manylinux1_x86_64.whl https://pypi.python.org/packages/68/1d/f5a4b9265c8696fa98c9a91ef2c921221cae8f4a614445fe158a2b2d60f2/porridge-0.1.4-cp27-none-manylinux1_x86_64.whl
68933 dedupe-1.6.7-cp27-cp27m-manylinux1_i686.whl https://pypi.python.org/packages/12/ab/54e107aecfb7b868c8925c4937f5e959e1aa627d0a1302cc3f8badba0057/dedupe-1.6.7-cp27-cp27m-manylinux1_i686.whl
67148 wsgi_lineprof-0.3.0-cp27-cp27m-manylinux1_x86_64.whl https://pypi.python.org/packages/d8/2a/3dbe75e2de52e19ea7a2cc8bc7596fba4ce4ed8b40fb92c7ba039d06145b/wsgi_lineprof-0.3.0-cp27-cp27m-manylinux1_x86_64.whl
66031 cymem-1.31.2-cp27-cp27m-manylinux1_x86_64.whl https://pypi.python.org/packages/e2/ef/e1ebaf55e478e1d34c6ac9b0d2fa8678e180542d3bc9398af9e95b546df1/cymem-1.31.2-cp27-cp27m-manylinux1_x86_64.whl
63569 sip-4.19.2-cp35-cp35m-manylinux1_x86_64.whl https://pypi.python.org/packages/19/5a/c50a0ccd211f864972da0f2e101722508ffb4e008347fd2a56a99beab9d0/sip-4.19.2-cp35-cp35m-manylinux1_x86_64.whl
58237 bcrypt-3.1.3-cp26-cp26m-manylinux1_i686.whl https://pypi.python.org/packages/58/24/a07d55c72180ff2d896dd0e94638e40c4d3a5fa6b4e367d6cd6c3e9919ba/bcrypt-3.1.3-cp26-cp26m-manylinux1_i686.whl
56730 lazy_object_proxy-1.2.2-cp26-cp26m-manylinux1_x86_64.whl https://pypi.python.org/packages/23/f9/f90b2953f858920959c25ab29cd856eb121f3267ce91cc27a4a4ecdf3712/lazy_object_proxy-1.2.2-cp26-cp26m-manylinux1_x86_64.whl
47998 pyquicklz-1.4.1-cp27-cp27m-manylinux1_i686.whl https://pypi.python.org/packages/4f/d5/387ffa9031eba8ad7cb52cb64b848bf251ed855b343bc0b23bf9c960e298/pyquicklz-1.4.1-cp27-cp27m-manylinux1_i686.whl
47650 greenstack-0.6-cp26-cp26m-manylinux1_x86_64.whl https://pypi.python.org/packages/9f/ae/9bc8f3e0aa9eed8b83a29af76952a54033826019a250ce5aef4af212d407/greenstack-0.6-cp26-cp26m-manylinux1_x86_64.whl
42905 PyQtPurchasing-5.8-5.8.0-cp34.cp35.cp36.cp37-abi3-manylinux1_x86_64.whl https://pypi.python.org/packages/55/b7/3b603ce8685522ae11f894052e223ce467a6c8678b94a6e2cc9db194b637/PyQtPurchasing-5.8-5.8.0-cp34.cp35.cp36.cp37-abi3-manylinux1_x86_64.whl
41175 greenlet-0.4.12-cp26-cp26m-manylinux1_x86_64.whl https://pypi.python.org/packages/da/52/74bbc91f8bbdb0c3b475f294bb2b51deafad78efcd69eef2c853bc007c90/greenlet-0.4.12-cp26-cp26m-manylinux1_x86_64.whl
37825 affinegap-1.10-cp27-cp27m-manylinux1_i686.whl https://pypi.python.org/packages/23/e5/4bb4d9e9341dfb62709fb42638a2e13199d3ef40b7fb32e14e379fe19626/affinegap-1.10-cp27-cp27m-manylinux1_i686.whl
35904 murmurhash-0.26.4-cp27-cp27m-manylinux1_x86_64.whl https://pypi.python.org/packages/7c/a9/0899d646b3806248ec93af2ee696fc72414781c4a5a1233fe51813960cbd/murmurhash-0.26.4-cp27-cp27m-manylinux1_x86_64.whl
23811 pandas_plink-1.1.6-cp27-cp27m-manylinux1_x86_64.whl https://pypi.python.org/packages/7a/c0/d8da30f905d1c8b947ccf8d72e9ba0cf3094baeab7ca672c6cfe9744f86e/pandas_plink-1.1.6-cp27-cp27m-manylinux1_x86_64.whl
23305 pycrypt-0.5.1-cp26-cp26m-manylinux1_i686.whl https://pypi.python.org/packages/6b/e6/65264364bfbd3a76e1476c94b4a41ff0962e875ed43fa3a29af7182cca95/pycrypt-0.5.1-cp26-cp26m-manylinux1_i686.whl
22590 json_cherry_pick-0.3.22-cp26-cp26m-manylinux1_i686.whl https://pypi.python.org/packages/2f/dd/94bb6995931237bebf91fa80f393baa2d1105b03d1d23c76c217c92a0b61/json_cherry_pick-0.3.22-cp26-cp26m-manylinux1_i686.whl
19338 sabyenc-3.0.2-cp27-cp27m-manylinux1_i686.whl https://pypi.python.org/packages/ea/f3/891da926c8728c132485e777554737ee0ebc085cff3c1b212920dec0eacf/sabyenc-3.0.2-cp27-cp27m-manylinux1_i686.whl
19263 arpreq-0.3.1-cp26-cp26m-manylinux1_i686.whl https://pypi.python.org/packages/77/34/cfbce43c41afd8773a12c9bfd9ad1e15c90c847ea387abe9cd2d151a52a9/arpreq-0.3.1-cp26-cp26m-manylinux1_i686.whl
18864 asyncwatch-0.1-cp35-cp35m-manylinux1_x86_64.whl https://pypi.python.org/packages/d9/76/d93a7a0457af70b60cbcdb8db3abdd52b45382a69d4dd4bc845cefb16fa2/asyncwatch-0.1-cp35-cp35m-manylinux1_x86_64.whl
11389 pyterminalsize-0.1.0-cp26-cp26mu-manylinux1_x86_64.whl https://pypi.python.org/packages/28/c3/542789262a7aafc935e85ab83cc8d1732401a843db0fd10100b6213d4356/pyterminalsize-0.1.0-cp26-cp26mu-manylinux1_x86_64.whl
2314 yftest-1.0.0-cp27-cp27mu-manylinux1_x86_64.whl https://pypi.python.org/packages/b8/73/0431ecd2ac0701e195c7fc99ab3b131fa32e45c5e8fa8bbb59742eabe2a9/yftest-1.0.0-cp27-cp27mu-manylinux1_x86_64.whl
```

asaitgalin:
Hi! Just a friendly ping ;) Really looking forward to seeing this pull-request is merged.

nimish:
Very interested in seeing this in pex and eventually in pants. Not being able to use manylinux wheels without a hacky self-built cache is what takes up the majority of our build time (numpy, pandas, scipy stack)

Is there anything I can do to help this along?

dan-blanchard:
I've fixed the merge conflicts here again, but I'm honestly not sure what the necessary next steps are at this point. We discussed possibly having another reviewer chime in about the CLI args. My initial idea behind them was to just make them match the ones `pip` uses, but then I added the `--tags` argument as @kwlzn and I discussed, so you could do either one.

nimish:
@dan-blanchard as a mere user, I don't have a strong opinion on the tags thing--I just want something that will work for the platforms I care about (cp35-cp35m-manylinux1) that's integrated into the main release, so that I can eventually get it into at least my pants install.

That said, it looks like explicitly defining the full platform tag (analogous to gcc platform triples) is the way to go rather than assume some decomposition will work (combinatorial explosion + implicit pickup of new platforms from combinations). The space of platforms of built wheels is pretty sparsely populated at least from my casual look at pypi so it's not an issue to manually specify each platform at least for the initial version of this feature.

@kwlzn can I ask if there's anything stopping this from being merged? Apologies if I am intruding, I am just a happy user of pex

dan-blanchard:
Rebased so this has latest master changes in it.

dudeche:
Would love to see this merged.  @kwlzn @dan-blanchard what are the next steps?

dudeche:
FWIW -- we're using this workaround for the time being: https://github.com/asottile/no-manylinux1

kwlzn:
should have some cycles this week to get this in, but want to make sure we have a clear integration strategy for pants first - will revisit soon.

kramasamy:
@dan-blanchard - we like this feature. Wondering when would you merge this?

dan-blanchard:
@kramasamy, it's out of my hands at this point. I don't have much time to dedicate to this anymore, and the conflicts introduced by #394 getting merged before this are more complex than I can handle on my own.

kwlzn:
I was hoping for a second unbiased opinion or some compromise to tie-break the unresolved "options design" stalemate in this PR, but assuming that isn't going to materialize I think we should figure out a path forward in the name of progress as this seems to be a pain point for several users.

My view is still that overloading `--platform` here with an alternate extended form is the most ideal from a UX and integration standpoint. pex != pip, so we really shouldn't be striving to exactly replicate pip's UI - but instead do what makes the most sense for the project itself and it's consumers. I've actually never been a fan of the options split/ergonomics as used by `pip` which this PR gleans from and would be less than thrilled about forklifting that directly into pex.

With that said, I had eagerly landed #394 to improve the pex + bazel integration story while also hoping to force the issue a bit here so that the union of these two changes would make the design challenge more apparent. Since this was my doing, I'd be happy to do the remaining work to carry this PR over the finish line if we can just agree on a strategy for that.

So in the interest of progress, how would you feel about starting with an overloaded form of `--platform` as the primary interface to this change? Then, if after some experimentation we decide that that API is just too ergonomically terrible for whatever reason, we can revisit adding the single-shot options split as a separate PR?

If that works, I could likely get something out in the next couple of days.

kramasamy:
@kwlzn - fine with us.

dan-blanchard:
@kwlzn I'm fine with making the overloaded `--platform` the primary interface. At this point, I'd just like to have this functionality.

srkukarni:
echoing @dan-blanchard comments. Would be nice to to see this feature in action :)

sitaktif:
I thought I'd drop an alternate idea, which is to create a new argument name (e.g. `--qualifier` or `--facet`...) so that we don't overload the `--platform` argument.

`--platform foo --platform bar` would be equivalent to `--qualifier foo-any-any-any --qualifier bar-any-any`. `--platform` and `--qualifier` could potentially be mutually exclusive, but not necessarily.

This approach has the disadvantage of creating a new option, but on the other hand, does not overload the `--platform` argument and does not break as many ties with `pip` and hence might be more future-proof.

srkukarni:
Is there any chance of getting this one anytime soon?
Thanks!

kbourgoin:
Any chance of seeing this happen anytime soon? It'd be a really great feature to have.

kwlzn:
I've forked and extended this PR over at https://github.com/pantsbuild/pex/pull/399 and it's mostly there but I've unfortunately run of out of time to devote to this for the moment. if someone wanted to help carry #399 over the finish line, that'd be helpful.

dan-blanchard:
@kwlzn, I fixed all the merge conflicts, got rid of the extra command line args, overloaded everything into the platform string (like `manylinux1-x86_64-36-cp-cp36m`), made the `ID_PY` interpreter identification bits reuse the new `pep425tags` code for increased accuracy and consistency, and made the test failures more informative—I was seeing weird failure while fixing the conflicts which were impossible to address without passing `-vvvv` to the pex command.

srkukarni:
@dan-blanchard @kwlzn After Dan's fixes, any additional work that needs to be done for this pr(other than testing to make sure that it works)?

dan-blanchard:
All the tests pass when I run them locally with tox, so I'm not sure what's causing the failure on Travis. 3.3 is failing in a really weird way, but 3.3 was end-of-lifed at the end of September, so maybe that's okay.

kramasamy:
@dan-blanchard - Wondering if this is ready to be merged?

kramasamy:
@dan-blanchard - also some docs about how to use it will be nice.

kramasamy:
@dan-blanchard - i tested this PR to build pex files in mac and run in ubuntu 14.04. I used the producer_tutorial.py and consumer_tutorial.py examples in https://github.com/streamlio/pulsar-python-tutorial 

For compiling pex files, I used bazel rules for pex from https://github.com/streamlio/bazel_rules_pex
where I modified PEX_SRC in pex/pex_rules.bzl to use your branch locally. I was able to compile the producer_tutorial and consumer_tutorial successfully using the following bazel rules

```
pex_binary(
    name = "consumer",
    srcs = ["consumer.py"],
    platforms = [
        "manylinux1_x86_64-27-cp-cp27m",
        "macosx_10_12_intel-27-cp-cp27m",
    ],
    reqs = [
        "pulsar_client",
    ],
)

pex_binary(
    name = "producer",
    srcs = ["producer.py"],
    platforms = [
        "manylinux1_x86_64-27-cp-cp27m",
        "macosx_10_12_intel-27-cp-cp27m",
    ],
    reqs = [
        "pulsar_client",
    ],
)
```

When I unzip the pex files created, I do find the pulsar_client for both the platforms manylinux1 and macos as shown here 
```
ls -al
total 32
drwxr-xr-x 5 root root 4096 Oct 21 17:29 .
drwx------ 7 root root 4096 Oct 21 17:29 ..
drwxr-xr-x 4 root root 4096 Oct 21 17:29 .bootstrap
drwxr-xr-x 4 root root 4096 Oct 21 17:29 .deps
-rw-r--r-- 1 root root  568 Oct 21 10:04 PEX-INFO
-rw-r--r-- 1 root root  737 Oct 21 10:04 __main__.py
-rw-r--r-- 1 root root  917 Oct 21 10:04 __main__.pyc
drwxr-xr-x 2 root root 4096 Oct 21 17:29 examples
root@57f53c09120c:~/t# ls -al .deps
total 16
drwxr-xr-x 4 root root 4096 Oct 21 17:29 .
drwxr-xr-x 5 root root 4096 Oct 21 17:29 ..
drwxr-xr-x 3 root root 4096 Oct 21 17:29 pulsar_client-1.20.0-cp27-cp27m-macosx_10_12_intel.whl
drwxr-xr-x 3 root root 4096 Oct 21 17:29 pulsar_client-1.20.0-cp27-cp27m-manylinux1_x86_64.whl
root@57f53c09120c:~/t# ls -al .deps/
total 16
drwxr-xr-x 4 root root 4096 Oct 21 17:29 .
drwxr-xr-x 5 root root 4096 Oct 21 17:29 ..
drwxr-xr-x 3 root root 4096 Oct 21 17:29 pulsar_client-1.20.0-cp27-cp27m-macosx_10_12_intel.whl
drwxr-xr-x 3 root root 4096 Oct 21 17:29 pulsar_client-1.20.0-cp27-cp27m-manylinux1_x86_64.whl
```
I compile the pex in MacOS X and trying to run in ubuntu14.04 docker container.  When I run it, I get the following error - 

```
./producer_tutorial.pex 
Failed to execute PEX file, missing compatible dependencies for:
pulsar-client
```
Wondering what I am doing wrong?

kramasamy:
@dan-blanchard - additional data point. I compiled the pex files in ubuntu14.04 and tried to run in ubuntu14.04 itself and I get the same error

```
./producer.pex
Failed to execute PEX file, missing compatible dependencies for:
pulsar-client
```

kramasamy:
@dan-blanchard - made it work. The key change is to use 'mu' instead of 'm'

```
"manylinux1_x86_64-27-cp-cp27m",
```

```
"manylinux1_x86_64-27-cp-cp27mu",
```

and it works. Compiled on a mac and ran it in ubuntu.

kramasamy:
@dan-blanchard - tried several tests, included several platforms

```
 "manylinux1_x86_64-27-cp-cp27m",
 "manylinux1_x86_64-27-cp-cp27mu",
 "manylinux1_x86_64-33-cp-cp33m",
 "manylinux1_x86_64-34-cp-cp34m",
 "manylinux1_x86_64-35-cp-cp35m",
 "manylinux1_x86_64-36-cp-cp36m",
 "macosx_10_12_intel-27-cp-cp27m",
 "macosx_10_12_x86_64-36-cp-cp36m",
```
ran in macosx, ubuntu14.04, centos7 and tried with python2.7 and python3.5 - with single compiled pex and it works!

dan-blanchard:
@kramasamy I think this is good to go, but it would be nice to know why that test fails only on Travis. I've added the docs you requested and made that `Failed to execute PEX file, missing compatible dependencies for` message more informative so it actually tells you the `--platform` you're running on so you can make sure it matches with what you generated things for.

dan-blanchard:
One issue that is currently present in this branch is that it does not appear to handle version-specific requirements, like `'enum34; python_version < "3"'`. If you try to build a PEX for `cryptography` using Python 3 that targets 2.7, you'll find the PEX is missing those. Something somewhere is using the current interpreter that shouldn't.

kramasamy:
@dan-blanchard - for improving the error message - it is a great help. I am not an expert in the pex code. If some experts like @kwlzn or @jsirois can provide some idea where to look, I am happy to fix.

dan-blanchard:
I managed to track the issue down. Turned out to require another bit of monkeypatching because they were being filtered out inside pkg_resources/packaging.

kramasamy:
@dan-blanchard - nice work!

dan-blanchard:
The REPL test failures are now reproducible in the sense that I can also get them by running `tox`, but not in the sense that running the exact same command from my virtualenv does not cause a failure.

kramasamy:
@dan-blanchard - any luck in fixing the travis issue?

dan-blanchard:
@kramasamy No, I think one of the maintainers will have to poke around for that. I had some time to dedicate to this again, but I'm busy with other things again. I also feel like it is going to take me much longer to figure out than one of the maintainers.

kbourgoin:
So is this PR dead again?

sitaktif:
@kwlzn @dan-blanchard a lot of effort has been invested in this work (can't thank you enough for that) and it looks like we're almost there, plus there seems to be a lot of interest from `pex` users, which makes me think getting this over the finish line could be worth the investment! wdyt? :-)

dan-blanchard:
Rebased to fix conflict again.

dan-blanchard:
ALL THE TESTS PASS NOW.

dan-blanchard:
Oh joy, more conflicts to fix. 😭

dan-blanchard:
Okay, conflicts fixed again. All tests still pass.

kbourgoin:
Please, please, please can we get this merged? It not being released is holding up things and forcing us to use a branch when we should be able to use a released version of PEX. It's there, tests pass, and it works. I get the maintainers don't need this use case, but clearly several users do.

dan-blanchard:
Since there are many other users like ourselves who need this feature, we've gone ahead and forked pex. [manypex](https://pypi.org/project/manypex/) is available on PyPI. The only difference between it and 1.3.2 is the addition of this manylinux PR. We try to keep things in sync with pex as new versions come out for as long as it takes to get this functionality into mainline pex.

Also, note that the distribution name is different, but the package and CLI names did not change. That way you don't need to change anything about your workflow to use it.

anthrotype:
@dan-blanchard thanks for this, though it's a pity you had to resort to a fork. It's still not clear to me why this PR hasn't been merged into upstream yet. Is it just lack of interest or are there any outstanding issues that haven't been addressed?

cc @kwlzn

anthrotype:
@stuhood that's good, thank you for the reply! Enjoy your break :)

kwlzn:
good news and bad news here.

the bad news: it turns out that #461 has been silently masking test failures in master for several weeks now. as a result, this PR (and transitively, the manypex fork), have active test failures/breakage that is revealed if you drill into the output of individual test shards.

the good news: I've got the changes from #399 retrofitted on top of this PR, which include fixes for the failing tests and other breakage.

we're working through #461 and the underlying master failures now. once master is green again, the plan is to land this PR as-is (breaking master) and then fast-follow with the #399 retrofit (repairing master). I'll post the #399 retrofit PR for review soon (as a separate PR). but please avoid any pushes to this PR in the meantime.

but beyond that, we're hoping to cut the first official pex release with manylinux support (likely pex 1.4.0) soon! thanks for the patience and perseverance on this and in particular to @dan-blanchard and team.

kwlzn:
the final, combined manylinux implementation is now reviewable here: #480

kwlzn:
this and #480 are now in and released via pex 1.4.0 on pypi.

I added a quick usage blurb here: https://github.com/pantsbuild/pex/issues/281#issuecomment-389352166

please take a moment to kick the tires on this new release and file any issues you might find. enjoy!

PR #321, opened 2016-11-30
Lazily initialize the `Variables` object to avoid implicit code execution on import.

### LINKED ISSUE #289: importing pex.variables will may look at disk and update state
https://github.com/pantsbuild/pex/blob/master/pex/variables.py#L36

When a module imports variables it creates an ENV object which checks the env variables and stores them.  It also may read files from disk.  In most cases this is probably an unexpected side effect.  Since many modules inside pex import variables the user is often more removed from this effect since they may not be directly importing variables.

Proposed solution is to change Variables to load values lazily on first access.  This will be a change in behavior and may impact users who implicitly rely on this behavior when using pex as a library.

sixninetynine:
Fixes #289

--

The idea here is that we just rename the `Variables` class to `_Variables` and proxy it with the new `Variables` object.

My only trepidation around this change is that it can make _inspecting_ the object somewhat annoying (having to do `dir(obj.proxied)` instead of just `dir(obj)` etc), but it at least fixes the problem of implicit code execution on import highlighted by @digwanderlust)

Tox tests pass locally.

### DIFF
--- a/pex/variables.py
+++ b/pex/variables.py
@@ -13,25 +13,8 @@ from .common import die
 __all__ = ('ENV', 'Variables')
 
 
-class Variables(object):
+class _Variables(object):
   """Environment variables supported by the PEX runtime."""
-
-  @classmethod
-  def process_pydoc(cls, pydoc):
-    if pydoc is None:
-      return 'Unknown', 'Unknown'
-    pydoc = pydoc.splitlines()
-    variable_type = pydoc[0]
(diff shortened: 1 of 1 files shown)


sixninetynine:
happy new year! 😉

sixninetynine:
ping @jsirois @kwlzn 

👼

sixninetynine:
oh hey -- any opinions on this?

PR #335, opened 2016-12-30
--constraints in pex

toumorokoshi:
this is an implementation of adding constraint support to pex:

https://pip.pypa.io/en/stable/user_guide/#constraints-files

As the "-c" keyword is already used, this implementation opts for "--constraint" by itself.

It works pretty well:

```
pex requests --constraints constraints.txt -o /tmp/requests.pex
```

with the constraints file
```
requests==2.12.1
aiohttp==1.2.0
```

resulted in a pex that had requests versioned 2.12.1, but not aiohttp.

I really wanted to start the conversation with this PR: I'm new to the Pex code and didn't know the best place to implement this, so I wanted to get feedback on an approach before doing a bunch of work to add tests.

One issue along the way: conflicts arose when attempting to reference the same package twice:

```
pex requests requests==2.12.1 -o /tmp/requests.pex
```
This was caused by a lack of semantic equality in the WheelPackage, so I added that too. Happy to pick that up in another PR/Issue if that's more appropriate.

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -341,6 +341,16 @@ def configure_clp():
            'times.')
 
   parser.add_option(
+      '--constraints',
+      dest='constraint_files',
+      metavar='FILE',
+      default=[],
+      type=str,
+      action='append',
+      help='Add requirements from the given requirements file.  This option can be used multiple '
+           'times.')
+
+  parser.add_option(
       '-v',
       dest='verbosity',
       default=0,
@@ -476,6 +486,13 @@ def build_pex(args, options, resolver_option_builder):
 
   for requirements_txt in options.requirement_files:
     resolvables.extend(requirements_from_file(requirements_txt, resolver_option_builder))
+
+  for constraints_txt in options.constraint_files:
+    constraints = []
+    for r in requirements_from_file(constraints_txt, resolver_option_builder):
+      r.is_constraint = True
+      constraints.append(r)
+    resolvables.extend(constraints)
 
   resolver_kwargs = dict(interpreter=interpreter, platform=options.platform)
 
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -347,7 +347,7 @@ def configure_clp():
       default=[],
       type=str,
       action='append',
-      help='Add requirements from the given requirements file.  This option can be used multiple '
+      help='Add constraints from the given constraints file.  This option can be used multiple '
            'times.')
 
   parser.add_option(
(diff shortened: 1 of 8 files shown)


pex/bin/pex.py
```diff
@@ -477,6 +487,13 @@ def build_pex(args, options, resolver_option_builder):
   for requirements_txt in options.requirement_files:
     resolvables.extend(requirements_from_file(requirements_txt, resolver_option_builder))
 
+  for constraints_txt in options.constraint_files:
+    constraints = []
+    for r in requirements_from_file(constraints_txt, resolver_option_builder):
```
toumorokoshi: pip states the constraints format is identical to requirements:

https://pip.pypa.io/en/stable/user_guide/#constraints-files

so I figured the way to guarantee that in pex was to share the requirements code.

pex/bin/pex.py
```diff
@@ -341,6 +341,16 @@ def configure_clp():
            'times.')
 
   parser.add_option(
+      '--constraints',
+      dest='constraint_files',
+      metavar='FILE',
+      default=[],
+      type=str,
+      action='append',
+      help='Add requirements from the given requirements file.  This option can be used multiple '
```
kwlzn: should this read `Add constraints from the given constraints file` instead?

pex/bin/pex.py
sixninetynine: seems like a good tidbit to put into a comment

toumorokoshi:
@kwlzn @sixninetynine thanks for reviewing! added unit tests.

tests/test_resolver.py
```diff
@@ -73,6 +73,48 @@ def test_resolvable_set():
...
+  rs.merge(c, [package])
+
+  assert rs.packages() == [(c, set([package]), None, True)]
+
+
+def test_resolvable_set_constraint_and_non_constraint():
+  builder = ResolverOptionsBuilder()
+  rs = _ResolvableSet()
+  constraint = ResolvableRequirement.from_string('foo', builder)
+  constraint.is_constraint = True
```
kwlzn: maybe I'm missing something, but I'd expect to see a test here that:

1) adds a constraint for `foo==$SOME_VERSION`.
2) adds a non-constraint requirement for `foo`.
3) given multiple visible `foo` versions (some higher, some lower than $SOME_VERSION), selects the appropriate one based on the constraint.

and then maybe another variant of that which doesn't apply the constraint and would end up with the very latest version of foo available under the same conditions (sans the constraint)?

seem reasonable?

toumorokoshi: yes, that makes sense. the latter sounds like a unit test that should already exist (since that's just testing standard requirement behavior), but I can add both.

toumorokoshi:
@kwlzn how does this look?

kwlzn:
@toumorokoshi thanks for the PR! and sorry for the delay on reviewing this - I've been out sick for a couple of days.

PR #343, opened 2017-01-16
Add dummy flush method for clean interpreter exit with python3.6

### LINKED ISSUE #342: Exit code 120 with python3.6
When trying to move to `python3.6` I noticed that our py.test builds are failing because the exit code of any PEX run is 120. This doesn't seem happen with `python3.5`. I'm observing this on arch linux as well as in a `ubuntu14.04` docker image.

To reproduce, create a pex file then launch it, exit with CTRL+D and check the return code.

```
$pex -o mypex
$./mypex
Python 3.6.0 (default, Dec 24 2016, 08:03:08) 
[GCC 6.2.1 20160830] on linux
(InteractiveConsole)
>>> 
$echo $?
120
```

sveitser:
This fixes #342 for me.

### DIFF
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -31,6 +31,9 @@ class DevNull(object):
   def write(self, *args, **kw):
     pass
 
+  def flush(self):
+    pass
+
 
 class PEX(object):  # noqa: T000
   """PEX, n. A self-contained python environment."""


kwlzn:
thanks for the PR @sveitser

Issue #347, opened 2017-01-20
Add python 3.6 coverage to tox.ini / .travis.yml


jsirois:
The 3.6 release is out there, people are using it and we've already hit bugs not caught by CI.

PR #348, opened 2017-01-20
Add py36 environments for tox/travis.

jsirois:
Addresses #347


jsirois:
OK - 36 stuff appears to go green but is actually failing for lack of python 3.6 on travis - working on that and will follow up with a diff to `.travis.yml`.

PR #350, opened 2017-01-21
Add support for controlling prerelease resolution.

### LINKED ISSUE #28: Filter prerelease versions in package iterators by default
Pip has the `--pre` flag -- we should allow similar behavior.

jsirois:
This change both allows control over whether prereleases are allowed
when resolving requirements and changes the default from allowing
prereleases to disallowing them - in line with pip behavior.

Support includes a pip-alike `--pre` cli option, support for `--pre` in
requirements.txt files, and API support for consumers of the pex
library, notably pants.

This closes #28.

### DIFF
--- a/pex/package.py
+++ b/pex/package.py
@@ -3,6 +3,7 @@
 
 import os
 
+from packaging.specifiers import SpecifierSet
 from pkg_resources import EGG_NAME, parse_version, safe_name, safe_version
 
 from .archiver import Archiver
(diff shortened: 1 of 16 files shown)


jsirois:
I'm happy to update `CHANGES.rst` once we've setlled on a release version which would seem to depend on how the new `allow_prereleases`/`--pre` knob is rolled out. As-is the default is behavior changing, so I think we technically would need to release `2.0.0`. If the default were changed to including pre-releases, we could release `1.2.0`.  Either way, I'm pretty sure we can't release as `1.1.21` per semver.

pex/version.py
```diff
@@ -3,5 +3,11 @@
 
 __version__ = '1.1.20'
 
-SETUPTOOLS_REQUIREMENT = 'setuptools>=5.7,<31.0'
+PACKAGING_REQUIREMENT = 'packaging>=16.8'
+
+# NB: We exlude 7.0b1 since it will use `packaging` if present and we install a modern version of
```
jsirois: Fixed.

pex/package.py
```diff
@@ -67,18 +68,27 @@ def raw_version(self):
...
     link_name = safe_name(self.name).lower()
     if link_name != requirement.key:
       return False
-    return self.raw_version in requirement
+
+    # TODO(John Sirois): This is a bit roundabout. Modern setuptools `Requirement` objects expose a
+    # `specifiers` field which is a `SpecifierSet`, but we can't be sure we have that. We can be
+    # sure though we have the `Requirement.specs` list of `(op, version)` tuples in all versions of
+    # setuptools we depend upon (we use `Requirement.specs` elsewhere in the codebase as well).
+    # Kill this re-parsing once our lower-bound setuptools dependency give access to
```
jsirois: Agreed. I'll take that up in a follow-up.

kwlzn:
re: version bumps, I'm good with whatever you think is the most reasonable (2.x.x or 1.2.x). imo, the fact that we were resolving pre-release versions is sort of in the gray area of "bug" territory. ostensibly, we shouldn't have been doing that - so correcting that "bug" here and thus aligning with the spirit of the "correct" design for 1.x may be sufficient grounds for excepting this to a 1.2.x release (and explicitly noting the flag to restore the broken behavior) may be fine. I doubt anyone is relying on this behavior in practice - but may be wrong. ultimately, I had always imagined a 2.x would carry a much larger set of breaking changes, but either way not that big of a deal - will let you decide.

jsirois:
Yeah - I have the same sentiment re `2.x`. Though the correct answer, it has perceived connotations that will be wrong. I'll follow up with the `CHANGES.rst` in a new PR.

PR #352, opened 2017-01-23
Prepare the 1.2.0 release.

jsirois:


pex/version.py
```diff
@@ -1,13 +1,8 @@
...
-PACKAGING_REQUIREMENT = 'packaging>=16.8'
-
-# NB: We exclude 7.0b1 since it will use `packaging` if present and we install a modern version of
-# packaging that may not be compatible. Versions before 7.0b1 don't use `packaging` and versions
-# after use a vendored `packaging`.
-SETUPTOOLS_REQUIREMENT = 'setuptools>=5.7,!=7.0b1<31.0'
+__version__ = '1.2.0'
 
+PACKAGING_REQUIREMENT = 'packaging>=16.8,<17.0'
+SETUPTOOLS_REQUIREMENT = 'setuptools>=30.0,<35.0'
```
kwlzn: I noticed in the [changelog]((http://setuptools.readthedocs.io/en/latest/history.html#v34-0-0) that the 34.0.0 version of setuptools is the first release that forgoes it's previously vendored deps in favor of typical dep linkages - but it seems there's a slight chicken/egg scenario here based on their strong advice to always use pip to install that made me wonder if this will break things for pex? may be good to do some extra tire kicking around `setuptools==34.0.0` in particular.

also, I wonder if it'd make sense to relax the lower end of this range a bit - perhaps early in the 2x.0.0 series (e.g. `setuptools>=20.0,<35.0`) - just to ensure flexibility for other composed libs in the same environment which may not be prepared for 30.x (which was just released in Dec '16).

jsirois: > ... made me wonder if this will break things for pex?

I think you meant pants, and agreed - not sure. Suggestions incorporated.

PR #355, opened 2017-01-23
Ensure `packaging` dependency is self-contained.

### LINKED ISSUE #354: Pex is no longer self-contained
In particular, its dependency on packaging is not included in the `.bootstrap` which can lead to errors when used as a library.

jsirois:
This dependency was not included in the pex `.bootstrap/` previously
leading to issues with use of pex as a library. We use the vendored
version from setuptools for now - which it leaks through its API. Notes
are added in the right spots to deal with setuptools de-vendoring of
`packaging` in its newest releases when we upgrade setuptools next.

Closes #354

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,16 @@
 Release Notes
 =============
 
+1.2.1
+-----
+
+This release is a quick fix for a bootstrapping bug that inadvertently went out in 1.2.0 (Issue
+#354).
+
+* Ensure `packaging` dependency is self-contained. (#355)
+  `#355 <https://github.com/pantsbuild/pex/pull/355>`_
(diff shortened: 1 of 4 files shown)


jsirois:
Now:
```
$ tox -e py27-package
$ zipinfo -1 dist/pex27 | grep packaging/specifiers
.bootstrap/pkg_resources/_vendor/packaging/specifiers.py
.bootstrap/pkg_resources/_vendor/packaging/specifiers.pyc
.deps/setuptools-33.1.1-py2.py3-none-any.whl/pkg_resources/_vendor/packaging/specifiers.py
```

I also verified setuptools 20.0.0 - our lower bound - has a vendored `packaging/specifiers.py`.

PR #357, opened 2017-01-30
Fix CacheControl import.

dsully:
Tested locally.


dsully:
@jsirois would love to get this in a release ASAP, as we can't use pex + CacheControl as a dependency right now.

Thanks

jsirois:
This fix is now out in https://pypi.python.org/pypi/pex/1.2.2

dsully:
Thanks for the quick turn-around!

Issue #178, opened 2015-11-13
pex throws a resolution error when using cache, succeeds otherwise


kwlzn:
when passing two requirements of the form (unpinned, pinned) for the same package (e.g. 'thrift thrift==0.9.1'), pex has different behavior in resolution depending on whether or not the cache is used. this also manifests in consumers of pex (i.e. pants) when dependencies land in a similar order.

repro:

```
[illuminati build]$ pex --version
pex 1.0.3
[illuminati build]$ rm -rf ~/.pex
[illuminati build]$ pex thrift thrift==0.9.1 -o test.pex
[illuminati build]$ pex thrift thrift==0.9.1 -o test.pex
Could not satisfy all requirements for thrift:
    thrift, thrift==0.9.1
```

this also repros under 1.1.0 and for different packages:

```
[illuminati build]$ pex --version
pex 1.1.0
[illuminati build]$ rm -rf ~/.pex
[illuminati build]$ pex psutil psutil==3.2.1 -o test.pex
[illuminati build]$ pex psutil psutil==3.2.1 -o test.pex
Could not satisfy all requirements for psutil:
    psutil, psutil==3.2.1
```

and is also affected by ordering:

```
[illuminati build]$ rm -rf ~/.pex
[illuminati build]$ pex psutil==3.2.1 psutil -o test.pex
[illuminati build]$ pex psutil==3.2.1 psutil -o test.pex
[illuminati build]$ pex psutil==3.2.1 psutil -o test.pex
[illuminati build]$ pex psutil psutil==3.2.1 -o test.pex
Could not satisfy all requirements for psutil:
    psutil, psutil==3.2.1
```

jsirois:
Some of the research here is relevant: https://github.com/pantsbuild/pants/issues/2533

arnib:
I've been hitting this with our pants setup as well.
We have tests depending on moto, which in turn has a dependency on boto with an unpinned version.
The application has a dependency on boto with a pinned version.

This results in a error from pants/pex, which looks like it could be solved:
`
Exception message: Could not satisfy all requirements for boto==2.38.0:
    boto==2.38.0, boto>=2.20.0(from: moto==0.4.10)
`

kwlzn:
@arnib do you have any data or a repro you can share that indicates these issues are related? afaict, #178 only happens when one of the deps is completely unpinned - however in your case it appears boto is pinned somewhere (`boto==2.38.0`).

fwiw:

```
[...]

PR #362, opened 2017-02-17
Fix corner case in cached dependency resolution

butlern:
Ref: https://github.com/pantsbuild/pex/issues/178

  The error is as follows:

  1. Two resolvables, one pinned, the other not.
  2. First pex run with clean cache works.
  3. Change the pinned version.
  4. Second pex run fails with Unsatisfiable.

  The problem is that the non-pinned resolvable locates the cached dep
  and only returns it. Since that version doesn't match the pinned
  version, we fail. This fixes that behavior by returning both the
  cached version and all uncached pypi packages for unpinned
  resolvables.

  We change the pinned version to test the use case where you have a
  cached version of a dep and then you transitively pull in another
  requirement that either has a tighter bound or a pin on a version not
  found in the cache.

### DIFF
--- a/pex/resolver.py
+++ b/pex/resolver.py
@@ -3,6 +3,7 @@
 
 from __future__ import print_function
 
+import itertools
 import os
 import shutil
 import time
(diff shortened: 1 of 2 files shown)


pex/resolver.py
```diff
@@ -248,8 +248,12 @@ def package_iterator(self, resolvable, existing=None):
...
-        if packages:
-          return packages
+        # Return both cached packages and the pypi packages. The edge case is if a inexact
+        # resolvable finds a cached version, it needs to return that cached package AND all pypi
+        # packages because the cached version might not match another resolvable which has a tighter
+        # bound or an exact version.
+        return packages + super(CachingResolver, self).package_iterator(resolvable,
+                                                                        existing=existing)
 
     return super(CachingResolver, self).package_iterator(resolvable, existing=existing)
```
butlern: So something like this?

```
  def package_iterator(self, resolvable, existing=None):
    iterator = Iterator(fetchers=[Fetcher([self.__cache])])
    packages = self.filter_packages_by_interpreter(
        resolvable.compatible(iterator), self._interpreter, self._platform)

    if packages:
      if resolvable.exact:
        return packages

      if self.__cache_ttl:
        packages = self.filter_packages_by_ttl(packages, self.__cache_ttl)
      else:
        packages = []

    return packages + super(CachingResolver, self).package_iterator(resolvable, existing=existing)
```

kwlzn: well, you could go a step further and guard the initialization of `packages` altogether by checking for the conditions in which it'd be used - was sort of imagining something along the lines of:

```
  def package_iterator(self, resolvable, existing=None):
    packages = []
    if self.__cache_ttl or resolvable.exact:
      iterator = Iterator(fetchers=[Fetcher([self.__cache])])
      packages = self.filter_packages_by_interpreter(
        resolvable.compatible(iterator),
        self._interpreter,
        self._platform
      )

      if resolvable.exact and packages:
        return packages

      packages = self.filter_packages_by_ttl(packages, self.__cache_ttl)

    return itertools.chain(
      packages,
      super(CachingResolver, self).package_iterator(resolvable, existing=existing)
    )
```

but I'm probably just splitting hairs at this point - fine with any of the above.

butlern: Gotcha, yah that looks good to me. I'll push up the change. For now I'll do a separate autosquashable commit unless you'd prefer it be a separate commit. Local tests run fine with that change btw :)

kwlzn:
merged - thanks for the PR @butlern !

PR #365, opened 2017-02-27
Fix bug in cached dependency resolution with exact resolvable

### LINKED ISSUE #178: pex throws a resolution error when using cache, succeeds otherwise

butlern:
* Add unit test and patch code to pass unit test

  Bug description:
  1. Exact resolvable and inexact resolvable
  2. First run completes, pushes exact resolvable into cache
  3. Second run fails with Unsatisfiable

  Reproduce:
  $ pex psutil psutil==3.2.1 -o test.pex
  $ pex psutil psutil==3.2.1 -o test.pex
  Could not satisfy all requirements for psutil:
      psutil, psutil==3.2.1

  The bug manifest itself in #178 because by default --cache-ttl is
  set to None. This doesn't change those semantics. The default is still
  None, so we simply don't return an exact resolvable found in cache
  if the --cache-ttl isn't set. Doing so also fixes another bug where
  --cache-ttl is set, there is an exact resolvable found in cache, but
  it is expired.

  It should be noted that this patch could affect the performance of
  default pex runs which specify an exact resolvable but don't set
  --cache-ttl.

@kwlzn this should fix #178 now, however, it changes the previous behavior by not returning an exact resolvable found in cache which might affect default performance. I was a little hesitant to set a default for --cache-ttl. Let me know if you think I should do that.

### DIFF
--- a/pex/resolver.py
+++ b/pex/resolver.py
@@ -239,7 +239,6 @@ class CachingResolver(Resolver):
 
   # Short-circuiting package iterator.
   def package_iterator(self, resolvable, existing=None):
-    packages = []
     iterator = Iterator(fetchers=[Fetcher([self.__cache])])
     packages = self.filter_packages_by_interpreter(
       resolvable.compatible(iterator),
--- a/pex/resolver.py
+++ b/pex/resolver.py
@@ -240,20 +240,15 @@ class CachingResolver(Resolver):
   # Short-circuiting package iterator.
   def package_iterator(self, resolvable, existing=None):
     packages = []
-    if self.__cache_ttl or resolvable.exact:
-      iterator = Iterator(fetchers=[Fetcher([self.__cache])])
-      packages = self.filter_packages_by_interpreter(
-        resolvable.compatible(iterator),
-        self._interpreter,
-        self._platform
-      )
+    iterator = Iterator(fetchers=[Fetcher([self.__cache])])
+    packages = self.filter_packages_by_interpreter(
+      resolvable.compatible(iterator),
+      self._interpreter,
+      self._platform
+    )
 
-      if packages:
-        if resolvable.exact:
-          return packages
-
-        if self.__cache_ttl:
-          packages = self.filter_packages_by_ttl(packages, self.__cache_ttl)
+    if packages and self.__cache_ttl:
+      packages = self.filter_packages_by_ttl(packages, self.__cache_ttl)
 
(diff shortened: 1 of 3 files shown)


pex/resolver.py
```diff
@@ -240,20 +240,15 @@ def __init__(self, cache, cache_ttl, *args, **kw):
...
-      )
-
-      if packages:
-        if resolvable.exact:
-          return packages
-
-        if self.__cache_ttl:
-          packages = self.filter_packages_by_ttl(packages, self.__cache_ttl)
+    iterator = Iterator(fetchers=[Fetcher([self.__cache])])
+    packages = self.filter_packages_by_interpreter(
```
kwlzn: should probably kill the `packages = []` above btw.

butlern:
Sounds good, removed packages=[] and set default ttl to 3600.

kwlzn:
@butlern merged. thanks for the PR!

Issue #366, opened 2017-02-28
Ambiguous Resolvable raised on transitive dep


butlern:
This always fails:

```
$ tox -e py27-package
$ ./dist/pex27 botocore awscli==1.11.54 -o test.pex
Traceback (most recent call last):
  File ".bootstrap/_pex/pex.py", line 339, in execute
  File ".bootstrap/_pex/pex.py", line 267, in _wrap_coverage
  File ".bootstrap/_pex/pex.py", line 299, in _wrap_profiling
  File ".bootstrap/_pex/pex.py", line 382, in _execute
  File ".bootstrap/_pex/pex.py", line 440, in execute_entry
  File ".bootstrap/_pex/pex.py", line 458, in execute_pkg_resources
  File "/Users/nate/.pex/install/pex-1.2.3-py2.py3-none-any.whl.97a3e8c0f7213b5f56931db5cbda8074a8f4973a/pex-1.2.3-py2.py3-none-any.whl/pex/bin/pex.py", line 578, in main
    pex_builder = build_pex(reqs, options, resolver_options_builder)
  File "/Users/nate/.pex/install/pex-1.2.3-py2.py3-none-any.whl.97a3e8c0f7213b5f56931db5cbda8074a8f4973a/pex-1.2.3-py2.py3-none-any.whl/pex/bin/pex.py", line 527, in build_pex
    resolveds = resolver.resolve(resolvables)
  File "/Users/nate/.pex/install/pex-1.2.3-py2.py3-none-any.whl.97a3e8c0f7213b5f56931db5cbda8074a8f4973a/pex-1.2.3-py2.py3-none-any.whl/pex/resolver.py", line 205, in resolve
    raise self.Error('Ambiguous resolvable: %s' % resolvable)
pex.resolver.Error: Ambiguous resolvable: botocore
```

It looks like this happens because botocore is unbounded and https://github.com/pantsbuild/pex/blob/master/pex/resolver.py#L201 grabs the first SourcePackage from the list of packages which happens to be the latest, 1.5.18. However, awscli==1.11.54 transitively depends upon botocore==1.5.17 but this isn't found until the second round of `while resolvables` so botocore has already been added to https://github.com/pantsbuild/pex/blob/master/pex/resolver.py#L215 and then this gets raised https://github.com/pantsbuild/pex/blob/master/pex/resolver.py#L204-L205

[...]

PR #367, opened 2017-02-28
Potential fix for Ambiguous Resolvable bug in transitive dependency resolution

butlern:
If an inexact dependency is specified, and a transitive dependency
  with either a pin or a tighter bound is found later, Ambiguous
  Resolvable can be raised. This is because the latest package is added
  to processed_packages and then by the time the transitive dep runs
  through the while resolvables loop, it's version doesn't match the
  one found in processed_packages and Ambiguous Resolvable is raised.

  This patch fixes it by 1) not raising an error if the packages don't
  match and 2) after all resolvables are processed, loop back through
  distributions and recreate the correct list of distributions from
  resolvable_set.packages().

  Fixes: #366

### DIFF
--- a/pex/resolver.py
+++ b/pex/resolver.py
@@ -200,10 +200,8 @@ class Resolver(object):
         assert len(packages) > 0, 'ResolvableSet.packages(%s) should not be empty' % resolvable
         package = next(iter(packages))
         if resolvable.name in processed_packages:
-          # TODO implement backtracking?
-          if package != processed_packages[resolvable.name]:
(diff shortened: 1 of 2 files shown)


kwlzn:
@butlern apologies for the delay in reviewing this and other pex PRs - have been swept up in end-of-quarter activities, but will carve out some time in the next couple days.

butlern:
Hey no worries. I didn't want to have to bother you specifically. I sent that email to pants-devel because I didn't know if there was a single gatekeeper (you) or a group.

PR #373, opened 2017-04-02
Pass allow_prerelease to other iterators (Static, Caching)

toddgardner:
Not sure if this is the best way to do this or test this, but while investigating another bug I noticed caching resolver always ignored prerelease versions, as it constructed it's own iterator and didn't take into account the options passed on resolve. This fixes that and tests it.

### DIFF
--- a/pex/resolver.py
+++ b/pex/resolver.py
@@ -35,12 +35,13 @@ class Unsatisfiable(Exception):
 class StaticIterator(IteratorInterface):
   """An iterator that iterates over a static list of packages."""
 
-  def __init__(self, packages):
+  def __init__(self, packages, allow_prereleases=None):
     self._packages = packages
+    self._allow_prereleases = allow_prereleases
 
   def iter(self, req):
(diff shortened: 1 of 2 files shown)


toddgardner:
I found another issue of the same nature, where allow_prerelease wasn't being passed to StaticIterator either, this meant two requirements like `pantsbuild.pants>=1.3.0.dev1` and `pantsbuild.pants==1.3.0.dev14` could never be satisfied, since the static iterator would exclude all prereleases regardless of allow_prereleases.

I rolled it into this change since it touched the same code.

Note this will have to be updated if #374 is accepted as it includes defaults that should be updated.

bloomonkey:
+1
I think I'm getting bitten by this too - a dependency of one of my dependencies is a pre-release.

jsirois:
@toddgardner - #374 is now merged, please update.

toddgardner:
Updated and added some tests for the defaults.

Issue #372, opened 2017-04-02
Resolving pre-release when explicitly requested


toddgardner:
So upgrading to pants 1.3.0 dev releases, I ran into an issue with an whl we have as a pre-release, pex is more picky with these than pip at the moment:

```
$ ls wheelhouse/pydicom-1.0.0a1-py2-none-any.whl 
wheelhouse/pydicom-1.0.0a1-py2-none-any.whl
$ pip install pydicom==1.0.0a1 -f wheelhouse/ --no-index
Collecting pydicom==1.0.0a1
Installing collected packages: pydicom
Successfully installed pydicom-1.0.0a1
$ pex pydicom==1.0.0a1 -f wheelhouse/ --no-index
Could not satisfy all requirements for pydicom==1.0.0a1:
    pydicom==1.0.0a1
```

Looking through pep 440 and pip code, it looks like if any specifier in a set is inclusive and specifies a pre-release, then pre-releases should be accepted.

It seems like that could just be a check here: https://github.com/pantsbuild/pex/blob/d70f2c4e679770a586c765eaaa5ce6466bbc8879/pex/package.py#L82 if operator is inclusive and version is prerelease?

kwlzn:
fixed in #374

PR #374, opened 2017-04-02
Change prerelease default to None

toddgardner:
This is a fix for #372

Reading the packaging code there was a semantic difference between "prereleases=None" and "prereleases=False"; this sets the default from False to None, like it is in pip; that way, you'll get a prerelease version if you explicitly specify it (with an inclusive operator).


toddgardner:
This needs to be tweaked after #373 is merged and can probably use some more end to end tests of the defaults then also.

jsirois:
This will go out with #375

Issue #223, opened 2016-03-30
Pex file looses the executable permission of a binary file if packaged 


sumitsanghrajka:
I have a use case where the pex file I am generating has to package another statically compiled binary. 
When the pex file is exploded the binary file has lost the executable permission. Is there anyway to preserve the permissions?

kwlzn:
are you sure the binary is executable before being composed into the pex? how are you including the binary? how are you extracting/invoking it?

we have several cases where we package binaries as resources in libraries and then compose those libraries into a pex  - and AFAICT this has never been a problem.

a repro would be helpful to understand the exact situation you're facing.

snyaggarwal:
Still happening with version 1.2.6:
Reproducible code is in foo.zip.
foo/yo/bar.sh has executable permissions.
After unzip run -- pex . -e bar.console:run -o foobar.pex -v
the pex file wont have executable permissions on bar.sh

[foo.zip](https://github.com/pantsbuild/pex/files/943355/foo.zip)

Is there anyway to preserve the permissions?

PR #381, opened 2017-04-24
making sure permissions retains while building pex binary

snyaggarwal:
Address issue #223

### DIFF
--- a/pex/common.py
+++ b/pex/common.py
@@ -338,7 +338,7 @@ class Chroot(object):
     with contextlib.closing(zipfile.ZipFile(filename, mode)) as zf:
       for f in sorted(self.files()):
         filepath = os.path.join(self.chroot, f)
-        perm = oct(os.stat(filepath)[stat.ST_MODE])[-3:]
         zf.write(filepath, arcname=f, compress_type=zipfile.ZIP_DEFLATED)
+        perm = os.stat(filepath).st_mode
         fileinfo = zf.getinfo(f)
-        fileinfo.external_attr = int(perm) << 16
+        fileinfo.external_attr = perm << 16
--- a/pex/common.py
+++ b/pex/common.py
@@ -290,7 +290,7 @@ class Chroot(object):
     safe_copy(abs_src, abs_dst, overwrite=False)
     # TODO: Ensure the target and dest are the same if the file already exists.
 
-  def write(self, data, dst, label=None, mode='wb'):
+  def write(self, data, dst, label=None, mode='wb', perm=None):
     """Write data to ``chroot/dst`` with optional label.
 
     Has similar exceptional cases as ``Chroot.copy``
@@ -298,8 +298,11 @@ class Chroot(object):
     dst = self._normalize(dst)
     self._tag(dst, label)
     self._ensure_parent(dst)
-    with open(os.path.join(self.chroot, dst), mode) as wp:
+    _file = os.path.join(self.chroot, dst)
+    with open(_file, mode) as wp:
       wp.write(data)
+    if perm:
+      os.chmod(_file, perm)
 
   def touch(self, dst, label=None):
     """Perform 'touch' on ``chroot/dst`` with optional label.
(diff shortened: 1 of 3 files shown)


pex/common.py
```diff
@@ -334,4 +337,8 @@ def delete(self):
   def zip(self, filename, mode='wb'):
     with contextlib.closing(zipfile.ZipFile(filename, mode)) as zf:
       for f in sorted(self.files()):
-        zf.write(os.path.join(self.chroot, f), arcname=f, compress_type=zipfile.ZIP_DEFLATED)
+        filepath = os.path.join(self.chroot, f)
+        perm = oct(os.stat(filepath)[stat.ST_MODE])[-3:]
```
jsirois: FWICT there is no need to `oct` the value here into a string only to `int` it below. Just `perm = os.stat(filepath).st_mode`, then `fileinfo.external_attr = perm << 16`.

pex/util.py
```diff
@@ -177,6 +177,9 @@ def cache_distribution(cls, zf, source, target_dir):
           with contextlib.closing(zf.open(name)) as zi:
             with safe_open(os.path.join(target_dir_tmp, target_name), 'wb') as fp:
               shutil.copyfileobj(zi, fp)
+              fileinfo = zf.getinfo(zi.name)
+              perm = fileinfo.external_attr >> 16
+              os.chmod(fp.name, int(str(perm), 8))
```
jsirois: Just `os.chmod(fp.name, fileinfo.external_attr >> 16)` - the ->str->int dance is not needed unless I'm missing something.

pex/util.py
```diff
@@ -177,6 +177,9 @@ def cache_distribution(cls, zf, source, target_dir):
           with contextlib.closing(zf.open(name)) as zi:
             with safe_open(os.path.join(target_dir_tmp, target_name), 'wb') as fp:
               shutil.copyfileobj(zi, fp)
+              fileinfo = zf.getinfo(zi.name)
```
jsirois: Maybe extract these 2 lines into a utility function in this file that pex_builder.py can re-use.

jsirois:
Also note that CI failed import sort order checks. The fix is automated, just: `tox -eisort-run`

snyaggarwal:
When do you think this can be merged?

pex/common.py
```diff
@@ -290,16 +290,19 @@ def link(self, src, dst, label=None):
...
+  def write(self, data, dst, label=None, mode='wb', perm=None):
     """Write data to ``chroot/dst`` with optional label.
 
     Has similar exceptional cases as ``Chroot.copy``
     """
     dst = self._normalize(dst)
     self._tag(dst, label)
     self._ensure_parent(dst)
-    with open(os.path.join(self.chroot, dst), mode) as wp:
+    _file = os.path.join(self.chroot, dst)
```
kwlzn: it'd be great to avoid the use of `_` prefixed variable names here with `_file` (e.g. `file_path` would be better/more descriptive). theres no great reason nor benefit to it as far as I can tell and it's visually confusing with the private function/method convention..

eryshev:
Do you have any updates on that? 
It breaks some important packages as pyspark for example that relies on spark-submit executable scripts.

snyaggarwal:
Completely forgot about this. I will take a look.

Sunny Aggarwal
Mobile: +91-9560331114
Skype: saggarwal89

Twitter <https://twitter.com/#!/snyaggarwal> | LinkedIn
<http://in.linkedin.com/in/snyaggarwal> | GitHub
<https://github.com/snyaggarwal>

fhoering:
I did some tests as we are blocked by this and I would like to get this merged.
It seems enough to only copy the permissions to the PEX_ROOT cache.

I created another PR with a unit test:
https://github.com/pantsbuild/pex/pull/703/files

jsirois:
#703 implemented this so closing. Thanks for pushing this along back in the day!

PR #386, opened 2017-05-06
Pass platform for SourceTranslator

thesamet:
Similarly to WheelTranslator and EggTranslator, pass the platform to SourceTranslator. Without it, SourceTranslator takes the current platform.

### DIFF
--- a/pex/resolver_options.py
+++ b/pex/resolver_options.py
@@ -172,7 +172,10 @@ class ResolverOptions(ResolverOptionsInterface):
         translators.append(EggTranslator(interpreter=interpreter, platform=platform))
       elif package is SourcePackage:
         installer_impl = WheelInstaller if WheelPackage in self._precedence else EggInstaller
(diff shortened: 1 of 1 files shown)


kwlzn:
@thesamet what's the point of passing this? afaik, sdists are compatible with all platforms - so the compatibility check is moot. and since there's no inherent cross-compilation ability, specifying a platform that isn't the current platform for `SourceTranslator` strikes me as odd.

thesamet:
I am actually implementing cross compilation: I created a shell script that runs Python 2 inside a docker container and providing that as an interpreter to pants (and hence to pex). So for me, current platform is "macosx-10.11-intel" but the target platform is "linux-x86_64". The source translator is invoking bdist on the source targets and creates a linux-specific wheel. Then, the compatibility check here https://github.com/pantsbuild/pex/blob/master/pex/translator.py#L114 fails without this PR, since self._platform would be OS X and not Linux.

kwlzn:
@thesamet ah, gotcha.

PR #394, opened 2017-07-24
Add first-class support for multi-interpreter and multi-platform pex construction.

### LINKED ISSUE #68: Make --platform/--interpreter multi-options

kwlzn:
This revives and extends the prior attempt in #201 to forklift the multi-interpreter/platform resolve behavior from pants directly into pex to better support general usage via `pex(1)` (and transitively, the Heron team's Bazel+pex integration).

Fixes #68.

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -297,14 +297,14 @@ def configure_clp_pex_environment(parser):
       type=str,
       action='append',
       help='The platform for which to build the PEX. This option can be passed multiple times '
-           'to create a multi-platform compatible pex. Default: %default')
+           'to create a multi-platform compatible pex. Default: current platform.')
 
   group.add_option(
       '--interpreter-cache-dir',
       dest='interpreter_cache_dir',
       default='{pex_root}/interpreters',
       help='The interpreter cache to use for keeping track of interpreter dependencies '
-           'for the pex tool. [Default: ~/.pex/interpreters]')
+           'for the pex tool. Default: `~/.pex/interpreters`.')
 
   parser.add_option_group(group)
 
@@ -493,7 +493,7 @@ def get_interpreter(python_interpreter, interpreter_cache_dir, repos, use_wheel)
 
 
 def _lowest_version_interpreter(interpreters):
-  """Given an iterable of interpreters, return the one with the lowest version."""
+  """Given a list of interpreters, return the one with the lowest version."""
   lowest = interpreters[0]
   for i in interpreters[1:]:
     lowest = lowest if lowest < i else i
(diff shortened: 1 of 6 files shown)


pex/bin/pex.py
```diff
@@ -460,39 +466,54 @@ def installer_provider(sdist):
...
       interpreter = resolve(interpreter, WHEEL_REQUIREMENT)
 
     return interpreter
 
 
-def build_pex(args, options, resolver_option_builder):
-  with TRACER.timed('Resolving interpreter', V=2):
-    interpreter = interpreter_from_options(options)
+def _lowest_version_interpreter(interpreters):
+  """Given an iterable of interpreters, return the one with the lowest version."""
```
jsirois: More than an iterable. Just say list?

pex/bin/pex.py
```diff
@@ -460,39 +466,54 @@ def installer_provider(sdist):
...
+  lowest = interpreters[0]
+  for i in interpreters[1:]:
+    lowest = lowest if lowest < i else i
```
jsirois: This implies the interpreter objects implement `<` which suggests `sorted(interpreters)[0]` (or -1, I'm never sure!). If this thinking is correct, then the comment above is moot.

pex/bin/pex.py
```diff
@@ -290,8 +293,11 @@ def configure_clp_pex_environment(parser):
   group.add_option(
       '--platform',
       dest='platform',
-      default=Platform.current(),
-      help='The platform for which to build the PEX.  Default: %default')
+      default=[],
```
kwlzn: ah, how would you feel about just replacing the `%default` here with a literal description?

the problem with defaulting to `[Platform.current()]` here is that the `action=append` mode is additive only, so without implementing a custom callback action for this option you'd end up resolving/building for the local platform every time whether or not you actually intended to.

pex/bin/pex.py
kwlzn: I considered that and attempted it, but given that `PythonInterpreter` only implements `__lt__` I was worried that `sorted` could misbehave - so stuck to use of the bare `<` operator only here.

I tried to add `functools.@total_ordering` to `PythonInterpreter`, which should've made `sorted` 100% safe, but that failed in the 2.6 shard due to not being present - so stuck with a 2.6-safe approach for the time being.

pex/bin/pex.py
jsirois: Makes sense, sgtm

pex/bin/pex.py
jsirois: Gotcha, makes sense.

pex/bin/pex.py
kwlzn: fixed.

pex/bin/pex.py
kwlzn: fixed.

PR #399, opened 2017-08-10
[testing]

kwlzn:


dan-blanchard:
#316 is actually pretty much ready to go now, so this should go away.

kwlzn:
there are several changes made here after you punted on #316 previously that need to be either re-incorporated into #316, or this needs to succeed #316. note the size difference between the two patches.

dan-blanchard:
This seems to have most of the same merge conflicts my PR had before I addressed them. My PR was bigger before the changes. 

It's unfortunate that we were both unknowingly working on it at the same time. I did not expect you to make a separate branch since GitHub now lets maintainers just add commits to open PR branches.

Anyway, whatever gets working manylinux support in PEX would be great.

PR #406, opened 2017-08-31
Kill `rbt` since we no longer use ReviewBoard.

jsirois:


jsirois:
Looks like this needs #407 to land 1st to get a green CI.

PR #407, opened 2017-08-31
Kill explicit `setuptools` install with `tox`.

jsirois:
Travis python environments are now modernized enough that this is not
needed.


jsirois:
I may TBR this and #406 after this since they're just housekeeping around the project that does not affect the finished product.

PR #408, opened 2017-08-31
Kill support for python 2.6.

### LINKED ISSUE #405: Drop python 2.6 support and testing
Seems about time to do this. This would help reduce legacy compatibility code and open the door for 2.7 and higher features. Chime in if you disagree / have a use case that still needs 2.6 support.

jsirois:
Closes #405

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,10 +19,6 @@ matrix:
       env: TOXENV=isort-check
 
     - language: python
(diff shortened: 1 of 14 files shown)


jsirois:
@kwlzn PTAL, added new code and tests to finally resolve the unicode/bytes issues across 2.7,3.x,pypy.

PR #409, opened 2017-08-31
Leverage `pex.common.open_zip`.

jsirois:
Also use in one spot in the `pex.testing` module where we weren't
before.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@
 # "machines".
 # See: http://docs.travis-ci.com/user/workers/container-based-infrastructure/
 sudo: false
+# The pypy test shard fails with "too many open files" under Trusty so we pin to Precise.
+dist: precise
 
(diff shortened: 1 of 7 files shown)


jsirois:
Discovered working too many open files issue under pypy in #408 CI job.

jsirois:
No clues. I just noticed the beta Trust banner on the failing pypy shard and tried pinning back to Precise. I'll try a spin against each tomorrow with `ulimit -n` tomorrow though. I expect the Trusty image will have a lower value.

jsirois:
Turns out the newer Travis Trusty image has a 30000 open file limit and Precise (the former default) 524288, so this seems to explain the new pypy failures. I did an audit of pex code file leaks and only turned up one instance of zip files in tests - fixed here, and 1 instance in the `InstallerBase` preamble. That case though is executed in a throwaway subprocess.

PR #411, opened 2017-09-01
Leverage `subprocess32` when available.

jsirois:
This allows for an upgrade in subprocess robustness when available that
Pants will leverage for one. Also formalize the pex optional
dependencies with `extras_require`, mainly for documentation purposes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,10 @@ matrix:
 
     - language: python
       python: "2.7"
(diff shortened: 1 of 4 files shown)


jsirois:
I'm not sure I understand the full negative implications - if any - of the `extras_require` portion of this change. If there are any I'm happy to retract that bit.

PR #416, opened 2017-09-10
Performance improvements for builder

jimi-c:
1) The first improvement inserts a break after setting wrote_setuptools
   to true. This saves a lot of seemingly unnecessary calls to walk_data,
   which signficantly improves performance.
2) The second improvement allows build to use a file-like object (ie. BytesIO())
   instead of always writing to disk.

This is my first dive into the PEX code, so I'm unsure if breaking out of the for loop for the first improvement might break anything, but in my testing I haven't seen any problems (and this reduces a typical call for build() on my system from 0.4s to 0.1s).


jimi-c:
Well after seeing the test results, it's clear this breaks a lot. The second fix does seem to work without issues though, so I may submit it as a separate PR.

PR #417, opened 2017-09-15
Create --pex-path argument for pex cli and load pex path into pex-info metadata 

CMLivingston:
Problem:
Pex environments rely on the PEX_PATH environment variable to resolve modules from other pexes when composing them into a single pex. Python scripts that re-execute (like a server listening for fs changes, a Jupyter notebook)  throw an ImportError upon re-exec due to environment scrubbing that removes this PEX_PATH information. 

Solution:
Add a --pex-path argument to the pex client and plumb the data through to the pex-info metadata. Resolve pexes to compose into the current pex being built by reading from the new pex_path property of the pex-info metadata. This will ensure a self-contained environment that does not lose context on a python script re-exec. 

Relates to: https://github.com/pantsbuild/pants/issues/4682

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -216,10 +216,10 @@ def test_pex_re_exec_failure():
           env = os.environ.copy()
           env['RAN_ONCE'] = '1'
           subprocess.call([sys.executable] + sys.argv, env=env)
-          sys.exit() 
+          sys.exit()
         '''))
 
-    # set up env for pex build
+    # set up env for pex build with PEX_PATH in the environment
     env = os.environ.copy()
     env['PEX_PATH'] = pex_path
     env['RAN_ONCE'] = '0'
@@ -232,8 +232,8 @@ def test_pex_re_exec_failure():
 
     # run test.py with composite env
     stdout, rc = run_simple_pex(pex_out_path, [test_file_path], env=env)
-    #raise StandardError(stdout)
-    #assert rc == 0
+
+    assert rc == 0
     assert stdout == 'Hello world\n'
 
 
@@ -249,7 +249,7 @@ def test_pex_path_arg():
     res2.assert_success()
     pex_path = os.path.join(output_dir, 'pex2.pex') + ":" + os.path.join(output_dir, 'pex1.pex')
 
-    # create test file test.py that attmepts to import modules from pex1/pex2
+    # create test file test.py that attempts to import modules from pex1/pex2
     test_file_path = os.path.join(output_dir, 'test.py')
     with open(test_file_path, 'w') as fh:
       fh.write(dedent('''\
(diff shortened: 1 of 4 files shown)


tests/test_integration.py
```diff
@@ -186,3 +186,102 @@ def test_pex_multi_resolve():
...
+    # necessary to break execution upon successful re-exec
+    env['RAN_ONCE'] = '0'
+    # set the pex arg for test.py
+    pex_out_path = os.path.join(output_dir, 'out.pex')
+    env['PEX'] = pex_out_path
+    # build out.pex composed from pex1/pex1
+    run_pex_command(['--disable-cache',
+      '--pex-path={}'.format(pex_path),
+      'wheel',
+      '-o', pex_out_path], env=env)
```
kwlzn: ditto for the `env` here.

tests/test_integration.py
```diff
@@ -186,3 +186,102 @@ def test_pex_multi_resolve():
     assert len(included_dists) == 4
     for dist_substr in ('-cp27-', '-cp36-', '-manylinux1_x86_64', '-macosx_'):
       assert any(dist_substr in f for f in included_dists)
+
+
+@pytest.mark.xfail
```
kwlzn: would be good to pass a `reason=str` arg to xfail for context on why this is expected to fail. linking a ticket etc that describes the issue would be sufficient.

e.g.

```
@pytest.mark.xfail(reason='See http://github.com/...')
```

tests/test_integration.py
```diff
@@ -186,3 +186,102 @@ def test_pex_multi_resolve():
...
+          env = os.environ.copy()
+          env['RAN_ONCE'] = '1'
+          subprocess.call([sys.executable] + sys.argv, env=env)
+          sys.exit()
+        '''))
+
+    # set up env for pex build with PEX_PATH in the environment
+    env = os.environ.copy()
+    env['PEX_PATH'] = pex_path
+    env['RAN_ONCE'] = '0'
```
kwlzn: you could kill the need to set `RAN_ONCE = 0` by doing a `if 'RAN_ONCE' in os.environ:` as the check in the test script in both tests.

tests/test_integration.py
```diff
@@ -186,3 +186,102 @@ def test_pex_multi_resolve():
...
+        import flask
+        import sys
+        import os
+        import subprocess
+        if os.environ['RAN_ONCE'] == '1':
+          print('Success!')
+        else:
+          env = os.environ.copy()
+          env['RAN_ONCE'] = '1'
+          subprocess.call([sys.executable] + [env['PEX']] + sys.argv, env=env)
```
kwlzn: rather than passing the `PEX` var via the env, you could just hardcode it in the script content via parameterization with `pex_out_path`.

tests/test_integration.py
```diff
@@ -186,3 +186,102 @@ def test_pex_multi_resolve():
...
+    env['PEX'] = pex_out_path
+    # build out.pex composed from pex1/pex1
+    run_pex_command(['--disable-cache',
+      '--pex-path={}'.format(pex_path),
+      'wheel',
+      '-o', pex_out_path], env=env)
+
+    # run test.py with composite env
+    stdout, rc = run_simple_pex(pex_out_path, [test_file_path], env=env)
+
```
kwlzn: and it seems you can kill the `env` here too when the need to set `RAN_ONCE` and `PEX` goes away.

PR #421, opened 2017-10-09
Improve PEX_PATH handling in pex.py

CMLivingston:
Replace all usages of self._vars.PEX_PATH with a unified pex path composed from merging pex_path in PEX-INFO and PEX_PATH set in the environment. 

Problem:
pex_info.pex_path was being overwritten in _activate() in pex.py by the path set in the environment. 

Solution: 
Implement a util function to merge the paths into a single pex path for usage throughout pex.py, giving priority to the PEX-INFO pex path if both are set.

### DIFF
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -68,16 +68,16 @@ class PEX(object):  # noqa: T000
 
       # set up the local .pex environment
       pex_info = self._pex_info.copy()
-      pex_info_pex_path = pex_info.pex_path
-      overrides_pex_path = self._pex_info_overrides.pex_path
+      combined_pex_paths = self._merge_split(pex_info.pex_path, self._vars.PEX_PATH)
       pex_info.update(self._pex_info_overrides)
-      pex_info.pex_path = ":".join(filter(None, [pex_info_pex_path, overrides_pex_path]))
       self._envs.append(PEXEnvironment(self._pex, pex_info))
-      # by this point, pex_info.pex_path will have a single pex path merged from pex_path
-      # in PEX-INFO and pex_path set in the environment
-      if pex_info.pex_path:
+      # N.B. By this point, `combined_pex_paths` will contain a single pex path
+      # merged from pex_path in PEX-INFO and PEX_PATH set in the environment.
+      # `PEX_PATH` entries written into `PEX-INFO` take precedence over those set
+      # in the environment.
+      if combined_pex_paths:
         # set up other environments as specified in pex_path
-        for pex_path in filter(None, pex_info.pex_path.split(os.pathsep)):
+        for pex_path in combined_pex_paths:
           pex_info = PexInfo.from_pex(pex_path)
           pex_info.update(self._pex_info_overrides)
           self._envs.append(PEXEnvironment(pex_path, pex_info))
@@ -136,7 +136,7 @@ class PEX(object):  # noqa: T000
   def site_libs(cls):
     site_libs = cls._get_site_packages()
     site_libs.update([sysconfig.get_python_lib(plat_specific=False),
-                      sysconfig.get_python_lib(plat_specific=True)])
+      sysconfig.get_python_lib(plat_specific=True)])
     # On windows getsitepackages() returns the python stdlib too.
     if sys.prefix in site_libs:
       site_libs.remove(sys.prefix)
@@ -280,7 +280,7 @@ class PEX(object):  # noqa: T000
       patch_dict(sys.modules, modules)
 
     old_sys_path, old_sys_path_importer_cache, old_sys_modules = (
-        sys.path[:], sys.path_importer_cache.copy(), sys.modules.copy())
+      sys.path[:], sys.path_importer_cache.copy(), sys.modules.copy())
     new_sys_path, new_sys_path_importer_cache, new_sys_modules = self.minimum_sys(inherit_path)
 
     new_sys_path.extend(self._merge_split(self._pex_info.pex_path, self._vars.PEX_PATH))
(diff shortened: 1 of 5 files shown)


pex/pex.py
```diff
@@ -68,11 +68,15 @@ def _activate(self):
...
       pex_info.update(self._pex_info_overrides)
+      pex_info.pex_path = ":".join(filter(None, [pex_info_pex_path, overrides_pex_path]))
       self._envs.append(PEXEnvironment(self._pex, pex_info))
-
+      # by this point, pex_info.pex_path will have a single pex path merged from pex_path
+      # in PEX-INFO and pex_path set in the environment
       if pex_info.pex_path:
-        # set up other environments as specified in PEX_PATH
+        # set up other environments as specified in pex_path
         for pex_path in filter(None, pex_info.pex_path.split(os.pathsep)):
```
kwlzn: how about instead of assigning to temp vars and then joining + mutating `pex_info.pex_path` here, just combining these into a new temporary `list` and then handling that directly?

also, would it make more sense to use `self._vars.PEX_PATH` as the source of the env var vs `self._pex_info_overrides.pex_path`? that's how it worked prior to your last change.

e.g.

```
# N.B. `PEX_PATH` entries written into `PEX-INFO` take precedence over those set in the environment.
combined_pex_path = self._merge_split(pex_info.pex_path, self._vars.PEX_PATH)
if combined_pex_path:
  for pex_path in combined_pex_path:
    ...
```

pex/pex.py
```diff
@@ -490,6 +494,10 @@ def cmdline(self, args=()):
     cmds.extend(args)
     return cmds
 
+  def _merge_split(self, *paths):
+    merged_paths = ":".join(filter(None, paths))
+    return filter(None, merged_paths.split(':'))
+
```
kwlzn: could reduce this to a one liner list comprehension:

```
def _merge_split(self, *paths):
  return [p for p in ':'.join(paths).split(':') if p]
```

with the `if p` bit doing the same as the `filter(None, ...)` above.

pex/pex.py
```diff
@@ -68,11 +68,15 @@ def _activate(self):
 
       # set up the local .pex environment
       pex_info = self._pex_info.copy()
+      pex_info_pex_path = pex_info.pex_path
+      overrides_pex_path = self._pex_info_overrides.pex_path
       pex_info.update(self._pex_info_overrides)
```
kwlzn: I assume that you found that this `pex_info.update()` call does not actually combine env vars + PEX-INFO, but instead clobbers one or the other?

tests/test_integration.py
```diff
@@ -278,3 +278,52 @@ def test_pex_path_arg():
...
+    res1 = run_pex_command(['--disable-cache', 'requests', '-o', pex1_path])
+    res1.assert_success()
+    pex2_path = os.path.join(output_dir, 'pex2.pex')
+    res2 = run_pex_command(['--disable-cache', 'flask', '-o', pex2_path])
+    res2.assert_success()
+    pex_path = ':'.join(os.path.join(output_dir, name) for name in ('pex1.pex', 'pex2.pex'))
+
+    # create a pex for environment PEX_PATH
+    pex3_path = os.path.join(output_dir, 'pex3.pex')
+    res3 = run_pex_command(['--disable-cache', 'sqlalchemy', '-o', pex3_path])
```
kwlzn: also, iirc `sqlalchemy` is a pretty heavy dep. how about something lighter, like `bitarray`, `enum34`, `mox` etc?

pex/pex.py
CMLivingston: Yup! Env vars clobber the existing vars set by pex info.

pex/pex.py
```diff
@@ -68,11 +68,16 @@ def _activate(self):
 
       # set up the local .pex environment
       pex_info = self._pex_info.copy()
+      combined_pex_paths = [pex_info.pex_path, self._vars.PEX_PATH]
       pex_info.update(self._pex_info_overrides)
+      pex_info.pex_path = ':'.join(filter(None, combined_pex_paths))
```
CMLivingston: I figure it makes more sense to use a temp var here rather than construct a new pex info API to preserve the overrides where we want pex_info vars overwritten. Doing something like `pex_info.merge_pex_paths_from_env(self._vars.PEX_PATH)` wouldn't be possible within this context as it would require more temp vars/code to ensure that oex_info.pex_path comes first in the unified path.

kwlzn: `pex_info.merge_pex_paths_from_env(self._vars.PEX_PATH)` should totally work, and it would be a cleaner API here in particular because it doesn't involve you as a caller having to remember to put the content of `pex_info.pex_path` at the head before setting it - it would just merge it for you.

the implementation would look approximately like this:

```
from pex.util import merge_split

class PexInfo(...):
  ...
  def merge_pex_path(self, pex_path):
    """Merges a new PEX_PATH definition into the existing one (if any).
    :param string pex_path: The PEX_PATH to merge.
    """
    if not pex_path:
      return

    self.pex_path = ':'.join(merge_split(self.pex_path, pex_path))
```

bonus points if you provide a flag on `merge_split` to enable inline deduping.

pex/util.py
```diff
@@ -245,3 +245,15 @@ def iter_pth_paths(filename):
         if extras_dir_case_insensitive not in known_paths and os.path.exists(extras_dir):
           yield extras_dir
           known_paths.add(extras_dir_case_insensitive)
+
+
+def merge_split(*paths):
+  """
+  Merge paths into a single path delimited by colons and split on colons to return
```
kwlzn: the docstring should start on the opening line like:

```
"""Docstring starts here.

...
"""
```

otherwise it won't appear correctly in IDEs/`help()`/etc.

pex/util.py
```diff
@@ -245,3 +245,15 @@ def iter_pth_paths(filename):
         if extras_dir_case_insensitive not in known_paths and os.path.exists(extras_dir):
           yield extras_dir
           known_paths.add(extras_dir_case_insensitive)
+
+
+def merge_split(*paths):
+  """
```
kwlzn: a quick test for this in `test_util.py` would probably be good.

PR #424, opened 2017-10-10
Fix handling of pre-release option.

### LINKED ISSUE #423: The `--pre` option is not passed from command-line options down to API call
When a `--pre` option is used from command line it is stored on an instance of `ResolverOptionsBuilder` only, but that instance is not passed to the call to `resolve_multi()`.
Neither is its `_allow_prerelease` attribute passed to the keyword argument of `resolve_multi()`.
This results in the creation of a new `ResolverOptionsBuilder` instance in `resolve()` (called from `resolve_multi()`) and the exclusion of pre-release packages despite the `--pre` command-line option.

The specific case where this hidden behavior was discovered requires a dependency at a specific version (e.g., `dep==1.2.3b`), while another package requires the same dependency with no version specified (e.g., `dep`). The resulting error from pex is:
```
Could not satisfy all requirements for dep==1.2.3b1:
    dep==1.2.3b1, dep
```

I have a fix for this ready and will send a PR.

zvezdan:
Pex takes a `--pre` option from the command line, but it does not pass
it correctly during build. We need to propagate it to resolver to get
the correct behavior.

Fixes #423

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -565,7 +565,8 @@ def build_pex(args, options, resolver_option_builder):
                                 interpreters=interpreters,
                                 platforms=options.platform,
                                 cache=options.cache_dir,
-                                cache_ttl=options.cache_ttl)
+                                cache_ttl=options.cache_ttl,
+                                allow_prereleases=resolver_option_builder.prereleases_allowed)
 
       for dist in resolveds:
         log('  %s' % dist, v=options.verbosity)
--- a/pex/resolver_options.py
+++ b/pex/resolver_options.py
@@ -117,6 +117,31 @@ class ResolverOptionsBuilder(object):
         [precedent for precedent in self._precedence if precedent is not SourcePackage])
     return self
 
+  # TODO: Make this whole interface more Pythonic.
+  #
+  # This method would be better defined as a property allow_prereleases.
+  # Unfortunately, the existing method below already usurps the name allow_prereleases.
+  # It is an existing API that returns self as if it was written in an attempt to allow
+  # Java style chaining of method calls.
+  # Due to that return type, it cannot be used as a Python property setter.
+  # It's currently used in this manner:
+  #
+  #     builder.allow_prereleases(True)
+  #
+  # and we cannot change it into @allow_prereleases.setter and use in this manner:
+  #
+  #     builder.allow_prereleases = True
+  #
+  # without affecting the existing API calls.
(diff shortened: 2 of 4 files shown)


zvezdan:
Can someone please re-run the jobs.
Only PyPy and Python 3.4 jobs failed.
Python 3.4 practically did not start, while PyPy job failed in a way that does not seem related to this change.
Both of these jobs have this info at the top of the log:
```
This job ran on our Precise environment, which is in the process of being decommissioned.
Please read about the status of the rollout on the blog, and take note that you can
explicitly stay on Precise by specifying dist: precise in your .travis.yml.
```
Since all other language versions are passing I think that re-run or fix of Travis configuration may enable this change to pass the tests completely. Before submitting this PR, I tested with `tox` and tests for 2.7, 3.6, style, and isort all passed in my local environment. I do not have 3.4 or PyPy to test.

pex/bin/pex.py
```diff
@@ -565,7 +565,8 @@ def build_pex(args, options, resolver_option_builder):
                                 interpreters=interpreters,
                                 platforms=options.platform,
                                 cache=options.cache_dir,
-                                cache_ttl=options.cache_ttl)
+                                cache_ttl=options.cache_ttl,
+                                allow_prereleases=resolver_option_builder._allow_prereleases)
```
kwlzn: actually, one quick ask: this access of `_allow_prereleases` violates the private convention. would you mind exposing `_allow_prereleases` here as a public attribute of the `ResolverOptionsBuilder` class and using that instead? a quick `@property def allow_prereleases(): return self._allow_prereleases` would work.

zvezdan: Sure. That's easy. The only reason I did not do it initially is because I did not know what would be the code owner's preferences. I'll do it and resubmit.

zvezdan: Actually, I just realized that there was another reason. The existing method:
```
  def allow_prereleases(self, allowed):
    self._allow_prereleases = allowed
    return self
```
is not exactly a property setter, because it returns `self`.
I'll try to work around the existing interface and see what feels the best.

zvezdan:
@kwlzn I added the change and the comment describing why the property has a different name/API to-do note. If you feel I should take any of the new comments out of the patch, please let me know.
Thanks for the quick review!

kwlzn:
@zvezdan thanks - merged.

PR #427, opened 2017-10-20
Add interpreter constraints option and use constraints to search for compatible interpreters at exec time

CMLivingston:
Problem:
PEX does not support interpreter compatibility ranges for selecting an interpreter to build/run a PEX with. Twitter is aiming at killing off PEX_PYTHON overriding and replacing with a PEX_PYTHON_PATH in /etc/pexrc, which will contain a path to directories containing canonical interpreters for searching against at pex bootstrap re-exec time. Interpreter constraints on a PEX should inform the searching of PEX_PYTHON_PATH (if present in the env) and error out if no compatible interpreter can be found, replacing the searching of $PATH as a fallback.

Solution:
Add an interpreter constraints option to pex CLI and write these constraints to PEX-INFO metadata. Add an env variable to variables.py to support the setting of PEX_PYTHON_PATH. When the bootstrapper pex re-execs at runtime, read constraints from PEX-INFO and attempt to resolve an interpreter to exec with from the PPP. If no interpreter can be found in PPP, error out indicating that no match can be found. 

Result:
PEXes will be able to be built and executed with specified interpreter constraints. Users will be able to define and control interpreter selection in managed development environments by setting the PEX_PYTHON_PATH so that PEXes are executed with supported interpreters only.


pex/interpreter_constraints.py
```diff
@@ -0,0 +1,58 @@
+# Copyright 2014 Pants project contributors (see CONTRIBUTORS.md).
```
CMLivingston: This new library is extent of the intersecting interpreter constraints logic that makes sense to move out of Pants and into Pex. It will be cleanly consumable by Pants and the only modification I made is to add a meet_all_constraints kwarg for meeting all filters passed instead of at least one, which is backward compatible with the Pants use cases. 

I added a unit test for the function I wrote myself, parse_interpreter_constraints. I operated on the assumption that the other methods of this lib have been tested or are functioning properly so I did not add any new tests for them but I am more than happy to. I also realize tests in this changeset (pex_bootstrapper, interpreter_constraints tests) are relatively sparse but I figured I would post for review to get feedback before putting significant time into them.

pex/bin/pex.py
```diff
@@ -525,6 +531,11 @@ def build_pex(args, options, resolver_option_builder):
       for interpreter in options.python or [None]
     ]
 
+  if options.interpreter_constraints:
+    safe_constraints = options.interpreter_constraints.strip("'").strip('"')
```
CMLivingston: I added this to avoid issues with using < chars on the command line; this issue came up during integration testing.

pex/bin/pex.py
```diff
@@ -544,6 +555,8 @@ def build_pex(args, options, resolver_option_builder):
   pex_info.always_write_cache = options.always_write_cache
   pex_info.ignore_errors = options.ignore_errors
   pex_info.inherit_path = options.inherit_path
+  if options.interpreter_constraints:
```
CMLivingston: I elected to keep interpreter constraints as a single string here and in the pex info API (instead of a list of args) so that we can take constraints defined in the similar format in BUILD files and pipe them into a pex's pex info with minimal/no processing.

tests/test_integration.py
```diff
@@ -326,3 +328,114 @@ def test_pex_path_in_pex_info_and_env():
...
+      '--interpreter-constraints=">3"',
+      '-o', pex_out_path])
+    if sys.version_info[0] == 3:
+      res.assert_success()
+    else:
+      assert res.return_code == 102
+
+
+@pytest.mark.skipif("hasattr(sys, 'pypy_version_info')")
+def test_interpreter_resolution_with_pex_python_path():
```
CMLivingston: This test case was tricky to figure out given the nature of the test and the need for multiple system interpreters that could vary between systems. I'm not sure if this will work in CI or if there is a better way to go about testing with dependence on system Python binaries, but I gave it my best shot and I am open to feedback/further iteration.

CMLivingston: I changed the interpreters for testing from system-specific Python installs to the Python binaries located in .tox directory. Hopefully this cleans up CI failures due to missing interpreters (when using the PythonInterpreter.all() approach).

pex/interpreter_constraints.py
```diff
@@ -0,0 +1,90 @@
...
+    pex, using the Requirement-style format, e.g. ``'CPython>=3', or just ['>=2.7','<3']``
+    for requirements agnostic to interpreter class.
+  :param meet_all_constraints: whether to match against all filters.
+    Defaults to matching interpreters that match at least one filter.
+  """
+  for match in _matching(interpreters, filters, meet_all_constraints):
+    yield match
+
+
+def parse_interpreter_constraints(constraints_string):
```
CMLivingston: I added this `parse_interpreter_constraints` method + a test when I originally implemented the interpreter constraints PEX cli option as a single comma-separated string. Upon further iteration, I changed this option to be a list that accepts multiple re-uses of the same --interpreter-constraint=X flag to construct a list of constraints (similar to declaring --python=Y multiple times) and subsequently made this method obsolete. I kept it in the PR in case we find later on that it would be good to define interpreter constraints as a single string, but AFAIK that will not be the case and I will remove this.

pex/interpreter_constraints.py
```diff
@@ -0,0 +1,90 @@
...
+    return None
+  lowest = interpreters[0]
+  for i in interpreters[1:]:
+    lowest = lowest if lowest < i else i
+  return lowest
+
+
+def highest_version_interpreter(interpreters):
+  """Given a list of interpreters, return the one with the highest version.
+     Returns the leftmost interpreter if all versions are equal.
```
CMLivingston: `interpreter.py` currently only supports `__lt__`, so I kept the same version filtering functionality here but it selects equal version interpreters in reverse order than the previous `lowest_version_interpreter` that bin/pex.py has been using for interpreter selection. Can implement `__gt__` if we think its necessary.

pex/interpreter_constraints.py
```diff
@@ -0,0 +1,68 @@
+# Copyright 2014 Pants project contributors (see CONTRIBUTORS.md).
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+# A library of functions for filtering Python interpreters based on compatibility constraints
+
+from .interpreter import PythonIdentity
+
+
+def _matches(interpreter, filters, meet_all_constraints=False):
```
CMLivingston: I added meet_all_constraints because for some reason the Pants usage of this code (currently in Pants codebase) only requires a match against one of the filters to be considered a "matching interpreter".

pex/pex_bootstrapper.py
```diff
@@ -56,28 +61,100 @@ def find_in_path(target_interpreter):
...
+  """
+  compatible_interpreters = _find_compatible_interpreters(
+    pex_python_path, compatibility_constraints)
+
+  target = None
+  if compatible_interpreters:
+    target = min(compatible_interpreters).binary
+  if not target:
+    die('Failed to find compatible interpreter for constraints: %s'
+        % str(compatibility_constraints))
```
kwlzn: seems like this `die` should be raised `if not compatible_interpreters` vs `if not target`. e.g.

```
compatible_interpreters = ...
if not compatible_interpreters:
  die(...)
target = min(compatible_interpreters).binary
```

unless its possible for `min(...).binary` to also be `''` or `None`?

pex/pex_bootstrapper.py
```diff
@@ -56,28 +61,100 @@ def find_in_path(target_interpreter):
...
+    if pex_python_path:
+      TRACER.log('Detected PEX_PYTHON_PATH, re-exec to %s' % target)
+    else:
+      TRACER.log('Re-exec to interpreter %s that matches constraints: %s' % (target,
+        str(compatibility_constraints)))
+    ENV.SHOULD_EXIT_BOOTSTRAP_REEXEC = True
+    os.execve(target, [target] + sys.argv, ENV.copy())
+
+
+def maybe_reexec_pex(compatibility_constraints=None):
```
kwlzn: this probably shouldn't default to `None`.

pex/pex_bootstrapper.py
```diff
@@ -56,28 +61,100 @@ def find_in_path(target_interpreter):
...
+  if ENV.SHOULD_EXIT_BOOTSTRAP_REEXEC:
+    return
+  if ENV.PEX_PYTHON and ENV.PEX_PYTHON_PATH:
+    # both vars are defined, fall through to PEX_PYTHON_PATH resolution (give precedence to PPP)
+    pass
+  elif ENV.PEX_PYTHON:
+    # preserve PEX_PYTHON re-exec for backwards compatibility
```
kwlzn: cleaner?

```
if ENV.PEX_PYTHON and not ENV.PEX_PYTHON_PATH:
  _handle_pex_python(...)
```

pex/pex_bootstrapper.py
```diff
@@ -56,28 +61,100 @@ def find_in_path(target_interpreter):
...
+    _handle_pex_python(ENV.PEX_PYTHON, compatibility_constraints)
+
+  if not compatibility_constraints:
+    # if no compatibility constraints are specified, we want to match against
+    # the lowest-versioned interpreter in PEX_PYTHON_PATH if it is set
+    if not ENV.PEX_PYTHON_PATH:
+      # no PEX_PYTHON_PATH, PEX_PYTHON, or interpreter constraints, continue as normal
+      return
+
+  _handle_general_interpreter_selection(ENV.PEX_PYTHON_PATH, compatibility_constraints)
```
kwlzn: it might be better/more readable to make the `_handle*` functions simply return a python interpreter path vs directly re-exec - and then you could have only one `os.execve` at the bottom of this function to actually re-exec against the selected interpreter.

this whole function could probably then be simplified to something along the lines of:

```
def maybe_reexec_pex(...):
  if ENV.SHOULD_EXIT_BOOTSTRAP_REEXEC:
    return

  selected_interpreter = None
  if ENV.PEX_PYTHON and not ENV.PEX_PYTHON_PATH:
    selected_interpreter = _select_pex_python_interpreter(...)
  elif ENV.PEX_PYTHON_PATH:
    selected_interpreter = _select_interpreter(...)

  if selected_interpreter:
    ENV.delete('PEX_PYTHON')
    ENV.delete('PEX_PYTHON_PATH')
    ENV.SHOULD_EXIT_BOOTSTRAP_REEXEC = True
    os.execve(selected_interpreter, [selected_interpreter] + sys.argv[1:], ENV.copy())  
```

pex/pex_bootstrapper.py
```diff
@@ -56,28 +61,100 @@ def find_in_path(target_interpreter):
...
+
+
 def bootstrap_pex(entry_point):
   from .finders import register_finders
   register_finders()
-  maybe_reexec_pex()
+  pex_info = get_pex_info(entry_point)
+  with TRACER.timed('Bootstrapping pex with maybe_reexec_pex', V=2):
```
kwlzn: the `get_pex_info` here should probably fall under the `TRACER.timed` block.

and a more concise/descriptive string here for the timer might be "Interpreter selection", probably with `V=3` to distinguish it in debug logging.

pex/pex_info.py
```diff
@@ -274,11 +289,13 @@ def update(self, other):
       raise TypeError('Cannot merge a %r with PexInfo' % type(other))
     self._pex_info.update(other._pex_info)
     self._distributions.update(other.distributions)
+    self._interpreter_constraints.extend(other.interpreter_constraints)
```
kwlzn: this should probably dedupe somehow, perhaps via internal storage as a `set`?

pex/testing.py
```diff
@@ -277,3 +278,25 @@ def combine_pex_coverage(coverage_file_iter):
 
   combined.write()
   return combined.filename
+
+
+def bootstrap_python_installer():
+  install_location = os.getcwd() + '/.pyenv_test'
```
kwlzn: `os.path.join(os.getcwd(), '.pyenv_test')`

pex/testing.py
```diff
@@ -277,3 +278,25 @@ def combine_pex_coverage(coverage_file_iter):
...
+  if not os.path.exists(install_location) or not os.path.exists(install_location + '/bin/pyenv'):
```
kwlzn: ditto.

pex/testing.py
```diff
@@ -277,3 +278,25 @@ def combine_pex_coverage(coverage_file_iter):
...
+    else:
+      raise RuntimeError("Helper method could not clone pyenv from git")
+
+
+def ensure_python_interpreter(version):
+  bootstrap_python_installer()
+  install_location = os.getcwd() + '/.pyenv_test/versions/' + version
+  if not os.path.exists(install_location):
+    os.environ['PYENV_ROOT'] = os.getcwd() + '/.pyenv_test'
+    subprocess.call([os.getcwd() + '/.pyenv_test/bin/pyenv', 'install', version])
```
kwlzn: `os.path.join` throughout this function.

pex/variables.py
```diff
@@ -232,6 +232,18 @@ def PEX_PYTHON(self):
     return self._get_string('PEX_PYTHON', default=None)
 
   @property
+  def PEX_PYTHON_PATH(self):
+    """String
+
+    A colon-seperated string containing paths to binaires of blessed Python interpreters
```
kwlzn: `binaries`

pex/variables.py
```diff
@@ -300,6 +312,27 @@ def PEX_IGNORE_RCFILES(self):
...
+    This is neccesary because that function relies on checking against variables present in the
+    ENV to determine whether to re-exec or return to current execution. When we introduced
+    interpreter constraints to a pex, these constraints can influence interpreter selection without
+    the need for a PEX_PYTHON or PEX_PYTHON_PATH ENV variable set. Since re-exec previously checked
+    for PEX_PYTHON or PEX_PYTHON_PATH but not constraints, this would result in a loop with no
+    stopping criteria. Setting SHOULD_EXIT_BOOTSTRAP_REEXEC will let the runtime know to break out
+    of a second execution of maybe_reexec_pex if neither PEX_PYTHON or PEX_PYTHON_PATH are set but
+    interpreter constraints are specified.
+    """
+    value = self._environ.get('SHOULD_EXIT_REEXEC', 'False')
```
kwlzn: would expect the environ var to map 1:1 to the property name here.

pex/variables.py
```diff
@@ -300,6 +312,27 @@ def PEX_IGNORE_RCFILES(self):
...
+    return False if value == 'False' else True
```
kwlzn: would be better to not rely on the string value here as anything other than indicator of truthyness, i.e.:

```
return bool(self._environ.get(..., False))
```

kwlzn:
noting that this also seems to be failing on the `pypy` shard after a few restarts.

pex/interpreter_constraints.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def check_requirements_are_well_formed(constraints):
+  # Check that the compatibility requirements are well-formed.
+  for req in constraints:
+    try:
+      PythonIdentity.parse_requirement(req)
+    except ValueError as e:
+      from .common import die
+      die("Compatibility requirements are not formatted properly: %s" % str(e))
```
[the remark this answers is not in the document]
CMLivingston: The ValueError raised by the exception covers this.

pex/interpreter_constraints.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def matched_interpreters(interpreters, filters, meet_all_constraints=False):
+  """Given some filters, yield any interpreter that matches at least one of them, or all of them
+     if meet_all_constraints is set to True.
+
+  :param interpreters: a list of PythonInterpreter objects for filtering
+  :param filters: A sequence of strings that constrain the interpreter compatibility for this
+    pex, using the Requirement-style format, e.g. ``'CPython>=3', or just ['>=2.7','<3']``
+    for requirements agnostic to interpreter class.
```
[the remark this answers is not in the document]
CMLivingston: Sure, filter is used due to the precedence set by Pants. The interpreter constraints lib is forklifted out of Pants and readily consumable if we elect to refactor usages in Pants.

pex/pex_info.py
CMLivingston: Python sets are not JSON serializable but I can use sets to reduce duplications and cast them as lists.

pex/bin/pex.py
```diff
@@ -525,6 +529,12 @@ def build_pex(args, options, resolver_option_builder):
       for interpreter in options.python or [None]
     ]
 
+  if options.interpreter_constraint:
+    constraints = options.interpreter_constraint
+    # TODO: add check to see if constraints are mutually exclusive (bad) so no time is wasted
+    check_requirements_are_well_formed(constraints)
+    interpreters = list(matched_interpreters(interpreters, constraints, meet_all_constraints=True))
+
```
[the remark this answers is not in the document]
CMLivingston: I like the idea of a combination of 1 and 2. Disallow both options as using both seems like an anti-pattern and fall back to PPP/PATH when --constraints is passed without --python.

pex/pex_bootstrapper.py
```diff
@@ -56,28 +61,95 @@ def find_in_path(target_interpreter):
...
 
-def maybe_reexec_pex():
-  from .variables import ENV
-  if not ENV.PEX_PYTHON:
-    return
+def _find_compatible_interpreters(pex_python_path, compatibility_constraints):
+  """Find all compatible interpreters on the system within the supplied constraints and use
+     PEX_PYTHON_PATH env variable if it is set. If not, fall back to interpreters on $PATH.
+  """
+  if pex_python_path:
```
UnrememberMe: same codes as pex/bin/pex.py? If so, we should extract this to a library.

pex/interpreter_constraints.py
```diff
@@ -0,0 +1,42 @@
...
+     if meet_all_constraints is set to True.
+
+  :param interpreters: a list of PythonInterpreter objects for filtering
+  :param constraints: A sequence of strings that constrain the interpreter compatibility for this
+    pex, using the Requirement-style format, e.g. ``'CPython>=3', or just ['>=2.7','<3']``
+    for requirements agnostic to interpreter class.
+  :param meet_all_constraints: whether to match against all filters.
+    Defaults to matching interpreters that match at least one filter.
+  """
+  for match in _matching(interpreters, constraints, meet_all_constraints):
```
UnrememberMe: Seems like we can simplify to 
```
def matched_interpreters(interpreters, constraints, meet_all_constraints=False):
  _matching(interpreters, constraints, meet_all_constraints=meet_all_constraints)
```

tests/test_pex_bootstrapper.py
```diff
@@ -28,3 +28,22 @@ def test_get_pex_info():
...
+  assert interpreters[0].binary == pex_python_path.split(':')[1]
+
+  interpreters = find_compatible_interpreters(pex_python_path, ['<3'])
+  assert interpreters[0].binary == pex_python_path.split(':')[0]
+
+  interpreters = find_compatible_interpreters(pex_python_path, ['<2'])
+  assert not interpreters
+
+  interpreters = find_compatible_interpreters(pex_python_path, ['>4'])
+  assert not interpreters
```
kwlzn: how about adding a few more permutations here for more granular checking based around real world cases, like  a 2 vs 3 split or selecting '2.7.10' out of a PATH with '2.7.9', '2.7.10', and '2.7.11' etc?

tests/test_integration.py
```diff
@@ -326,3 +328,217 @@ def test_pex_path_in_pex_info_and_env():
...
+  with temporary_dir() as td:
+    pex_out_path = os.path.join(td, 'pex.pex')
+    res = run_pex_command(['--disable-cache',
+      '-o', pex_out_path])
+    res.assert_success()
+
+    stdin_payload = b'import sys; print(sys.executable); sys.exit(0)'
+    stdout, rc = run_simple_pex(pex_out_path, stdin=stdin_payload)
+    assert rc == 0
+    assert str(sys.executable).encode() in stdout
```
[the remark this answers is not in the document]
CMLivingston: Unfortunately, for the py36-requests env, that errors out with `TypeError: a bytes-like object is required, not 'str'`

pex/pex_bootstrapper.py
```diff
@@ -56,28 +61,93 @@ def find_in_path(target_interpreter):
...
+    interpreters = []
+    for binary in pex_python_path.split(os.pathsep):
+      try:
+        interpreters.append(PythonInterpreter.from_binary(binary))
+      except Executor.ExecutionError:
+        pass
```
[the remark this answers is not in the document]
CMLivingston: sgtm

pex/pex_info.py
CMLivingston: sounds good

tests/test_pex_bootstrapper.py
CMLivingston: I believe this is already testing filtering based on a 2/3 split here, right? The current PPP has both 2 and 3 in it. I can certainly add testing for various other minor versions though.

tests/test_integration.py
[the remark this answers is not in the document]
kwlzn: gotcha.

tests/test_pex_bootstrapper.py
kwlzn: thinking more along the lines of fully exercising all 3 fields of `major.minor.patch` here vs just major versions.

pex/bin/pex.py
```diff
@@ -525,6 +530,28 @@ def build_pex(args, options, resolver_option_builder):
...
+  # NB: this block is neccesary to load variables from the pexrc runtime configuration.
+  # It is only neccesary to cover the case where the pexrc file is located in the same
+  # directory as the output pex. The other supported pexrc locations (/etc/pexrc) do not
+  # require special treatment for reading at build time.
+  if options.pex_name:
+    if '/' in options.pex_name:  # the output pex is not in the working directory
+      pexrc_dir = os.path.dirname(options.pex_name)
+    else:
+      pexrc_dir = os.getcwd()
+  else:
```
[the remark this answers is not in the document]
CMLivingston: So making this change actually broke the ability to read from the pexrc at runtime. Seems like the author expects .pexrc to be in the same directory as the pex that is running, which makes sense to me, at least for directory-level configurations. 

The runtime breakage has to do with os.path.dirname(sys.argv[0]) being the location of the running pex which is different than os.getcwd() at runtime (the root of the pex project I am running these tests in).  I think this could be a quirk of running all of this in pytest, however it might make sense to leave it as is because if we change to getcwd() it will require users to either have a .pexrc from wherever they decide to execute a pex from (which could be annoyling/problematic if absolute/long relative paths are used), or always run a pex from the directory the pex is in, with the expectation that the .pexrc is in it. 

Given that getcwd() broke the runtime config reading, I think I should leave it as is and make an issue + TODO to circle back, as I think this is beyond the scope of these changes and is more specific to the case where .pexrc is used. Thoughts?

pex/variables.py
```diff
@@ -33,11 +33,34 @@ def iter_help(cls):
       variable_type, variable_text = cls.process_pydoc(getattr(value, '__doc__'))
       yield variable_name, variable_type, variable_text
 
+  @classmethod
+  def from_rc(cls, rc):
+    """Read pex runtime configuration variables from a pexrc file.
+
+       :param rc: an absolute path to a pexrc file
+       :return ret_vars: a dict object containing key values pairs found in processed pexrc files
```
kwlzn: these and docstring continuations should indent to align with the opening `"""`, e.g.

```
"""The docstring summary which might
spill onto a new line.

Extended summary would go here.

:param etc: Would go here.
"""
```

pex/bin/pex.py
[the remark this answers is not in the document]
kwlzn: makes sense. I think whatever the case tho, this block of code here:

```
+  # NB: this block is necessary to load variables from the pexrc runtime configuration.
 +  # It is only necessary to cover the case where the pexrc file is located in the same
 +  # directory as the output pex. The other supported pexrc locations (/etc/pexrc) do not
 +  # require special treatment for reading at build time.
 +  if options.pex_name:
 +    if '/' in options.pex_name:  # the output pex is not in the working directory
 +      pexrc_dir = os.path.dirname(options.pex_name)
 +    else:
 +      pexrc_dir = os.getcwd()
 +    pexrc = os.path.join(pexrc_dir, '.pexrc')
 +  else:
 +    pexrc = '.pexrc'
```

feels pretty hacky and like it's largely intended to make tests pass. to me, this sort of logic belongs in `Variables` instead so that the behavior of PP/PPP discovery is uniform across all callers. this block also doesn't consider `~/.pexrc` like `Variables` as used today does and will look at `.pexrc` 2x in the case that `options.pex_name` isn't defined. I'm also not quite sure why we would consider the `.pexrc` relative to a new output path.

so as long as we can clean that aspect up, I don't feel strongly about `os.getcwd()` vs `os.path.dirname(sys.argv[0])`.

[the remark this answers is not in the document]
CMLivingston: Ok. I made some changes that I think might be along the line of what you are thinking of, however please let me know if theres a better way to do this. This was as simple as I could manage. 

The .pexrc relative to output path is for when you have an absolute path to the output pex specified for the `-o /path/to/pex/pex` flag of the PEX CLI. This hacky-ness of this is largely due to the fact that I am trying to process rc files that are intended to be used at runtime only. This grew out of the existing integration tests placing the pexrc files in the same directory as the output pex via `os.path.join`. I think it still makes sense to have this in place for the case where the output pex is not the cwd.

Anyhow, I took all of that out of bin/pex.py and moved it into `Variables`.

pex/variables.py
```diff
@@ -34,14 +34,26 @@ def iter_help(cls):
...
+    :return ret_vars: a dict object containing key values pairs found in processed pexrc files
     """
     ret_vars = {}
-    for filename in ['/etc/pexrc', rc, os.path.join(os.path.dirname(sys.argv[0]), '.pexrc')]:
+    rc_locations = ['/etc/pexrc',
+                    rc,
+                    os.path.join(os.path.dirname(sys.argv[0]), '.pexrc'),
+                    os.getcwd()]
+    if not rc:
+      rc_locations.append('~/.pexrc')
```
CMLivingston: The only time `~/.pexrc` is the default is if we are calling the Variables constructor; the `from_rc` api requires a string so I added support for adding `~/.pexrc` if the input is blank. I suppose I could make it a default arg or pass it in in PEX CLI...

CMLivingston: Added it as default arg.

kwlzn: seems to me like `~/.pexrc` needs to be considered in all cases (i.e. as a static member of `rc_locations`) or else we aren't reading from all of the places that pexrc configuration can exist. (essentially, I don't see any existing supporting usage of `Variables` in pex/pants to warrant keeping the `rc` param here the way it is).

pex/variables.py
```diff
@@ -33,11 +33,46 @@ def iter_help(cls):
...
+    read a pexrc at build time in the case where a pexrc is located in the same directory as a the
+    output pex.
+
+    :return ret_vars: a dict object containing key values pairs found in processed pexrc files
+    """
+    ret_vars = {}
+    rc_locations = ['/etc/pexrc',
+                    rc,
+                    os.path.join(os.path.dirname(sys.argv[0]), '.pexrc'),
+                    os.getcwd()]
```
kwlzn: afaict, this should reference an rcfile, not the current working dir directly?

but maybe it's not actually needed?

pex/bin/pex.py
```diff
@@ -525,6 +530,16 @@ def build_pex(args, options, resolver_option_builder):
       for interpreter in options.python or [None]
     ]
 
+  if options.interpreter_constraint:
+    # Overwrite the current interpreter as defined by sys.executable.
+    # NB: options.python and interpreter constraints cannot be used together, so this will not
+    # affect usages of the interpreter(s) specified by the "--python" command line flag.
+    constraints = options.interpreter_constraint
+    validate_constraints(constraints)
+    rc_variables = Variables.from_rc(build_time_rc_dir=os.path.dirname(options.pex_name))
```
kwlzn: its still puzzling to me why we need to look at a `.pexrc` relative to the output pex name for `PPP` configuration at build time.. can you clarify that with a comment here?

if this is purely for tests, we should probably re-evaluate.

pex/variables.py
```diff
@@ -231,6 +250,18 @@ def PEX_PYTHON(self):
     """
     return self._get_string('PEX_PYTHON', default=None)
 
+  @property
+  def PEX_PYTHON_PATH(self):
+    """String
+
+    A colon-separated string containing binaires of blessed Python interpreters
```
kwlzn: `s/binaires/paths/`

CMLivingston: I will remove `rc`.

pex/bin/pex.py
CMLivingston: The use case is when the output pex is not in cwd, i.e. `-o dir/output.pex`. In that case, we should be supporting the case where the .pexrc is in `dir` as per a precedent set in the tests, however I see this being practical beyond testing situations. 

Previously, the check was for os.path.dirname(sys.argv[0]) to find this .pexrc in the same dir as the output pex at runtime, which works fine because sys.argv[0] is the pex itself, however, at build time, sys.argv[0] is the pex cli binary, which can be different than the dir of the output pex i.e. `pex requests -o dir/output.pex`. In this case, where I have .pexrc in `dir`, special accommodations are needed to make sure that the pex build routine can find that pexrc at build time. 

Like you say, this precedent is set largely by the tests, however having a pexrc in the same location as the pex itself seems like a reasonable case to cover to me.

pex/pex_bootstrapper.py
```diff
@@ -56,28 +61,105 @@ def find_in_path(target_interpreter):
...
+    interpreters, compatibility_constraints, meet_all_constraints=True))
 
-  target_python = ENV.PEX_PYTHON
+
+def _select_pex_python_interpreter(target_python, compatibility_constraints):
   target = find_in_path(target_python)
+
   if not target:
     die('Failed to find interpreter specified by PEX_PYTHON: %s' % target)
+  elif compatibility_constraints:
```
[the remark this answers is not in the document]
CMLivingston: the `e` string contains this requirement.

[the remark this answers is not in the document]
CMLivingston: `_matching` got folded into the `matched_interpreters` method so I will remove the method.

pex/pex_bootstrapper.py
```diff
@@ -56,28 +61,105 @@ def find_in_path(target_interpreter):
...
+
+  """
+  if ENV.SHOULD_EXIT_BOOTSTRAP_REEXEC:
+    return
+
+  selected_interpreter = None
+  with TRACER.timed('Selecting runtime interpreter based on pexrc', V=3):
+    if ENV.PEX_PYTHON and not ENV.PEX_PYTHON_PATH:
+      # preserve PEX_PYTHON re-exec for backwards compatibility
+      selected_interpreter = _select_pex_python_interpreter(ENV.PEX_PYTHON,
```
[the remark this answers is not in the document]
CMLivingston: I agree about the simplification. We discussed leaving it in as a way to give us flexibility when migrating the /etc/pexrc configs in prod and in MDE, as well as provide backwards interoperability for pex users who want to continue using a single python interpreter. However, IIRC, I think the main reason was to give us flexibility with the whole migration of /etc/pexrc and to give us a safety net in case we find that PEX_PYTHON_PATH ends up disastrously breaking things or causing other unexpected problems. 

I like the idea of a safety net and backwards compatibility (at least for now), and I think it wouldn't be a bad idea to keep it in for now until we have success using PPP internally, maybe with a reasonable subset of Python users successfully running and deploying py3 PEX both locally and in prod? What do you think? @UnrememberMe any thoughts here?

pex/variables.py
CMLivingston: I think it is needed for the case where the pex you are invoking and the .pexrc is in the cwd.
Also, idk why the `pi = interpreter or PythonInterpreter.get()` line was already changed to that, I am not sure why this reverted, but I will re-add, sorry about that.

pex/bin/pex.py
```diff
@@ -289,6 +291,25 @@ def configure_clp_pex_environment(parser):
            'can be passed multiple times to create a multi-interpreter compatible pex. '
            'Default: Use current interpreter.')
 
+  group.add_option(
+    '--interpreter-constraint',
+    dest='interpreter_constraint',
```
UnrememberMe: IIRC, a-b-c will be automatically convert to a_b_c for variable names, so you don't necessarily need to specify dest here.

pex/bin/pex.py
```diff
@@ -289,6 +291,25 @@ def configure_clp_pex_environment(parser):
...
+    default=[],
+    type='str',
+    action='append',
+    help='A constraint that determines the interpreter compatibility for '
+         'this pex, using the Requirement-style format, e.g. "CPython>=3", or ">=2.7" '
+         'for requirements agnostic to interpreter class. This option can be passed multiple '
+         'times.')
+
+  group.add_option(
+    '--rcfile',
```
UnrememberMe: suggestion: for consistency, change option to` --rc-file`

pex/interpreter_constraints.py
```diff
@@ -0,0 +1,42 @@
...
+
+def validate_constraints(constraints):
+  # TODO: add check to see if constraints are mutually exclusive (bad) so no time is wasted
+  # Check that the compatibility requirements are well-formed.
+  for req in constraints:
+    try:
+      PythonIdentity.parse_requirement(req)
+    except ValueError as e:
+      die("Compatibility requirements are not formatted properly: %s" % str(e))
+
```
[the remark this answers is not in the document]
UnrememberMe: second. Including the failed string will help user to debug.

pex/pex_bootstrapper.py
```diff
@@ -56,28 +61,107 @@ def find_in_path(target_interpreter):
...
     ENV.delete('PEX_PYTHON')
-    os.execve(target, [target_python] + sys.argv, ENV.copy())
+    ENV.delete('PEX_PYTHON_PATH')
+    ENV.SHOULD_EXIT_BOOTSTRAP_REEXEC = True
+    cmdline = [selected_interpreter] + sys.argv[1:]
+    TRACER.log('Re-executing: cmdline="%s", sys.executable="%s", PEX_PYTHON="%s", '
+               'PEX_PYTHON_PATH="%s", COMPATIBILITY_CONSTRAINTS="%s"'
+               % (cmdline, sys.executable, ENV.PEX_PYTHON, ENV.PEX_PYTHON_PATH,
+                  compatibility_constraints))
+    os.execve(selected_interpreter, cmdline, ENV.copy())
```
UnrememberMe: won't this become `MyPython MyPython original_params` since on line 152, cmdline has already prepended selected_interpreter?

pex/interpreter_constraints.py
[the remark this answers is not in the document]
CMLivingston: the `e` string contains this requirement. Sorry if this comment shows up elsewhere, I have no clue why it is not persisting in github.

pex/pex_bootstrapper.py
CMLivingston: the first arg of `cmdline` is ignored in the os.execve API (not sure why exactly, but it is).

pex/interpreter_constraints.py
lgvital: Should this get bumped to 2017?

pex/interpreter_constraints.py
```diff
@@ -0,0 +1,37 @@
...
+# A library of functions for filtering Python interpreters based on compatibility constraints
+
+from .common import die
+from .interpreter import PythonIdentity
+from .tracer import TRACER
+
+
+def validate_constraints(constraints):
+  # TODO: add check to see if constraints are mutually exclusive (bad) so no time is wasted
```
lgvital: Did this TODO exist already when you moved it? If not, can we start an issue for it to track?

pex/interpreter_constraints.py
```diff
@@ -0,0 +1,37 @@
...
+def matched_interpreters(interpreters, constraints, meet_all_constraints=False):
+  """Given some filters, yield any interpreter that matches at least one of them, or all of them
+     if meet_all_constraints is set to True.
+
+  :param interpreters: a list of PythonInterpreter objects for filtering
+  :param constraints: A sequence of strings that constrain the interpreter compatibility for this
+    pex, using the Requirement-style format, e.g. ``'CPython>=3', or just ['>=2.7','<3']``
+    for requirements agnostic to interpreter class.
+  :param meet_all_constraints: whether to match against all filters.
+    Defaults to matching interpreters that match at least one filter.
```
lgvital: Should we add a line to this docstring for what exactly we're yielding?

pex/pex_bootstrapper.py
```diff
@@ -56,28 +61,107 @@ def find_in_path(target_interpreter):
...
+
+  """
+  if ENV.SHOULD_EXIT_BOOTSTRAP_REEXEC:
+    return
+
+  selected_interpreter = None
+  with TRACER.timed('Selecting runtime interpreter based on pexrc', V=3):
+    if ENV.PEX_PYTHON and not ENV.PEX_PYTHON_PATH:
+      # preserve PEX_PYTHON re-exec for backwards compatibility
+      # TODO: Kill this off completely in favor of PEX_PYTHON_PATH
```
lgvital: Start an issue for this TODO?

pex/pex_info.py
```diff
@@ -123,6 +123,8 @@ def __init__(self, info=None):
                        '%s of type %s' % (info, type(info)))
     self._pex_info = info or {}
     self._distributions = self._pex_info.get('distributions', {})
+    # cast as set because pex info from json must store interpreter_constraints as a list
```
lgvital: Not sure I understand this comment, or maybe it's a typo. You're casting it as a set, but interpreter constraints needs to be stored as a list?

CMLivingston: Reading in a pex info using `from_pex` or `from_json` will instantiate a PexInfo object with json loaded as a dict. Interpreter constraints get serialized to json as a list b.c. json does not support Python sets, so casting it here is to cover the case where we read in a populated pex-info from json (which will be a list), however we want to store it internally as a set.

pex/pex_bootstrapper.py
CMLivingston: It is on the line below this one.

pex/interpreter_constraints.py
CMLivingston: ✅

pex/interpreter_constraints.py
CMLivingston: ✅

pex/variables.py
```diff
@@ -33,11 +33,40 @@ def iter_help(cls):
       variable_type, variable_text = cls.process_pydoc(getattr(value, '__doc__'))
       yield variable_name, variable_type, variable_text
 
-  def __init__(self, environ=None, rc='~/.pexrc', use_defaults=True):
+  @classmethod
+  def from_rc(cls, rc=None):
+    """Read pex runtime configuration variables from a pexrc file.
+
+    :param rc: an absolute path to a pexrc file.
+    :return ret_vars: a dict object containing key values pairs found in processed pexrc files
```
kwlzn: specifying an arg name for a return doesn't make sense.

should look like:

```
:return: A dict of key value pairs found in processed pexrc files.
:rtype: dict
```

pex/variables.py
```diff
@@ -33,11 +33,40 @@ def iter_help(cls):
...
+    """
+    ret_vars = {}
+    rc_locations = ['/etc/pexrc',
+                    '~/.pexrc',
+                    os.path.join(os.path.dirname(sys.argv[0]), '.pexrc'),
+                    os.path.join(os.getcwd(), '.pexrc')]
```
kwlzn: did this actually end up being needed given the new `--rcfile` flag? if so, why?

seems to me like tests were passing without it before?

tests/test_integration.py
```diff
@@ -326,3 +328,224 @@ def test_pex_path_in_pex_info_and_env():
...
+    assert str(pex_python_path.split(':')[0]).encode() in stdout
+
+
+def test_pex_python():
+  ensure_python_interpreter('2.7.10')
+  ensure_python_interpreter('3.6.3')
+  with temporary_dir() as td:
+    pexrc_path = os.path.join(td, '.pexrc')
+    with open(pexrc_path, 'w') as pexrc:
+      pex_python = os.getcwd() + '/.pyenv_test/versions/3.6.3/bin/python3.6'
```
kwlzn: this should get the path from `ensure_python_interpreter` vs hardcoding.

tests/test_integration.py
```diff
@@ -326,3 +328,224 @@ def test_pex_path_in_pex_info_and_env():
...
+    stdin_payload = b'import sys; print(sys.executable); sys.exit(0)'
+    stdout, rc = run_simple_pex(pex_out_path, stdin=stdin_payload)
+    assert rc == 0
+    correct_interpreter_path = pex_python.encode()
+    assert correct_interpreter_path in stdout
+
+    # test PEX_PYTHON with incompatible constraints
+    pexrc_path = os.path.join(td, '.pexrc')
+    with open(pexrc_path, 'w') as pexrc:
+      pex_python = os.getcwd() + '/.pyenv_test/versions/2.7.10/bin/python2.7'
```
kwlzn: this should get the path from `ensure_python_interpreter` vs hardcoding.

tests/test_variables.py
```diff
@@ -84,27 +85,30 @@ def test_pex_get_kv():
 
 
 def test_pex_from_rc():
-  with named_temporary_file(mode='w') as pexrc:
+  with open('.pexrc', 'w') as pexrc:
```
kwlzn: I think the `named_temporary_file` before was slightly safer?

PR #441, opened 2018-01-17
Change PEX re-exec variable from ENV to os.environ

CMLivingston:
Currently, if an application packaged as a pex builds and runs a product pex as part of its functionality (i.e. a Pants pex that builds and runs pexes), the interpreter selection logic in a product pex will short-circuit because the flag telling it to do so has already been set in the initial bootstrapping logic of the main app's pex. Consequently, when the main app runs a product pex as part of its functionality, the interpreter selection in accordance with a pexrc for the product pex is skipped because the environment (`os.environ`) now contains the `SHOULD_EXIT_BOOTSTRAP_REEXEC` flag set to `"True"` from the bootstrap phase of the main app's pex. 

This is because `SHOULD_EXIT_BOOTSTRAP_REEXEC` (the flag governing whether or not to re-exec) was being set in an ENV object variable, which is composed into the environment of the re-execution call `os.execve` via the `env` parameter.

The global instance of ENV is unique to each isolated Pex environment, but for this use case we need environment data that can be shared across Pex environments. This changes in this PR switch the implementation of `SHOULD_EXIT_BOOTSTRAP_REEXEC` from ENV to os.environ, where changes to this variable will be visible across different pex environments running from the same Python process.

### DIFF
--- a/pex/pex_bootstrapper.py
+++ b/pex/pex_bootstrapper.py
@@ -113,7 +113,7 @@ def _select_interpreter(pex_python_path, compatibility_constraints):
   # TODO: https://github.com/pantsbuild/pex/issues/430
   target = min(compatible_interpreters).binary
 
-  if os.path.exists(target) and os.path.realpath(target) != os.path.realpath(sys.executable):
+  if os.path.exists(target):
     return target
 
 
@@ -133,7 +133,9 @@ def maybe_reexec_pex(compatibility_constraints):
   Python interpreter to re-exec this pex with.
 
   """
-  if ENV.SHOULD_EXIT_BOOTSTRAP_REEXEC:
+  if os.environ.get('SHOULD_EXIT_BOOTSTRAP_REEXEC'):
+    # We've already been here and selected an interpreter. Continue to execution.
+    del os.environ['SHOULD_EXIT_BOOTSTRAP_REEXEC']
     return
 
   selected_interpreter = None
@@ -147,7 +149,7 @@ def maybe_reexec_pex(compatibility_constraints):
     elif ENV.PEX_PYTHON_PATH:
       selected_interpreter = _select_interpreter(ENV.PEX_PYTHON_PATH, compatibility_constraints)
 
-  if selected_interpreter:
+  if selected_interpreter and os.path.realpath(selected_interpreter) != os.path.realpath(sys.executable):
     cmdline = [selected_interpreter] + sys.argv
     TRACER.log('Re-executing: cmdline="%s", sys.executable="%s", PEX_PYTHON="%s", '
                'PEX_PYTHON_PATH="%s", COMPATIBILITY_CONSTRAINTS="%s"'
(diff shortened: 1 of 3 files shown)


pex/pex_bootstrapper.py
```diff
@@ -133,7 +133,9 @@ def maybe_reexec_pex(compatibility_constraints):
   Python interpreter to re-exec this pex with.
 
   """
-  if ENV.SHOULD_EXIT_BOOTSTRAP_REEXEC:
+  if os.environ.get('SHOULD_EXIT_BOOTSTRAP_REEXEC'):
```
kwlzn: should be able to `os.environ.pop('XX', None)` to retrieve + del in a single call here.

tests/test_integration.py
```diff
@@ -567,3 +567,95 @@ def test_entry_point_targeting():
 
     stdout, rc = run_simple_pex(pex_out_path)
     assert 'usage: autopep8'.encode() in stdout
+
+
+def test_interpreter_selection_using_os_environ_for_bootstrap_reexec():
+  """
+  This test is a special test meant to verify proper function of the
```
kwlzn: "This test is a special test" ?

tests/test_integration.py
```diff
@@ -567,3 +567,95 @@ def test_entry_point_targeting():
...
+    # Select pexrc interpreter versions based on test environemnt.
+    # The parent interpreter is the interpreter we expect the parent pex to 
+    # execute with. The child interpreter is the interpreter we expect the 
+    # child pex to execute with. 
+    if (sys.version_info[0], sys.version_info[1]) == (3, 6):
+      parent_interpreter_version = '3.6.2'
+    else:
+      parent_interpreter_version = '2.7.10'
+    if (sys.version_info[0], sys.version_info[1]) == (3, 6):
+      child_interpreter_version = '3.6.3'
```
kwlzn: could probably combine these into one block like:

```
if (sys.version_info[0], sys.version_info[1]) == (3, 6):
  parent_interp...
  child_interp...
else:
  parent..
  child...
```

since they both use the same if clause.

pex/pex_bootstrapper.py
```diff
@@ -133,30 +133,31 @@ def maybe_reexec_pex(compatibility_constraints):
...
+    cmdline = [target] + sys.argv
     TRACER.log('Re-executing: cmdline="%s", sys.executable="%s", PEX_PYTHON="%s", '
                'PEX_PYTHON_PATH="%s", COMPATIBILITY_CONSTRAINTS="%s"'
                % (cmdline, sys.executable, ENV.PEX_PYTHON, ENV.PEX_PYTHON_PATH,
                   compatibility_constraints))
     ENV.delete('PEX_PYTHON')
     ENV.delete('PEX_PYTHON_PATH')
-    ENV.SHOULD_EXIT_BOOTSTRAP_REEXEC = True
-    os.execve(selected_interpreter, cmdline, ENV.copy())
+    os.environ['SHOULD_EXIT_BOOTSTRAP_REEXEC'] = '1'
```
jsirois: This was the equivalent of just deleting the line since os.execve is used below with an explicit env passed to the subprocess. Since that env is formed from ENV.copy(), which does not see mutations to os.environ, the `'SHOULD_EXIT_BOOTSTRAP_REEXEC'` based logic was nuked in this PR. This PR's problem description should be taken in account when fixing

PR #444, opened 2018-02-09
RFC: Prepend, not append, bundled dists to sys.path

illicitonion:
If you build with `--inherit-path`, right now the dependencies
explicitly bundled in the pex are put on sys.path *after* the existing
$PYTHONPATH.

This is bad, because it can lead to things you expect to be pinned (or
transitive dependencies which you expect to be pinned) to leak in from
the system with incorrect versions.

Instead, put things we explicitly bundle at fixed versions _early_ on
the $PYTHONPATH, so that the system entries are just fallbacks, rather
than overrides.

If this looks reasonable in concept, I'll clean it up a little and add an integration test.
I could also put this behind a flag if that would be preferred.

### DIFF
--- a/pex/environment.py
+++ b/pex/environment.py
@@ -132,7 +132,9 @@ class PEXEnvironment(Environment):
   def activate(self):
     if not self._activated:
       with TRACER.timed('Activating PEX virtual environment from %s' % self._pex):
+        before = sys.path[:]
         self._working_set = self._activate()
+        sys.path = sys.path[len(before):] + before
       self._activated = True
 
     return self._working_set


pex/environment.py
```diff
@@ -132,7 +132,9 @@ def can_add(self, dist):
   def activate(self):
     if not self._activated:
       with TRACER.timed('Activating PEX virtual environment from %s' % self._pex):
+        before = sys.path[:]
         self._working_set = self._activate()
+        sys.path = sys.path[len(before):] + before
```
kwlzn: this feels like an odd place to arbitrarily mutate `sys.path`. ideally all sys.path mutation would take place in/around https://github.com/pantsbuild/pex/blob/master/pex/pex.py#L271

illicitonion:
@kwlzn Moved the code to the place where `sys.path` is actually being modified. WDYT?

@baroquebobcat Yeah - I feel like if you're using pex files, you're trying to lean towards hermeticity, so this is the correct default... How would you feel about `--inherit-path` no longer being a boolean flag, but instead taking values `false`, `prefer`, and `fallback`? We can make `true`-like values still parse as `prefer` for backwards compatibility...

baroquebobcat:
> How would you feel about --inherit-path no longer being a boolean flag, but instead taking values false, prefer, and fallback? We can make true-like values still parse as prefer for backwards compatibility...
hm. I think I like that. It gives folks a way to debug sys.path issues when using `--inherit-path` without digging into pex.

illicitonion:
@kwlzn @baroquebobcat Updated and added tests, please take a look :)

pex/bin/pex.py
```diff
@@ -620,7 +625,13 @@ def make_relative_to_root(path):
...
 
 def main(args=None):
-  args = args or sys.argv[1:]
+  args = args[:] if args else sys.argv[1:]
+  # inherit-path used to be a boolean arg (so either was absent, or --inherit-path)
+  # Now it takes a string argument, so --inherit-path is invalid.
+  # Fix up the args we're about to parse to preserve backwards compatibility.
+  for i, arg in enumerate(args):
+    if arg == '--inherit-path':
+      args[i] = arg + '=false'
```
baroquebobcat: Should this be `=prefer`?

baroquebobcat:
I'm not sure how finicky it'd be but the defaulting to prefer on empty might be worth adding a test for.

pex/bin/pex.py
illicitonion: Yes, yes it should! Good catch!

illicitonion:
Added a test :)

pex/bin/pex.py
```diff
@@ -620,7 +625,13 @@ def make_relative_to_root(path):
...
 
 def main(args=None):
-  args = args or sys.argv[1:]
+  args = args[:] if args else sys.argv[1:]
+  # inherit-path used to be a boolean arg (so either was absent, or --inherit-path)
+  # Now it takes a string argument, so --inherit-path is invalid.
+  # Fix up the args we're about to parse to preserve backwards compatibility.
+  for i, arg in enumerate(args):
+    if arg == '--inherit-path':
+      args[i] = arg + '=prefer'
```
kwlzn: potentially more readable/idiomatic to just craft a new list and replace the old args inline?

```
def transform_legacy_arg(arg):
  if arg == '--inherit-path':
    return '--inherit-path=prefer'
  return arg

args = args or sys.argv[1:]
args = [transform_legacy_arg(arg) for arg in args]
```

pex/environment.py
```diff
@@ -192,6 +193,18 @@ def _activate(self):
...
+              # This ensures that bundled versions of libraries will be used before system-installed
+              # versions, in case something is installed in both, helping to favor hermeticity in
+              # the case of non-hermetic PEX files (i.e. those with inherit_path=True).
+              #
+              # If the path is not already in sys.path, site.addsitedir will append (not prepend)
+              # the path to sys.path. But if the path is already in sys.path, site.addsitedir will
+              # leave sys.path unmodified, but will do everything else it would do. This is not part
+              # of its advertised contract (which is very vague), but has been verified to be the
+              # case by inspecting its source for both cpython 2.7 and cpython 3.7.
+              sys.path = [dist.location] + sys.path
```
kwlzn: maybe mutate vs replace here in case anything else has an existing ref to the original `sys.path`?

```
sys.path.insert(0, dist.location)
```

this is more in-line with `addsitedir`'s `sys.path.append()` too.

pex/bin/pex.py
illicitonion: Done

pex/environment.py
illicitonion: Done

PR #449, opened 2018-02-23
scrub path when not inheriting

sveitser:
Since the last release `inherit_path` is a (non-empty) string and the path is therefore no longer scrubbed.

FYI, I realize it would be preferable to have a regression test for this and to run the tests prior to opening a PR, however I'm having hard time getting the tests to run behind the Chinese firewall.


illicitonion:
Thanks for the fix! I was just investigating this same problem, and I've got some test updates on my own branch - I'll merge this and then add the test fixes :)

Issue #452, opened 2018-03-01
Improve pex resolver caching for local dist resolves


CMLivingston:
This is an upstream issue from: https://github.com/pantsbuild/pants/issues/5449

Currently, the pex resolver handles resolving dists and caches them using the name of the wheel in the pex resolver cache. This can be problematic if two locally-built (and locally resolved via a find-links fetcher), setup.py-based Python projects have the same name and version info in their respective setup.py's. This could lead to undefined behavior in the event that there is a collision in the pex resolver cache.

The scope of this ticket is to improve the pex resolver caching logic for the use case where dists are resolved from local fetcher directories, allowing for the resolution of uncached dists in the event that two dists have the same name and version but have different contents. This solution should be generalizable to any case where the cached dist contents differ from an equally versioned dist with the same name.

jsirois:
> This can be problematic if two locally-built (and locally resolved via a find-links fetcher), setup.py-based Python projects have the same name and version info in their respective setup.py's.

The upstream issue seems most sanely solved by never mutating a published version. IE, (pseudocode) use `python setup.py egg_info --tag-build $(target.transitive_invalidation_hash()) sdist` (for example) to always publish a unique version.

If that works for upstream (downstream?) is there any further motiviation for this feature?

cosmicexplorer:
The analogy to jvm `-SNAPSHOT` deps you made in [your other comment](https://github.com/pantsbuild/pex/pull/520#issuecomment-402815847) seems really reasonable to me. I don't know if there is need specifically for this, personally -- when I first started looking at this I was convinced it was necessary for pantsbuild/pants#6022, but since we're just using `--tag-build` as you describe in that PR I think this issue can be resolved in pants alone (or at least, the integration test in that PR which tests invalidation passes). I will probably close #453 after spending some more time getting the pants PR I mentioned to pass CI.

CMLivingston:
This makes sense to me, no further motivation. Closing this issue.

PR #453, opened 2018-03-02
Improve pex resolver caching for local dist resolves

CMLivingston:
Solution for: https://github.com/pantsbuild/pex/issues/452

## Problem:
Currently, the pex resolver handles resolving dists and caches them using the name of the wheel in the pex resolver cache. This can be problematic if two locally-built (and locally resolved via a find-links fetcher) setup.py-based Python projects have the same name and version info in their respective setup.py files. This could lead to undefined behavior in the event that there is a collision in the pex resolver cache.

## Solution:
Use os.stat file size in bytes to create cache keys for wheel files in the resolver cache based on their contents. Create subdirectories in the cache dir based on these cache keys and store the wheel file in its unique directory. Create a `current` directory in the cache dir for crawling the latest packages, and compare the cache key of the latest version of the package to the cache key of the first non-matching *local* package found by the `Resolver` package iterator. (This is key for managing performance impact; if we do not filter out packages that do not use the `file://` protocol, the current implementation will parse and filter all sdists and bdists found on pypi and remote links, and these remote links will not/should not have the version content conflict as described by the core problem this PR aims to solve). If the keys do not match, invalidate the latest matching wheel in the cache, and cache the new wheel and use it to resolve. 

## Result:
Wheels with the same name and version but different contents will by treated differently by the pex resolver cache. This implies that sdists for local Python projects (crawled from a find-links fetcher to a file system location) will appear differently to the resolver cache when the underlying dist contents change, triggering a cache invalidation when the project contents change but the dist name and version do not. 

## Performance impacts

Here is some preliminary data on the impact of build time caching performance on requirements of small, medium-small, medium, and large sizes, averaged over 3 runs for each.

```
Pex `CachedResolver#resolve` performance (in seconds)

Master branch: 

q: 0.0771693388621
q (cached run): 0.0517366727193

django==1.6: 0.714973926544
django==1.6 (cached run): 0.196810007095

pandas: 1.91385062536
pandas (cached run): 0.865164756775

tensorflow: 6.24201631546
tensorflow (cached run): 1.94899066289


Feature branch with improved caching strategy:

q: 0.0724590619405
q (cached run): 0.0489939848582

django==1.6: 0.831711292267
django==1.6 (cached run): 0.183348735174

pandas: 2.54747994741
pandas (cached run): 0.964469353358

tensorflow: 7.79748272896
tensorflow (cached run): 4.38255461057
```

Note that runtime caching performance will by completely unaffected by any of these chages. The impact would only be on the time it takes to build a pex.

### DIFF
--- a/pex/resolver.py
+++ b/pex/resolver.py
@@ -162,6 +162,7 @@ class Resolver(object):
         StaticIterator(existing, allow_prereleases=self._allow_prereleases))
     else:
       existing = resolvable.packages()
+
     return self.filter_packages_by_interpreter(existing, self._interpreter, self._platform)
 
   def build(self, package, options, uncached_whl_pkg=None):
@@ -189,9 +190,10 @@ class Resolver(object):
 
     def get_possibly_conflicting_wheel_package(cached_wheel_pkg, all_packages):
       def predicate(x):
-        return (x != cached_wheel_pkg and
+        return (hash(x) != hash(cached_wheel_pkg) and
                x.name == cached_wheel_pkg.name and
-               x.raw_version == cached_wheel_pkg.raw_version)
+               x.raw_version == cached_wheel_pkg.raw_version and
+               'file://' in x.url)
       for pkg in all_packages:
         # return the first uncached wheel package found by Resolver#package_iterator
         if predicate(pkg):
--- a/pex/resolver.py
+++ b/pex/resolver.py
@@ -17,7 +17,7 @@ from .fetcher import Fetcher
 from .interpreter import PythonInterpreter
 from .iterator import Iterator, IteratorInterface
 from .orderedset import OrderedSet
-from .package import Package, distribution_compatible
+from .package import Package, distribution_compatible, WheelPackage
 from .platforms import Platform
 from .resolvable import ResolvableRequirement, resolvables_from_iterable
 from .resolver_options import ResolverOptionsBuilder
@@ -165,7 +165,7 @@ class Resolver(object):
       existing = resolvable.packages()
     return self.filter_packages_by_interpreter(existing, self._interpreter, self._platform)
 
-  def build(self, package, options, uncached_pkg=None):
+  def build(self, package, options, uncached_whl_pkg=None):
     context = options.get_context()
     translator = options.get_translator(self._interpreter, self._platform)
     with TRACER.timed('Fetching %s' % package.url, V=2):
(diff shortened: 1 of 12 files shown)


pex/resolver.py
```diff
@@ -187,6 +188,18 @@ def resolve(self, resolvables, resolvable_set=None):
     processed_packages = {}
     distributions = {}
 
+    def get_possibly_conflicting_wheel_package(cached_wheel_pkg, all_packages):
+      def predicate(x):
+        return (hash(x) != hash(cached_wheel_pkg) and
+               x.name == cached_wheel_pkg.name and
+               x.raw_version == cached_wheel_pkg.raw_version and
+               'file://' in x.url)
```
UnrememberMe: move the hash comparison to the last item so that we have chance to short circulate without the need to run a more expensive hash function.

pex/resolver.py
```diff
@@ -187,6 +188,18 @@ def resolve(self, resolvables, resolvable_set=None):
...
+      for pkg in all_packages:
```
UnrememberMe: `return next(p for p in all_packages if predicate(p), None)`
next should be faster than the for loop.

pex/resolver.py
```diff
@@ -272,22 +302,45 @@ def package_iterator(self, resolvable, existing=None):
...
+        dist = uncached_dist
+
     # if distribution is not in cache, copy
-    target = os.path.join(self.__cache, os.path.basename(dist.location))
+    dist_filename = os.path.basename(dist.location)
+    cached_location = os.path.join(self.__cache, '%s.%s' % (
+      dist_filename, self.get_cache_key(dist.location)))
+    safe_mkdir(cached_location)
+
+    target = os.path.join(cached_location, dist_filename)
```
UnrememberMe: could do something like 
```
for location in [cached_location, self.ensure_current_dir()]:
  target = os.path.join(location, dist_filename)
  ...
```

pex/resolver.py
```diff
@@ -247,16 +263,27 @@ def filter_packages_by_ttl(cls, packages, ttl, now=None):
     return [package for package in packages
         if package.remote or package.local and (now - os.path.getmtime(package.local_path)) < ttl]
 
+  @classmethod
+  def get_cache_key(cls, dist_location):
+    return os.stat(dist_location).st_size
```
illicitonion: Rather than size, can we use something more robust like a content digest?

Looking at django 1.6 (which is ~7MB) which you profiled:

```
tmp$ time md5 Django-1.6.11-py2.py3-none-any.whl
MD5 (Django-1.6.11-py2.py3-none-any.whl) = e43d8c4147a821205b554aebc1da62d5

real	0m0.025s
user	0m0.019s
sys	0m0.008s
tmp$ time shasum -a 1 Django-1.6.11-py2.py3-none-any.whl
0d7d9071158294930ae86b776eb9ff9011eb1d7b  Django-1.6.11-py2.py3-none-any.whl

real	0m0.076s
user	0m0.047s
sys	0m0.021s
tmp$ time shasum -a 256 Django-1.6.11-py2.py3-none-any.whl
6f2cc848b2b72adf53a6f3f21be049c477e82c408bce7cedb57efeb0984bde24  Django-1.6.11-py2.py3-none-any.whl

real	0m0.100s
user	0m0.062s
sys	0m0.023s
```

I'm definitely happy to add 50ms per 7M dependency to pex building. I could also live with an extra 200ms.

CMLivingston: Thanks for taking a look! I agree that a hash digest would certainly be more robust. I changed it to size purely to squeeze out some perf, but it may make sense to switch back. I'll wait for some additional feedback on the perf impact.

CMLivingston:
This was addressed in #6022. Closing.

Issue #456, opened 2018-04-05
Implement long-term solution for cross-interpreter requirement handling the in pex resolver


CMLivingston:
When a setup.py file specifies `install_requires` requirements with interpreter-specific constraints, the pex resolver cannot parse and handle these constraints to produce a cross-interpreter pex. See https://github.com/Julian/jsonschema/blob/master/setup.py#L43 for an example of such a constraint. 

Interpreter constraints like this are typically for stdlib backport libraries and the like that only make sense to include in a universal dist, so this problem is not as significant when the pex is targeting a specific interpreter version. In the targeted case, the resolver only needs to ignore requirements that are pinned to non-targeted interpreter versions. The short term workaround for the targeted-interpreter case is to allow the caller of `pex.resolver#resolve` to supply a blacklist dict that maps package name string -> interpreter constraint string, allowing the resolver to effectively skip them as implemented in https://github.com/pantsbuild/pex/pull/457. 

The scope of this ticket is to remove the blacklist implementation and replace it with smarter handling for the "pex-targeting-X-interpreter-version" case, specifically where all requirements are compared against interpreter constraints for a pex. Then at run time, the pex environment selectively activates packaged dependencies in accordance with a given pex's interpreter constraints. If no constraints are specified and the pex is intended to run with Python 2 or 3, then there should be support for determining this at runtime and handling env activation accordingly as well. 

This approach would solve the cross-interpreter case. This is a similar problem to the one surfaced in #455, however that issue deals with cross-platform specifications, but this is effectively the same concept.

kmannislands:
Is there any workaround for this currently available?
I tried something like:
```sh
$ pex jupyter --python=python2.7 --python=python3.4 -o jup_pex.pex
```
Which results in:
```
**** Failed to install functools32-3.2.3-2 (caused by: NonZeroExit("received exit code 1 during execution of `['/Users/kieran/.pyenv/versions/3.4.8/bin/python3.4', '-', 'bdist_wheel', '--dist-dir=/var/folders/7d/cnv90sjj7rx4nhq38nvx44nr0000gn/T/tmpZEG5fd']` while trying to execute `['/Users/kieran/.pyenv/versions/3.4.8/bin/python3.4', '-', 'bdist_wheel', '--dist-dir=/var/folders/7d/cnv90sjj7rx4nhq38nvx44nr0000gn/T/tmpZEG5fd']`",)
):
stdout:

stderr:
This backport is for Python 2.7 only.


[...]

PR #457, opened 2018-04-05
Add blacklist handling for skipping requirements in pex resolver

CMLivingston:
## Problem
When resolving for a specific interpreter version, the pex resolver fails to resolve universal requirements that transitively require libraries with interpreter constraints that conflict with the target interpreter version (see: https://github.com/Julian/jsonschema/blob/master/setup.py#L43 ).

## Solution
The short term solution is to allow the caller of pex.resolver#resolve to supply a blacklist dict that maps package name -> interpreter constraint, enabling the resolver to skip blacklisted requirement names at resolve time if the target interpreter conforms with the interpreter constraint. The long-term solution is to be addressed by https://github.com/pantsbuild/pex/issues/456. 

## Result
It is now possible for systems that build pex files (Pants) via the pex api to blacklist certain requirements that are redundant (backports) or otherwise should not be included in the output pex.

### DIFF
--- a/pex/resolver.py
+++ b/pex/resolver.py
@@ -151,10 +151,11 @@ class Resolver(object):
     return [package for package in packages
         if package.compatible(interpreter.identity, platform)]
 
-  def __init__(self, allow_prereleases=None, interpreter=None, platform=None):
+  def __init__(self, allow_prereleases=None, interpreter=None, platform=None, pkg_blacklist=None):
     self._interpreter = interpreter or PythonInterpreter.get()
     self._platform = platform or Platform.current()
     self._allow_prereleases = allow_prereleases
+    self._blacklist = pkg_blacklist or {}
 
   def package_iterator(self, resolvable, existing=None):
     if existing:
@@ -180,6 +181,11 @@ class Resolver(object):
         'Could not get distribution for %s on platform %s.' % (package, self._platform))
     return dist
 
+  def resolvable_is_blacklisted(self, resolvable_name):
+    if not resolvable_name in self._blacklist:
+      return False
+    return self._interpreter.identity.matches(self._blacklist[resolvable_name])
+
   def resolve(self, resolvables, resolvable_set=None):
     resolvables = [(resolvable, None) for resolvable in resolvables]
     resolvable_set = resolvable_set or _ResolvableSet()
@@ -193,6 +199,12 @@ class Resolver(object):
         if resolvable in processed_resolvables:
           continue
         packages = self.package_iterator(resolvable, existing=resolvable_set.get(resolvable.name))
+
+        # TODO: Remove blacklist strategy in favor of smart requirement handling
+        # https://github.com/pantsbuild/pex/issues/456
+        if self.resolvable_is_blacklisted(resolvable.name):
+          processed_resolvables.add(resolvable)
+          continue
         resolvable_set.merge(resolvable, packages, parent)
         processed_resolvables.add(resolvable)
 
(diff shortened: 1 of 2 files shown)


pex/resolver.py
```diff
@@ -180,6 +181,11 @@ def build(self, package, options):
         'Could not get distribution for %s on platform %s.' % (package, self._platform))
     return dist
 
+  def resolvable_is_blacklisted(self, resolvable_name):
+    if not resolvable_name in self._blacklist:
+      return False
+    return self._interpreter.identity.matches(self._blacklist[resolvable_name])
```
UnrememberMe: A possible one-liner
```
return (
  resolvable_name in self._blacklist and
  self._interpreter.identity.matches(self._blacklist[resolvable_name])
)
```

pex/resolver.py
```diff
@@ -151,10 +151,11 @@ def filter_packages_by_interpreter(cls, packages, interpreter, platform):
     return [package for package in packages
         if package.compatible(interpreter.identity, platform)]
 
-  def __init__(self, allow_prereleases=None, interpreter=None, platform=None):
+  def __init__(self, allow_prereleases=None, interpreter=None, platform=None, pkg_blacklist=None):
     self._interpreter = interpreter or PythonInterpreter.get()
     self._platform = platform or Platform.current()
     self._allow_prereleases = allow_prereleases
+    self._blacklist = pkg_blacklist or {}
```
UnrememberMe: Not sure if it matters here.  This line carries risk of `pkg_blacklist` being updated outside of the function and impact execution inside since pkg_blacklist becomes a shared object.  Maybe `self._blacklist = {}.update(pkg_blacklsit or {})` will be safer?

pex/resolver.py
```diff
@@ -193,6 +199,12 @@ def resolve(self, resolvables, resolvable_set=None):
         if resolvable in processed_resolvables:
           continue
         packages = self.package_iterator(resolvable, existing=resolvable_set.get(resolvable.name))
+
+        # TODO: Remove blacklist strategy in favor of smart requirement handling
```
UnrememberMe: maybe
``` 
if not self.resolvable_is_blacklisted(resolvable.name):
  resolvable_set.merge(resolvable, packages, parent)
processed_resolvables.add(resolvable)
```
seems cleaner to me.

pex/resolver.py
```diff
@@ -329,6 +342,10 @@ def resolve(requirements,
     ``context``.
   :keyword allow_prereleases: (optional) Include pre-release and development versions.  If
     unspecified only stable versions will be resolved, unless explicitly included.
+  :keyword pkg_blacklist: (optional) A blacklist dict (str->str) that maps package name to
```
UnrememberMe: would love to have a better documentation on why a package needed to be blacklisted

pex/resolver.py
```diff
@@ -193,7 +200,11 @@ def resolve(self, resolvables, resolvable_set=None):
         if resolvable in processed_resolvables:
           continue
         packages = self.package_iterator(resolvable, existing=resolvable_set.get(resolvable.name))
-        resolvable_set.merge(resolvable, packages, parent)
+
+        # TODO: Remove blacklist strategy in favor of smart requirement handling
```
kwlzn: this TODO is good, but it'd be great to make the temporary nature of this change more evident in the docstring(s) below where developers are more likely to look. Would be good to include the ticket in there as well.

tests/test_resolver.py
```diff
@@ -349,3 +352,32 @@ def test_resolvable_set_built():
   updated_rs.merge(rq, [binary_pkg])
   assert updated_rs.get('foo') == set([binary_pkg])
   assert updated_rs.packages() == [(rq, set([binary_pkg]), None, False)]
+
+def test_resolver_blacklist():
+  project = make_sdist(name='project', version='1.0.0')
```
kwlzn: this probably needs a `setup.py`-sourced transitive dep to blacklist as a complete test case?

PR #469, opened 2018-05-12
Add tracebacks to IntegResults.

kwlzn:


kwlzn:
TBR'ing this based on a positive improvement here: https://travis-ci.org/kwlzn/pex/jobs/378028725#L322

Issue #470, opened 2018-05-13
integration tests in CI use a travis provided python that doesn't contain setuptools


kwlzn:
discovered while working through #461:

```
=================================== FAILURES ===================================
_________________ test_interpreter_constraints_to_pex_info_py3 _________________
    @pytest.mark.skipif(NOT_CPYTHON_36)
    def test_interpreter_constraints_to_pex_info_py3():
      with temporary_dir() as output_dir:
        # target python 3
        pex_out_path = os.path.join(output_dir, 'pex_py3.pex')
        res = run_pex_command(['--disable-cache',
          '--interpreter-constraint=>3',
          '-o', pex_out_path])
>       res.assert_success()
tests/test_integration.py:354: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = IntegResults(output=['Saving PEX file to /tmp/tmphjizs6w3/pex_py3.pex'], return_code=None, exception=RuntimeError('Fai...\' % self._interpreter.binary\nRuntimeError: Failed to find setuptools via /usr/bin/python3.2mu while building pex!\n')
    def assert_success(self):
      if not (self.exception is None and self.return_code is None):
        raise AssertionError('integration test failed: return_code=%s, exception=%r, traceback=%s' % (
>         self.return_code, self.exception, self.traceback
        ))
E       AssertionError: integration test failed: return_code=None, exception=RuntimeError('Failed to find setuptools via /usr/bin/python3.2mu while building pex!',), traceback=Traceback (most recent call last):
E         File "/home/travis/build/kwlzn/pex/.tox/py36-requests/lib/python3.6/site-packages/pex/testing.py", line 241, in run_pex_command
E           main(args=args)
E         File "/home/travis/build/kwlzn/pex/.tox/py36-requests/lib/python3.6/site-packages/pex/bin/pex.py", line 669, in main
E           pex_builder.build(tmp_name)
E         File "/home/travis/build/kwlzn/pex/.tox/py36-requests/lib/python3.6/site-packages/pex/pex_builder.py", line 472, in build
E           self.freeze(bytecode_compile=bytecode_compile)
E         File "/home/travis/build/kwlzn/pex/.tox/py36-requests/lib/python3.6/site-packages/pex/pex_builder.py", line 456, in freeze
E           self._prepare_bootstrap()
[...]

Issue #461, opened 2018-05-11
master, travis are silently broken


kwlzn:
after some mysterious local failures that weren't reproducing in CI, I noticed that several of the recent travis runs have green test shards that actually have silently failing tests.

e.g. PR #316 in https://travis-ci.org/pantsbuild/pex/builds/368145168 the 3.6 shard in https://travis-ci.org/pantsbuild/pex/jobs/368145178 shows:

```
======== 16 failed, 170 passed, 12 skipped, 1 xfailed in 111.28 seconds ========
```

tox emits an error, but then there appear to be some missing commands at the end of the run which seem suspect:

```
ERROR: InvocationError for command '/home/travis/build/pantsbuild/pex/.tox/py36/bin/py.test' (exited with code 1)
___________________________________ summary ____________________________________
ERROR:   py36: commands failed
/home/travis/.travis/job_stages: line 95: uname: command not found
/home/travis/.travis/job_stages: line 105: date: command not found
/home/travis/.travis/job_stages: line 87: -: syntax error: operand expected (error token is "-")
cache.2
store build cache
Done. Your build exited with .
/home/travis/.travis/job_stages: line 163: pgrep: command not found
/home/travis/.travis/job_stages: line 163: grep: command not found
/home/travis/.travis/job_stages: line 164: awk: command not found
```

cc @CMLivingston @jsirois

kwlzn:
looks like this started in https://github.com/pantsbuild/pex/pull/427 likely via the `.travis.yml` changes - @CMLivingston would you mind taking a look?

CMLivingston:
Will do.

PR #472, opened 2018-05-13
Skip two failing integration tests.

kwlzn:
Temporary relief for #461, #470


kwlzn:
will TBR on greenish CI

PR #473, opened 2018-05-13
Use pytest instead of py.test per upstream recommendation, #dropthedot

scop:
http://blog.pytest.org/2016/whats-new-in-pytest-30/
https://twitter.com/hashtag/dropthedot


Eric-Arellano:
This is now stale: this script has been deleted and tox.ini is fixed.

PR #474, opened 2018-05-13
Fix .travis.yml path setting for testing purposes

CMLivingston:
Fix for issue #470.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,12 +8,6 @@ dist: precise
 
 # TRAVIS_PYTHON_VERSION
 
-env:
-  global:
-    # This is necessary to have consistent testing of methods that rely on falling back to PATH for
-    # selecting a python interpreter.
-    - PATH=$PWD/.tox/py36/bin:$PWD/.tox/py27/bin:$PATH
-
 cache:
   directories:
     - .pyenv_test
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -344,7 +344,6 @@ def test_interpreter_constraints_to_pex_info_py2():
     assert set(['>=2.7', '<3']) == set(pex_info.interpreter_constraints)
 
 
-#@pytest.mark.skip('https://github.com/pantsbuild/pex/issues/470')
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -2,18 +2,12 @@
 # which should provide faster startup times and beefier
 # "machines".
 # See: http://docs.travis-ci.com/user/workers/container-based-infrastructure/
-sudo: false
+sudo: true
 # The pypy test shard fails with "too many open files" under Trusty so we pin to Precise.
-dist: precise
+dist: trusty
```
kwlzn: would probably be better to make this change in a separate PR, I think.

tests/test_integration.py
```diff
@@ -343,17 +349,19 @@ def test_interpreter_constraints_to_pex_info_py2():
...
-  with temporary_dir() as output_dir:
-    # target python 3
-    pex_out_path = os.path.join(output_dir, 'pex_py3.pex')
-    res = run_pex_command(['--disable-cache',
-      '--interpreter-constraint=>3',
-      '-o', pex_out_path])
-    res.assert_success()
-    pex_info = get_pex_info(pex_out_path)
-    assert ['>3'] == pex_info.interpreter_constraints
+  py3_interpreter = ensure_python_interpreter('3.6.3')
```
kwlzn: rather than mocking out `os.getenv` here, how about just temporarily setting the `PATH` env var with `twitter.common.contextutil.environment_as` for the scope of this test?

tests/test_integration.py
```diff
@@ -343,17 +349,19 @@ def test_interpreter_constraints_to_pex_info_py2():
     assert set(['>=2.7', '<3']) == set(pex_info.interpreter_constraints)
 
 
-@pytest.mark.skip('https://github.com/pantsbuild/pex/issues/470')
 def test_interpreter_constraints_to_pex_info_py3():
```
kwlzn: `test_pex_python` in `test_integration.py` needs the same treatment as this test - both were disabled.

PR #495, opened 2018-05-23
Update RELEASE.rst pip usage for cache avoidance

kwlzn:
This is more aggressively cached by pip as of the new warehouse backend for PyPI.

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -51,7 +51,7 @@ Dogfood
 
 ::
 
-    $ pip install --upgrade pex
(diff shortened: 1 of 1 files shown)


CMLivingston:
What do you mean by "aggressively" here?

kwlzn:
> What do you mean by "aggressively" here?

as in, the new cache settings used upstream will prevent `pip install upgrade` from immediately seeing a newly published version (for up to an ~hour) unless the cache is explicitly ignored. previously, this was instantaneous. hence, the caching policy is more "aggressively" utilizing the cache.

PR #496, opened 2018-05-23
Isolate integration tests onto their own shard

CMLivingston:
Use pytest keyword collection to omit integration tests by default. Use a new tox env to pass a positional arg that overrides the default POSARGS for py.test in tox.ini. 

I only did {py27, py26, pypy}-requests because that is what runs when you run a local `tox` command, but I am happy to add more.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -34,7 +34,7 @@ whitelist_externals = open
 deps =
     tox
 commands =
-    #tox -e pypy-requests -- -k "tests/test_integration.py" {posargs:-vvs}
+    tox -e pypy-requests -- -k "tests/test_integration.py" {posargs:-vvs}
     tox -e py27-requests -- -k "tests/test_integration.py" {posargs:-vvs}
     tox -e py36-requests -- -k "tests/test_integration.py" {posargs:-vvs}
 
--- a/tox.ini
+++ b/tox.ini
@@ -34,9 +34,9 @@ whitelist_externals = open
 deps =
     tox
 commands =
-    tox -e pypy-integration
     tox -e py27-integration
     tox -e py36-integration
+    tox -e pypy-integration
 
 [testenv:pypy-integration]
 deps =
--- a/tox.ini
+++ b/tox.ini
@@ -34,9 +34,9 @@ whitelist_externals = open
 deps =
     tox
 commands =
-    tox -e pypy-requests -- -k "tests/test_integration.py" {posargs:-vvs}
-    tox -e py27-requests -- -k "tests/test_integration.py" {posargs:-vvs}
-    tox -e py36-requests -- -k "tests/test_integration.py" {posargs:-vvs}
+    tox -e pypy-integration
+    tox -e py27-integration
+    tox -e py36-integration
 
 [testenv:pypy-integration]
 deps =
--- a/tox.ini
+++ b/tox.ini
@@ -6,7 +6,7 @@ envlist =
 
 [testenv]
 commands =
-    py.test {posargs:-vvs -k "not test_integration"}
+    py.test -k "not tests/test_integration.py" {posargs:-vvs}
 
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -2,11 +2,12 @@
 skip_missing_interpreters = True
 minversion = 1.8
 envlist =
-	py{py,27,36}-requests,style,isort-check
+	py{py,27,36}-requests,style,isort-check,integration-tests
 
 [testenv]
 commands =
-    py.test {posargs:-vvs}
+    py.test {posargs:-vvs -k "not test_integration"}
```
kwlzn: probably want to use e.g. `--ignore=tests/test_integration.py` form here vs `-k`.

I think that should also go outside of the posargs default, too. otherwise passing any kind of positional args will override this and cause integration tests to run where they would not normally.

tox.ini
```diff
@@ -29,6 +30,15 @@ deps =
     subprocess: subprocess32
 whitelist_externals = open
 
+[testenv:integration-tests]
+deps =
+    tox
+commands =
+    # meta
```
kwlzn: it's not clear what this comment means.

tox.ini
```diff
@@ -29,6 +30,15 @@ deps =
...
+    tox "-k test_integration -vvsx" -e pypy-requests
+    tox "-k test_integration -vvsx" -e py27-requests
+    tox "-k test_integration -vvsx" -e py36-requests
```
kwlzn: would drop the `-x` here. also clearer form here might be passthru arg style, e.g.

```
tox -e py27-requests -- -k test_integration -vvs
```

would also be good to preserve the ability to handle passthru args when executing this environment, so maybe a further improvement here could be:

```
tox -e py27-requests -- -k test_integration {posargs:-vvs}
...
```

it might also be good to explicitly specify `-k tests/test_integration.py` to avoid inadvertently selecting any tests that contain the substring `test_integration`.

tox.ini
CMLivingston: I used the default for the purpose of overriding with a positional arg. The issue is that `py.test --ignore=tests/test_integration.py -k tests/test_integration.py -vvs` results in no tests being run. Any ideas?

kwlzn: > The issue is that py.test --ignore=tests/test_integration.py -k tests/test_integration.py -vvs results in no tests being run. 

ah, I see - because the `integration-tests` env itself relies on this for its own execution. `-k not tests/test_integration.py` sounds good, but probably outside of the `posargs`?

tox.ini
```diff
@@ -29,6 +30,14 @@ deps =
     subprocess: subprocess32
 whitelist_externals = open
 
+[testenv:integration-tests]
+deps =
+    tox
+commands =
+    tox -e pypy-requests -- -k "tests/test_integration.py" {posargs:-vvs}
+    tox -e py27-requests -- -k "tests/test_integration.py" {posargs:-vvs}
+    tox -e py36-requests -- -k "tests/test_integration.py" {posargs:-vvs}
```
kwlzn: running 3x serial runs of our heaviest test suite on the same shard may present a long pole in terms of the total CI runtime.. wonder if it would make sense to do one shard per interpreter + test_integration.py invocation here so that these can be parallelized? e.g. break out `pypy-integration`, `py27-integration`, `py36-integration` envs w/ backing shards.

wdyt?

tox.ini
```diff
@@ -29,6 +30,32 @@ deps =
...
```
kwlzn: this aggregate env should probably just call `tox -e pypy-integration` / `tox -e py27-integration` / `tox -e py36-integration` now that those exist?

(but probably in the order of 27, 36, pypy?)

CMLivingston: Will do.

PR #502, opened 2018-05-26
Set higher ulimit in travis.yml.

### LINKED ISSUE #497: pypy shards in travisCI can fail with 'too many open files' errors
this needs more investigation, but intermittently only the "pypy" shards in travis will fail arbitrarily with "too many open files" errors.

we had previously pinned to the travis `precise` image because of this, but because of deprecation prompting for that image we'd jumped forward here: https://github.com/pantsbuild/pex/pull/476

an initial pass may want to try some simple `ulimit` tuning.

kwlzn:
Fixes #497

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,7 +77,8 @@ install:
   - pip install -U tox "setuptools<34"
 
 before_script:
-  - ulimit -n 32768
+  - ulimit -Hn 10000
+  - ulimit -Sn 10000
 
 script:
   - tox -v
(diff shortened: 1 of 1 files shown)


kwlzn:
till TBR on 3 successful green re-runs of the pypy shard for CI relief

kwlzn:
reviving and TBR'ing this - since I think it's an innocuous enough change and seems to at least help with the pypy open files problem.

kwlzn:
> I was able to fix this problem locally by passing -n instead

ah, good eye - that's precisely what I had meant to do here.

PR #504, opened 2018-05-26
Bump pytest version.

kwlzn:


kwlzn:
TBR'ing this one in the hopes of some potential relief for #497

kwlzn:
so much for that idea.

Issue #510, opened 2018-06-06
PythonInterperter should support creating bare interpreters.


jsirois:
Discovered working pantsbuild/pants#5910. The only remaining public APIs (the constructor is fapp not public) all route through `from_binary` which creates a `PythonInterpreter` with all the extras installed in that interpreter's `site_packages`, etc. As such, creating a bare interpreter requires something like:
```python
interpreter_with_extras = PythonInterpreter.from_binary(binary_path)
bare_interpreter = PythonInterpreter(binary_path, interpreter_with_extras.identity, extras=None)
```

It's perhaps enough to plumb an `include_site_extras=True` parameter through `from_binary` that, when passed as false, could produce a bare interpreter.

Issue #509, opened 2018-06-05
PythonInterpreter.from_binary has order-dependent caching


jsirois:
Noticed working https://github.com/pantsbuild/pants/pull/5910. The method signature allows for optional interpreter path extras, but these are not included in the cache key. As a result, once an interpreter is found for a given binary - with say no extras - later attempts to instantiate an interpreter from that same binary with extras will silently fail to include the requested extras.

PR #517, opened 2018-06-15
Introduce a controlled env for pex testing.

jsirois:
Although this could be used as a stepping stone to containerized
Travis CI, in the short term it helps me have a working test
environment. The `pex.testing.ensure_python_interpreter` fails to
install CPython 2.7.9, 2.7.10, etc. on Arch Linux atm.

### DIFF
--- a/dtox.sh
+++ b/dtox.sh
@@ -10,13 +10,13 @@ BASE_INPUT=(
 base_hash=$(cat ${BASE_INPUT[@]} | git hash-object -t blob --stdin)
 
 function base_id() {
-  docker images -q -f label=${base_hash} pantsbuild/pex
+  docker images -q -f label=base_hash=${base_hash} pantsbuild/pex:base
 }
 
 if [[ -z "$(base_id)" ]]; then
   docker build \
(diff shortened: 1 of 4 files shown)


jsirois:
Supports work on #509 and #510.

jsirois:
N.B: This allows me to run `test_interpreter.py` green where I couldn't before via:
```
./dtox.sh -epy27-requests -- -vktest_interpreter.py
```

dtox.sh
```diff
@@ -0,0 +1,45 @@
...
+    --build-arg GROUP=$(id -gn) \
+    --build-arg GID=$(id -g) \
+    --tag pantsbuild/pex:user \
+    --label user_hash=${user_hash} \
+    "${ROOT}/docker/user"
+fi
+
+exec docker run \
+  --tty \
+  --rm \
```
kwlzn: might this need a `-i` here too for e.g. pdb? or does `--tty` imply that?

jsirois: Yup, I had not thought about interactive tox commands. Added.

PR #518, opened 2018-06-15
Fix `PythonInterpreter` caching and ergonomics.

### LINKED ISSUE #509: PythonInterpreter.from_binary has order-dependent caching

### LINKED ISSUE #510: PythonInterperter should support creating bare interpreters.

jsirois:
The `PythonInterpreter.from_binary` method now caches by a full stable
key over all input parameters fixing previously order-dependent caching.
In addition, an `include_site_extras` argument is added to allow the
caller to request a bare interpreter (by passing `False`) with none of
the extra distributions installed in the interpeter `site-packages`
directory.

Fixes #509
Fixes #510

### DIFF
--- a/pex/interpreter.py
+++ b/pex/interpreter.py
@@ -51,7 +51,9 @@ print(
   )
 )
 
-setuptools_path = None
+"""
+
+EXTRAS_PY = b"""\
 try:
   import pkg_resources
 except ImportError:
@@ -69,7 +71,7 @@ for requirement_str, location in requirements.items():
 """
 
 
-def _generate_identity_source():
+def _generate_identity_source(include_site_extras):
   # Determine in the most platform-compatible way possible the identity of the interpreter
   # and its known packages.
   encodables = (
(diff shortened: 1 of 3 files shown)


pex/interpreter.py
```diff
@@ -366,14 +375,22 @@ def from_env(cls, hashbang):
...
 
   @classmethod
-  def from_binary(cls, binary, path_extras=None):
+  def from_binary(cls, binary, path_extras=None, include_site_extras=True):
+    """
+
+    :param binary:
+    :param path_extras:
+    :param include_site_extras:
+    :return:
```
kwlzn: missing values for docstrings here

jsirois: Facepalm. Fixed!

PR #520, opened 2018-07-04
[WIP] Improve resolver caching by using a hash of the wheel contents

### LINKED ISSUE #452: Improve pex resolver caching for local dist resolves

cosmicexplorer:
*Fixes #452, closes #453. WIP right now because I need to make it pass CI.*

This is branched directly off of #453, but uses a checksum of the dist as a cache key when using the `CachingResolver` (instead of using the file size as in #453). There was a rebase against master, so the lineage was lost.

I have synced with @CMLivingston and he is ok with me picking this up. I'm pushing this now because (1) it's done (see the comments in #453) (2) it may be part of the issue I'm investigating on the osx ci shard in pantsbuild/pants#6022.

I do not know how to produce the perf comparisons for building pexes as in the OP of #453 -- I have just reached out to Chris to see if he can tell me how to do that. I obviously wouldn't want to merge this until we can produce those comparisons (there was a concern in #453 that hashing the file contents could affect perf), but I don't know if there's a canonical way to do that (if there's not, I'll make a non-canonical way).

Let me know if there is any further work that needs to be done on this PR besides the perf comparison (and passing CI).

### DIFF
--- a/pex/compiler.py
+++ b/pex/compiler.py
@@ -7,6 +7,7 @@ from .compatibility import to_bytes
 from .executor import Executor
 from .util import named_temporary_file
 
+
 _COMPILER_MAIN = """
 from __future__ import print_function
 
--- a/pex/compiler.py
+++ b/pex/compiler.py
@@ -7,7 +7,6 @@ from .compatibility import to_bytes
 from .executor import Executor
 from .util import named_temporary_file
 
-
 _COMPILER_MAIN = """
 from __future__ import print_function
 
--- a/pex/fetcher.py
+++ b/pex/fetcher.py
@@ -35,13 +35,7 @@ class FetcherBase(AbstractClass):
 
 
 class Fetcher(FetcherBase):
-  def __init__(self, urls, include_subdirs=False):
-    if include_subdirs:
-      subdir_urls = []
-      for url in urls:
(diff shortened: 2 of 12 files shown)


jsirois:
See my note here: https://github.com/pantsbuild/pex/issues/452#issuecomment-402815459
But do we need to do this? It's the equivalent of adding 1st class support for always re-resolving `-SNAPSHOT` dependencies for the jvm, which, at least historically, we've opposed as supporting an anti-pattern.

cosmicexplorer:
Closed as per the discussion in #452.

PR #521, opened 2018-07-09
Optionally validate entry point at build time

### LINKED ISSUE #508: pex should error at build-time if entrypoint target doesn't exist
current behavior is to let you build a pex that cannot execute:

```
[omerta ~]$ pex requests -e qqqqqqqqqqqq -o test.pex
[omerta ~]$ ./test.pex
  File ".bootstrap/_pex/pex.py", line 367, in execute
  File ".bootstrap/_pex/pex.py", line 293, in _wrap_coverage
  File ".bootstrap/_pex/pex.py", line 325, in _wrap_profiling
  File ".bootstrap/_pex/pex.py", line 410, in _execute
  File ".bootstrap/_pex/pex.py", line 468, in execute_entry
  File ".bootstrap/_pex/pex.py", line 473, in execute_module
  File "/Users/kwilson/Python/CPython-2.7.13/lib/python2.7/runpy.py", line 182, in run_module
    mod_name, loader, code, fname = _get_module_details(mod_name)
  File "/Users/kwilson/Python/CPython-2.7.13/lib/python2.7/runpy.py", line 107, in _get_module_details
    raise error(format(e))
ImportError: No module named qqqqqqqqqqqq
```

this is set in the `PEXBuilder` here: https://github.com/pantsbuild/pex/blob/master/pex/pex_builder.py#L227-L241

just above this, we do a similar check for "scripts" that exist in any distribution in the pex:
https://github.com/pantsbuild/pex/blob/master/pex/pex_builder.py#L216

using the finder utilities here:
https://github.com/pantsbuild/pex/blob/master/pex/finders.py#L269-L292

which use the `pkg_resources.Distribution` `IResourceProvider` interface here: https://setuptools.readthedocs.io/en/latest/pkg_resources.html#iresourceprovider

to check for scripts inside of any distributions in the pex. it's possible we can use methods from this same interface to at a minimum validate the namespace of a given entrypoint in thirdparty deps. but validating that the entrypoint callable exists seems a little trickier - I think that may require importing, which could be tricky (in e.g. the case of a multi-platform pex.. and have side effects, etc).

one safer alternative there may be to load the source file as a resource and then `ast.parse` it - then walk the AST looking for matching `FunctionDef` (or other callable?) nodes: https://docs.python.org/3/library/ast.html#abstract-grammar (this may have limitations for multi-interpreter pexes tho).

we would also need to cover the "exists in raw sources in the pex" case (possibly by building up a synthetic `Distribution` around it and using the same approach)?

wisechengyi:
Resolves #508

### Problem

Current behavior allows building a pex that cannot execute:

```
[omerta ~]$ pex requests -e qqqqqqqqqqqq -o test.pex
[omerta ~]$ ./test.pex
  File ".bootstrap/_pex/pex.py", line 367, in execute
  File ".bootstrap/_pex/pex.py", line 293, in _wrap_coverage
  File ".bootstrap/_pex/pex.py", line 325, in _wrap_profiling
  File ".bootstrap/_pex/pex.py", line 410, in _execute
  File ".bootstrap/_pex/pex.py", line 468, in execute_entry
  File ".bootstrap/_pex/pex.py", line 473, in execute_module
  File "/Users/kwilson/Python/CPython-2.7.13/lib/python2.7/runpy.py", line 182, in run_module
    mod_name, loader, code, fname = _get_module_details(mod_name)
  File "/Users/kwilson/Python/CPython-2.7.13/lib/python2.7/runpy.py", line 107, in _get_module_details
    raise error(format(e))
ImportError: No module named qqqqqqqqqqqq
```

### Solution

Verify entry point at build time. E.g. `a.b.c:m` means we will try to do `from a.b.c import m` in a separate process.

### Result

```
$ find hello
hello
hello/test.py
hello/tree.py

# Invalid module
$ python bin.py -D hello -e invalid.module  -o x.pex --validate-entry-point
ImportError: No module named invalid.module
  File "bin.py", line 758, in <module>
    main()
  File "bin.py", line 743, in main
    pex_builder.build(tmp_name, verify_entry_point=options.validate_ep)
  File "/Users/yic/workspace/pex/pex/pex_builder.py", line 529, in build
    self.freeze(bytecode_compile=bytecode_compile, verify_entry_point=verify_entry_point)
  File "/Users/yic/workspace/pex/pex/pex_builder.py", line 514, in freeze
    self._verify_entry_point()
  File "/Users/yic/workspace/pex/pex/pex_builder.py", line 263, in _verify_entry_point
    raise self.InvalidEntryPoint('Failed to:`{}`'.format(import_statement))
pex.pex_builder.InvalidEntryPoint: Failed to:`import invalid.module`

# invalid method
$ python bin.py -D hello -e test:invalid_method  -o x.pex --validate-entry-point
ImportError: cannot import name invalid_method
  File "bin.py", line 758, in <module>
    main()
  File "bin.py", line 743, in main
    pex_builder.build(tmp_name, verify_entry_point=options.validate_ep)
  File "/Users/yic/workspace/pex/pex/pex_builder.py", line 529, in build
    self.freeze(bytecode_compile=bytecode_compile, verify_entry_point=verify_entry_point)
  File "/Users/yic/workspace/pex/pex/pex_builder.py", line 514, in freeze
    self._verify_entry_point()
  File "/Users/yic/workspace/pex/pex/pex_builder.py", line 263, in _verify_entry_point
    raise self.InvalidEntryPoint('Failed to:`{}`'.format(import_statement))
pex.pex_builder.InvalidEntryPoint: Failed to:`from test import invalid_method`

# without the flag, invalid method still works
$ python bin.py -D hello -e test:invalid_method  -o x.pex 
```


wisechengyi:
Hi @kwlzn please kindly take an initial look for sanity. I am still working on tests, so will ping when it's fully ready.

pex/pex_builder.py
```diff
@@ -224,6 +226,69 @@ def set_script(self, script):
...
+        methods = filter(lambda x: isinstance(x, ast.FunctionDef) and x.name == ep_method, tree.body)
+        if methods:
+          final_candidates.append(c)
+
+    if not final_candidates:
+      raise self.InvalidEntryPoint("No method name `{}` has been found in {}."
+                                   .format(ep_method, file_candidates))
+
+  def _gather_file_candidates(self, entry_point_elements):
+    # 1. Find the files matching a.b.c:m, which is a/b/c.py
```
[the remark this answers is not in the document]
wisechengyi: Importing the code feels quite sketchy, even out of process, because they could have bad side effect or even security concerns.

How do you feel about putting verifying entry point behind an option, stating that it would only check for static method in
* `a/b/c.py`
* `a/b/c/__init__.py`
and allow people to skip static check by disabling it?

pex/pex_builder.py
```diff
@@ -238,6 +303,7 @@ def set_entry_point(self, entry_point):
     The entry point may also be specified via ``PEXBuilder.set_executable``.
     """
     self._ensure_unfrozen('Setting an entry point')
+    self._verify_entry_point(entry_point)
```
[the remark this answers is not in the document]
wisechengyi: Any pointer for a better place?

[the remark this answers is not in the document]
jsirois: In [`freeze`](https://github.com/pantsbuild/pex/blob/master/pex/pex_builder.py#L444), where you know there can be no more code or requirement additions.

pex/pex_builder.py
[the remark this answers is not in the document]
jsirois: If validation is opt-in, it seems like import would be the way to go. The user is going to run the pex after all and suffer the side-effects soon anyhow when they do so! Basically, this is forced to be a half measure one way or the other, so the choice comes down to avoiding side-effects at build time (they'll happen at runtime) for the X% of cases that have this pathology, or not supporting Y% of cases that have bizarre dynamic module generation. I'm really not sure what these percentages are or what is the better way to go. Import just seems attractive since it's robust: ie it's what pex does at runtime. Parsing is new logic.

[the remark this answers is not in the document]
wisechengyi: Yeah that's fair. Importing could also get around having to workaround ast parsing between python 2 and 3.

So I am going to do the importing, and then put it behind an option with default disabled.

pex/pex_builder.py
```diff
@@ -455,11 +509,13 @@ def freeze(self, bytecode_compile=True):
...
     self._prepare_bootstrap()
     self._prepare_main()
+    if verify_entry_point:
+      self._verify_entry_point()
     if bytecode_compile:
       self._precompile_source()
     self._frozen = True
 
-  def build(self, filename, bytecode_compile=True):
+  def build(self, filename, bytecode_compile=True, verify_entry_point=False):
```
wisechengyi: This is the interface pants will call, correct?

changing it in a way to keep backwards compatibility.

pex/pex_builder.py
```diff
@@ -224,6 +226,58 @@ def set_script(self, script):
...
+      ep_module = ep_split[0]
+      import_statement = 'import {}'.format(ep_module)
+    elif len(ep_split) == 2:
+      ep_module = ep_split[0]
+      ep_method = ep_split[1]
+      import_statement = 'from {} import {}'.format(ep_module, ep_method)
+    else:
+      raise self.InvalidEntryPoint("Fail to parse: `{}`".format(entry_point))
+
+    args = [self._interpreter.binary, '-c', import_statement]
```
jsirois: Execution via subprocess was consolidated to [`Executor`](https://github.com/pantsbuild/pex/blob/c07c70f31d8070f26f45b31f84f84741ed2593a1/pex/executor.py#L87-L98), so this code should use that instead of `subprocess` directly.

A bit more true to the runtime environment the entry point will be executed in would be something like:
```python
with named_temporary_file() as fp:
  fp.write(import_statement)
  fp.close()
  result = PEX(self.chroot().path()).run([fp.name], env={'PEX_INTERPRETER': '1'})
  if result != 0:
    raise self.InvalidEntryPoint('Failed to validate entrypoint {}'.format(entry_point))
```
This though unfortunately creates a circular PEX <-> PEXBuilder dep. Better I think to lift the check to the PEX constructor which could be plumbed both from pex.bin.pex and from API users (Pants).

pex/pex_builder.py
jsirois: Yes, but see above.

pex/pex.py
```diff
@@ -520,3 +522,31 @@ def run(self, args=(), with_chroot=False, blocking=True, setsid=False, **kwargs)
...
+    # Only module is specified
+    if len(ep_split) == 1:
+      ep_module = ep_split[0]
+      import_statement = 'import {}'.format(ep_module)
+    elif len(ep_split) == 2:
+      ep_module = ep_split[0]
+      ep_method = ep_split[1]
+      import_statement = 'from {} import {}'.format(ep_module, ep_method)
+    else:
+      raise InvalidEntryPoint("Fail to parse: `{}`".format(entry_point))
```
jsirois: Failure or Failed would make more sense here.

pex/bin/pex.py
```diff
@@ -741,7 +753,7 @@ def main(args=None):
     pex_builder.freeze()
 
     log('Running PEX file at %s with args %s' % (pex_builder.path(), cmdline), v=options.verbosity)
-    pex = PEX(pex_builder.path(), interpreter=pex_builder.interpreter)
+    pex = PEX(pex_builder.path(), interpreter=pex_builder.interpreter, verify_entry_point=options.validate_ep)
```
jsirois: >100

pex/bin/pex.py
```diff
@@ -386,6 +386,15 @@ def configure_clp_pex_entry_points(parser):
       help='Set the entry point as to the script or console_script as defined by a any of the '
            'distributions in the pex.  For example: "pex -c fab fabric" or "pex -c mturk boto".')
 
+  group.add_option(
+      '--validate-entry-point',
+      dest='validate_ep',
+      default=False,
+      action='store_true',
+      help='Validate the entry point by importing it in separate process. Warning: this could have '
+           'side effect. For example, entry point `a.b.c:m` will translate to '
```
jsirois: 'side effects.

pex/bin/pex.py
```diff
@@ -732,6 +741,9 @@ def main(args=None):
       tmp_name = options.pex_name + '~'
       safe_delete(tmp_name)
       pex_builder.build(tmp_name)
+      if options.validate_ep:
```
jsirois: You could be more uniform, more correct and fail faster with:
```python
pex_builder.freeze()
pex = PEX(pex_builder.path(), interpreter=pex_builder.interpreter, verify_entry_point=options.validate_ep)

if options.pex_name is not None:
  ...
```
As things stand, in the options.pex_name branch, the pex is zipped up before the entrypoint validation which could be wasted work. Worse though is that the builder's custom interpreter, is any, is not used. This could be a problem if the current running interpreter does not match what the built pex needs and syntax differs (a 2/3 split).

pex/pex.py
```diff
@@ -520,3 +522,31 @@ def run(self, args=(), with_chroot=False, blocking=True, setsid=False, **kwargs)
                                     stderr=kwargs.pop('stderr', None),
                                     **kwargs)
     return process.wait() if blocking else process
+
+  def do_entry_point_verification(self):
```
jsirois: Given comments above re uniformity in performing the validation, I'd make this _private.

wisechengyi:
looks like `PEX.run` would ignore `-c` arg.
```
$ python '/var/folders/yt/pxdd3gp96530xh_116kskmkh0000gn/T/tmp8rQWfk' -c 'print 123'
hello
```
executed the entry point (print hello) rather than `-c` which prints 123.

pex/pex.py
wisechengyi: Corrected.

pex/bin/pex.py
wisechengyi: Corrected.

pex/bin/pex.py
wisechengyi: Corrected.

pex/bin/pex.py
wisechengyi: Moved things up to fail fast.

pex/pex.py
wisechengyi: Corrected.

wisechengyi:
Updated. Tests incoming.

wisechengyi:
Tests added. Ready for final review.

PR #540, opened 2018-08-18
Preserve perms of files copied to pex chroots.

### LINKED ISSUE #536: File perms are not preserved for `pex.common.Chroot` in copy mode.
In particular, if an executable resource is added to a pex chroot, it is not executable from within the chroot.

jsirois:
Failing tests are added that now pass.

Fixes #536

### DIFF
--- a/pex/common.py
+++ b/pex/common.py
@@ -25,7 +25,7 @@ def die(msg, exit_code=1):
 def safe_copy(source, dest, overwrite=False):
   def do_copy():
     temp_dest = dest + uuid4().hex
-    shutil.copyfile(source, temp_dest)
+    shutil.copy(source, temp_dest)
     os.rename(temp_dest, dest)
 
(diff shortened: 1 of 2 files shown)


pex/common.py
```diff
@@ -348,7 +348,7 @@ def __str__(self):
   def delete(self):
     shutil.rmtree(self.chroot)
 
-  def zip(self, filename, mode='wb'):
+  def zip(self, filename, mode='w'):
```
kwlzn: opening a zip in non-binary mode strikes me as a bit odd.. what's the reasoning behind this?

jsirois: It's not even a thing! That was a bug. Zipfile only supports a subset of modes - this is not an open pass-thru.

jsirois: To be more clear, the one user passed `mode='a'`, avoiding the bug.

PR #541, opened 2018-08-18
Expand the range of supported setuptools.

jsirois:
We now just exclude the island where setuptools ran a disruptive
de-vendoring experiment.

### DIFF
--- a/pex/version.py
+++ b/pex/version.py
@@ -5,8 +5,6 @@ __version__ = '1.4.5'
 
 # Versions 34.0.0 through 35.0.2 (last pre-36.0.0) de-vendored dependencies which causes problems
 # for pex code so we exclude that range.
-__exclusions = '!=34.*,!=35.*'
-SETUPTOOLS_REQUIREMENT = 'setuptools>=20.3,<41,{exclusions}'.format(exclusions=__exclusions)
-del __exclusions
+SETUPTOOLS_REQUIREMENT = 'setuptools>=20.3,<41,!=34.*,!=35.*'
 
 WHEEL_REQUIREMENT = 'wheel>=0.26.0,<0.32'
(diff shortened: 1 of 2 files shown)


jsirois:
I manually tested `pex --pre` with setuptools versions inside and outside the excluded range, only getting failures in the excluded range. I used https://github.com/pypa/setuptools/blob/master/CHANGES.rst to find the exclusion range in the first place.

The motivation here is to let Pants use a newer version of setuptools so that it can run under python 3.7. See: https://github.com/pantsbuild/pants/issues/6363

pex/version.py
```diff
@@ -3,9 +3,10 @@
...
 
-# NB: If we upgrade to setuptools>=34 pex's bootstrap code in `PEXBuilder` will need an update to
-# include the `packaging` package in the `.bootstrap/` code since we use
-# `packaging.specifiers.SpecifierSet` - indirectly - through `pkg_resources.Requirement.specifier`.
-SETUPTOOLS_REQUIREMENT = 'setuptools>=20.3,<34.0'
+# Versions 34.0.0 through 35.0.2 (last pre-36.0.0) de-vendored dependencies which causes problems
+# for pex code so we exclude that range.
+__exclusions = '!=34.*,!=35.*'
+SETUPTOOLS_REQUIREMENT = 'setuptools>=20.3,<41,{exclusions}'.format(exclusions=__exclusions)
+del __exclusions
```
kwlzn: the use and explicit `del` of `__exclusions` struck me as odd here. any reason to not keep the exclusions around as a constant (like `SETUPTOOLS_EXCLUSIONS`) - or just elide and hardcode in the string literal?

jsirois: Agreed - I'll inline. This was half remembering wickmanisms and half wishing for assignment blocks:
```
x = block:
  foo
  # NB: bar is very very worth exposition
  bar

  'baz'

assert x == 'baz'
```

PR #545, opened 2018-08-25
Support -c for plat spec dists in multiplat pexes.

### LINKED ISSUE #537: `pex -c` does not work for multiplatform pexes when the target dist is platform specific
If the console script is housed in a dist that is platform-specific and the pex is multiplatform, then the same script is found more than once and pex creation fails with something like:
```
RuntimeError: Ambiguous script specification pants matches multiple entry points:pants = pants.bin.pants_loader:main pants = pants.bin.pants_loader:main
```
Pex should probably allow dups from a single requirement key if the found scripts are bytewise identical as they are in this example case.

See https://github.com/pantsbuild/pants/pull/6349 for an example of getting bit by this.

jsirois:
Fixes #537

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,7 +61,7 @@ matrix:
       env: TOXENV=pypy-integration
 
 install:
-  - pip install -U tox "setuptools<34"
+  - pip install -U tox "setuptools>=36"
 
 script:
   - ./scripts/ci.sh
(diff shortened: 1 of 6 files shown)


jsirois:
Eeenteresting. So setuptools 36.2.0+ take `install_requires` with environment markers and move them to anonymous `extras_requires`: [see here](https://github.com/pypa/setuptools/pull/1081)). This 'feature' is only documented [here in the wheel docs](https://wheel.readthedocs.io/en/stable/index.html#defining-conditional-dependencies). I'll back off the un-needed requirement upgrades - just a pytest bump was needed - and file an issue to handle this new wrinkle.

PR #546, opened 2018-08-30
Fix for super() usage on "old style class" ZipFile

cbrinker:
I was experiencing an error of "TypeError: must be type, not classobj"
on the line using `super()._extract_member()` below.
This error appears to be due to the `zipfile.ZipFile` class never inheriting from
`object` which is a requirement since py2.2 (2001).

Since the use of `super()` in `PermPreservingZipFile()` depends on
`object` being somewhere in the inheritance tree, I am including it in
the class definition of `PermPreservingZipFile()`. My testing of this patch
locally has proven to solve the "TypeError" I was seeing.

### DIFF
--- a/pex/common.py
+++ b/pex/common.py
@@ -78,7 +78,7 @@ class MktempTeardownRegistry(object):
 _MKDTEMP_SINGLETON = MktempTeardownRegistry()
 
 
-class PermPreservingZipFile(zipfile.ZipFile):
+class PermPreservingZipFile(zipfile.ZipFile, object):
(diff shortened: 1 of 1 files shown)


jsirois:
CI is continuing to fail on 1 shard despite me re-running it a few times. It does not likely to be related to the change, but one thing that seems a bit off is multiple-inheriting from object instead of using old-style super to go along with the old-style baseclass. IE:
```python
result = PermPreservingZipFile._extract_member(self, member, targetpath, pwd)
```

jsirois:
Alrighty - got this green with another shard restart :/ Thanks @cbrinker !

PR #552, opened 2018-09-24
Add OSX shards.

jsirois:


jsirois:
Not for review - lots of work to do here 1st - just kicking CI tires.

PR #554, opened 2018-09-25
Fixup `PEX.demote_bootstrap`: fully unimport.

### LINKED ISSUE #550: Pex 1.4.6 builds a failing pex
pex 1.4.6 will allow you to build a pex against the [responses](https://github.com/getsentry/responses) library, but blows up when you try to actually import it.

This works fine outside of a pex environment and with pex 1.4.5.

```
 ~/sandbox $ . venv/bin/activate
(venv)  ~/sandbox $ python -V
Python 2.7.15
(venv)  ~/sandbox $ pex --version
pex 1.4.6
(venv)  ~/sandbox $ pex responses
Python 2.7.15 (default, May  1 2018, 16:44:08)
[GCC 4.2.1 Compatible Apple LLVM 9.1.0 (clang-902.0.39.1)] on darwin
(InteractiveConsole)
>>> import responses
  File "/private/var/folders/mk/bz4w0f3505190h3n4x21_jlc0000gn/T/tmpReh1br/.deps/responses-0.9.0-py2.py3-none-any.whl/responses.py", line 41, in <module>
    import mock as std_mock
  File "/private/var/folders/mk/bz4w0f3505190h3n4x21_jlc0000gn/T/tmpReh1br/.deps/mock-2.0.0-py2.py3-none-any.whl/mock/__init__.py", line 2, in <module>
    import mock.mock as _mock
  File "/private/var/folders/mk/bz4w0f3505190h3n4x21_jlc0000gn/T/tmpReh1br/.deps/mock-2.0.0-py2.py3-none-any.whl/mock/mock.py", line 71, in <module>
    _v = VersionInfo('mock').semantic_version()
  File "/private/var/folders/mk/bz4w0f3505190h3n4x21_jlc0000gn/T/tmpReh1br/.deps/pbr-4.2.0-py2.py3-none-any.whl/pbr/version.py", line 462, in semantic_version
    self._semantic = self._get_version_from_pkg_resources()
  File "/private/var/folders/mk/bz4w0f3505190h3n4x21_jlc0000gn/T/tmpReh1br/.deps/pbr-4.2.0-py2.py3-none-any.whl/pbr/version.py", line 439, in _get_version_from_pkg_resources
    import pkg_resources
  File "/private/var/folders/mk/bz4w0f3505190h3n4x21_jlc0000gn/T/tmpReh1br/.bootstrap/pkg_resources/__init__.py", line 77, in <module>
    from . import py31compat
ImportError: cannot import name py31compat
```

jsirois:
Previously only root 3rdparty packages copied into the bootstrap were
unimported leaving subpackages in-place. This could lead to a mismatch
in 3rdparty package API expectations when a re-imported 3rdparty root
package from the user PEX sys.path was from a different version of the
3rdparty package.

Fixes #550

### DIFF
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -419,6 +419,9 @@ class PEX(object):  # noqa: T000
         TRACER.log('Un-importing third party bootstrap dependency %s from %s'
                    % (mod, bootstrap_path))
         sys.modules.pop(mod)
+        submod_prefix = mod + '.'
+        for submod in [m for m in sys.modules.keys() if m.startswith(submod_prefix)]:
+          sys.modules.pop(submod)
     sys.path.pop(bootstrap_path_index)
     sys.path.append(bootstrap_path)
 
--- a/tests/test_pex.py
+++ b/tests/test_pex.py
@@ -6,7 +6,6 @@ import subprocess
 import sys
 import textwrap
 from contextlib import contextmanager
-from pex.interpreter import PythonInterpreter
 from types import ModuleType
 
 import pytest
@@ -15,6 +14,7 @@ from twitter.common.contextutil import temporary_file
 from pex.bin.pex import get_interpreter
 from pex.compatibility import PY2, WINDOWS, nested, to_bytes
 from pex.installer import EggInstaller, WheelInstaller
+from pex.interpreter import PythonInterpreter
 from pex.pex import PEX
 from pex.pex_builder import PEXBuilder
 from pex.pex_info import PexInfo
--- a/tests/test_pex.py
+++ b/tests/test_pex.py
@@ -6,6 +6,7 @@ import subprocess
 import sys
 import textwrap
 from contextlib import contextmanager
+from pex.interpreter import PythonInterpreter
 from types import ModuleType
 
 import pytest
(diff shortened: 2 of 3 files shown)


jsirois:
NB: The new test fails before the fix.

jsirois:
@DrumMachineLearning, @zsimic & @zvezdan: if any of you all could try this out prior to release I'd be grateful.

jsirois:
#555 tracks a hotfix release for this.

tests/test_pex.py
```diff
@@ -373,3 +374,29 @@ def test_pex_run_custom_setuptools_useable():
       )
       rc = PEX(pex.path()).run()
       assert rc == 0
+
+
+def test_pex_run_conflicting_custom_setuptools_useable():
+  # Here we use an older setuptools to build the pex which has a newer setuptools requirement.
```
Eric-Arellano: Are we testing these specific setuptool versions, or more generally setuptools conflicting? If someone changed from 40.4.3 to 44.2.1 in a year, would that be testing something different?

Would be helpful to add a quick note saying "These specific version numbers matter due to API changes" etc. vs "The particular version numbers do not matter, rather simply having different versions matters."

pex/pex.py
```diff
@@ -419,6 +419,9 @@ def demote_bootstrap(cls):
         TRACER.log('Un-importing third party bootstrap dependency %s from %s'
                    % (mod, bootstrap_path))
         sys.modules.pop(mod)
+        submod_prefix = mod + '.'
+        for submod in [m for m in sys.modules.keys() if m.startswith(submod_prefix)]:
```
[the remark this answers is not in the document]
jsirois: I'm pretty sure I can't. Python 3 complains of concurrent modification - need a snapshot of the underlying dict keys.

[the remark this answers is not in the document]
Eric-Arellano: Ah you're totally right. Didn't make the connection you're changing the size of `sys.modules`

tests/test_pex.py
jsirois: I was trying to point out these specific versions numbers by including the comment diff. IE: these two specific versions work for the test since the exhibit the illustrated API diff.

tests/test_pex.py
```diff
@@ -373,3 +374,29 @@ def test_pex_run_custom_setuptools_useable():
...
+                                                include_site_extras=False)
+    dists = resolve(['setuptools==40.4.3'], cache=resolve_cache)
+    with temporary_dir() as temp_dir:
+      pex = write_simple_pex(
+        temp_dir,
+        'from pkg_resources import appdirs, py31compat',
+        dists=dists,
+        interpreter=interpreter
+      )
+      rc = PEX(pex.path()).run()
```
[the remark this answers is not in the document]
jsirois: This is re-using the convention in the surrounding file.

tests/test_pex.py
Eric-Arellano: Okay makes sense. Consider leading with a comment like "Setuptools changed one of its APIs in version x." Then the rest you have below and diff. 

It's fine as is if you don't want to change, but might make things even more clear.

jsirois: OK - I'm not even sure what version it changed in! I just took a low and a high from the pex supported range. I think I'll leave as is.

jsirois:
Thanks a bunch for the review Eric!

PR #558, opened 2018-09-27
Ignore concurrent-rename failures.

### LINKED ISSUE #557: non-zip_safe pex extraction has a race on renaming
If a particular non-zip_safe pex is run multiple times in parallel, there can be a race unpacking it into the unsafe_zip_cache_dir.

https://github.com/pantsbuild/pex/blob/1600d5b8d4a3a85fa8c6e17d0525da794a1f915a/pex/environment.py#L38-L52

Possible fixes
- add a file lock
- use a construct like 'safe_concurrent_creation` in pants does

https://github.com/pantsbuild/pants/blob/5d850fcadeb3da5e3a3fa075595f9e3bf2e32a47/src/python/pants/util/dirutil.py#L361

stuhood:
Fixes #557.

### DIFF
--- a/pex/environment.py
+++ b/pex/environment.py
@@ -3,6 +3,7 @@
 
 from __future__ import absolute_import, print_function
 
+import errno
 import itertools
 import os
 import site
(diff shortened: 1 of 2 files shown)


pex/environment.py
```diff
@@ -49,7 +50,11 @@ def force_local(cls, pex, pex_info):
           safe_rmtree(explode_tmp)
           raise
       TRACER.log('Renaming %s to %s' % (explode_tmp, explode_dir))
-      os.rename(explode_tmp, explode_dir)
+      try:
```
jsirois: How about using this which is slightly more correct (it cleans up on rename failure) and already tested - so you can maybe drop your test:
https://github.com/pantsbuild/pex/blob/47c26d746046a8ab21bd3cb3bb782ce5f018a369/pex/common.py#L160-L172

jsirois:
You're getting hit by #561, I'll see if I can fix that up.

PR #559, opened 2018-09-27
Sort requirements before dumping to PEX-INFO

AlexHill:
I have a goal of making pex builds reproducible given the right conditions, and this is one easy first step. Because at various points during requirement resolution the dependencies are put into sets and dicts, the order is unreliable and changes from build to build.

### DIFF
--- a/pex/pex_info.py
+++ b/pex/pex_info.py
@@ -301,7 +301,7 @@ class PexInfo(object):
 
   def dump(self, **kwargs):
     pex_info_copy = self._pex_info.copy()
-    pex_info_copy['requirements'] = list(self._requirements)
+    pex_info_copy['requirements'] = list(sorted(self._requirements))
     pex_info_copy['interpreter_constraints'] = list(self._interpreter_constraints)
     pex_info_copy['distributions'] = self._distributions.copy()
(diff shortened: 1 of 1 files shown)


jsirois:
You're getting hit by #561, I'll see if I can fix that up.

AlexHill:
Thanks, I was wondering!

jsirois:
Do you mind rebasing / merging to pick up the CI fix and adding a test? In isolation this change serves no obvious purpose; so it could easily be removed in the future with no context present to say why we sort.

AlexHill:
Sure can.

Out of interest, has the topic of reproducible PEX builds come up before? For my purposes, I think exposing a flag to turn off pre-compilation in addition to this change would do the trick once this change is merged. I'll open another PR for that shortly.

jsirois:
> ... For my purposes, I think exposing a flag to turn off pre-compilation in addition to this change would do the trick.

Sounds good to me. You probably found this already, but the flag would plumb here:
https://github.com/pantsbuild/pex/blob/e45b07bc3d3bb36d28ce00bd4ecaf6416fe5feee/pex/pex_builder.py#L462-L470

AlexHill:
This is fixed now, thanks!

PR #563, opened 2018-09-28
Add support for `-m` and `-c` in interpreter mode.

### LINKED ISSUE #547: Execute pex archive at runtime with -m parameter
We would like to create a pex without a fixed entry point and provide different entry points at runtime.
The doc indicates that this can be achieved  by tuning the environment variable  PEX_MODULE

> PEX_MODULE
Override the entry point into the PEX file. Can either be a module, e.g. ‘SimpleHTTPServer’, or a specific entry point in module:symbol form, e.g. “myapp.bin:main”.
>
https://pex.readthedocs.io/en/stable/api/index.html#module-pex.variables

This indeed works but is a bit counter intuitive. 
It would be nice to provide the entry point with the -m argument similar to how the python interpreter works but this doesn't work
`./mypexfile.pex -m mymodule arg0 arg1 arg2`

Is this something that can be implemented or it there a technical restriction not to to this ? In theory as we already propagate the arguments in the __main__ file we could also remove the beginning and redirect to the right module ?

jsirois:
This emulates the behavior of the python interpreter as we already do for
`-` and <file>.

Fixes #547

### DIFF
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -431,18 +431,29 @@ class PEX(object):  # noqa: T000
     TRACER.log('  * - paths that do not exist or will be imported via zipimport')
 
   def execute_interpreter(self):
-    if sys.argv[1:]:
-      program = sys.argv[1]
-      try:
-        if program == '-':
-          content = sys.stdin.read()
-        else:
-          with open(program) as fp:
-            content = fp.read()
-      except IOError as e:
-        die("Could not open %s in the environment [%s]: %s" % (program, sys.argv[0], e))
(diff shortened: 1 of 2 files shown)


jsirois:
Thanks Stu. I think we're good. Previously we supported 2 of exactly 4 execution modes of cpython:
```
$ pyyhon -h
usage: python [option] ... [-c cmd | -m mod | file | -] [arg] ...
...
```

fhoering:
I had a look. Perfect. Thanks for the fix.

PR #565, opened 2018-09-29
Add OSX shards.

### LINKED ISSUE #479: add an OSX shard
this would be useful for OSX-specific tests and generally for improved runtime platform coverage.

jsirois:
Fixes #479


jsirois:
I've filed #566 for 3.7 support. That will now be straightforward to add - and if I switch the linux shards to using pyenv, it will be at ~ no time cost since we're caching the pyenv already.

PR #567, opened 2018-09-29
Introduce 3.7 to CI and release.

### LINKED ISSUE #566: Add CPython 3.7 to CI
It's been out for a while and with pyenv support working for OSX, worst case we could also use it on linux to get 3.7

jsirois:
Fixes #566


.travis.yml
```diff
@@ -1,133 +1,192 @@
...
+# use TOXENV and instead pass the toxenv via -e on the command line. This helps ensure we share
+# caches as much as possible (eg: all linux python 2.7.15 shards share a cache).
 matrix:
   include:
     - <<: *x-linux-shard
-      python: "2.7"
       name: TOXENV=style
-      script: ./scripts/ci.sh style
+      env: *x-py27
+      script: tox -ve style
```
Eric-Arellano: Are there long flag names for this? I'm not sure what `-v` and `-e` are supposed to mean.

tests/test_compiler.py
```diff
@@ -45,6 +46,9 @@ def test_compile_success():
       compiled_abspath = os.path.join(root, compiled)
       with open(compiled_abspath, 'rb') as fp:
         fp.read(4)  # Skip the magic header.
+        if sys.version_info[:2] >= (3, 7):
```
Eric-Arellano: consider adding `compatibility.PY37`, so this would be the more readable `if compatibility.PY37:`

tests/test_integration.py
```diff
@@ -32,6 +32,12 @@
 from pex.util import DistributionHelper, named_temporary_file
 
 
+def make_env(**kwargs):
+  env = os.environ.copy()
+  env.update((k, str(v)) for k, v in kwargs.items())
```
Eric-Arellano: This works fine for Py2 vs Py3 interop? `str` in Py2 will have `v` be bytes, but Py3 unicode.

I see it's test code so probably fine.

Eric-Arellano: Wish we had `future` library in PEX. I know we discussed not spending time rewriting Py2 vs Py3 compatibility to use the Py3-first approach, but perhaps it would be worth installing to allow new and refactored code to use the `future` backports.

jsirois: > This works fine for Py2 vs Py3 interop?

.travis.yml
jsirois: Yeah - I'll roll this up into an env var. I was avoiding doin that just because changing an env var kills the caches and the shards go from 4 minutes to ~20. Better to do that now though before this hits master to seal a good re-usable master cache set in for future PRs.

tests/test_compiler.py
jsirois: I was reluctant to do that since compatibility is a prod module and the only use would be here in a test.

.travis.yml
jsirois: No long opts for these, but fixed with a documented factor-up.

tests/test_compiler.py
Eric-Arellano: That sounds fine to keep as is.

If we were to make the change, we would also need to add a variable for every major version, e.g. `PY34` and `PY35`, which is outside the scope of this PR.

Eric-Arellano:
Thanks John, changes look good! 

One concern - it looks like we're only testing Py27 and Py37, correct? If so, we might want to consider testing at least 3.4 or 3.5 as well, because the behavior is different with inconsistent dictionary ordering.

jsirois:
No, we're testing 2.7, 3.4, 3.5, 3.6, 3.7 + pypy (on linux). On OSX only 2.7 and 3.7, but that's for expediency.

jsirois:
Alright - switching to ${TOX} caused unexpected issues. A bit of thrash since you last looked, but the upshot is reverting to using scripts/ci.sh with the pypi workaround. I moved the tox -v and -e comments there.

Issue #523, opened 2018-07-13
OSX's python 2.7.10 interpreter reports a bad local platform, bdists fail to resolve


jsirois:
In particular, under the OSX system 2.7.10 / xcode 8.3 (on TravisCI), the reported platform is `macosx-10.12-intel` and the resulting calculated tags are:
```
[...]

PR #568, opened 2018-10-01
WIP: Failing test for #523

jsirois:


pex/environment.py
```diff
@@ -120,13 +120,14 @@ def __init__(self, pex, pex_info, interpreter=None, **kw):
     self._interpreter = interpreter or PythonInterpreter.get()
     self._inherit_path = pex_info.inherit_path
     self._supported_tags = []
+
+    platform = Platform.current()
     super(PEXEnvironment, self).__init__(
       search_path=[] if pex_info.inherit_path == 'false' else sys.path,
+      platform=platform.platform,
```
Eric-Arellano: `platform.platform` is a bit confusing 🤔. I think it's outside the scope of this PR to rename the `Platform` object and/or its field `.platform`, but in the meantime this variable could be renamed perhaps, or maybe a brief note what the difference is.

pex/interpreter.py
```diff
@@ -261,8 +261,10 @@ def __str__(self):
     )
 
   def __repr__(self):
-    return 'PythonIdentity(%r, %s, %s, %s)' % (
-      self._interpreter,
+    return 'PythonIdentity(%r, %r, %r, %r, %r, %r)' % (
```
Eric-Arellano: Oh how I wish we could use f-strings. 

I would say we should change this to new style string interpolation, but I know PEX always uses old style so this is fine.

pex/resolver.py
```diff
@@ -164,10 +164,52 @@ class Resolver(object):
...
+  def _expand_and_maybe_adjust_platform(interpreter, platform=None):
+    # Adjusts `platform` if it is 'current' and does not match the given `interpreter` platform.
+    cur_plat = Platform.current()
+
+    given_platform = Platform.create(platform or 'current')
+    if cur_plat.platform != given_platform.platform:
+      # IE: Say we're on OSX and platform was 'linux-x86_64' or 'linux_x86_64-cp-27-cp27mu'.
+      return given_platform
+
+    ii = interpreter.identity
```
Eric-Arellano: The brevity is fine right here, but by line 186 I had to look back and ask what is `ii` again? 

Is the intent of `ii` to make things shorter, or is there also a cost to accessing `interpreter.identity` that you're trying to cache to avoid recomputing every time? If only the former, then I think it would be more readable to use `interpreter.identity` everywhere, such as `interpreter.identity.abbr_impl`.

pex/resolver.py
```diff
@@ -164,10 +164,52 @@ class Resolver(object):
...
+
+    # Otherwise we need to adjust the platform to match a local interpreter different from the
+    # currently executing interpreter.
+    adjusted_platform = Platform(platform=cur_plat.platform,
+                                 impl=ii.abbr_impl,
+                                 version=ii.impl_ver,
+                                 abi=ii.abi_tag)
+
+    TRACER.log("""
+  Modifying given platform of {given_platform!r}:
```
Eric-Arellano: This string interpolation is very readable. Well used!

tests/test_environment.py
```diff
@@ -95,3 +108,57 @@ def test_load_internal_cache_unzipped():
...
+                        stderr=subprocess.PIPE)
+      stdout, stderr = process.communicate()
+      assert 0 == process.returncode, (
+        'Process failed with exit code {} and stderr:\n{}'.format(process.returncode, stderr)
+      )
+
+      # Verify this all worked under the previously problematic pkg_resources-reported platform.
+      release, _, _ = platform.mac_ver()
+      major_minor = '.'.join(release.split('.')[:2])
+      assert 'macosx-{}-intel'.format(major_minor) == stdout.strip()
```
Eric-Arellano: IIRC, in Python3 `stdout` will be in bytes. If the tests fail, then you'll have to decode.

jsirois: You are correct, but I went with the simpler `b'macosx-{}-intel'...` since `b'' == ''` in python 2.

pex/environment.py
jsirois: Explicitified

pex/interpreter.py
jsirois: It does not - there is a mix. As a rule, pex or any other project, I try to stick to the conventions in the file unless on a pure cleaning change.

pex/resolver.py
jsirois: This is a bit of the style bit mentioned above but I've fixed. For context, Pex was written by @wickman and later maintained by @kwlzn and the predominant style is one they share which is favoring brevity in local scopes where you have the handful of lines right there for all the context you need.

tests/test_environment.py
jsirois: ...and bytes doesn't support format under python3 - fixed to encode to bytes.

PR #571, opened 2018-10-02
Minimize interpreter bootstrapping in tests.

jsirois:
Master is currently timing out trying to fetch and build 7 interpreters.
Since We have no need for seven, constrain the interpreter count. Along
the way fix a flaky test that exposed production code not setting up all
interpreters.

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -573,14 +573,14 @@ def setup_interpreter(interpreter, interpreter_cache_dir, repos, use_wheel):
 
 def build_pex(args, options, resolver_option_builder):
   with TRACER.timed('Resolving interpreters', V=2):
-    def to_python_interpreter(interpreter):
-      if os.path.exists(interpreter):
-        return PythonInterpreter.from_binary(interpreter)
+    def to_python_interpreter(full_path_or_basename):
+      if os.path.exists(full_path_or_basename):
+        return PythonInterpreter.from_binary(full_path_or_basename)
       else:
-        pi = PythonInterpreter.from_env(interpreter)
-        if pi is None:
-          die('Failed to find interpreter: %s' % interpreter)
-        return pi
+        interpreter = PythonInterpreter.from_env(full_path_or_basename)
+        if interpreter is None:
+          die('Failed to find interpreter: %s' % full_path_or_basename)
+        return interpreter
 
     interpreters = [to_python_interpreter(interp) for interp in options.python or [sys.executable]]
 
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -557,19 +557,7 @@ def resolve_interpreter(cache, fetchers, interpreter, requirement):
     return interpreter.with_extra(egg.name, egg.raw_version, egg.path)
 
 
-def get_interpreter(python_interpreter, interpreter_cache_dir, repos, use_wheel):
-  interpreter = None
-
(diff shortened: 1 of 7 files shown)


Eric-Arellano:
I'm not very familiar with the PEX directory structure. Is this all test code, or is any source code?

Will review next few minutes!

jsirois:
The only production code change is in pex/bin/pex.py here: https://github.com/pantsbuild/pex/pull/571/files#diff-178808e5641b15e64b2d09f744976f03

pex/bin/pex.py
```diff
@@ -585,13 +573,16 @@ def get_interpreter(python_interpreter, interpreter_cache_dir, repos, use_wheel)
...
-                      options.interpreter_cache_dir,
-                      options.repos,
-                      options.use_wheel)
-      for interpreter in options.python or [None]
-    ]
+    def to_python_interpreter(interpreter):
+      if os.path.exists(interpreter):
+        return PythonInterpreter.from_binary(interpreter)
+      else:
+        pi = PythonInterpreter.from_env(interpreter)
```
Eric-Arellano: I immediately thought of 3.14 haha. I would have wanted to name this variable `interpreter`, but see its already taken. Which raises the question, how is this interpreter different than the parameter? Is there a name we could use like `interpreter_from_env` that reflects the semantic difference?

tests/test_pex_bootstrapper.py
```diff
@@ -34,39 +34,25 @@ def test_get_pex_info():
...
 
-  interpreters = find_compatible_interpreters(pex_python_path, ['>3'])
-  assert interpreters[0].binary == pex_python_path.split(':')[3]  # 3.4.2
+  def find_interpreters(*constraints):
+    return [interp.binary for interp in find_compatible_interpreters(pex_python_path, constraints)]
 
-  interpreters = find_compatible_interpreters(pex_python_path, ['<3'])
-  assert interpreters[0].binary == pex_python_path.split(':')[0]  # 2.7.9
+  assert [py35, py36] == find_interpreters('>3')
+  assert [py27] == find_interpreters('<3')
```
Eric-Arellano: Nice, I really like how readable these tests are.

pex/bin/pex.py
jsirois: Explicitified.

jsirois:
This has brought the cold cache CI times on OSX down from 50+ minutes to 30- minutes - so looks like it'll do.

tests/test_pex_bootstrapper.py
```diff
@@ -34,39 +34,25 @@ def test_get_pex_info():
...
 
-  interpreters = find_compatible_interpreters(pex_python_path, ['>3.5.4'])
-  assert interpreters[0].binary == pex_python_path.split(':')[5]  # 3.6.2
+  assert [py36] == find_interpreters('>{}'.format(PY35))
```
CMLivingston: Part of the reason I made these tests at this level of granularity was to assure that patch-level version constraints between constraints with the same major versions are respected, however I would imagine that is covered in testing elsewhere so this change is fine with me.

jsirois: Understood. Yeah, the underlying comparison is done by `PythonIdentity.matches` which is actually not unit-tested, but it could and should be. The testing here was just way too expensive.

Issue #569, opened 2018-10-01
Pex docs unclear for python scrubs


asyncjake:
Hey there. Spoiler alert, I'm the title scrub

Part of an epic I'm on at work was to make python deployable, micro-service style. PEX fits the bill to the penny and it's currently working great. But, it relies so heavily on the project's local setup.py that I spent a few weeks debugging and triaging things that really should've been documented. Like if you bundle your project in a pex, to run -c blah.py, that blah.py *must* be declared in the scripts array. Or to run a web service like gunicorn, you have to bundle gunicorn in your install_requires, as gunicorn can't enter the pex and find the flask/django/whatever module to run

As part of Hacktober, I'm happy to update the docs with a "This is what pex expects from a setup" approach, or perhaps a section of information on "Integrating PEX into your CI/CD". Of course I'd reference and recommend Pants, but it was dramatic overkill for our use case, and invoking pex directly was precisely what we needed

Would a docs PR be welcome? I already have it documented in house and I promise to make it quality, just figured I'd start the discussion before going in on it

jsirois:
Bottom line - docs are very welcome.

Below I have questions intended to suss out if you found the docs we already provide.

> ... project in a pex, to run -c blah.py, that blah.py must be declared in the scripts array

Hopefully you saw the docs here:
```
$ pex --help | grep -A4 console-script
    -c SCRIPT_NAME, --script=SCRIPT_NAME, --console-script=SCRIPT_NAME
                        Set the entry point as to the script or console_script
                        as defined by a any of the distributions in the pex.
                        For example: "pex -c fab fabric" or "pex -c mturk
                        boto".
```
As a python scrub though `script` is probably too familiar a word whereas `console_script` might catch your eye as different and yet both are referring to `setup.py` entries of the same name. I can see the confusion and higher level docs would be welcome. Take a gander [here](https://pex.readthedocs.io/en/stable/buildingpex.html#pex-c) though in case you missed it and apply more-clear-making-words here:

https://github.com/pantsbuild/pex/blob/master/docs/buildingpex.rst#L209-L234

> Or to run a web service like gunicorn, you have to bundle gunicorn in your install_requires, as gunicorn can't enter the pex and find the flask/django/whatever module to run

This I'm not sure I understand. Is this an instance of not zip safe?
```
$ pex --help | grep -A8 zip-safe
    --zip-safe, --not-zip-safe
                        Whether or not the sources in the pex file are zip
[...]

PR #574, opened 2018-10-03
Documentation for #569

jakethedev:
Adding information on gunicorn with pex, plus a note on a common error to encounter with `pex -c`

### DIFF
--- a/docs/buildingpex.rst
+++ b/docs/buildingpex.rst
@@ -231,6 +231,11 @@ Even scripts defined by the "scripts" section of a distribution can be used, e.g
                  {bal,hit,hits,new,extend,expire,rm,as,approve,reject,unreject,bonus,notify,give-qual,revoke-qual}
                  ...
     mturk: error: too few arguments
+    
+Note: If you run ``pex -c`` and come across an error similar to 
+``pex.pex_builder.InvalidExecutableSpecification: Could not find script 'mainscript.py' in any distribution within PEX!``, 
+double-check your setup.py and ensure that ``mainscript.py`` is included 
+in your setup's ``scripts`` array.
 
 
 Saving .pex files
--- a/docs/recipes.rst
+++ b/docs/recipes.rst
@@ -1,0 +1,51 @@
+.. _recipes:
+
+PEX Recipes and Notes
+=====================
+
+This is a space for info beyond the purpose or building of PEX files.
+
+Gunicorn and PEX
+----------------
+
+Normally, to run a wsgi-compatible application with Gunicorn, you'd just
+point Gunicorn at your application, tell Gunicorn how to run it, and you're
+ready to go - but if your application is shipping as a PEX file, you'll have
+to bundle Gunicorn as a dependency and set Gunicorn as your entry point. Gunicorn
+can't enter a PEX file to retrieve the wsgi instance, but that doesn't prevent
+the PEX from invoking Gunicorn.
+
+This retains the benefit of zero `pip install`'s to run your service, but it
+requires a bit more setup. First, ensure Gunicorn is packaged as a depenency, 
+the following assumes Flask as the wsgi framework, Django setup should be 
+similar:
+
+.. code-block:: bash
+
+    $ pex flask gunicorn myapp -e gunicorn.app.wsgiapp -o ~/service.pex
+    
+The dependency on Gunicorn can live in your setup.py as well:
+    
+.. code-block:: python
+
+    setup(
+        ...
+        modules=[
+          myapp
+        ],
+        install_requires=[
+          ...other deps...,
+          flask,
+          gunicorn
+        ]
+    )
+
+Once your pex file is created, you need to make sure to pass your wsgi app
+instance name to the CLI at runtime for Gunicorn to know how to hook into it,
+configuration can be passed in the same way:
+
+.. code-block:: bash
+
+  $ service.pex myapp:appinstance -c /path/to/gunicorn_config.py
+
+And there you have it, a fully portable python web service. 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -31,4 +31,5 @@ Guide:
 
    whatispex
    buildingpex
+   recipes
    api/index


docs/recipes.rst
```diff
@@ -0,0 +1,51 @@
...
+This retains the benefit of zero `pip install`'s to run your service, but it
+requires a bit more setup. First, ensure Gunicorn is packaged as a depenency, 
+the following assumes Flask as the wsgi framework, Django setup should be 
+similar:
+
+.. code-block:: bash
+
+    $ pex flask gunicorn myapp -e gunicorn.app.wsgiapp -o ~/service.pex
+    
+The dependency on Gunicorn can live in your setup.py as well:
```
jsirois: Although true, if the app has 0 lines of code that import from gunicorn this is not a clean recommendation. Consider dropping.

docs/buildingpex.rst
```diff
@@ -231,6 +231,11 @@ Even scripts defined by the "scripts" section of a distribution can be used, e.g
                  {bal,hit,hits,new,extend,expire,rm,as,approve,reject,unreject,bonus,notify,give-qual,revoke-qual}
                  ...
     mturk: error: too few arguments
+    
+Note: If you run ``pex -c`` and come across an error similar to 
+``pex.pex_builder.InvalidExecutableSpecification: Could not find script 'mainscript.py' in any distribution within PEX!``, 
+double-check your setup.py and ensure that ``mainscript.py`` is included 
+in your setup's ``scripts`` array.
```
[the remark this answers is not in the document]
jakethedev: Interesting, I didn't realize there was a way to invoke specific functions that way, that's super cool. I'll update the note to include the console_scripts approach and a link to the packaging docs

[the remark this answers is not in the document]
jsirois: > Interesting, I didn't realize there was a way to invoke specific functions that way

Nitpicking your words to make sure things are _crystal clear_:
You can specify an entry point in 3 ways when building a pex:
1. -c <named console script>: these are entrypoint functions granted names in setup.py
2. -c <named script>: these are actual scripts granted names in setup.py
3. -e <entrypoint>: this is an arbitrary entrypoint function; ie: `myapp.submodule:nullary_function_name`

When building a pex, method 3 is the only way to have a specific arbitrary function invoked. Method 1 only allows pre-named functions ("console scripts") to be invoked.

docs/recipes.rst
jakethedev: Fair point, it is a bit of a hack. I added the option as it's a cleaner approach for build systems. In terms of code, the app definitely doesn't rely on Gunicorn - especially with Flask included as well - but from an ops view Gunicorn is required for the app to go live, and it's easier to have a standard PEX invocation at compile time that leans more on setup.py than to have special build configuration for each wsgi app.

Would it be more appropriate to shorten that bit to just line 27, and instead of a code block, have a caveat that it may make CI/CD easier but is not recommended when the app isn't using Gunicorn imports?

jsirois: Also a fair point, but I think at the wrong scope for these docs. I'd leave it out. You've already taught the reader how to use _pex_ to create a self-contained pex with gunicorn / wsgi container included.

If they want to use a pants BUILD file with a `python_binary` definition with the extra deps, write a side-car setup.py, a setup.py that depends on the original, or even modify the original setup.py except use [`extras_require`](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies) instead of `install_requires` (say with an extra requirement list named 'deployable': `pex myapp[deployable] -e gunicorn.app.wsgiapp -o ~/service.pex), then they can probably figure that out.

docs/buildingpex.rst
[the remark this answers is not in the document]
jakethedev: I appreciate the clarity, I've been going on this project in a breadth-first way so it's nice to have that spelled out

docs/recipes.rst
jakethedev: Great points, I dig it - redacted that section

docs/recipes.rst
```diff
@@ -0,0 +1,33 @@
...
+the PEX from invoking Gunicorn.
+
+This retains the benefit of zero `pip install`'s to run your service, but it
+requires a bit more setup as you must ensure Gunicorn is packaged as a dependency. 
+The following snippets assume Flask as the wsgi framework, Django setup should be 
+similar:
+
+.. code-block:: bash
+
+    $ pex flask gunicorn myapp -e gunicorn.app.wsgiapp -o ~/service.pex
```
jsirois: Sorry about this - one last edit please: `pex flask gunicorn myapp -c gunicorn -o ~/service.pex`.

I do not know gunicorn and the `-e` seemed suspiciously manual for something that is presumably gunicorn bread and butter. My discovery process:
```
$ pex gunicorn -o gunicorn.pex
$ unzip -qc gunicorn.pex .deps/gunicorn-19.9.0-py2.py3-none-any.whl/gunicorn-19.9.0.dist-info/entry_points.txt

    [console_scripts]
    gunicorn=gunicorn.app.wsgiapp:run
    gunicorn_paster=gunicorn.app.pasterapp:run

    [paste.server_runner]
    main=gunicorn.app.pasterapp:paste_server
$ unzip -qc gunicorn.pex .deps/gunicorn-19.9.0-py2.py3-none-any.whl/gunicorn/app/wsgiapp.py | tail
    """\
    The ``gunicorn`` command line runner for launching Gunicorn with
    generic WSGI applications.
    """
    from gunicorn.app.wsgiapp import WSGIApplication
    WSGIApplication("%(prog)s [OPTIONS] [APP_MODULE]").run()


if __name__ == '__main__':
    run()
```

jakethedev: Oh... I spent an embarrassing amount of time researching/asking questions regarding executing a specific module under gunicorn. And I'm glad it can be this simple. I hadn't seen a line of docs on `entry_points` until our chat here, even after digging deep into the python docs on setuputils. Now knowing about it, it's aggressively obvious in the Gunicorn setup file :face_with_head_bandage: 

This PR has been a journey of enlightenment.

So that totally works like a charm and simplifies a huge pile of assumptions I've had about the way I've been hot-wiring Gunicorn. Update commit inbound.

jakethedev:
No prob, thanks for filling in the blanks for some of this stuff!

kavink:
This is really good. But wondering if there is a complete beginners example using pants/pex, flask and gunicorn?  Because I have been building my flask app like this`./pants binary src/python/myapp:server` , but I’m unable to figure out how to include gunicorn and make flask app run in it

jakethedev:
Hey @kavink - gunicorn can't handle running a zip file (to my knowledge), so you have to bundle flask and gunicorn with your application at build time for it to work. This can be done in-line with pex, or by adding gunicorn and flask directly to the install_requires of your setup.py - but I'm not sure what that looks like in a pants build situation

From there, I don't know how pants sets the entry point for the application, but you'll want to set it so it's `gunicorn myapp:server` to have it run properly. Using pex directly, you build the pex with `-c gunicorn`, then run `./app.pex myapp:server` to pass the myapp.server instance to the gunicorn command

PR #575, opened 2018-10-04
Fix PEXBuilder.clone.

### LINKED ISSUE #570: [Errno 13] Permission denied when executing with bdist_pex
It seems that [this review](https://github.com/jsirois/pex/commit/8bba41f39198215ea1378ab5c325dde7deaba358#diff-4895aa16756686554f1906f607a1296e) released in version 1.4.6 has introduced a regression when executing pex with the bdist_pex target

```python setup.py bdist_pex
running bdist_pex
Writing environment pex into dist/test-0.1.pex
error: [Errno 13] Permission denied: '/tmp/tmpaumuj9wb/.deps/tensorflow-1.10.1-cp36-cp36m-manylinux1_x86_64.whl/tensorflow/libtensorflow_framework.so'
```

The package that fails is tensorflow.
Minimal setup.py:
```
#!/usr/bin/env python3

import setuptools
import os

setuptools.setup(
    name='test',
    version='0.1',
    install_requires=["tensorflow==1.10.1"],
    packages=setuptools.find_packages())
```

It works when just calling the command line tool pex:
`pex tensorflow==1.10.1`

Can you have a look ?

jsirois:
Previously clone copied distributions twice. One copy was made by the
underlying `Chroot`, and then another was manually performed. The
manual copy was only needed to update the internal `_distributions`
set, so just do that.

A test that failed before the fix is added to ensure distributions with
unwriteable files can be cloned.

Fixes #570

### DIFF
--- a/tests/test_bdist_pex.py
+++ b/tests/test_bdist_pex.py
@@ -124,7 +124,7 @@ def test_unwriteable_contents():
                                'bdist_pex',
                                '--pex-args=--disable-cache --no-pypi -f {}'
                               .format(os.path.join(my_app_project_dir, 'dist'))])
-        subprocess.check_call(['zipinfo', 'dist/uses_my_app-0.0.0.pex'])
+
         with open_zip('dist/uses_my_app-0.0.0.pex') as zf:
           unwriteable_sos = [path for path in zf.namelist()
                              if path.endswith('my_app/unwriteable.so')]
--- a/tests/test_bdist_pex.py
+++ b/tests/test_bdist_pex.py
@@ -2,12 +2,14 @@
 # Licensed under the Apache License, Version 2.0 (see LICENSE).
 
 import os
+import stat
 import subprocess
 import sys
 from textwrap import dedent
 
 from twitter.common.contextutil import pushd
 
+from pex.common import open_zip
 from pex.testing import temporary_content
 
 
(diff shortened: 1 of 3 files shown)


jsirois:
@fhoering if you have time to give this a review I'd be grateful. Thanks for finding this - it unearthed something surprising!

tests/test_bdist_pex.py
```diff
@@ -81,3 +83,52 @@ def test_pex_args_shebang_with_spaces():
 
 def test_pex_args_shebang_without_spaces():
   assert_pex_args_shebang('#!/usr/bin/python')
+
+
+def test_unwriteable_contents():
+  my_app_setup_py = dedent("""
```
Eric-Arellano: I usually have to do this to get dedent working as I intend:
```python
dedent("""\
.....
.....
""")
```

It likes this is working without the `\` in the first line though.

tests/test_bdist_pex.py
```diff
@@ -81,3 +83,52 @@ def test_pex_args_shebang_with_spaces():
...
+        packages=['my_app'],
+        include_package_data=True,
+        package_data={'my_app': ['unwriteable.so']},
+      )
+    """)
+
+  UNWRITEABLE_PERMS = 0o400
+  with temporary_content({'setup.py': my_app_setup_py,
+                          'my_app/__init__.py': '',
+                          'my_app/unwriteable.so': 42},
```
Eric-Arellano: What does 42 mean? Consider defining as a constant to avoid magic numbers.

jsirois: It's not magic - it's random:
https://github.com/pantsbuild/pex/blob/cbf690845d40da54b3c7a3c2704c55a4a9dc614b/pex/testing.py#L68-L80
I think FORTY_TWO_BYTES_OF_RANDOM_DATA is perhaps over the top in this case.

tests/test_bdist_pex.py
jsirois: Yes, without the `\` I get an initial newline which is fine for debugging output.

tests/test_bdist_pex.py
Eric-Arellano: Oh goodness we not trying to write Java 😂 But something shorter like `random_data` I think would be helpful.

jsirois: Hrm - thats just as bad almost.
```
random_data = 42
with temporary_content({'setup.py': my_app_setup_py,
                          'my_app/__init__.py': '',
                          'my_app/unwriteable.so': random_data},
...
```
This is a lie. We are not writing `42` as the random data in `my_app/unwriteable.so`, we're writing 42 bytes of random data.

I'll switch to ''. But note!: it's just as magic and yet most folks don't balk at it. There is something to be said for familiarity with the testing libs in use (`temporary_content` in this case). You can't spell _everything_ out. This is coming from someone who likes explicit more than most! All this said you are in the position of not knowing this codebase and being kind enough to review, so I completely understand the comments.

PR #581, opened 2018-10-08
Cleanup `PexInfo` and `PythonInterpreter`.

jsirois:
Kill an unused type in `PexInfo` as well as our last remaining use of
`pkg_resources.get_platform`. Also kill unused `COMPATIBLE_SETUPTOOLS`
constants in `PythonInterpreter`.

### DIFF
--- a/pex/pex_info.py
+++ b/pex/pex_info.py
@@ -6,7 +6,6 @@ from __future__ import absolute_import, print_function
 import json
 import os
 import warnings
-from collections import namedtuple
 
 from .common import open_zip
 from .compatibility import PY2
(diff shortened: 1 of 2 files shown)


pex/pex_info.py
```diff
@@ -53,13 +50,15 @@ class PexInfo(object):
   @classmethod
   def make_build_properties(cls, interpreter=None):
     from .interpreter import PythonInterpreter
-    from pkg_resources import get_platform
+    from .platforms import Platform
```
Eric-Arellano: This relative import initially made me uncomfortable because I know they can be problematic for things like refactoring, and the Py2 syntax breaks on Py3. But I reviewed the book Python Cookbook and they say it's totally valid + you're copying the precedent, so I think it's fine.

You're also using the Py3 syntax, which resolves the biggest concern I would have.

pex/pex_info.py
```diff
@@ -53,13 +50,15 @@ class PexInfo(object):
...
 
     pi = interpreter or PythonInterpreter.get()
+    plat = Platform.current()
+    platform_name = plat.platform
```
Eric-Arellano: Thanks for adding this variable name. Definitely makes things more clear than `platform: plat.platform`

PR #582, opened 2018-10-08
Support environment markers during pex activation.

### LINKED ISSUE #456: Implement long-term solution for cross-interpreter requirement handling the in pex resolver

jsirois:
We've had support for environment markers on the resolve side for a
while and with just a little plumbing we can now support multi-python
pexes with environment-specific requirements.

Fixes #456

### DIFF
--- a/pex/environment.py
+++ b/pex/environment.py
@@ -161,6 +161,9 @@ class PEXEnvironment(Environment):
     # Resolve them one at a time so that we can figure out which ones we need to elide should
     # there be an interpreter incompatibility.
     for req in reqs:
+      if req.marker and not req.marker.evaluate():
+        TRACER.log('Skipping activation of `%s` due to environment marker de-selection' % req)
+        continue
       with TRACER.timed('Resolving %s' % req, V=2):
         try:
           resolveds.update(working_set.resolve([req], env=self))
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -661,10 +661,11 @@ def build_pex(args, options, resolver_option_builder):
                                 allow_prereleases=resolver_option_builder.prereleases_allowed,
                                 use_manylinux=options.use_manylinux)
 
-      for dist in resolveds:
-        log('  %s' % dist, v=options.verbosity)
-        pex_builder.add_distribution(dist)
-        pex_builder.add_requirement(dist.as_requirement())
+      for resolved_dist in resolveds:
+        log('  %s -> %s' % (resolved_dist.requirement, resolved_dist.distribution),
+            v=options.verbosity)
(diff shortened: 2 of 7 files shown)


pex/environment.py
```diff
@@ -161,6 +161,9 @@ def _resolve(self, working_set, reqs):
     # Resolve them one at a time so that we can figure out which ones we need to elide should
     # there be an interpreter incompatibility.
     for req in reqs:
+      if req.marker and not req.marker.evaluate():
+        TRACER.log('Skipping activation of `%s` due to environment marker de-selection' % req)
```
Eric-Arellano: With logging, it's preferred to do the below style, because it allows for lazy string interpolation (e.g. turning off logging won't load any of that code). Refer to https://stackoverflow.com/a/4149231.
```
TRACER.log('Skipping activation of `%s` due to environment marker de-selection', req)
```

If you have multiple things to interpolate, then it would look like:
```
TRACER.log("%s %s %s", v1, v2, v3)
```

I don't think we use this style anywhere in PEX so I understand if you wouldn't want to start introducing it now, but wanted to bring it up so we're aware.

jsirois: Thanks. I'm aware; however, `TraceLogger` is not a `logging.Logger` and does not have a lazy formatting API:
https://github.com/pantsbuild/pex/blob/c69213893bdbe7aa1648799d9d1a94753748942c/pex/tracer.py#L66

PR #586, opened 2018-10-08
Update pyenv if neccesary

CMLivingston:
During my local testing, I hit an exception that the pyenv versions required for certain tests do not exist in the pyenv registry. I needed to `git pull` to fix this, and this change runs a git pull in pyenv every time there is an update to pyenv/pyenv/master.

There is probably a better way to do this - perhaps running `pyenv install --list` and searching for the versions we want would be sufficient and obviate the need for a `rev-list` or `pull`. However, the method used in this change is a bit more dynamic and adaptable to new versions we might wish to test with in the future (but also adds some overhead for git ops).

cc: @jsirois

### DIFF
--- a/pex/testing.py
+++ b/pex/testing.py
@@ -323,6 +323,8 @@ def bootstrap_python_installer(dest):
       break
   else:
     raise RuntimeError("Helper method could not clone pyenv from git after 3 tries")
+  # Create an empty file indicating the fingerprint of the correct set of test interpreters.
+  open(os.path.join(dest, INTERPRETER_SET_FINGERPRINT), 'w').close()
 
 
 # NB: We keep the pool of bootstrapped interpreters as small as possible to avoid timeouts in CI
@@ -334,6 +336,22 @@ PY35 = '3.5.6'
 PY36 = '3.6.6'
 
 _VERSIONS = (PY27, PY35, PY36)
+# This is the filename of a sentinel file that sits in the pyenv root directory.
+# Its purpose is to indicate whether pyenv has the correct interpreters installed
+# and will be useful for indicating whether we should trigger a reclone to update
+# pyenv.
+INTERPRETER_SET_FINGERPRINT = '_'.join(_VERSIONS) + '_pex_fingerprint'
+
+
+def pyenv_has_test_interpreters(pyenv_root):
+  """
+  Return a boolean indicating whether the current pyenv installation can support the current
+  set of interpreters as specified by the presence of a fingerprint file with a filename based on
+  `_VERSIONS`.
+  """
+  if not os.path.exists(os.path.join(pyenv_root, INTERPRETER_SET_FINGERPRINT)):
+    return False
+  return True
 
 
 def ensure_python_distribution(version):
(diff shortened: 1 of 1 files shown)


pex/testing.py
```diff
@@ -347,6 +347,28 @@ def ensure_python_distribution(version):
 
   if not os.path.exists(pyenv):
     bootstrap_python_installer(pyenv_root)
+  else:
+    # Update pyenv if necessary.
```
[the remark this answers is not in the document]
jsirois: Bestest - just make the magic file-name be a deterministic combination of _VERSIONS so we don't even need to open the file to check staleness of the .pyenv clone.

pex/testing.py
```diff
@@ -334,6 +336,22 @@ def bootstrap_python_installer(dest):
 PY36 = '3.6.6'
 
 _VERSIONS = (PY27, PY35, PY36)
+# This is the filename of a sentinel file that sits in the pyenv root directory.
+# Its purpose is to indicate whether pyenv has the correct interpreters installed
+# and will be useful for indicating whether we should trigger a reclone to update
+# pyenv.
+INTERPRETER_SET_FINGERPRINT = '_'.join(_VERSIONS) + '_pex_fingerprint'
```
jsirois: How about `_INTERPRETER_SET_FINGERPRINT` to respect the local precedent of private constants?

pex/testing.py
```diff
@@ -323,6 +323,8 @@ def bootstrap_python_installer(dest):
       break
   else:
     raise RuntimeError("Helper method could not clone pyenv from git after 3 tries")
+  # Create an empty file indicating the fingerprint of the correct set of test interpreters.
+  open(os.path.join(dest, INTERPRETER_SET_FINGERPRINT), 'w').close()
```
jsirois: We have `touch`:
https://github.com/pantsbuild/pex/blob/a4eac2fcf138f1145a033bbbbcd71497d1242b4c/pex/common.py#L196-L199

pex/testing.py
```diff
@@ -345,7 +363,7 @@ def ensure_python_distribution(version):
   pyenv = os.path.join(pyenv_root, 'bin', 'pyenv')
   pip = os.path.join(interpreter_location, 'bin', 'pip')
 
-  if not os.path.exists(pyenv):
+  if not (os.path.exists(pyenv) and pyenv_has_test_interpreters(pyenv_root)):
```
jsirois: Since the touch file is only touched post pyenv clone, you just need the 1 check. Given that, just inlining makes sense:
```python
if not os.path.exists(os.path.join(pyenv_root, INTERPRETER_SET_FINGERPRINT)):
  bootstrap_python_installer(pyenv_root)
```

pex/testing.py
```diff
@@ -340,7 +340,7 @@ def bootstrap_python_installer(dest):
 # Its purpose is to indicate whether pyenv has the correct interpreters installed
 # and will be useful for indicating whether we should trigger a reclone to update
 # pyenv.
-INTERPRETER_SET_FINGERPRINT = '_'.join(_VERSIONS) + '_pex_fingerprint'
+_INTERPRETER_SET_FINGERPRINT = '_'.join(_VERSIONS) + '_pex_fingerprint'
 
 
 def pyenv_has_test_interpreters(pyenv_root):
```
jsirois: You can kill this function now.

Issue #585, opened 2018-10-08
Release 1.5.0


jsirois:
This release will contain a ~public API break / new feature:
+ Implement long-term solution for cross-interpreter requirement handling the in pex resolver #456

I've filed #584 to track tightening up our semver contract / public API for release purposes.

jsirois:
Released:
+ https://github.com/pantsbuild/pex/releases/tag/v1.5.0
+ https://pypi.org/project/pex/1.5.0/

Issue #584, opened 2018-10-08
Document / implement a formal semver release system.


jsirois:
We have not held to semver wrt defining the public API or handling deprecations. It may be that only Pants uses PEX via the API, but this is dubious. We should come up with some way for external parties to assess if the API they are using is public so that they can rely on semver for their PEX requirement range.

See: https://semver.org/

jsirois:
Although Pants no longer uses any API except the Pex CLI API, the need for semver still exists for CLI changes like adding subcommands 1st class (instead of via the side-door of PEX_TOOLS, etc.).

jsirois:
I think the release announcement here has served to close this issue: https://github.com/pex-tool/pex/releases/tag/v2.2.0

It probably also deserves a doc page, but I won't block closing this on that. The doc re-vamp will be quite large and towards the close of this year.

PR #587, opened 2018-10-08
Prepare the 1.5.0 release.

### LINKED ISSUE #583: Release 1.4.9
This will be a bugfix release for 1.4.8 (#560) immediately followed by a 1.5.0 release to introduce the new functionality / ~semi-public API change in #456.

On the docket:
+ Fix resolve regressions introduced by the 1.4.8. (#580)

jsirois:
Closes #583
Work towards #585

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,11 +1,15 @@
 Release Notes
 =============
 
-1.4.9
+1.5.0
 -----
 
-This is a hotfix release for 1.4.8 that fixes a regression in interpreter setup that could lead to
-resolved distributions failing to build or install.
+This release fixes pexes such that they fully support environment markers, the canonical use case
+being a python 2/3 pex that needs to conditionally load one or more python 2 backport libs when
+running under a python 2 interpreter only.
+
+A fix for a regression introduced in 1.4.8 in interpreter setup that could lead to resolved
+distributions failing to build or install is also included.
 
 * Support environment markers during pex activation. (#582)
   `PR #582 <https://github.com/pantsbuild/pex/pull/582>`_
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -1,6 +1,27 @@
 Release Notes
 =============
 
+1.4.9
+-----
+
+This is a hotfix release for 1.4.8 that fixes a regression in interpreter setup that could lead to
+resolved distributions failing to build or install.
+
+* Support environment markers during pex activation. (#582)
```
jsirois: Hrm... wasn't thinking. This commit should not be in the release.
Thinking now how I want to handle.

jsirois: Alright - I guess the only real option is to just do 1.5.0. I'll bump the version and change the descriptions.

jsirois: Alright - fixed up. Sorry about this folks!

jsirois:
It does. This was what I was weighing. As noted in the [1.5.0 ticket](https://github.com/pantsbuild/pex/issues/585) we break "API"s without bumping our major version - which is a problem. The key bit here is defining our API formally. Informally, everyone uses the CLI as the API and only Pants uses PEX as a library - so this only breaks Pants - which is fine, I'll fix that on upgrade. That said - we can't be sure Pants is the only API user, thus #584

jsirois:
The only other reasonable option here is to revert 5f1f00f767 and then release 1.4.9. Perhaps I should just do that.

jsirois:
At the cost of thrashing reviewers, I'll close this and proceed with a revert, new 1.4.9 PR, revert-revert, 1.5.0 PR

Eric-Arellano:
I agree reverting 5f1f00f is the best option 👌

Issue #572, opened 2018-10-02
ImportError: No module named _ctypes


ConnorRules:
It looks like the ctypes library is required by the pex bootstrapper. Our current python environment does not have that library. Is there any way to package ctypes in my pex build?

jsirois:
There isn't. The actual ctypes library is part of the python stdlib, but it relies on the native library libffi. See here for some details: https://bugs.python.org/issue31652

This is unfortunate - we only use ctypes on linux and in the context of determining if the current linux host is manylinux1 compatible. We could spend some effort to make importing lazy such that we only trigger this error if you're actually running pex on linux _and_ you want manylinux support.

ConnorRules:
Interesting. During initial debugging of this error I did install libffi to no avail. The only thing that worked was copying the ctypes.so and ctypes library directory into my python environment. 

Trying to find a good workaround. The python environment I'm working with is prepackaged with some enterprise software. So manipulating the modules in it is not preferred and for some reason they didn't pack ctypes with it even though that's standard.

Thanks for the reply

jsirois:
> Interesting. During initial debugging of this error I did install libffi to no avail.

I should have been more clear: you need to have libffi(-dev) installed to have libffi headers available before you _build_ python. Unfortunately python will happily build without ctypes support. This is why you had to monkey with copying.

If you don't actually need manylinux1 support, then I think this issue stands as a bug/enhancement. We definitely could be lazy accessing the tiny bit of code that needs to consult ctypes. If you then built pexes with `--no-manylinux`, you should be good.

Laevos:
Hey @jsirois, I've been working with @ConnorRules on this issue. A `--no-manylinux` build option in pants would definitely solve this for us. Alternatively, would it be at all possible to just wrap the import of ctypes in [glibc.py#L4](https://github.com/pantsbuild/pex/blob/1600d5b8d4a3a85fa8c6e17d0525da794a1f915a/pex/glibc.py#L4) in a try block? Thanks!

jsirois:
To be clear `--no-manylinux` already exists. I'm just suggesting that its only valid for us to lazily import the ctypes code (or try/except as you suggest) if `--no-manylinux`.

Laevos:
[...]

PR #591, opened 2018-10-09
 572: Allow import of ctypes to be skipped if use_manylinux is false

Laevos:
As described in #572, ctypes is only required if use_manylinux is going to be set to True in the PEX. This allows PEXes to be used in environments where ctypes is not available.


jsirois:
Pex no longer uses copied tag calculation code but the blessed PyPA `packaging` project to determine tags. As such this PR no longer applies as-written. The fix should not be needed in modern Pex (2.x) since modern packaging accpets ctypes not being available:
https://github.com/pantsbuild/pex/blob/6b5f82a7b2730a3247e4fbf9da36cafa75536b28/pex/vendor/_vendored/packaging/packaging/tags.py#L549-L557

Issue #319, opened 2016-11-25
No such file or directory error


Hateman31:
Hello!
I built pex-file from my project ( https://github.com/Hateman31/RD).
I run 'pex RD -o RD.pex'. When i do "python RD.pex" i get next error:
Traceback (most recent call last):

> `  File ".bootstrap\_pex\pex.py", line 320, in execute
  File ".bootstrap\_pex\pex.py", line 78, in _activate
  File ".bootstrap\_pex\environment.py", line 135, in activate
  File ".bootstrap\_pex\environment.py", line 179, in _activate
  File ".bootstrap\_pex\environment.py", line 124, in update_candidate_distributions
  File ".bootstrap\_pex\environment.py", line 110, in load_internal_cache
  File ".bootstrap\_pex\environment.py", line 97, in write_zipped_internal_cache
  File ".bootstrap\_pex\util.py", line 176, in cache_distribution
  File ".bootstrap\_pex\common.py", line 125, in safe_open
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Users\\MyPC\\.pex\\install\\Kivy-1.9.1-cp34-none-win32.w
hl.eede4757567c219d7f99978091a446b258778859.aff7e4a3de1a4c60b2fb907744e38d83\\Kivy-1.9.1-cp34-none-win32.whl\\Kivy-1.9.1
.data/data/share/kivy-examples/demo/kivycatalog/container_kvs/AnchorLayoutContainer.kv'`

What i do wrong?
Thanks!

jsirois:
Windows is the issue here, pex has no support for it. The only hint I think we give is here: https://github.com/pantsbuild/pex/blob/master/setup.py#L38-L40

We have had some folks spend some time to get parts of ex prepped for Windows, but no concerted effort which would necessarily end in an AppVeyor CI job being wired up so we stay green and, further, continued involvement by win devs to keep things moving forward. Let us know if you're the person to help kick this off, that would be very welcome!

chrisjbremner:
@Hateman31 it looks like the path in your error message is 261 characters long, which exceeds the [maximum path length in Windows](https://docs.microsoft.com/en-us/windows/desktop/FileIO/naming-a-file#maximum-path-length-limitation). See https://github.com/pantsbuild/pex/pull/596 for a proposed fix.

PR #596, opened 2018-10-11
Proposal: Add extended path prefix for Windows filepath

chrisjbremner:
I recently ran across an issue similar to #319, where I received a `FileNotFoundError` while trying to run my pex file that I had created. As can be seen in the traceback in that issue, the error occurred [here](https://github.com/pantsbuild/pex/blob/39bcfaaa1e64c7edd57c8514c8086e624dbd1c56/pex/util.py#L178), during the `safe_open` command. 

After a little research, I found that the issue most likely stems from the [maximum path length limitations in Windows](https://docs.microsoft.com/en-us/windows/desktop/FileIO/naming-a-file#maximum-path-length-limitation), where paths cannot be 260 characters or above. In the example given in the issue, the path is 261 characters long. This is shown in the example below, where I am able to open a filepath that is 259 characters long, but not 260:

<details>
<summary>Maximum path length test</summary>

```python
In [1]: len(os.path.abspath(os.path.curdir + 'a'*227))
Out[1]: 260

In [2]: len(os.path.abspath(os.path.curdir + 'a'*226))
Out[2]: 259

In [3]: with open(os.path.abspath(os.path.curdir + 'a'*227),'wb') as f:
    ...:     pass
    ...:
---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
<ipython-input-42-dda94ce7e14c> in <module>()
----> 1 with open(os.path.abspath(os.path.curdir + 'a'*227),'wb') as f:
      2     pass

FileNotFoundError: [Errno 2] No such file or directory: 'C:\\filepath_maximum_le
ngth_example\\.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
aa'

In [4]: with open(os.path.abspath(os.path.curdir + 'a'*226),'wb') as f:
    ...:     pass
    ...:

```
</details>

This limitation can be mitigated by prefixing the filepath with `\\?\`, which allows files paths up to 32767 characters long. This is illustrated by the example below:

<details>
<summary>Maximum path length test with "\\?\" prefix</summary>

```python
In [1]: with open(os.path.abspath(os.path.curdir + 'a'*227),'wb') as f:
    ...:     pass
    ...:
---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
<ipython-input-42-dda94ce7e14c> in <module>()
----> 1 with open(os.path.abspath(os.path.curdir + 'a'*227),'wb') as f:
      2     pass

FileNotFoundError: [Errno 2] No such file or directory: 'C:\\filepath_maximum_le
ngth_example\\.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
aa'

In [2]: with open(u'\\\\?\\' + os.path.abspath(os.path.curdir + 'a'*227),'wb')
    ...:  as f:
    ...:     pass
```
</details>

There was discussion about accounting for this within cpython [here](https://bugs.python.org/issue18199), but it was decided that it wouldn't be added since Windows 10 Anniversary would allow long file paths. Unfortunately, this behavior seems to be opt-in in Windows 10, by [modifying a group policy](https://stackoverflow.com/a/27694470/7810672). In addition, those who still use Windows 7 don't have this option either.

In this PR, I took a shot at modifying `safe_open` to include the `\\?\` prefix to account for longer filepaths. I then tested this by trying to run the pex that had previously failed, which I was now able to do after the change was made. I also confirmed that I was able to properly build pex files (which I could do prior to making this change).

Ideally, this behavior would be incorporated as part of cpython, but failing that, it'd be nice if every path that is used within `pex` would have this `\\?\` prefix appended to it, but I realize that it would be cumbersome to implement everywhere. Therefore, I propose that the change is implemented in `safe_open` so that users can at least run a pex file on Windows, even if some of the other peripherals/cleanup may not work as intended.

### DIFF
--- a/pex/common.py
+++ b/pex/common.py
@@ -138,8 +138,9 @@ def safe_open(filename, *args, **kwargs):
   ``safe_open`` ensures that the directory components leading up the
   specified file have been created first.
   """
-  safe_mkdir(os.path.dirname(filename))
-  return open(filename, *args, **kwargs)  # noqa: T802
+  filepath = safe_path(filename)
+  safe_mkdir(os.path.dirname(filepath))
+  return open(filepath, *args, **kwargs)  # noqa: T802
 
 
 def safe_delete(filename):
@@ -155,6 +156,23 @@ def safe_rmtree(directory):
   """Delete a directory if it's present. If it's not present, no-op."""
   if os.path.exists(directory):
     shutil.rmtree(directory, True)
+
+
+def safe_path(path):
+  """Make a path safe for Windows.
+
+  Windows has a max path length of 260 character by default, which can be
+  expanded to 32767 characters by appending a "\\?\" prefix:
+  https://docs.microsoft.com/windows/desktop/FileIO/naming-a-file#maximum-path-length-limitation
+
+  If running Windows, add extended-path prefix if it's not already present. If not running
+  Windows, path is not modified.
+  """
+  if os.name == "nt":
+    extended_path_prefix = u"\\\\?\\"
+    if not path.startswith(extended_path_prefix):
+      return extended_path_prefix + path
+  return path
 
 
 def rename_if_empty(src, dest, allowable_errors=(errno.EEXIST, errno.ENOTEMPTY)):
--- a/pex/common.py
+++ b/pex/common.py
@@ -171,7 +171,7 @@ def safe_path(path):
   if os.name == "nt":
     extended_path_prefix = u"\\\\?\\"
     if not path.startswith(extended_path_prefix):
-      return extended_path_prefix + path
+      return os.path.abspath(extended_path_prefix + path)
   return path
 
 


chrisjbremner:
@kwlzn do you have a suggested reviewer? It looks like @mikekap did a lot of the Windows work, but it seems as though he may not be an active maintainer here anymore?

chrisjbremner:
I am going to close this as stale. The suggestion for getting around the issue presented in this PR is to use the "Disable path length limit" option, which can be selected when installing Python on Windows, or by later modifying the group policy as shown in the link above.

Issue #598, opened 2018-10-12
unhandled AttributeError during pex bootstrapping with PEX_PATH


kwlzn:
I don't have a minimal repro yet, but in some internal code that uses PEX_PATH with two other pex files that contain namespace packages we seem to be seeing the following unhandled exception on startup:

{code}
Traceback (most recent call last):
  File ".bootstrap/_pex/pex.py", line 364, in execute
  File ".bootstrap/_pex/pex.py", line 90, in _activate
  File ".bootstrap/_pex/environment.py", line 147, in activate
  File ".bootstrap/_pex/environment.py", line 198, in _activate
  File ".bootstrap/_pex/environment.py", line 64, in update_module_paths
AttributeError: '_NamespacePath' object has no attribute 'insert'
{code}

I think this points to a setuptools version problem, but I've ensured uniform alignment across setuptools 33.1.1 with no luck.

jsirois:
This is not a setuptools symbol, its a python 3 symbol via https://www.python.org/dev/peps/pep-0420/
On my machine, for example I find:
```
$ find /usr/lib/python{2.7,3.4,3.5,3.6,3.7}/ -type f -name "*.py" | xargs grep "class _NamespacePath"
/usr/lib/python3.4/importlib/_bootstrap.py:class _NamespacePath:
/usr/lib/python3.5/importlib/_bootstrap_external.py:class _NamespacePath:
/usr/lib/python3.6/importlib/_bootstrap_external.py:class _NamespacePath:
/usr/lib/python3.7/importlib/_bootstrap_external.py:class _NamespacePath:
```

These things become the `__path__` value only when python3-style implicit namespace packages are detected (interveining dirs with no `__init__.py`): https://www.python.org/dev/peps/pep-0420/
For normal packages, `__path__` is a plain old `list` - which has insert.

So... I think you must be running:
1. Under python3
2. In a project that purposefully (or inadvertantly?!) uses implicit namespace packages.

jsirois:
As a small piece of the test puzzle, I have confirmed that a wheel built from:
1. An old style `pkg_resources` namespace tree
2. Using a setup.py that declares `namespace_packages`

Generates a `*nspkg.pth` file that is imported on dist activation, from which new style `_NamespacePath`.

kwlzn:
> So... I think you must be running:
>
> Under python3
> In a project that purposefully (or inadvertantly?!) uses implicit namespace packages.

[...]

PR #599, opened 2018-10-12
Repair unhandled AttributeError during pex bootstrapping.

kwlzn:
A quick, surface level fix for #598.

### DIFF
--- a/pex/environment.py
+++ b/pex/environment.py
@@ -64,6 +64,8 @@ class PEXEnvironment(Environment):
         try:
           module.__path__.insert(0, module_dir)
         except AttributeError:
+          # TODO: This is a temporary bandaid for an unhandled AttributeError which results
+          # in a startup crash. See https://github.com/pantsbuild/pex/issues/598 for more info.
           TRACER.log(
             'Failed to insert %s: %s.__path__ of type %s does not support insertion!' % (
               module_dir,
--- a/pex/environment.py
+++ b/pex/environment.py
@@ -58,10 +58,19 @@ class PEXEnvironment(Environment):
     TRACER.log('Adding to the head of sys.path: %s' % new_code_path)
     sys.path.insert(0, new_code_path)
     for name, module in sys.modules.items():
-      if hasattr(module, "__path__"):
+      if hasattr(module, '__path__'):
         module_dir = os.path.join(new_code_path, *name.split("."))
         TRACER.log('Adding to the head of %s.__path__: %s' % (module.__name__, module_dir))
-        module.__path__.insert(0, module_dir)
+        try:
+          module.__path__.insert(0, module_dir)
+        except AttributeError:
+          TRACER.log(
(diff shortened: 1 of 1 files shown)


pex/environment.py
```diff
@@ -58,10 +58,19 @@ def update_module_paths(cls, new_code_path):
...
     for name, module in sys.modules.items():
-      if hasattr(module, "__path__"):
+      if hasattr(module, '__path__'):
         module_dir = os.path.join(new_code_path, *name.split("."))
         TRACER.log('Adding to the head of %s.__path__: %s' % (module.__name__, module_dir))
-        module.__path__.insert(0, module_dir)
+        try:
+          module.__path__.insert(0, module_dir)
+        except AttributeError:
+          TRACER.log(
```
jsirois: Why is it ok to log and not?:
`# Force subsequent imports to come from the .pex directory rather than the .pex file.`

This begs the question as to whether the enclosing `update_module_paths` function is needed at all.

kwlzn: the aim is purely to queue a quick bandaid in the pex->pants->internal release lifecycle to unblock our use of pex here, tbh. I think implicitly allowing breakage of the ~2/100+ affected modules w/ logging beats a hard crash preventing no use until a deeper fix can be made (and subsequently released etc).

this may or may not be reasonable, but just citing my current stance. :)

jsirois: The stance is reasonable, the state the code is left in does not read as reasonable. Can you call the shot in a comment with the real fix issue pointed to?

kwlzn: ah, of course.

jsirois:
@kwlzn I've created a 1.5.2 release ticket in #602 and included this change. I assume you want to merge this and get a release for Pants to consume?

cosmicexplorer:
@jsirois yes, this addresses an issue showing up internally that we haven't been able to nail down yet (in the interest of getting this fix in).

pex/environment.py
jsirois: FWIW - I think I sussed what's going on here. Details added to #598. I may be able to get a real fix in here. I'll stare at this code in light of PEP-420 and report an estimate.

jsirois: This looks like it will take a few hours to fix the underlying issue with a test coverging the implicit namespace package case.

kwlzn:
@jsirois propped this up purely as a quick drive-by bandaid to help expedite internal unblocking (and ticketed/discussed internally) - but no strong preference on course of action here.

pex/environment.py
kwlzn: sgtm - thank you!

cosmicexplorer:
I dropped the ball on making sure this made it into a pex release, and this is an active blocker for an internal use case -- are there any blockers to merging this as of now? @kwlzn @jsirois I would be willing to do whatever is necessary to make the pex release work, or I could just handle the pants part.

PR #605, opened 2018-10-19
Pex exits with correct code when using entrypoint

illicitonion:
Currently the exit code from an entrypoint is swallowed, so running pex
always exits 0.

Tested with:
```
tox -e py27-package
dist/pex27 -e pytest:main pytest -D /path/to/failing/test/dir
```

before this PR, this exits 0, after it exits 1.

I'm not sure how to properly integration test this, as it is in the
outer-most wrapper layer of the pex binary, so looks like it's bypassed
by the way we tend to test things?

### DIFF
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -278,8 +278,7 @@ class PEX(object):  # noqa: T000
 
   def _wrap_coverage(self, runner, *args):
     if not self._vars.PEX_COVERAGE and self._vars.PEX_COVERAGE_FILENAME is None:
-      runner(*args)
-      return
+      return runner(*args)
 
     try:
       import coverage
@@ -296,7 +295,7 @@ class PEX(object):  # noqa: T000
     cov.start()
 
     try:
-      runner(*args)
+      return runner(*args)
     finally:
       TRACER.log('Stopping coverage')
       cov.stop()
@@ -310,8 +309,7 @@ class PEX(object):  # noqa: T000
 
   def _wrap_profiling(self, runner, *args):
     if not self._vars.PEX_PROFILE and self._vars.PEX_PROFILE_FILENAME is None:
-      runner(*args)
-      return
+      return runner(*args)
 
     pex_profile_filename = self._vars.PEX_PROFILE_FILENAME
     pex_profile_sort = self._vars.PEX_PROFILE_SORT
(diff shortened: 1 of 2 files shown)


tests/test_integration.py
```diff
@@ -1022,10 +1022,12 @@ def pex_with_entrypoints(entry_point):
...
     def do_something():
-      return run_setup
+      try:
+        from setuptools.sandbox import run_setup
+        return 0
+      except:
+        return 1
 
     if __name__ == '__main__':
       do_something()
```
jsirois: You beat me to it. As you see fit: https://github.com/pantsbuild/pex/pull/606
The do_something() should probably be used to sys.exit.

tests/test_integration.py
```diff
@@ -837,6 +837,27 @@ def test_pex_manylinux_runtime():
...
+      assert False
+    """
+  )
+
+  with temporary_content({'tester.py': test_stub}) as output_dir:
+    pex_path = os.path.join(output_dir, 'test.pex')
+    tester_path = os.path.join(output_dir, 'tester.py')
+    results = run_pex_command(['pytest==3.9.1',
+                               '-e', 'pytest:main',
+                               '-D', output_dir,
```
jsirois: Don't embed the source to test in the pex, instead maybe just: `subprocess.call([pex_path, os.path.join(output_dir, 'tester.py')])`.

tests/test_integration.py
```diff
@@ -854,7 +854,7 @@ def test_fail():
                                '-o', pex_path])
     results.assert_success()
 
-    assert subprocess.call([pex_path, tester_path]) == 1
+    assert subprocess.call([pex_path, os.path.realpath(tester_path)]) == 1
```
jsirois: Lovely, eh?

pex/pex.py
```diff
@@ -348,7 +346,9 @@ def execute(self):
       self.patch_sys(pex_inherit_path)
       working_set = self._activate()
       self.patch_pkg_resources(working_set)
-      self._wrap_coverage(self._wrap_profiling, self._execute)
+      exit_code = self._wrap_coverage(self._wrap_profiling, self._execute)
```
jsirois, line 349: This was never correct. Who knows is the entry point function even returns an int? The real problem was ever using the pytest:main entrypoint. We should have been using `-c pytest` or else `-m pytest`. Pex needs to provide some guidance in its --help and docs, but using an entrypoint function (foo.bar:baz) implies you know the function does its own sys.exits to indicate failure or else raises. If it just returns some value, whether an int or a string or a what have you, PEX really can't just presume and pas that to sys.exit.

jsirois: The right path may be to keep this behavior but actually document it. Pex could say "If you pass an entrypoint _function_ and it returns anything, we'll assume that's an int that represents the exit code of the process that should be used."

It looks like sys.exit behavior is a bit whacky and gracefully supports the present use :/:
```
$ python -c 'import sys; sys.exit("Hello")'; echo $?
Hello
1
$ python -c 'import sys; sys.exit([])'; echo $?
[]
1
$ python -c 'import sys; sys.exit(["a"])'; echo $?
['a']
1
$ python -c 'import sys; sys.exit(True)'; echo $?
1
$ python -c 'import sys; sys.exit(False)'; echo $?
0
```

PR #612, opened 2018-10-26
Vendor setuptools and wheel.

### LINKED ISSUE #607: Vendor setuptools / wheel.
We rely on wheel and setuptools (pkg_resources) APIs heavily. Supporting changes in these APIs leads to complex and brittle code - we should vendor what we need and simplify associated code and interpreter setup in the pex build phase.

jsirois:
This change removes pex dependence on setuptools and wheel resolution
at build time in favor of vendoring pinned versions.

Two changes make this possible:
1. The vendor and pex/vendor modules are introduced to support
   vendoring and activation of vendored dependencies.
2. The third_party module is introduced to better delineate use of
   third party code and runtime vendoring in pexes built by the pex
   tool.

Using 1 above, `tox -evendor` can be run to vendor new dependencies
when we change versions. Vendored dependencies are activated in just
two contexts:
1. By pex.bin.pex:main
2. By tests.

This is enabled by 2 above and moving `bdist_pex` away from the PEX api
to building via the CLI (ie: pex.bin.pex).

The vendored dependencies needed by the pex runtime to bootstrap are
now re-vendored by PEXBuilder via the third_party module which also knows
how to talk to vendor in this context only.

The over-arching theme is better delineation of pex the tool
("build-time") vs pexes ("run-time"). Further changes can make this key
distinction more visible by breaking out `tool` and `runtime` packages
which should allow further slimming of created pex bootstraps.

Fixes #607


jsirois:
Not for review yet - just making sure CI is green. I want to do one more pass on the 1st commit before asking folks to review a large diff.

jsirois:
Reviewers: this is a large change - my apologies for that.
The 420761d is the result of running `tox -vendor` introduced in 681a668 and can be skipped for the purposes of review - purely mechanical addition of pinned `setuptools` and `wheel` via `pip install`.

Probably most worthy of scrutiny are the '-sE' additions to python invocations and the elimination of interpreter extras collection from the host site-packages. These were build-time leaks that I could see no reason for. The runtime allows `PEX_INHERIT_PATH` / `--inherit-path` via other means.

pex/third_party.py
```diff
@@ -0,0 +1,62 @@
...
+# Copyright 2018 Pants project contributors (see CONTRIBUTORS.md).
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+import importlib
+import os
+
+from .tracer import TRACER
+
+
+class LazyImport(object):
```
stuhood: A docstring explaining why this is necessary (or something at the usage site) would be good.

pex/third_party.py
```diff
@@ -0,0 +1,62 @@
...
+      self._module = importlib.import_module(self._module_name)
+      TRACER.log('Lazy-loaded {}'.format(self._module), V=3)
+    return getattr(self._module, item)
+
+
+pkg_resources = LazyImport.for_module('pkg_resources')
+wheel_install = LazyImport.for_module('wheel.install', runtime=False)
+
+
+def vendor_runtime(chroot, dest_basedir, label):
```
stuhood: A docstring would be good.

pex/vendor/__main__.py
```diff
@@ -0,0 +1,42 @@
...
+from . import vendor_specs
+from ..common import safe_delete, safe_rmtree
+
+
+class VendorizeError(Exception):
+  """Indicates an error was encountered updating vendored libraries."""
+
+
+def vendorize(vendor_spec):
+  cmd = ['pip', 'install', '--upgrade', '--no-compile', '--target', vendor_spec.target_dir,
```
stuhood: Is it necessary to delete the old copy before doing this, or does the upgrade do the right thing? Would it be safer to delete and then always do a fresh install?

tests/test_environment.py
```diff
@@ -6,32 +6,31 @@
 from contextlib import contextmanager
 
 import pytest
-from twitter.common.contextutil import temporary_dir
 
-from pex import resolver
+from pex import resolver, vendor
```
stuhood: Is `vendor` intended to be a public API of PEX? If not, should it be prefixed with `_`?

jsirois: It is not, but this is a big problem right now with pex - there is no demarcation. I'll defer and keep the local style for now if you don't object. I had filed #584 a little while back to attack the problem.

pex/vendor/__main__.py
jsirois: Upgrade does the right thing and ~the world relies on this day in, day out.

pex/bin/pex.py
```diff
@@ -720,6 +662,10 @@ def main(args=None):
   if options.python and options.interpreter_constraint:
     die('The "--python" and "--interpreter-constraint" options cannot be used together.')
 
+  # We already adjusted the sys.path at the top of this file, but potentially re-adjust path to
+  # include wheel.
+  vendor.adjust_sys_path(include_wheel=options.use_wheel)
```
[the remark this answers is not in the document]
jsirois: Yes - it will with different values for include_wheel. It's probably not a big concern since it only happens at build time, but `vendor` and `third_party` could stand some tests and docs so I'll likely fix in my review feedback pass here.

pex/third_party.py
jsirois: Agreed, I'll take a pass at docs here explaining the problems unit tests present with imports.

pex/third_party.py
jsirois: Agreed, will fixup and add a test.

jsirois:
> I'd love to see a blurb somewhere about the steps involved with vendoring and bumping versions.

Agreed, I'll at least add a comment in version.py to run `tox -evendor` and commit.

> Also, it would be very good to integrate this with pants before landing it to make sure that there are no fundamental issues.

This I did. The only issue is all the red code - kills alot of Pants code.

jsirois:
I'm going to mark this closed but may re-open. The exposure of pkg_resources here - unshaded - via an undeclared dist dep - to users of pex as a library is evil. Unexpected things could happen if they also use setuptools and at different API compatibility. I'm working on a shaded approach...

PR #613, opened 2018-10-29
Fix pex force local to handle PEP 420.

### LINKED ISSUE #598: unhandled AttributeError during pex bootstrapping with PEX_PATH

jsirois:
Previously, pex would blow up trying to adjust import paths if a PEP-420
implicit namespace package was encountered. Add a test of the path
adjustment functionality both to confirm it was needed (it was) and that
the fix technique works with all forms of namespace packages by only
assuming they support `append`.

Fixes #598


jsirois:
NB: This is feathered off of #612 which was required to provide stable behavior of setuptools (pkg_resources) namespace packages which varied in at least 3 ways over versions of setuptools PEX previously claimed to support.

Reviewable changes are contained in beb77bd

jsirois:
Rebased against master now that vendoring change is in. This is just a CI burn, one more cleanup pass and I'll ping reviewers for another look.

jsirois:
Alright - ready for re-review. Thanks again in advance!

pex/bootstrap.py
```diff
@@ -0,0 +1,76 @@
...
+  def __init__(self, location):
+    self._location = location
+
+  def demote(self):
+    """Demote the bootstrap code to the end of the `sys.path` so it is found last.
+
+    :return: The list of un-imported bootstrap modules.
+    :rtype: list of :class:`types.ModuleType`
+    """
+    import sys  # Grab a hold of `sys` early since we'll be un-importing our module in this process.
```
Eric-Arellano: If I understand correctly, this won't work as you intend if you import at the module level, correct? Or is it here to be closer to the code its used with?

pex/bootstrap.py
```diff
@@ -0,0 +1,76 @@
...
+    path = getattr(module, '__file__', None)
+    if path and path.startswith(self._location):
+      return True
+
+    # A vendored package.
+    path = getattr(module, '__path__', None)
+    if path:
+      for path in path:
+        if path.startswith(self._location):
+          return True
```
Eric-Arellano: can replace with `any`
```
    if path and any(path.startswith(self._location) for path in path):
      return True
```

pex/environment.py
```diff
@@ -27,19 +31,73 @@
...
+          loaded = self._delegate.load_module(fullname)
+          loaded.__loader__ = self
+        return loaded
+
+    _REGISTERED = False
+
+    @classmethod
+    def _ensure_namespace_handler_registered(cls):
+      if not cls._REGISTERED:
+        import pkg_resources
```
Eric-Arellano: Same question on these imports—is it required to be here to work, or here for readability? There are a couple of imports below as well like this.

pex/environment.py
```diff
@@ -27,19 +31,73 @@
...
+      return caching_loader
+
+  @classmethod
+  def _install_pypy_zipimporter_workaround(cls, pex_file):
+    def pypy_zipimporter_workaround(path):
+      import os
+
+      if not path.startswith(pex_file) or '.' in os.path.relpath(path, pex_file):
+        # We only need to claim the pex zipfile root modules.
+        raise ImportError()
```
Eric-Arellano: If PEX users could ever see this error, would be helpful to add a brief error message here.

pex/bootstrap.py
Eric-Arellano: In code below, you use `for path_item in path`. I think that's good naming to use here as well.

pex/bootstrap.py
jsirois: The former. You get very strange errors where sys turns up `None` if you leave the imports at the level of the module, the module gets popped from `sys.modules` and then the Bootstrap class is later executed.

pex/bootstrap.py
jsirois: Agreed. What I had here was crazy. Fixed.

pex/environment.py
jsirois: They can't, this is PEP-302 path hook import protocol - raise ImportError if your hook doesn't handle the given path; otherwise, return an importer. That said - it caused you questions, so added a comment with pep link.

pex/environment.py
jsirois: This was different and was delayed to only happen when we new namespace packages were in-play and thus setuptools available. It looks like this necessity linkage is now lost... looking at a refactor to re-establish.

jsirois:
Noting the OSX errors look like the classic, befuddling `/private/var/folders` vs `/var/folders` - `os.path.realpath` needed - thing:
```
File "/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/tmp8ecs_u41/pex2/.bootstrap/pex/compatibility.py", line 81, in exec_function
    exec(ast, globals_map, locals_map)
  File "/private/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/tmp8ecs_u41/pex2/.deps/foo_bar_baz-0.0.1-py3-none-any.whl/bin/foobaz", line 6, in <module>
    sys.exit(baz.main())
```

pex/environment.py
jsirois: Re-established the link with better comments.

jsirois: Unfortunately, the link is really still just established by (better) comments. I could not re-establish a direct code link due to early-loading requirements of PEXEnvironments when many are involved (PEX_PATH). That case is particularly important since its needed for the original proximal cause of this bug this fixes in PEX.

jsirois:
Thanks for the reviews and suggestions folks! Green and merging.

PR #624, opened 2018-11-16
Vendor `setuptools` and `wheel`.

### LINKED ISSUE #607: Vendor setuptools / wheel.

jsirois:
This change removes pex dependence on `setuptools` and `wheel`
resolution at build time in favor of vendoring pinned versions.
Switching to a vendoring model allows us to eliminate code that dealt
with the vagaries of supporting a ~5 year span of setuptools organic
growth and API drift and it also allows us to fully isolate pex
build-time interpreters. Going forward more libraries could be vendored
easily at the cost of pex distribution size.

Three changes make this possible:
1. The `pex.vendor` module is introduced to centralize vendored
   distribution management including establishing vendored distribution
   requirements and setting up interpreters with vendored distribution
   extras.
2. A `pex.vendor.__main__` is introduced and run via `tox -e vendor` to
   actually perform vendoring of third-party distributions inside pex.
   This includes installing the distributions in the
   `pex/vendor/_vendored` subtree and re-writing their imports
   (shading).
3. The `pex.third_party` module is introduced to provide runtime import
   support for the vendored code using a
   [PEP-302](https://www.python.org/dev/peps/pep-0302/) `sys.meta_path`
   importer.

Important consequences of this change include:
1. The pex runtime that used to be vendored under `.bootstrap/_pex` is
   now vendored under `.bootstrap/pex` in order to support a fixed
   `pex.third_party` vendor shading import prefix. The PEX runtime
   bootstrap demotion code is made more aggressive, correct and complete
   as a result to support an old pex vendored runtime handing off to a
   newer pex distribution to support pexed pex. In addtion, the
   `pex.third_party` vendoring importer is re-used to expose `pex` as
   `_pex` for users taking advantage of `_pex.util.DistributionHelper`
   and its `access_zipped_assets` for a deprecation period up to pex
   2.0.0.
2. Since we no longer re-name `pex` to `_pex` in the runtime, all
   imports are now absolute.
3. The pex build-time interpreter cache is eliminated. Since we vendor
   the extras needed by interpreters, setup is now cheap, fast and
   hermetic and needs no cache. As a result the
   `--interpreter-cache-dir` flag now no-ops and is deprecated for
   removal in pex 2.0.0.

Additional changes of note include:
1. The `bdist_pex` distutils command now dogfoods the pex cli for better
   fidelity with standard pex tool usage.
2. The pex `setup.py` is now self-hosting and can be run by a python
   interpreter lacking `setuptools` and/or `wheel`; in fact, pex can be
   pexed via `python -sSE setup.py bdist_pex`.

Replaces #612
Fixes #607


jsirois:
NB: This is a re-write of #612 to avoid being a bad actor and 1/2 vendoring un-shaded code as that change did. I've added the original reviewers of #612 plus a few more. I would appreciate a hard look, but it needn't be by all reviewers.

Some review notes:
+ 31e1362e is the initial reviewable commit, c6b2b98b was machine generated by `tox -e vendor` and can be ignored.
+ Best to start with [`pex.third_party.install`](https://github.com/pantsbuild/pex/pull/624/commits/31e1362e7795a8a1a16828f3a638c50f785ccca6#diff-39284bca463a0d1012d56b8f0af0ecf5R345):
https://github.com/pantsbuild/pex/blob/31e1362e7795a8a1a16828f3a638c50f785ccca6/pex/third_party/__init__.py#L345-L387
 The docs contain an interpreter session (`tox -epy27-repl`) that best explain what the upshot of the shaded vendor code is. The repl session directly demonstrates not doing evil with exposed, unshaded third party code smuggled away inside pex and this is at the heart of the two week journey to fix the initial cut in #612.
+ Next read [`installer`](https://github.com/pantsbuild/pex/pull/624/commits/31e1362e7795a8a1a16828f3a638c50f785ccca6#diff-6a70203f99a5e3cb26cafb4dc44d02a5), [`pex_builder`](https://github.com/pantsbuild/pex/pull/624/commits/31e1362e7795a8a1a16828f3a638c50f785ccca6#diff-f307120da1f697f1f173558b18a358ca) and [`vendor`](https://github.com/pantsbuild/pex/pull/624/commits/31e1362e7795a8a1a16828f3a638c50f785ccca6#diff-ac8134aab21277347390dbca90819365). These work together somewhat closely with [`pex.third_party](https://github.com/pantsbuild/pex/pull/624/commits/31e1362e7795a8a1a16828f3a638c50f785ccca6#diff-39284bca463a0d1012d56b8f0af0ecf5) to support re-vendoring vendored code into the pex runtime and exposing vendored code directly to the `setup.py` of distributions we build at pex build time.
+ Move on to [`bdist_pex`](https://github.com/pantsbuild/pex/pull/624/commits/31e1362e7795a8a1a16828f3a638c50f785ccca6#diff-3e0edd1c569a2520c9109c7b9d56e344) which switches from using a `PEXBuilder` to use the pex cli.
+ Finally scan the remaining small production code changes and tests, which mainly need to call `vendor.setup_interpreter(...)` in many places where they previously relied upon the interpreter setup by tox with extra setuptools and wheel deps as well as relying on some leaky behavior now fixed.

Thank you in advance for taking a look.

pex/commands/bdist_pex.py
```diff
@@ -41,67 +46,80 @@ def _write(self, pex_builder, target, script=None):
...
-        self._write(pex_builder, target, script=script_name)
-    elif name in console_scripts:
-      # The package has a namesake entry point, so use it.
-      log.info('Writing %s to %s' % (name, target))
-      self._write(pex_builder, target, script=name)
-    else:
-      # The package has no namesake entry point, so build an environment pex.
-      log.info('Writing environment pex into %s' % target)
-      self._write(pex_builder, target, script=None)
+        cmd += ['-c', script_name]
```
Eric-Arellano: Are there long argument names for these, like `--compile` instead of `-c`?

pex/compatibility.py
```diff
@@ -28,6 +30,11 @@
 unicode_string = (str,) if PY3 else (unicode,)
 bytes = (bytes,)
 
+if PY2:
+  from collections import Iterable, MutableSet
+else:
+  from collections.abc import Iterable, MutableSet
+
```
Eric-Arellano: You can potentially avoid this code if you'd like. Python 3 still includes `Iterable, Mapping`, etc under `collections`, even though it's deprecated and `collections.abc` is preferred.

In Python 3.7, it leads to a deprecation warning that gets printed to console (although maybe we silence that like we do with Pants). In Python 3.8, it will be removed completely, which we're a few years away from that.

--

For Pants, we ultimately decided to just use std lib's `collections` backport—even with 3.8 eventually dropping support—under the logic that Pants won't be using Python 2 by the time 3.8 comes out. 

Because PEX will support Python 2 for a long, long time, I think you're probably right to add the backport.

pex/executor.py
```diff
@@ -56,47 +58,40 @@ def __init__(self, cmd, exc):
       self.exc = exc
 
   @classmethod
-  def open_process(cls, cmd, env=None, cwd=None, combined=False, **kwargs):
+  def open_process(cls, cmd, **kwargs):
```
Eric-Arellano: Why remove these two kwargs? Although they can off course still be passed through `**kwargs`, if they are used frequently I would argue explicit is better than implicit.

pex/installer.py
```diff
@@ -76,43 +56,37 @@ def _setup_command(self):
...
+exec(ast)
+""".format(root=root, mixins=self.mixins)
 
   def run(self):
     if self._installed is not None:
       return self._installed
 
     with TRACER.timed('Installing %s' % self._install_tmp, V=2):
-      command = [self._interpreter.binary, '-'] + self._setup_command()
+      command = [self._interpreter.binary, '-sE', '-'] + self._setup_command()
```
Eric-Arellano: Long flags possible for `-s` or -`E`?

pex/link.py
```diff
@@ -5,17 +5,11 @@
...
 import posixpath
-from collections import Iterable
 
-from .compatibility import PY3, WINDOWS, pathname2url
-from .compatibility import string as compatible_string
-from .compatibility import url2pathname
-from .util import Memoizer
-
-if PY3:
-  import urllib.parse as urlparse
```
Eric-Arellano: Good change moving to the `compatibility.py`.

pex/pex_builder.py
```diff
@@ -1,31 +1,33 @@
 # Copyright 2014 Pants project contributors (see CONTRIBUTORS.md).
 # Licensed under the Apache License, Version 2.0 (see LICENSE).
 
-from __future__ import absolute_import, print_function
```
Eric-Arellano: Why remove this?

pex/third_party/__init__.py
```diff
@@ -0,0 +1,392 @@
...
+class _Importable(namedtuple('_Importable', ['module', 'is_pkg', 'path', 'prefix'])):
+  _exposed = False  # noqa: We want instance variable access defaulting to cls here.
+
+  def expose(self):
+    self._exposed = True
+    importlib.import_module(self.module)
+    _tracer().log('Exposed {}'.format(self), V=3)
+
+  def loader_for(self, fullname):
+    _, _, target = fullname.partition(self.prefix + '.')
```
Eric-Arellano: Would be helpful to use `_my_ignored_var` style if you know what the first two are. Not a big deal.

pex/third_party/__init__.py
```diff
@@ -0,0 +1,392 @@
...
+      if isinstance(importer, cls):
+        yield importer
+
+  @classmethod
+  def _iter_installed_vendor_importers(cls, prefix, root, path_items):
+    for importer in cls._iter_all_installed_vendor_importers():
+      # All Importables for a given VendorImporter will have the same prefix.
+      if importer._importables and importer._importables[0].prefix == prefix:
+        if importer._root == root:
+          if {importable.path for importable in importer._importables} == set(path_items):
```
Eric-Arellano: Are you nesting the conditionals for readability? Because there are no else clauses, I think one multiline if with multiple `and`s could express intent better.

pex/compatibility.py
jsirois: I actually added the backport because I was getting the warning in earlier iterations of the code where I turned on warnings more broadly (for the _pex vendoring re-director code). It felt cheap to ignore it when we already have the `pex.compatibility` module for just this sort of thing.

pex/third_party/__init__.py
jsirois: Expressing intent and readability seem to be synonyms to me! Yes - this is for readability. The single expression form is where I started and it seemed both not easily parsed through and left no natural place for the comment on the 1st condition noting any `_Importable`'s prefix would do.

setup.py
```diff
@@ -3,37 +3,41 @@
...
-# Populates the following variables:
-#   __version__
-#   __setuptools_requirement
-#   __wheel_requirement
-__version__ = ''
-version_py_file = os.path.join(os.path.dirname(__file__), 'pex', 'version.py')
-with open(version_py_file) as version_py:
-  exec(compile(version_py.read(), version_py_file, 'exec'))
+from pex import third_party
+third_party.install(expose=['setuptools', 'wheel'])
```
benjyw: Why must these be exposed here? 

Also, is it possible to get here when `include_wheel=False` so that `wheel` would silently not be exposed?

pex/third_party/__init__.py
```diff
@@ -0,0 +1,391 @@
...
+        self._loaders.append(loader)
+        if self._warning:
+          warnings.warn('Found loader for `import {}`:\n\t{}'.format(fullname, self._warning),
+                        category=DeprecationWarning)
+        return loader
+    return None
+
+  def _expose(self, paths):
+    for importable in self._importables:
+      if importable.path in paths:
```
benjyw: Should it be an error to request to expose something that isn't in the importables?

pex/vendor/_vendored/setuptools/pkg_resources/_vendor/appdirs.py
```diff
@@ -0,0 +1,608 @@
...
+    from com.sun.jna.platform import win32
+
+    buf_size = win32.WinDef.MAX_PATH * 2
+    buf = array.zeros('c', buf_size)
+    shell = win32.Shell32.INSTANCE
+    shell.SHGetFolderPath(None, getattr(win32.ShlObj, csidl_name), None, win32.ShlObj.SHGFP_TYPE_CURRENT, buf)
+    dir = jna.Native.toString(buf.tostring()).rstrip("\0")
+
+    # Downgrade to short path name if have highbit chars. See
+    # <http://bugs.activestate.com/show_bug.cgi?id=85099>.
```
TansyArron: bugs.activestate.com does not exist. Unclear if the bug still does.

pex/third_party/__init__.py
```diff
@@ -0,0 +1,391 @@
...
+                    root=self._root,
+                    importables=self._importables))
+
+
+_ISOLATED = None
+
+
+@contextmanager
+def isolated():
+  """Returns a chroot for third_party isolated from the ``sys.path``.
```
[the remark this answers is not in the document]
jsirois: Agreed - fixed. It was creating a chroot per-use in prior iterations of the PR.

pex/third_party/__init__.py
```diff
@@ -0,0 +1,391 @@
...
+class VendorImporter(object):
+  """A `PEP-302 <https://www.python.org/dev/peps/pep-0302/>`_ meta_path importer for vendored code.
+
+  This importer redirects imports from its package to vendored code, optionally exposing the
+  vendored code by its un-prefixed module name as well.
+
+  For example, if the ``requests`` distribution was vendored, it could be imported using this
+  importer via ``import pex.third_party.requests`` as longs as:
+
+    * The requests distribution was housed under some importable path prefix inside the this
```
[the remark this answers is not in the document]
jsirois: Fixed.

pex/vendor/README.md
```diff
@@ -0,0 +1,37 @@
+PEX Vendored Distributions
+==========================
+
+PEX vendors distributions of critical third party code it uses at build-time and at run-time both
+to ensure predictable behavior and provide a zero-dependency library and cli tool to build upon
+without fear of dependency conflict by higher software layers.
+
```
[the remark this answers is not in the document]
jsirois: I'm not going to go there on patches because I agree very strongly with just never editing period. Added a comment that this is the point - vendor as is, rewrite imports mechanically - you want a fix contribute it upstream and re-vendor.

pex/commands/bdist_pex.py
```diff
@@ -41,67 +46,80 @@ def _write(self, pex_builder, target, script=None):
...
+    # sys.path so the subprocess can find us.
+    env = os.environ.copy()
+    env['PYTHONPATH'] = os.pathsep.join(sys.path)
 
-    reqs = [package_dir] + reqs
+    args = [sys.executable, '-s', '-m', 'pex.bin.pex', package_dir] + self.pex_args
 
-    with ENV.patch(PEX_VERBOSE=str(options.verbosity), PEX_ROOT=options.pex_root):
-      pex_builder = build_pex(reqs, options, options_builder)
+    # A hack to get the existing distutils logging level.
```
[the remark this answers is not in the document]
jsirois: It refers to the combination of 106 and 107 - I added a newline to highlight this and extracted a method to further narrow scope.

pex/commands/__init__.py
```diff
@@ -0,0 +1,2 @@
+# Copyright 2015 Pants project contributors (see CONTRIBUTORS.md).
```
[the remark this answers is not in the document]
jsirois: Updated for this new file, but the remaining instances were me adding a copyright post-facto to files introduced at the date used.

pex/pex_builder.py
jsirois: There is no use of `print`.

Eric-Arellano: In Pants, we require all 4 future imports regardless of use, but it seems PEX does not do that. 

I advocate for always requiring the 4 future imports because it makes sure we never forget it when necessary. Print is fairly obvious and easy to remember; division is a much trickier one and easy to forget. 

It’s outside the scope of this PR to change how we handle future imports in PEX overall, so fine to remove print here, although I think there is more benefit than harm in keeping it.

pex/commands/bdist_pex.py
```diff
@@ -41,67 +46,80 @@ def _write(self, pex_builder, target, script=None):
...
         pass
     elif raw_entry_points is not None:
       die('When entry_points is provided, it must be a string or dict.')
 
-    return {}
+    return ()
 
   def run(self):
+    parser, options_builder = configure_clp()
+    options, _ = parser.parse_args(self.pex_args)
```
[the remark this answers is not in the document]
jsirois: Per OptParse docs this is args - remaining unparsed args. I'll respond to this style suggestion below though.

pex/commands/bdist_pex.py
jsirois: Lengthened

pex/third_party/__init__.py
jsirois: I couldn't find a way to make this clear and uncluttered at the same time - switched to a test-then-chomp.

pex/third_party/__init__.py
```diff
@@ -0,0 +1,392 @@
...
+
+
+class VendorImporter(object):
+  """A `PEP-302 <https://www.python.org/dev/peps/pep-0302/>`_ meta_path importer for vendored code.
+
+  This importer redirects imports from its package to vendored code, optionally exposing the
+  vendored code by its un-prefixed module name as well.
+
+  For example, if the ``requests`` distribution was vendored, it could be imported using this
+  importer via ``import pex.third_party.requests`` as longs as:
```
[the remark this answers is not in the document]
jsirois: Fixed.

pex/vendor/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+  path = os.path.dirname(os.path.abspath(__file__))
+  for _ in _PACKAGE_COMPONENTS:
+    path = os.path.dirname(path)
+  return path
+
+
+class VendorSpec(collections.namedtuple('VendorSpec', ['key', 'version'])):
+  """Represents a vendored distribution.
+
+  NB: Vendored distributions should comply with the host distribution splatform constraints. In the
```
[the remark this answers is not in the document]
jsirois: Correct - fixed.

pex/installer.py
```diff
@@ -76,43 +56,37 @@ def _setup_command(self):
...
-      bootstrap_sys_paths.append(self.SETUP_BOOTSTRAP_PYPATH % {'path': path})
-      bootstrap_modules.append(self.SETUP_BOOTSTRAP_MODULE % {'module': module})
-    return '\n'.join(
-      [self.SETUP_BOOTSTRAP_HEADER] +
-      bootstrap_sys_paths +
-      bootstrap_modules +
-      [self.SETUP_BOOTSTRAP_FOOTER]
-    )
+sys.argv[0] = __file__
+exec(ast)
```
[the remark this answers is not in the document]
jsirois: The high-level context is that we're executing the `setup.py` of an sdist here and attempting to control its environment - injecting setuptools and (optionally) wheel into sys.path so that `bdist_egg` and (optionally) bdist_wheel work. As such we can't `python setup.py bdist_wheel` against their setup.py directly since we need to inject `sys.path.insert(...)` modifications (granted - another approach would be to use `PYTHONPATH` and not pass `-E` to the interpreter - but I'm sticking with the existing technique and better isolation here).

As to the exec - I'm not exactly sure why it was done this way. Declaring `__file__` before `exec` suffices to inject `__file__` into the exec'd code which I think could have been the only point - let the 3rdparty `setup.py` think its `setup.py` even though it's being slurped in as a subset of a virtual setup.py pumped through stdin.

Comment added.

pex/third_party/__init__.py
```diff
@@ -0,0 +1,391 @@
...
+    _ISOLATED = chroot
+  yield _ISOLATED.path()
+
+
+def uninstall():
+  """Uninstall all uninstallable :class:`VendorImporter`s and uninmport the modules they loaded."""
+  VendorImporter.uninstall_all()
+
+
+def prefix():
```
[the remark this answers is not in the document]
jsirois: Done.

pex/vendor/__main__.py
```diff
@@ -0,0 +1,199 @@
...
+          for _from, _to in modifications.items():
+            print('    {} -> {}'.format(_from, _to))
+
+
+if __name__ == '__main__':
+  if len(sys.argv) != 1:
+    print('Usage: {}'.format(sys.argv[0]), file=sys.stderr)
+    sys.exit(1)
+
+  root_dir = os.path.realpath(os.path.join(os.path.dirname(__file__), '..', '..'))
```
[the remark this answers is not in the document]
jsirois: Changed to `root_directory` and `import_prefix`.

pex/third_party/__init__.py
benjyw: Actually I think `_, _, thing_i_need = foo.partition(...)` is very idiomatic. In general, and particularly with `partition`.

jsirois: I left this at zag instead of zigging back to the original. A recurring theme is definitely that Eric wants more explicit names - long flag names where possible in bash scripts and subprocess calls, and favoring `_unused_but_named` over `_`. These two preferences are at odds with the pex codebase generally and I haven't pushed really hard on underlying motivations. I will note that it could be argued in the `_` case that the very point of `_` is "doesn't matter" and so these might be read as "don't read me": ie focus on the other code. As such, switching from `_` to `_distracting_you_a_little_by_naming_my_intent_if_actually_used` is ... just distracting.

jsirois:
Thanks folks! Merging.

benjyw:
Epic!

pex/third_party/__init__.py
benjyw: I prefer the `_` usage for precisely the reason you mention: giving something a name implies to the reader that it matters, when it doesn't. It's just more cognitive load for no reason.

Eric-Arellano: Agreed that I have through a couple reviews been pushing for more explicitness. Much of this derives from my lack of experience with PEX, making it sometimes challenging for me to figure out what certain CLI flags represent or an abbreviated variable stands for. Another motivation is personal style, that to me it seems more readable. I recognize, however, that with open source, codebase conventions should trump personal preferences.

You both raise solid arguments for `_` over `_unused_but_named`. For how I reason through code, the latter style helps me to infer what the function is doing. However, the former style is completely valid and there are strong arguments for why it is preferred. Especially because PEX uses the convention of `_`, I think you're both right that we should stick to that style, or refactor it as John did with the two uses I flagged.

The one point I will continue to advocate for is long CLI flags. Even if that is not the convention with PEX, I always think the readability of long flags trumps saving a few characters for short flags. When using the command line directly, I use short flags, but when writing it in a file saved to VCS, I advocate for long flags.

P.S. Congratulations on merging 🎉

benjyw: Yeah, I think those are different cases. The flags are meaningful. Using the long version helps the reader. But the unused variables are not, and naming them distracts the reader. 

The general rule of naming and comments I like to follow is "assume the reader knows the language and the std libraries as well as you, but doesn't know what you're trying to do with them".

pex/interpreter.py
```diff
@@ -308,41 +287,19 @@ def all(cls, paths=None):
     return cls.filter(cls.find(paths))
 
   @classmethod
-  def _parse_extras(cls, output_lines):
-    def iter_lines():
-      for line in output_lines:
-        try:
-          dist_name, dist_version, location = line.split()
```
benjyw: For historical reference, this line was broken, as `location` is a path, and might have embedded spaces. Should be `line.split(' ', 2)`.

PR #635, opened 2018-12-14
Setup interpreter extras in InstallerBase.

### LINKED ISSUE #632: PythonInterpreter should not need to be manually setup with vendored extras
This shows up as a wart when pex is used as a library. Folks have to do this:
```python
import pex
from pex.interpreter import PythonInterpreter


interpreter = pex.vendor.setup_interpreter(PythonInterpreter.get())
```

The vendor package is really not meant for public consumption, but populating interpreters with appropriate extras (setuptools & wheel) that pex has vendored for use with PEXBuilder should be automatic.

jsirois:
This is the single bit of pex build-time that needs interpreters setup
with extras. Encapsulate setup there where it's safe to rely on the
vendor package and simplify the pex API as a result as evidenced by test
simplification.

Fixes #632

### DIFF
--- a/pex/installer.py
+++ b/pex/installer.py
@@ -37,7 +37,10 @@ class InstallerBase(object):
     self._source_dir = source_dir
     self._install_tmp = install_dir or safe_mkdtemp()
     self._installed = None
-    self._interpreter = interpreter or PythonInterpreter.get()
+
+    from pex import vendor
+    self._interpreter = vendor.setup_interpreter(distributions=self.mixins,
+                                                 interpreter=interpreter or PythonInterpreter.get())
     if not self._interpreter.satisfies(self.mixins):
       raise self.IncapableInterpreter('Interpreter %s not capable of running %s' % (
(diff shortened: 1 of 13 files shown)


pex/installer.py
```diff
@@ -37,7 +37,10 @@ def __init__(self, source_dir, interpreter=None, install_dir=None):
     self._source_dir = source_dir
     self._install_tmp = install_dir or safe_mkdtemp()
     self._installed = None
-    self._interpreter = interpreter or PythonInterpreter.get()
+
+    from pex import vendor
+    self._interpreter = vendor.setup_interpreter(distributions=self.mixins,
```
stuhood: It feels vaguely sideeffecty to do this here... but I suppose that doing it outside has been errorprone so far.

jsirois: the `setup_interpreter` function returns a new PythonInterpreter with the same binary path but added extras - it's a bad name - no side-effects.

PR #638, opened 2018-12-21
When iterating over Zipfiles, always use the Unix file separator to fix a Windows issue

yorinasub17:
Hello!

I know pex doesn't officially support windows, but there seems to be a regression in pex 1.6.0 with windows with respect to the way the vendoring works. Specifically, the regex for iterating the zip path in the vendoring uses the os separator, which has two problems on windows (where the separator is `\`):

1. `\` is an escape sequence, so unless the separator is added twice in the string, it uses it to escape the character immediately after the separator, causing the regex compilation to fail.
1. The path in the zip file uses unix separators (`/`), so using the os separator is actually wrong.

This PR addresses this by updating the regex to replace windows separators (`\`) in the prefix var with the unix separator.

### DIFF
--- a/pex/third_party/__init__.py
+++ b/pex/third_party/__init__.py
@@ -117,12 +117,9 @@ class _ZipIterator(namedtuple('_ZipIterator', ['zipfile_path', 'prefix'])):
 
   def _filter_names(self, relpath, pattern, group):
     # We use '/' here instead of os.sep because the zip file format spec specifies that paths must use forward slashes.
-    # See section 4.4.17 of https://support.pkware.com/display/PKZIP/APPNOTE
-    prefix = self.prefix + ((relpath + '/') if relpath else '')
-    pat = re.compile(r'^{prefix}{pattern}$'
-                     .format(prefix=prefix,
-                             pattern=pattern))
-
+    # See section 4.4.17 of https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT
+    relpath_pat = '{}/'.format(relpath) if relpath else ''
+    pat = re.compile(r'^{}{}{}$'.format(self.prefix, relpath_pat, pattern))
     with contextlib.closing(zipfile.ZipFile(self.zipfile_path)) as zf:
       for name in zf.namelist():
         match = pat.match(name)
--- a/pex/third_party/__init__.py
+++ b/pex/third_party/__init__.py
@@ -118,7 +118,7 @@ class _ZipIterator(namedtuple('_ZipIterator', ['zipfile_path', 'prefix'])):
   def _filter_names(self, relpath, pattern, group):
     prefix = self.prefix + ((relpath + os.sep) if relpath else '')
     if os.sep == '\\':
-        prefix = prefix.replace('\\', '/')
+      prefix = prefix.replace('\\', '/')
     pat = re.compile(r'^{prefix}{pattern}$'
                      .format(prefix=prefix,
                              pattern=pattern))
--- a/pex/third_party/__init__.py
+++ b/pex/third_party/__init__.py
@@ -102,11 +102,10 @@ class _ZipIterator(namedtuple('_ZipIterator', ['zipfile_path', 'prefix'])):
     while path:
       # We use '/' here instead of os.sep because the zip file format spec specifies that paths
       # must use forward slashes. See section 4.4.17 of
-      # https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT
+      # https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT.
       if zipfile.is_zipfile(path):
-        return cls(zipfile_path=path, prefix='{}/'.format(prefix) if prefix else '')
-      path_basename = os.path.basename(path)
-      prefix = '{}/{}'.format(prefix, path_basename) if prefix else path_basename
+        return cls(zipfile_path=path, prefix='{}{}'.format(prefix, os.sep).replace(os.sep, '/'))
+      prefix = os.path.join(prefix, os.path.basename(path)).replace(os.sep, '/')
       path = os.path.dirname(path)
     raise ValueError('Could not find the zip file housing {}'.format(root))
 
@@ -122,8 +121,9 @@ class _ZipIterator(namedtuple('_ZipIterator', ['zipfile_path', 'prefix'])):
   def _filter_names(self, relpath, pattern, group):
     # We use '/' here instead of os.sep because the zip file format spec specifies that paths must
     # use forward slashes. See section 4.4.17 of
-    # https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT
-    relpath_pat = '{}/'.format(relpath.replace(os.sep, '/')) if relpath else ''
+    # https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT. Note that relpath is constructed
+    # using os.sep by the time it is passed in here, so we need to replace it with '/'.
+    relpath_pat = '' if not relpath else '{}{}'.format(relpath, os.sep).replace(os.sep, '/')
     pat = re.compile(r'^{}{}{}$'.format(self.prefix, relpath_pat, pattern))
     with contextlib.closing(zipfile.ZipFile(self.zipfile_path)) as zf:
       for name in zf.namelist():
--- a/pex/third_party/__init__.py
+++ b/pex/third_party/__init__.py
@@ -123,7 +123,7 @@ class _ZipIterator(namedtuple('_ZipIterator', ['zipfile_path', 'prefix'])):
     # We use '/' here instead of os.sep because the zip file format spec specifies that paths must
     # use forward slashes. See section 4.4.17 of
     # https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT
-    relpath_pat = '{}/'.format(relpath) if relpath else ''
+    relpath_pat = '{}/'.format(relpath.replace(os.sep, '/')) if relpath else ''
     pat = re.compile(r'^{}{}{}$'.format(self.prefix, relpath_pat, pattern))
     with contextlib.closing(zipfile.ZipFile(self.zipfile_path)) as zf:
       for name in zf.namelist():
--- a/pex/third_party/__init__.py
+++ b/pex/third_party/__init__.py
@@ -100,9 +100,13 @@ class _ZipIterator(namedtuple('_ZipIterator', ['zipfile_path', 'prefix'])):
     prefix = ''
     path = root
     while path:
+      # We use '/' here instead of os.sep because the zip file format spec specifies that paths
+      # must use forward slashes. See section 4.4.17 of
+      # https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT
       if zipfile.is_zipfile(path):
-        return cls(zipfile_path=path, prefix=prefix + os.sep if prefix else '')
-      prefix = os.path.join(prefix, os.path.basename(path))
+        return cls(zipfile_path=path, prefix='{}/'.format(prefix) if prefix else '')
+      path_basename = os.path.basename(path)
+      prefix = '{}/{}'.format(prefix, path_basename) if prefix else path_basename
       path = os.path.dirname(path)
     raise ValueError('Could not find the zip file housing {}'.format(root))
 
@@ -117,8 +121,8 @@ class _ZipIterator(namedtuple('_ZipIterator', ['zipfile_path', 'prefix'])):
 
   def _filter_names(self, relpath, pattern, group):
     # We use '/' here instead of os.sep because the zip file format spec specifies that paths must
-    # use forward slashes.
-    # See section 4.4.17 of https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT
+    # use forward slashes. See section 4.4.17 of
+    # https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT
     relpath_pat = '{}/'.format(relpath) if relpath else ''
     pat = re.compile(r'^{}{}{}$'.format(self.prefix, relpath_pat, pattern))
     with contextlib.closing(zipfile.ZipFile(self.zipfile_path)) as zf:
--- a/pex/third_party/__init__.py
+++ b/pex/third_party/__init__.py
@@ -116,8 +116,11 @@ class _ZipIterator(namedtuple('_ZipIterator', ['zipfile_path', 'prefix'])):
       yield package
 
   def _filter_names(self, relpath, pattern, group):
+    prefix = self.prefix + ((relpath + os.sep) if relpath else '')
+    if os.sep == '\\':
+        prefix = prefix.replace('\\', '/')
     pat = re.compile(r'^{prefix}{pattern}$'
-                     .format(prefix=self.prefix + ((relpath + os.sep) if relpath else ''),
+                     .format(prefix=prefix,
                              pattern=pattern))
 
     with contextlib.closing(zipfile.ZipFile(self.zipfile_path)) as zf:
--- a/pex/third_party/__init__.py
+++ b/pex/third_party/__init__.py
@@ -116,7 +116,8 @@ class _ZipIterator(namedtuple('_ZipIterator', ['zipfile_path', 'prefix'])):
       yield package
 
   def _filter_names(self, relpath, pattern, group):
-    # We use '/' here instead of os.sep because the zip file format spec specifies that paths must use forward slashes.
+    # We use '/' here instead of os.sep because the zip file format spec specifies that paths must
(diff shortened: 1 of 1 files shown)


yorinasub17:
@Eric-Arellano Thanks for the PR review! I am currently bogged down with my other work, but will take a look at this when I free up later in the month.

yorinasub17:
> Research and confirm that this assumption about zip files always using / is indeed 100% true. It would be great to find the spec claiming this and to link to this in the PR description. Because we don't have any CI testing for Windows, this would be a good way to verify correctness.

It appears that the "official spec" of zip files is the one published by PKWARE (See https://support.pkware.com/display/PKZIP/APPNOTE, [the reference to that in wikipedia](https://en.wikipedia.org/wiki/Zip_(file_format)), and [the reference to the appnote file from winzip](https://www.winzip.com/win/en/comp_info.html)). From there, I checked [the latest spec](https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT) and confirmed that all paths must use forward slashes (section 4.4.17):

> All slashes MUST be forward slashes '/' as opposed to backwards slashes '\' for compatibility with Amiga and UNIX file systems etc.

Based on the above, I'll go through the process of simplifying the code.

pex/third_party/__init__.py
```diff
@@ -116,8 +116,11 @@ def iter_root_packages(self, relpath):
       yield package
 
   def _filter_names(self, relpath, pattern, group):
+    # We use '/' here instead of os.sep because the zip file format spec specifies that paths must use forward slashes.
+    # See section 4.4.17 of https://support.pkware.com/display/PKZIP/APPNOTE
```
Eric-Arellano: It looks like this link works a bit better: https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT.

pex/third_party/__init__.py
```diff
@@ -116,8 +116,11 @@ def iter_root_packages(self, relpath):
...
+    prefix = self.prefix + ((relpath + '/') if relpath else '')
```
Eric-Arellano: I think there's an opportunity to refactor this code to hopefully make this more readable and to avoid string concatenation—`.format()` is preferred in the codebase.

```python
prefix_suffix = "{}/".format(relpath) if relpath else ""
pat = re.compile(r'^{}{}$'.format(self.prefix + prefix_suffix, pattern)
```

Please feel free to iterate on this if you can think of a better approach!

Eric-Arellano:
It also looks like CI is failing due to the style check. After making the above changes, run `tox -v -e style` to test if style will be pass.

pex/third_party/__init__.py
yorinasub17: Got it. Sorry, I got lazy and copy pasted what was there before instead of following the rule of leaving things better than you found them.

Let me know what you think of what I pushed.

yorinasub17:
@Eric-Arellano I believe I addressed all the comments, but **please don't merge this yet**. I haven't had a chance to test on my windows box yet and I have a suspicion the updated version won't work because I haven't touched the `prefix` var, which might contain `os.sep`.

pex/third_party/__init__.py
Eric-Arellano: Looks good! Thanks.

yorinasub17:
Sorry about the style error. I thought tox succeeded, but I was fooled because it actually failed on the `find` on my mac and I wasn't paying close attention to the logs:

```
using tox.ini: /Users/yoriy/packages/yoripex/tox.ini
using tox-3.4.0 from /Users/yoriy/.pyenv/versions/2.7.15/lib/python2.7/site-packages/tox/__init__.pyc
GLOB sdist-make: /Users/yoriy/packages/yoripex/setup.py
  /Users/yoriy/packages/yoripex$ /Users/yoriy/.pyenv/versions/2.7.15/bin/python2.7 /Users/yoriy/packages/yoripex/setup.py sdist --formats=zip --dist-dir /Users/yoriy/packages/yoripex/.tox/dist >/Users/yoriy/packages/yoripex/.tox/log/tox-0.log
style reusing: /Users/yoriy/packages/yoripex/.tox/style
style inst-nodeps: /Users/yoriy/packages/yoripex/.tox/dist/pex-1.6.0.zip
  /Users/yoriy/packages/yoripex$ /Users/yoriy/packages/yoripex/.tox/style/bin/python -m pip install --no-deps -U /Users/yoriy/packages/yoripex/.tox/dist/pex-1.6.0.zip >/Users/yoriy/packages/yoripex/.tox/style/log/style-12.log
  /Users/yoriy/packages/yoripex$ /Users/yoriy/packages/yoripex/.tox/style/bin/python -m pip freeze >/Users/yoriy/packages/yoripex/.tox/style/log/style-13.log
style installed: DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7.,gitdb==0.6.4,GitPython==0.3.2rc1,pep8==1.4.5,pex==1.6.0,pyflakes==0.7.2,smmap==0.9.0,twitter.checkstyle==0.1.0,twitter.common.app==0.3.0,twitter.common.collections==0.3.0,twitter.common.contextutil==0.3.0,twitter.common.dirutil==0.3.0,twitter.common.lang==0.3.0,twitter.common.log==0.3.0,twitter.common.options==0.3.0,twitter.common.process==0.3.0,twitter.common.string==0.3.0,twitter.common.util==0.3.0
style run-test-pre: PYTHONHASHSEED='1115460778'
style runtests: commands[0] | bash scripts/style.sh
  /Users/yoriy/packages/yoripex$ /bin/bash scripts/style.sh
find: ,: unknown primary or operator
__________________________________________________________________________ summary ___________________________________________________________________________
  style: commands succeeded
  congratulations :)
```

pex/third_party/__init__.py
```diff
@@ -116,10 +116,11 @@ def iter_root_packages(self, relpath):
       yield package
 
   def _filter_names(self, relpath, pattern, group):
-    pat = re.compile(r'^{prefix}{pattern}$'
-                     .format(prefix=self.prefix + ((relpath + os.sep) if relpath else ''),
-                             pattern=pattern))
-
+    # We use '/' here instead of os.sep because the zip file format spec specifies that paths must
+    # use forward slashes.
+    # See section 4.4.17 of https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT
```
Eric-Arellano: Nit, can you please move the start of this comment to the prior line, as they are related? And add a period to the end of this sentence.

Thanks!

yorinasub17:
@Eric-Arellano Thanks for your patience with me through this review.

Ok after testing on my windows box, it turns out the updated version doesn't work for two reasons:

- As suspected, I needed to fix the prefix as well.
- I also had forgotten that the `relpath` that is passed into the function is constructed with `os.sep`. This needs to be the case because the `_DirIterator` expects the path to contain the sep.

I attempted to address both of these, but it sort of reverts back to what I had originally and I am not sure if the current way is actually simpler. What do you think? Any ideas to improve this?

pex/third_party/__init__.py
```diff
@@ -100,9 +100,13 @@ def containing(cls, root):
...
     while path:
+      # We use '/' here instead of os.sep because the zip file format spec specifies that paths
+      # must use forward slashes. See section 4.4.17 of
+      # https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT
       if zipfile.is_zipfile(path):
-        return cls(zipfile_path=path, prefix=prefix + os.sep if prefix else '')
-      prefix = os.path.join(prefix, os.path.basename(path))
+        return cls(zipfile_path=path, prefix='{}/'.format(prefix) if prefix else '')
+      path_basename = os.path.basename(path)
+      prefix = '{}/{}'.format(prefix, path_basename) if prefix else path_basename
```
Eric-Arellano: What about this:
```python
prefix = os.path.join(prefix, os.path.basename(path)).replace(os.sep, '/')
```

If `os.sep` is already `/`, `replace()` will simply no-op. For example, in a Python REPL `"/".replace("/", "/")` returns back `/`.

pex/third_party/__init__.py
```diff
@@ -116,10 +120,11 @@ def iter_root_packages(self, relpath):
...
 
   def _filter_names(self, relpath, pattern, group):
-    pat = re.compile(r'^{prefix}{pattern}$'
-                     .format(prefix=self.prefix + ((relpath + os.sep) if relpath else ''),
-                             pattern=pattern))
-
+    # We use '/' here instead of os.sep because the zip file format spec specifies that paths must
+    # use forward slashes. See section 4.4.17 of
+    # https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT
+    relpath_pat = '{}/'.format(relpath.replace(os.sep, '/')) if relpath else ''
```
Eric-Arellano: This might be an improvement. Inverts the expression so that there's an early return and applies `replace()` to the final result, rather than just to `relpath`.

```python
relpath_pat = '' if not relpath else '{}{}'.format(relpath, os.sep).replace(os.sep, '/')
```

pex/third_party/__init__.py
```diff
@@ -100,9 +100,13 @@ def containing(cls, root):
     prefix = ''
     path = root
     while path:
+      # We use '/' here instead of os.sep because the zip file format spec specifies that paths
+      # must use forward slashes. See section 4.4.17 of
+      # https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT
       if zipfile.is_zipfile(path):
-        return cls(zipfile_path=path, prefix=prefix + os.sep if prefix else '')
-      prefix = os.path.join(prefix, os.path.basename(path))
+        return cls(zipfile_path=path, prefix='{}/'.format(prefix) if prefix else '')
```
Eric-Arellano: Is there any guarantee `prefix` will never already have `os.sep` in it? That's an earnest question - I couldn't figure it out from a very quick search for `_ZipIterator`.

Either way, for consistency and safety, might be better to change this line to this idiom:

```python
return cls(zipfile_path=path, prefix='' if not prefix else '{}{}'.format(prefix, os.sep).replace(os.sep, '/')
```

pex/third_party/__init__.py
```diff
@@ -116,10 +120,11 @@ def iter_root_packages(self, relpath):
       yield package
 
   def _filter_names(self, relpath, pattern, group):
-    pat = re.compile(r'^{prefix}{pattern}$'
-                     .format(prefix=self.prefix + ((relpath + os.sep) if relpath else ''),
-                             pattern=pattern))
-
+    # We use '/' here instead of os.sep because the zip file format spec specifies that paths must
```
Eric-Arellano: With all these suggestions, you'll also want to update both the comments to say `"We replace os.sep with '/' because..."`.

pex/third_party/__init__.py
yorinasub17: > Is there any guarantee prefix will never already have os.sep in it?

I think this can be guaranteed because `prefix` starts off as `''` outside the while loop, and only contains parts constructed with `os.path.basename`. `os.path.basename` takes the last path element, so it is equivalent to `arg.split(os.sep)[-1]`, which will mean the result won't contain the `os.sep` character.

That said, there is something that doesn't make sense to me about this while loop. This threw me off at first, but I believe the while loop is repeatedly walking up `root` directory by directory until it hits a zip file. E.g ignoring `prefix` for a minute, the loop boils down to:

```python
while path:
  if zipfile.is_zipfile(path):
    return
  # if path = foo/bar/baz, then os.path.dirname(path) = foo/bar
  path = os.path.dirname(path)
```

Given that, I think the `prefix` is actually *backwards*. If we go into the loop with `foo.zip/bar/baz`, assuming `foo.zip` is the zip file, `prefix` will become `baz/bar`, because after the first iteration, `prefix` will be basename of the path (which is `baz`), and the second iteration joins that with basename of the path again (the original was `prefix = os.path.join(prefix, os.path.basename(path))`, so `os.path.join('baz', os.path.basename('foo.zip/bar')) = 'baz/bar'`). Am I missing something with my logic here?

This must be working because the tests pass, but I wonder if this is actually an uncaught bug? I haven't had a chance to walk through all the test cases to know if there is any test case for this, or if the caller of `_ZipIterator` somehow guarantees this loop only goes one iteration deep.

Also this loop is an infinite loop if `root` is ever an absolute path that does not contain a zip file, because `os.path.dirname('/') == '/'`, so it will never be falsy.

jsirois: You're right about the bug - and it skates by because items are exactly one level deep. The `prefix = os.path.join(prefix, os.path.basename(path))` line should be `prefix = os.path.join(os.path.basename(path), prefix)`.

As far as the infinite loop goes, the one call of `containing` ensures `root` is not a dir (but not that it's not a file). It probably makes sense to either guard the infinite loop case or else assert early that root is either a zipfile or else not a file and not a directory (`foo.zip/bar/baz` is not either of these).

pex/third_party/__init__.py
```diff
@@ -100,9 +100,12 @@ def containing(cls, root):
     prefix = ''
     path = root
     while path:
+      # We replace os.sep with '/' here because the zip file format spec specifies that paths must
+      # use forward slashes. See section 4.4.17 of
+      # https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT.
       if zipfile.is_zipfile(path):
-        return cls(zipfile_path=path, prefix=prefix + os.sep if prefix else '')
-      prefix = os.path.join(prefix, os.path.basename(path))
+        return cls(zipfile_path=path, prefix='{}{}'.format(prefix, os.sep).replace(os.sep, '/'))
```
jsirois: The `prefix` starts as `''` and is built up from there on the line below. Since that line takes care of `/` it seems like there is no need to `replace` here. Also, you've now changed handling of the `''` case. That case used to trigger `cls(..., prefix='')` and now you `cls(..., prefix='/')`. I think you want:
```python
return cls(zipfile_path=path, prefix='{}/'.format(prefix) if prefix else '')
```

pex/third_party/__init__.py
Eric-Arellano: Great investigation @yorinasub17 and thanks @jsirois for confirming.

I think it'd be best to save this bug fix for a separate PR, as this one is focused on Windows support. If you're up for it, that'd be great for you to open this PR. No worries if you're busy - we  can do it otherwise.

pex/third_party/__init__.py
```diff
@@ -100,9 +100,12 @@ def containing(cls, root):
...
+      prefix = os.path.join(prefix, os.path.basename(path)).replace(os.sep, '/')
```
jsirois: os.path.join is os.sep aware and roughly the whole point of this PR is os.sep sensitivity was wrong. Since this loop builds up a prefix from path components not containing os.sep by construction (os.path.basename ensures this), why not be more direct and (I reversed the joining as well here per your discovery of that bug):
```python
prefix = '{}/{}'.format(os.path.basename(path), prefix)
```

pex/third_party/__init__.py
```diff
@@ -116,10 +119,11 @@ def iter_root_packages(self, relpath):
...
 
   def _filter_names(self, relpath, pattern, group):
-    pat = re.compile(r'^{prefix}{pattern}$'
-                     .format(prefix=self.prefix + ((relpath + os.sep) if relpath else ''),
-                             pattern=pattern))
-
+    # We replace os.sep with '/' here because the zip file format spec specifies that paths must
+    # use forward slashes. See section 4.4.17 of
+    # https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT.
+    relpath_pat = '' if not relpath else '{}{}'.format(relpath, os.sep).replace(os.sep, '/')
```
jsirois: Nit, but, like above, why use os.sep when you don't want it?
```python
relpath_pat = '' if not relpath else '{}/'.format(relpath.replace(os.sep, '/'))
```

pex/third_party/__init__.py
yorinasub17: Thanks for the confirmation. I made the fix for the prefix here, but have not addressed the infinite loop.

I won't be able to address it today though. I'll check back again when I have time and if it still hasn't been fixed yet, I'll open a PR.

pex/third_party/__init__.py
yorinasub17: Good catch. Fixed!

pex/third_party/__init__.py
yorinasub17: Fixed as well, although I needed a guard for the base case since otherwise it will become `basename/`, which breaks later when we append another `/`.

pex/third_party/__init__.py
yorinasub17: Fixed!

yorinasub17:
Thanks for your continued support on this @jsirois and @Eric-Arellano.

I made the changes requested and tested it out on my windows box and confirmed it still works.

Issue #639, opened 2018-12-22
Resolver appears to miss manylinux wheel


zmanji:
I have asked the pex resolver to resolve `protobuf==3.6.1` for linux and cpython 2.7 and I have enabled manylinux support. On [PyPI](https://pypi.org/project/protobuf/#files) I see the following wheels that could satisfy the constraint:

* `protobuf-3.6.1-cp27-cp27mu-manylinux1_x86_64.whl`
* `protobuf-3.6.1-py2.py3-none-any.whl`

I expect pex to pick the manylinux wheel over the catchall py2 wheel.

To reproduce:

Get the pex 1.6.0 pex and the following script:
```
#!/usr/bin/env python2.7

from pprint import pprint

from pex.package import WheelPackage
from pex.platforms import Platform
from pex.resolver import resolve

def main():
    linux = Platform('linux_x86_64', impl='cp', version='27')

    requirements = ['protobuf==3.6.1']

    resolved = resolve(
            requirements,
            platform=linux,
            precedence=(WheelPackage,),
            allow_prereleases=False,
            use_manylinux=True,
    )

    pprint(resolved)

if __name__ == '__main__':
    main()
```

Then run run it like so:
```
PEX_INTERPRETER=1 ./pex27.pex ./bug.py
```

Observe the following output:
```
[ResolvedDistribution(requirement=Requirement.parse('six>=1.9'), distribution=six 1.12.0 (/var/folders/4q/slynn4zs3tq5k9zl8411whx80000gn/T/tmphXfUqn/six-1.12.0-py2.py3-none-any.whl)),
 ResolvedDistribution(requirement=Requirement.parse('protobuf==3.6.1'), distribution=protobuf 3.6.1 (/var/folders/4q/slynn4zs3tq5k9zl8411whx80000gn/T/tmplcCuOj/protobuf-3.6.1-py2.py3-none-any.whl)),
 ResolvedDistribution(requirement=Requirement.parse('setuptools'), distribution=setuptools 40.6.3 (/var/folders/4q/slynn4zs3tq5k9zl8411whx80000gn/T/tmplqty7T/setuptools-40.6.3-py2.py3-none-any.whl))]
```

See that PEX has resolved `protobuf-3.6.1-py2.py3-none-any.whl` despite the platform being linux and manylinux support enabled.

Note, I don't think this is a platform issue because if I change the requirement to `grpcio==1.17.1` it successfully resolves a manylinux wheel:
```
[...]

PR #640, opened 2018-12-29
Follow PEP 425 suggestions on distribution preference.

zmanji:
PEP 425 describes how a single interpreter on a platform can support many
tags and a single package might have many distributions to satisfy those
supported tags [0]. It describes a tie breaking mechanism:

> Sometimes there will be more than one supported built distribution for a
> particular version of a package. For example, a packager could release a
> package tagged cp33-abi3-linux_x86_64 that contains an optional C extension
> and the same distribution tagged py3-none-any that does not. The index of the
> tag in the supported tags list breaks the tie, and the package with the C
> extension is installed in preference to the package without because that tag
> appears first in the list.

pex lacks this tie breaking logic currently and it is the root cause of
issue #639.

To address this issue, I have enhanced the resolver's sorter to deprioritize
packages that support the 'any' platform.

[0]: https://www.python.org/dev/peps/pep-0425/#id1


stuhood:
@zmanji : Looks like there is a style issue on the first shard... would you mind fixing that? Thanks!

zmanji:
@stuhood I believe it should pass now.

For reference the `style.sh` script appears to be broken on OSX and the script doesn't fail:
```
┌─ ~/code/pex (sort-platform-tag) (venv)                                                                                                                                                           12:39:08
└─ ◆ ./scripts/style.sh
find: ,: unknown primary or operator
┌─ ~/code/pex (sort-platform-tag) (venv)                                                                                                                                                           12:39:11
└─ ◆ echo $?
0
```

PR #643, opened 2019-01-08
Support PEP-503 name normalization.

### LINKED ISSUE #614: Pex cannot correctly resolve between multiple transitive dependencies
Reproduction:
`pex apache-airflow==1.10.0 --interpreter-constraint='CPython<3.7'`
Pex version:
pex 1.5.1

Problem: 
Above command gives output of 
```
Could not satisfy all requirements for flask-login==0.2.11:
    flask-login==0.2.11(from: apache-airflow==1.10.0), Flask-Login<0.5,>=0.3(from: apache-airflow==1.10.0->flask-appbuilder<2.0.0,>=1.11.1)
```

Airflow 1.10.0's setup.py denotes flask-login of 0.2.11. https://github.com/apache/incubator-airflow/blob/1.10.0/setup.py#L280

However, it also depends on flask-appbuilder with a constraint of `<2.0.0,>=1.11.1`.
Looking at two different versions of flask-appbuilder:
https://github.com/dpgaspar/Flask-AppBuilder/blob/v1.11.1/setup.py#L41 vs https://github.com/dpgaspar/Flask-AppBuilder/blob/v1.12.1/setup.py#L42

1.11.1 and 1.12.1 both meet the requirement of `<2.0.0,>=1.11.1` but they require incompatible versions of `flask-login`. Pex should be able to resolve to the earlier version and pick up `flask-login` 0.2.11.

### LINKED ISSUE #641: Dependency resolution can't handle PEP 503-style normalized package names
PyPI and setuptools support package name normalization as per PEP 503: https://www.python.org/dev/peps/pep-0503/#normalized-names

This means that a package like https://pypi.org/project/pytest-selenium/ can be installed both with `pip install pytest.selenium` and `pip install pytest-selenium`.

pex only accepts already normalized package names. If I put `pytest.selenium` in a `requirements.txt` file this won't work:

```
$ pex -o pex.bin -vvvv -r requirements.txt 
pex: Constructed RequestsContext context <pex.http.RequestsContext object at 0x107466910>
pex: Constructed RequestsContext context <pex.http.RequestsContext object at 0x10701c950>
pex: crawling link i=0 link=Link('file:///Users/nik/.pex/build') follow_links=False
pex: crawling link i=0 link=Link('https://pypi.org/simple/pytest-selenium/') follow_links=False
Could not satisfy all requirements for pytest.selenium:
    pytest.selenium
```

Same works with `pytest-selenium`.

For consistency with other build tools pex should support this as well. Arguably the PEP 503 behavior is weird (I only just learned about it today too) and probably bad design. But consistency across tools might trump bad design decisions. I came across this because a user was trying an existing requirements file that had always worked for them with pip and couldn't figure out why it failed in pex.

jsirois:
Fixes #641

### DIFF
--- a/pex/package.py
+++ b/pex/package.py
@@ -9,8 +9,8 @@ from pex.archiver import Archiver
 from pex.base import maybe_requirement
 from pex.link import Link
 from pex.pep425tags import get_supported
-from pex.third_party.pkg_resources import EGG_NAME, parse_version, safe_name, safe_version
-from pex.util import Memoizer
+from pex.third_party.pkg_resources import EGG_NAME, parse_version, safe_version
+from pex.util import Memoizer, normalized_name
 
 
 class Package(Link):
@@ -80,8 +80,8 @@ class Package(Link):
     :returns: True if the package matches the requirement, otherwise False
     """
     requirement = maybe_requirement(requirement)
-    link_name = safe_name(self.name).lower()
-    if link_name != requirement.key:
+    link_name = normalized_name(self.name)
+    if link_name != normalized_name(requirement.key):
       return False
 
     # NB: If we upgrade to setuptools>=34 the SpecifierSet used here (requirement.specifier) will
@@ -136,7 +136,7 @@ class SourcePackage(Package):
 
   @property
   def name(self):
-    return safe_name(self._name)
+    return normalized_name(self._name)
 
   @property
   def raw_version(self):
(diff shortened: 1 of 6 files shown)


jsirois:
Not for review yet - this needs a test.

JustinVenus:
Thanks a lot for working on this @jsirois ... I'm having a lot of success w/ this patch so far.  I've monkey patched pants-1.70/pex-1.3.2 like so https://gist.github.com/JustinVenus/2fece24acfb376553a629c805ae6eace

jsirois:
I'm not sure how I lost track of this one, but I've re-based. I'm going to sit on it though and prefer the work in #781 to fix this bug if that work is accepted.

jsirois:
Since #781 is now merged I'm going to drop this.

PR #646, opened 2019-01-18
Fix invalid regex and escape sequences causing DeprecationWarning

Eric-Arellano:
These bad escape sequences were causing `DeprecationWarning`s when running Pants with Python 3.

Invalid entries found through the regex `'re\..*\\'`, and equality to original string tested by comparing the two in a REPL.

### DIFF
--- a/pex/vendor/_vendored/wheel/wheel/metadata.py
+++ b/pex/vendor/_vendored/wheel/wheel/metadata.py
@@ -13,7 +13,7 @@ from .pkginfo import read_pkg_info
 
 # Wheel itself is probably the only program that uses non-extras markers
 # in METADATA/PKG-INFO. Support its syntax with the extra at the end only.
-EXTRA_RE = re.compile("""^(?P<package>.*?)(;\s*(?P<condition>.*?)(extra == '(?P<extra>.*?)')?)$""")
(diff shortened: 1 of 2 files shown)


pex/vendor/_vendored/wheel/wheel/metadata.py
```diff
@@ -13,7 +13,7 @@
 
 # Wheel itself is probably the only program that uses non-extras markers
 # in METADATA/PKG-INFO. Support its syntax with the extra at the end only.
-EXTRA_RE = re.compile("""^(?P<package>.*?)(;\s*(?P<condition>.*?)(extra == '(?P<extra>.*?)')?)$""")
+EXTRA_RE = re.compile(r"^(?P<package>.*?)(;\s*(?P<condition>.*?)(extra == '(?P<extra>.*?)')?)$")
```
jsirois: The vendored code should remain pristine as per: https://github.com/pantsbuild/pex/blob/master/pex/vendor/README.md

I'll follow up with a revert / re-vendor and look into adding read-only perm bits to the vendored files to make it harder to do this sort of thing when grepping for changes of a certain sort in the future.

jsirois: https://github.com/pantsbuild/pex/pull/651

PR #648, opened 2019-01-22
Revert "Fix invalid regex and escape sequences causing DeprecationWarning"

Eric-Arellano:
@jsirois pointed out in https://github.com/pantsbuild/pex/pull/646#discussion_r249837587 that this PR changed vendored code, which we should not be doing.

Instead, we will have a followup PR that only touches our own code.


jsirois:
As I said in https://github.com/pantsbuild/pex/pull/646/files#r249837587 I've got this. Still working on the `tox -evendor` code to flip perm bits to r/o.

PR #651, opened 2019-01-22
Make tox -evendor idempotent.

### LINKED ISSUE #649: The vendor script should support skip markers.
Right now, running `tox -evendor` modifies `pex/environment.py`:
```diff
$ git diff pex/environment.py
diff --git a/pex/environment.py b/pex/environment.py
index e161254..aff4d9e 100644
--- a/pex/environment.py
+++ b/pex/environment.py
@@ -32,12 +32,12 @@ from pex.util import CacheHelper, DistributionHelper
 
 def _import_pkg_resources():
   try:
-    import pkg_resources
+    import pex.third_party.pkg_resources as pkg_resources
     return pkg_resources, False
   except ImportError:
     from pex import third_party
     third_party.install(expose=['setuptools'])
-    import pkg_resources
+    import pex.third_party.pkg_resources as pkg_resources
     return pkg_resources, True
```

It should not touch these particular imports; so some mechanism to tell it to skip is needed. Something like a `# vendor:skip` should do.

### LINKED ISSUE #650: Add `tox -evendor` idempotency check to CI
This will ensure we never check in modifications (beyond automated import re-writes) to vendored code.

jsirois:
This affords adding a CI check to ensure we never commit vendored code
with modifications beyond import re-writes.

Fixes #649
Fixes #650


pex/vendor/__main__.py
```diff
@@ -153,15 +172,22 @@ class VendorizeError(Exception):
...
+           '--install-option', '--install-scripts={}'.format(script_dev_null),
            vendor_spec.requirement]
     result = subprocess.call(cmd)
+    shutil.rmtree(script_dev_null)
     if result != 0:
       raise VendorizeError('Failed to vendor {!r}'.format(vendor_spec))
 
-    # We know we can get these as a by-product of a pip install but never need them.
-    safe_rmtree(os.path.join(vendor_spec.target_dir, 'bin'))
+    # We know we can get this as a by-product of a pip install but never need it.
```
jsirois: NB: Instead of deleting `bin/` post-install, we now instruct install to use a tmp dir for bin now. This is because the scripts wheel writes to bin/ form part of the checked-in `RECORD` file digests and their contents can change (namely the `#!/path/to/python shebang`). Using a tempdir - a dir outside the `--target` install root - eliminates the `bin/` scripts from being enteed into the `RECORD`.

jsirois: Upon further review had to go a different route to deal with this robustly; thus: 7a72d65

jsirois:
@Eric-Arellano did you run `tox -e vendor-check`? Running `tox` alone does not run all of the environments configured in `tox.ini`.

As to perms, the idea was malformed. Git only differentiates the executable bit - you can't store read-only.

jsirois:
I'll mention the (partial) revert.

jsirois:
It's not a with-travis thing, it's an idempotency against the committed state thing. You should repro on you machine if you 1st commit your edit.

PR #662, opened 2019-01-31
Fix interpreter finding

OniOni:
## Problem:
We currently don't expect interpreter binaries to have a name such as `python3` or `python2`. However those are valid names for interpreters.

Ex:
```
$ python3 -c 'import sys; print(sys.executable)'
/Library/Frameworks/Python.framework/Versions/3.6/bin/python3
```

## Solution:
Add entry to support those filenames.

### DIFF
--- a/pex/interpreter.py
+++ b/pex/interpreter.py
@@ -343,7 +343,7 @@ class PythonInterpreter(object):
     return cls.CACHE[binary]
 
   @classmethod
-  def _matches_binary_name(basefile):
+  def _matches_binary_name(cls, basefile):
     return any(matcher.match(basefile) is not None for matcher in cls.REGEXEN)
 
   @classmethod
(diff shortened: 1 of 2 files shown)


pex/interpreter.py
```diff
@@ -351,7 +356,7 @@ def find(cls, paths):
     for path in paths:
       for fn in cls.expand_path(path):
         basefile = os.path.basename(fn)
-        if any(matcher.match(basefile) is not None for matcher in cls.REGEXEN):
+        if self._matches_binary_name(basefile):
```
Eric-Arellano: This needs to be `cls._matches_binary_name`

tests/test_interpreter.py
```diff
@@ -58,3 +58,19 @@ def test_interpreter_caching(self, test_interpreter1, test_interpreter2):
...
+      'python',
+      'Python',
+      'python2',
+      'python2.7',
+      'python3',
+      'python3.6'
+    )
+
+    for name in valid_binary_names:
+      assert interpreter.PythonInterpreter._matches_binary_name(name), 'Expected {} to be valid'.format(name)
```
Eric-Arellano: I believe linter is complaining this line is over 100 characters.

PR #663, opened 2019-02-05
Add User-Agent header when resolving via urllib

AndrewLvov:
At least some Python Index servers return a simple format of packages page, redirecting to <path>/<package> to <path>/+simple/<package>.

So there is a different result between `pex` run using `urllib` and using `requests` - in one case it errors.

### DIFF
--- a/pex/http.py
+++ b/pex/http.py
@@ -134,6 +134,8 @@ class Context(AbstractClass):
 class UrllibContext(Context):
   """Default Python standard library Context."""
 
+  USER_AGENT = 'pex/%s' % PEX_VERSION
+
   def open(self, link):
     return urllib_request.urlopen(link.url)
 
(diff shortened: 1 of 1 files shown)


pex/http.py
```diff
@@ -134,6 +134,8 @@ def resolve(self, link):
 class UrllibContext(Context):
   """Default Python standard library Context."""
 
+  USER_AGENT = 'pex/%s' % PEX_VERSION
```
jsirois: This should be lifted up to the base class since RequestsContext already defines the same and we really want to always send a single UA no matter the Context impl.

pex/http.py
```diff
@@ -134,6 +134,8 @@ def resolve(self, link):
...
+
   def open(self, link):
     return urllib_request.urlopen(link.url)
```
jsirois: This is the other urlopen that needs a UA.

pex/http.py
AndrewLvov: Hey, @jsirois , thank you. Moved to `Context`

pex/http.py
AndrewLvov: Done.

PR #667, opened 2019-02-11
Prepare the 1.6.2 release.

### LINKED ISSUE #652: Release 1.6.2
On the docket:
+ [X] pex 1.6.1 cannot build cryptography 2.5 #661
+ [x] Allow pex to be invoked using runpy (python -m pex). #637 
+ [x] Add User-Agent header when resolving via urllib #663

jsirois:
Fixes #652

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,24 @@
 Release Notes
 =============
 
+1.6.2
+-----
+
+* Support de-vendoring for installs. (#666)
+  `PR #666 <https://github.com/pantsbuild/pex/pull/666>`_
+
+* Add User-Agent header when resolving via urllib (#663)
+  `PR #663 <https://github.com/pantsbuild/pex/pull/663>`_
+
+* Fix interpreter finding (#662)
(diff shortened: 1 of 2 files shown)


CHANGES.rst
```diff
@@ -1,6 +1,24 @@
...
+1.6.2
+-----
+
+* Support de-vendoring for installs. (#666)
+  `PR #666 <https://github.com/pantsbuild/pex/pull/666>`_
+
+* Add User-Agent header when resolving via urllib (#663)
+  `PR #663 <https://github.com/pantsbuild/pex/pull/663>`_
+
+* Fix interpreter finding (#662)
```
Eric-Arellano: We aren’t allowed to rename this here, are we? I realized right after hitting merge I had wanted to rename the commit message to something like “Fix interpreter failing to find python{2,3} bin name”. It was too late to fix once merged into master. 

Would be great to fix here at least, but not sure we want to diverge commit message from release notes.

jsirois: There is no policy I'm aware of for pants or pex here, but I've been sticking to damage done via commit.

Issue #656, opened 2019-01-25
If an `--interpreter-constraint` is set, it should always be honored.


jsirois:
Right now these are only consulted when PEX_PYTHON or PEX_PYTHON_PATH are set:
https://github.com/pantsbuild/pex/blob/a16f0ecb21ce3526f7d243778507d68219c3a91c/pex/pex_bootstrapper.py#L141-L150

There is no reason for the constraints to not always apply.

blorente:
Took a stab at it in #668, where I have written some questions.

PR #668, opened 2019-02-11
Honor interpreter constraints even when PEX_PYTHON and PEX_PYTHON_PATH not set

blorente:
Solves issue #656, where if no `PEX_PYTHON_PATH` and no `PEX_PYHON` are configured, we still honour `--interpreter-constraint` flag when bootstrapping a PEX.

There is an integration test that worked only when applied the last commit, but there is a lot of nuance to running it that I'd like a hand with. In particular, I'd like to know if there's a way of:
- Specifying "please ignore system-wide settings for PEX_PYTHON_PATH" in a test **or** Specifying 
"please override PEX_PYTHON_PATH with `None`". I think the first one should be easy to implement with a command-line flag defined in the same place as `--rcfile`.
- Specifying "This test only runs on Python3.5", because I need two different interpreters to play with (one for the environment that the pex runs in, another one to pass to `--interpreter-constraint`).

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -430,8 +430,7 @@ def test_interpreter_constraints_honored_without_ppp_or_pp():
       PATH=":".join([
         os.path.dirname(py36_path),
         os.path.dirname(py35_path)
-      ]),
-      PEX_VERBOSE="1"
+      ])
     )
     res = run_pex_command(['--disable-cache',
       '--interpreter-constraint===%s' % PY35,
@@ -483,12 +482,19 @@ def test_interpreter_resolution_pex_python_path_precedence_over_pex_python():
 def test_plain_pex_exec_no_ppp_no_pp_no_constraints():
   with temporary_dir() as td:
     pex_out_path = os.path.join(td, 'pex.pex')
-    res = run_pex_command(['--disable-cache',
-      '-o', pex_out_path])
+    env = make_env(
+      PEX_IGNORE_RCFILES="1",
+      PATH=os.path.dirname(os.path.realpath(sys.executable))
+    )
+    res = run_pex_command([
+      '--disable-cache',
+      '-o', pex_out_path],
+      env=env
+    )
     res.assert_success()
 
     stdin_payload = b'import os, sys; print(os.path.realpath(sys.executable)); sys.exit(0)'
-    stdout, rc = run_simple_pex(pex_out_path, stdin=stdin_payload)
+    stdout, rc = run_simple_pex(pex_out_path, stdin=stdin_payload, env=env)
     assert rc == 0
     assert os.path.realpath(sys.executable).encode() in stdout
 
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -415,6 +415,36 @@ def test_interpreter_resolution_with_pex_python_path():
     else:
       assert str(pex_python_path.split(':')[0]).encode() in stdout
 
+@pytest.mark.skipif(IS_PYPY)
+def test_interpreter_constraints_honored_without_ppp_or_pp():
+
+  # Create a pex with interpreter constraints, but for not the default interpreter in the path.
+  with temporary_dir() as td:
+
+    py36_path = ensure_python_interpreter(PY36)
+
+    pex_out_path = os.path.join(td, 'pex.pex')
+    env = make_env(
+      PEX_IGNORE_RCFILES="1",
+      PATH=os.path.dirname(py36_path)
+    )
+    res = run_pex_command(['--disable-cache',
+      '--interpreter-constraint===%s' % PY36,
+      '-o', pex_out_path],
+      env=env
+    )
+    res.assert_success()
+
+    # We want to try to run that pex with no environment variables set
+    stdin_payload = b'import sys; print(sys.executable); sys.exit(0)'
+
+    stdout, rc = run_simple_pex(pex_out_path, stdin=stdin_payload, env=env)
+    print("BL: stdout = {}".format(stdout))
+    assert rc == 0
+
+    # If the constraints are honored, it will have run python3.6 and not python3.5
+    if sys.version_info[0] == 3:
+      assert ("Python %s" % PY36) in str(stdout).split("\n")[0]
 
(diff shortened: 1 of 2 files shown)


jsirois:
@blorente for point #2 you want this:
https://github.com/pantsbuild/pex/blob/edc636ce6a004fa1bfc762d74f01c9fd3a862429/pex/testing.py#L324-L366

For point #1 you should just be able to run pex with an `env=...` that has what you want. And part of what you want is maybe:
https://github.com/pantsbuild/pex/blob/edc636ce6a004fa1bfc762d74f01c9fd3a862429/pex/variables.py#L322-L328

blorente:
Depends on #673

blorente:
For some reason I can repro [failing test](https://github.com/pantsbuild/pex/blob/8bdd0c3aa1c16491a27fdd7328102fe69d6451b1/tests/test_integration.py#L467) (and others) locally in `master`, even after blowing up `./.tox`. I have no idea what's up or why they don't show up in other places. I will keep investigating further, but ideas are welcome.

blorente:
@jsirois, any idea why I can repro the failing tests in `master`? Meaning, how they made it past CI in the first place.

tests/test_integration.py
```diff
@@ -416,6 +417,40 @@ def test_interpreter_resolution_with_pex_python_path():
...
+    res.assert_success()
+
+    # We want to try to run that pex with no environment variables set
+    stdin_payload = b'import sys; print(sys.executable); sys.exit(0)'
+
+    stdout, rc = run_simple_pex(pex_out_path, stdin=stdin_payload, env=env)
+    print("BL: stdout = {}".format(stdout))
+    assert rc == 0
+
+    # If the constraints are honored, it will have run python3.5 and not python3.6
```
Eric-Arellano: Comment outdated.

Eric-Arellano: Consider also adding `Without constraints, we would expect it to use python3.5 as it is the minimum interpreter.`

tests/test_integration.py
```diff
@@ -416,6 +417,40 @@ def test_interpreter_resolution_with_pex_python_path():
       assert str(pex_python_path.split(':')[0]).encode() in stdout
 
 
+@pytest.mark.skipif(IS_PYPY)
+def test_interpreter_constraints_honored_without_ppp_or_pp():
+  # Create a pex with interpreter constraints, but for not the default interpreter in the path.
+  with temporary_dir() as td:
+    py36_path = ensure_python_interpreter(PY36)
+    py35_path = ensure_python_interpreter(PY35)
```
Eric-Arellano: Trivial nit: put py35 before py36, as the informal convention for Pants and Pex seems to be going from older to newer. Also applies to line 431 and 432.

Eric-Arellano:
@jsirois or @CMLivingston a review of this would be quite helpful. While I'm more familiar with Pex than a few months ago, I'm still not aware of all the expectations and its history.

tests/test_integration.py
```diff
@@ -416,6 +417,40 @@ def test_interpreter_resolution_with_pex_python_path():
...
+      '-o', pex_out_path],
+      env=env
+    )
+    res.assert_success()
+
+    # We want to try to run that pex with no environment variables set
+    stdin_payload = b'import sys; print(sys.executable); sys.exit(0)'
+
+    stdout, rc = run_simple_pex(pex_out_path, stdin=stdin_payload, env=env)
+    print("BL: stdout = {}".format(stdout))
```
CMLivingston: Nit: remove print statement

CMLivingston:
> @jsirois, any idea why I can repro the failing tests in `master`? Meaning, how they made it past CI in the first place.

In my experience, this happens because certain env conditions align locally that are not covered in CI, and the developer did not bother to run tox locally before pushing the branch.

tests/test_integration.py
```diff
@@ -416,6 +417,40 @@ def test_interpreter_resolution_with_pex_python_path():
...
+def test_interpreter_constraints_honored_without_ppp_or_pp():
+  # Create a pex with interpreter constraints, but for not the default interpreter in the path.
+  with temporary_dir() as td:
+    py36_path = ensure_python_interpreter(PY36)
+    py35_path = ensure_python_interpreter(PY35)
+
+    pex_out_path = os.path.join(td, 'pex.pex')
+    env = make_env(
+      PEX_IGNORE_RCFILES="1",
+      PATH=":".join([
```
jsirois: Prefer `os.pathsep` when forming `PATH`.

tests/test_integration.py
```diff
@@ -416,6 +417,40 @@ def test_interpreter_resolution_with_pex_python_path():
...
+    # We want to try to run that pex with no environment variables set
+    stdin_payload = b'import sys; print(sys.executable); sys.exit(0)'
+
+    stdout, rc = run_simple_pex(pex_out_path, stdin=stdin_payload, env=env)
+    print("BL: stdout = {}".format(stdout))
+    assert rc == 0
+
+    # If the constraints are honored, it will have run python3.5 and not python3.6
+    if sys.version_info[0] == 3:
+      assert ("Python %s" % PY36) in str(stdout).split("\n")[0]
```
jsirois: Printing out `sys.executable` should not yield "Python ...", It should yield the the value of `py36_path`. Bear in mind, OSX has funky tmp dir paths though so you'll want to `os.realpath` left and right hand sides in the assert comparison.

tests/test_integration.py
```diff
@@ -416,6 +417,40 @@ def test_interpreter_resolution_with_pex_python_path():
...
+
+    # We want to try to run that pex with no environment variables set
+    stdin_payload = b'import sys; print(sys.executable); sys.exit(0)'
+
+    stdout, rc = run_simple_pex(pex_out_path, stdin=stdin_payload, env=env)
+    print("BL: stdout = {}".format(stdout))
+    assert rc == 0
+
+    # If the constraints are honored, it will have run python3.5 and not python3.6
+    if sys.version_info[0] == 3:
```
jsirois: Why this guard? The test should work under a pytest interpreter of - say - python2.7 as well since you bootstrap a pyenv interpreter for 3.5 and 3.6.

PR #673, opened 2019-02-15
Honor PEX_IGNORE_RCFILES in to_python_interpreter()

blorente:
Currently, it will try to resolve a path to an interpreter in `PEX_PYTHON_PATH` even if `PEX_IGNORE_RCFILES` is set.

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -519,7 +519,7 @@ def build_pex(args, options, resolver_option_builder):
     # affect usages of the interpreter(s) specified by the "--python" command line flag.
     constraints = options.interpreter_constraint
     validate_constraints(constraints)
-    if not ENV.PEX_IGNORE_RCFILES:
+    if options.rc_file or not ENV.PEX_IGNORE_RCFILES:
       rc_variables = Variables.from_rc(rc=options.rc_file)
       pex_python_path = rc_variables.get('PEX_PYTHON_PATH', '')
     else:
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -519,8 +519,11 @@ def build_pex(args, options, resolver_option_builder):
     # affect usages of the interpreter(s) specified by the "--python" command line flag.
     constraints = options.interpreter_constraint
     validate_constraints(constraints)
-    rc_variables = Variables.from_rc(rc=options.rc_file)
-    pex_python_path = rc_variables.get('PEX_PYTHON_PATH', '')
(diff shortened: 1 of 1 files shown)


pex/bin/pex.py
```diff
@@ -519,8 +519,11 @@ def to_python_interpreter(full_path_or_basename):
     # affect usages of the interpreter(s) specified by the "--python" command line flag.
     constraints = options.interpreter_constraint
     validate_constraints(constraints)
-    rc_variables = Variables.from_rc(rc=options.rc_file)
-    pex_python_path = rc_variables.get('PEX_PYTHON_PATH', '')
+    if not ENV.PEX_IGNORE_RCFILES:
+      rc_variables = Variables.from_rc(rc=options.rc_file)
```
jsirois: The options.rc_file value is `None` by default. To be non-`None` pex had to be invoked from the cli with `--rcfile=[path]`. As such you're setting the precedent rule here of env beats flags. That does not seem kosher. Is that what you're intending here or have I read things wrong?

jsirois: Honestly, the flag really makes no sense. It shouldn't exist at all fwict. I'd be fine with considering this not your problem. I've filled https://github.com/pantsbuild/pex/issues/674

blorente: Even if the option is `None`, `Variable.from_rc` will try to load stuff from well-known files (i.e. `/etc/pexrc`, `~/.pexrc`...), which leaves us with no way of specifying here "I actually want you to ignore all pexrc files".

jsirois:
@CMLivingston can you provide any guidance here navigating the tangle between `options.rc_file`, `PEX_IGNORE_RCFILES` and Using `ENV` everywhere else and `Variables` directly here? This all needs straightening and crisping even if it doesn't happen in this PR.

pex/bin/pex.py
jsirois: Well, the condition could be `if options.rc_file or not ENV.PEX_IGNORE_RCFILES`.

blorente: So I'd still like a way of expressing "Please do not try to load pexrc files", so would you be okay with something like "If the flag is None, and the env variable PEX_IGNORE_RCFILES is set, please do not load any pexrc file? That would create slightly inconsistent semantics for the env variable, until the flag is removed.

jsirois: Yes. I think we're saying the same thing. SGTM.

Issue #655, opened 2019-01-25
The `--interpreter-constraint` option should OR


jsirois:
Right now, the full list of interpreter constraints is ANDed. This is boh redundant (the comma operator already can AND constraints inside a single constraint string) and restrictive. There are common important cases where disjoint interpreter constraints are desirable; eg: A "universal" pex that supports being run under python 2.7 or python 3.4-3.7. Pex itself is an example of this.

CMLivingston:
But why wouldn't the author user just widen the single constraint in this case? Is this for when they want an incompatible "gap" in the middle?

> 
A "universal" pex that supports being run under python 2.7 or python 3.4-3.7. Pex itself is an example of this.

Eric-Arellano:
In https://github.com/pantsbuild/pex/pull/678, I try to resolve this by following the advice from 
https://github.com/pantsbuild/pex/issues/654#issuecomment-457763458. If `PEX_PYTHON` or `PEX_PYTHON_PATH` are specified, we keep the default AND behavior, else we OR.

This feels wrong to me. I would expect as a user consistent behavior regardless of those env vars being set. However, it sounds like Twitter has some use cases where they need a list of constraints to AND.

What do you all think of this proposal: introduce a new setting/env var, e.g. `PEX_AND_INTERPRETER_CONSTRAINTS`, that will turn on the `AND` behavior regardless of what values `PEX_PYTHON` and `PEX_PYTHON_PATH` have. With deprecation policy, we can do something like default this to be true for two releases, perhaps.

cc @stuhood @jsirois @CMLivingston

CMLivingston:
We should be able to just OR them in both cases and things should be fine. It should not break Twitter usage. The only reason it makes sense to AND for PPP is because with PPP you will deterministically select the lowest compatible interpreter on PPP if we OR the constraints. This shouldn't be an issue if users expect this and set constraints on pants.ini and Python targets accordingly within Pants context.

PR #678, opened 2019-03-07
OR interpreter constraints when multiple given

Eric-Arellano:
## Problem
Resolves https://github.com/pantsbuild/pex/issues/655. 

We would expect a list of constraints, such as `['CPython>=2.7,<3', 'CPython>=3.4']`, to OR the separate list entries, i.e. allow an interpreter that satisfies any of these distinct requirement strings. Instead, we always AND lists of constraints.

## Solution
Always `OR` a list of interpreter constraints. If the user wants `AND`, they may do so within the single requirement string via `,`, e.g. `'CPython>=2.7,<3'`.

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -317,10 +317,10 @@ def configure_clp_pex_environment(parser):
     default=[],
     type='str',
     action='append',
-    help='A constraint that determines the interpreter compatibility for '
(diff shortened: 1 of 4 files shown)


Eric-Arellano:
https://github.com/pantsbuild/pants/commit/4d63f8ac3bfab48b5b78f3bef3b7e06a1e65616e closed this when it shouldn't have.

Issue #428, opened 2017-10-28
update pex.readthedocs.io


nathanhruby:
Hi Folks!

http://pex.readthedocs.io/en/latest/ seems to be from v1.0.3 circa 2014 and is missing a few details (like .pexrc support).  When you have a moment, could you update readthedocs to the latest version?

Thank you!

-n

jsirois:
@wickman do you hold the keys to the https://pex.readthedocs.io project? If so can you add jsirois (https://readthedocs.org/profiles/jsirois/) to the admin?

nathanhruby:
looks like this is fixed, thanks!

PR #680, opened 2019-03-10
Provide control over pex warning behavior.

### LINKED ISSUE #677: PEX runtime warnings should have a knob to squelch
I'm not sure the knob should default on or off, but some from of warning control makes sense since once a pex is built, it will emit this warning on every run which does not seem useful past the first time a human acknowledges the issue - which they may not even be able to fix easily if due to 3rdparty deps.

jsirois:
Previously warnings were emitted to stderr unconditionally, but now
both build-time and run-time controls are provided. At build-time
--no-emit-warnings can be specified and a pex will be produced that
does not emit warnings. At run-time, PEX_EMIT_WARNINGS and PEX_VERBOSE
can be used to toggle warning behavior.

Fixes #677

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -477,6 +477,15 @@ def configure_clp():
       help='Turn on logging verbosity, may be specified multiple times.')
 
   parser.add_option(
+      '--emit-warnings', '--no-emit-warnings',
+      dest='emit_warnings',
+      action='callback',
+      callback=parse_bool,
+      default=True,
+      help='Emit runtime UserWarnings on stderr. If false, only emit them when PEX_VERBOSE is set.'
+           'Default: emit user warnings to stderr')
+
+  parser.add_option(
       '--pex-root',
       dest='pex_root',
       default=None,
@@ -559,6 +568,7 @@ def build_pex(args, options, resolver_option_builder):
   pex_info.pex_path = options.pex_path
   pex_info.always_write_cache = options.always_write_cache
   pex_info.ignore_errors = options.ignore_errors
+  pex_info.emit_warnings = options.emit_warnings
   pex_info.inherit_path = options.inherit_path
   if options.interpreter_constraint:
     for ic in options.interpreter_constraint:
--- a/pex/pep425tags.py
+++ b/pex/pep425tags.py
@@ -10,7 +10,6 @@ import platform
 import re
 import sys
 import sysconfig
-import warnings
 
 from pex.glibc import have_compatible_glibc
 
@@ -19,11 +18,13 @@ logger = logging.getLogger(__name__)
 _OSX_ARCH_PAT = re.compile(r'(.+)_(\d+)_(\d+)_(.+)')
 
 
+# TODO(John Sirois): Kill this file and use wheel.pep425tags from our vendored wheel instead.
+# The `# Issue #1074` doesn't even refer to a pex issue :/.
 def get_config_var(var):
   try:
     return sysconfig.get_config_var(var)
   except IOError as e:  # Issue #1074
-    warnings.warn("{0}".format(e), RuntimeWarning)
+    logger.warn(str(e))
     return None
 
 
--- a/pex/pep425tags.py
+++ b/pex/pep425tags.py
@@ -1,6 +1,12 @@
 # This file was forked from the pip project master branch on 2016/12/05
 
-"""Generate and work with PEP 425 Compatibility Tags."""
+"""Generate and work with PEP 425 Compatibility Tags.
+
+NB: Several functions here have their code grabbed by `pex.interpreter._generate_identity_source`
+to use in runtime identification of python interpreter platform details. These functions are marked
+here and should maintain reliance on only stdlib support. Even then, any new imports added should
+be checked against those setup in `pex.interpreter.ID_PY_TMPL`.
+"""
 
 from __future__ import absolute_import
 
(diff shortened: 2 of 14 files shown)


pex/bin/pex.py
```diff
@@ -476,6 +476,15 @@ def configure_clp():
       callback=increment_verbosity,
       help='Turn on logging verbosity, may be specified multiple times.')
```
Eric-Arellano: I think we need to also modify the help message for `verbosity`. I remember over Slack you had thought `-vvvvvvvvv` != `PEX_VERBOSE=9`, but this line suggests to me that it resolves the same thing: https://github.com/pantsbuild/pex/pull/680/files#diff-178808e5641b15e64b2d09f744976f03R673.

If this is the case, then I recommend changing the message to something like `'Turn on logging verbosity, may be specified multiple times. This will override `--no-emit-warnings.'`.

pex/pep425tags.py
```diff
@@ -19,11 +18,13 @@
...
 
 
+# TODO(John Sirois): Kill this file and use wheel.pep425tags from our vendored wheel instead.
+# The `# Issue #1074` doesn't even refer to a pex issue :/.
 def get_config_var(var):
   try:
     return sysconfig.get_config_var(var)
   except IOError as e:  # Issue #1074
-    warnings.warn("{0}".format(e), RuntimeWarning)
+    logger.warn(str(e))
```
Eric-Arellano: To clarify, you use `logger.warn` here because the issue does not come from Pex but is copied code from another library? So you do not want `--no-emit-warnings` to silence this, right?

pex/variables.py
```diff
@@ -327,6 +327,17 @@ def PEX_IGNORE_RCFILES(self):
     """
     return self._get_bool('PEX_IGNORE_RCFILES', default=False)
 
+  @property
+  def PEX_EMIT_WARNINGS(self):
```
Eric-Arellano: This will update https://pex.readthedocs.io/en/stable/api/index.html#pex-variables-module on the next release, right?

tests/test_pex_warnings.py
```diff
@@ -0,0 +1,76 @@
...
+def pex_info_no_emit_warnings():
+  pex_info = PexInfo.default()
+  pex_info.emit_warnings = False
+  return pex_info
+
+
+skip_py2_warnings = pytest.mark.skipif(PY2,
+                                       reason="The warnings.catch_warnings mechanism doesn't work "
+                                              "properly under python 2.7 / pypy2, etc. across "
+                                              "multiple tests")
```
Eric-Arellano: It'd be helpful to specify why. Sounds like you investigated this and it has to do with C code. Any knowledge transfer would be helpful in case someone spends more time trying to figure this out.

tests/test_pex_warnings.py
```diff
@@ -0,0 +1,76 @@
...
+
```
Eric-Arellano: And helpful to reiterate this is not an issue with our code and things work IRL, only this is an issue with the testing mechanism, as you explained to me over Slack.

pex/pep425tags.py
jsirois: No. The source code of functions in this module are plucked at runtime to use in interpreter identification. As such, the functions must be stdlib-only. I'll add a comment about this pre-existing wonkiness.

pex/variables.py
jsirois: No. See #428. I'm hoping @wickman will hand over the keys. I'm sure he'd gladly do so but he's just hard to get a hold of.

pex/bin/pex.py
jsirois: I don't agree. There is a cross-product doc issue here and it seems to me just mentioning PEX_VERBOSE in the context of the PEX_EMIT_WARNINGS and --emit-warnings docs is enough. If you turn on PEX_VERBOSE, you've already asked for stderr to be polluted _in general_. Specific polluting options like \*emit\* are just subsets.

Eric-Arellano: Okay that's convincing enough.

tests/test_pex_warnings.py
jsirois: I have to disagree here. This will never be fixed in python 2.7 since it's eol and pypy appears to be trying to emulate CPython as much as possible - which makes sense. As such the detailed why is unimportant.

tests/test_pex_warnings.py
jsirois: I mean, I already say this.

If you have some exact wording changes you want me to toss in, I'll do it.

Eric-Arellano: Maybe something like: 
```
"The warnings.catch_warnings mechanism doesn't work "
"properly under python 2.7 / pypy2, etc. across "
"multiple tests. This is an issue with the testing mechanism due to"
"issues with the underlying C code, and it does not negatively impact"
"our warning code when used outside of these tests."
```

Make it clear no one should waste time on trying to fix this.

jsirois: OK - left out the C reference since that doesn't explain pypy and I still don't think the underlying reason is important. I did add language though that makes it clear this is only a unit-test failure that can't be fixed and is otherwise covered by ITs. I will point out we should never have to say - but this actually works - in comments. If we're allowing code to hit master that doesn't work - we have big problems. It should be assumed!

Issue #665, opened 2019-02-06
Release 1.6.3


jsirois:
On the docket:
  + [x] The `--interpreter-constraint` option should OR #655
  + [x] Provide control over pex warning behavior. #680

jsirois:
I split out other items previously on the docket to #681

jsirois:
Released:
+ https://github.com/pantsbuild/pex/releases/tag/v1.6.3
+ https://pypi.org/project/pex/#files

PR #682, opened 2019-03-10
Prepare the 1.6.3 release.

jsirois:
Work towards #665.


jsirois:
This depends on #680 being merged.

jsirois:
> Thanks John! I know we don’t follow SemVer, but we might want to call extra attention to #678 in the release announcement as it’s a significant API change without any prior warning.

Will do.

PR #692, opened 2019-03-28
Use all compatible versions when calculating tags.

### LINKED ISSUE #539: pex does not resolve abi3 wheels correctly
pex doesn't seem to resolve abi3 wheels correctly. I have a project that depends on [cryptography](https://pypi.org/project/cryptography/), which seems to have started publishing abi3 wheels on March 20 2018 ([v2.2](https://pypi.org/project/cryptography/2.2/#files))

At first I thought this was missing wheels from cryptography (see https://github.com/pyca/cryptography/issues/4404), but actually the python 3.4 abi3 wheel should cover python 3.6 for example (which I'm using).

I pex package my project like so:
```
pip --cache-dir dist wheel --wheel-dir dist .
pex --no-pypi --cache-dir dist --repo dist -cmyproject -omyproject.pex myproject
```

If I pin cryptography to [2.1.4](https://pypi.org/project/cryptography/2.1.4/#files) (last version that published non-abi3 wheels), pex packaging above works.

If I try latest cryptography version, `pip wheel` correctly finds `cryptography-2.3.1-cp34-abi3-macosx_10_6_intel.whl`, however pex complains with `Could not satisfy all requirements for cryptography`

My environment:
- OSX 10.13.6
- python 3.6.6
- pex 1.4.5
- pip 18.0

jsirois:
Previously, we'd pass a single version to `get_supported` when we knew
one which would lead to a single tag being generated for the abi3 case
instead of one per previous minor version.

As part of this change, we need a more modern version of the
pep425tags.py we had copied from pip that allows passing multiple
versions to `get_supported`. Ideally, we'd use the version embedded in
our vendored setuptools code and this is tracked by #696. Extra code
had been added to pep425tags to implement `get_supported_for_any_abi`
and handle manylinux. This is now inlined in platforms.py, which was
the only user,  the door to using undisturbed copies in #696.

Fixes #539

### DIFF
--- a/pex/pep425tags.py
+++ b/pex/pep425tags.py
@@ -1,6 +1,8 @@
 # NB: Copied from our vendored setuptools' version at:
 #   pex/vendor/_vendored/setuptools/setuptools/pep425tags.py
 # Modifications are marked with `# NB: Modified from ...`
+# TODO(John Sirois): Remove this file as part of https://github.com/pantsbuild/pex/issues/696
+# -------------------------------------------------------------------------------------------
 
 # This file originally from pip:
 # https://github.com/pypa/pip/blob/8f4f15a5a95d7d5b511ceaee9ed261176c181970/src/pip/_internal/pep425tags.py
--- a/pex/pep425tags.py
+++ b/pex/pep425tags.py
@@ -1,6 +1,9 @@
-# This file was forked from the pip project master branch on 2016/12/05
-# TODO(John Sirois): Kill this file and use wheel.pep425tags from our vendored wheel instead.
+# NB: Copied from our vendored setuptools' version at:
+#   pex/vendor/_vendored/setuptools/setuptools/pep425tags.py
+# Modifications are marked with `# NB: Modified from ...`
 
+# This file originally from pip:
+# https://github.com/pypa/pip/blob/8f4f15a5a95d7d5b511ceaee9ed261176c181970/src/pip/_internal/pep425tags.py
 """Generate and work with PEP 425 Compatibility Tags.
 
 NB: Several functions here have their code grabbed by `pex.interpreter._generate_identity_source`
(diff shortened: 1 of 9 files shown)


jsirois:
NB: Not for review - getting a full CI burn before adding tests and possibly streamlining `pex/platforms.py` modifications a bit.

jsirois:
This is now ready for review.

Eric-Arellano:
Starting review now. 

If you're able to, it would be helpful to break out a pre-cursor PR for the tweaks you make to the copied code, i.e. the indentation and the `NB` comments about what you changed from the copied code. Don't include the ABI3 changes. That way, this diff can be smaller and only include the changes you needed to make for ABI3 to work. Won't block on this though.

pex/pep425tags.py
```diff
@@ -12,345 +15,317 @@
 from __future__ import absolute_import
 
 import distutils.util
-import logging
+from distutils import log
```
Eric-Arellano: This appears diverged from original: https://github.com/pypa/pip/blob/8f4f15a5a95d7d5b511ceaee9ed261176c181970/src/pip/_internal/pep425tags.py#L5.

pex/pep425tags.py
```diff
@@ -12,345 +15,317 @@
...
+                    lambda: impl == 'cp',
+                    warn=(impl == 'cp')):
+            m = 'm'
+
+        # NB: Modified from ~ `from .extern import six; six.PY2`
+        PY2 = sys.version_info[0] == 2
+        if (get_flag('Py_UNICODE_SIZE',
+                     lambda: sys.maxunicode == 0x10ffff,
+                     expected=4,
+                     warn=(impl == 'cp' and PY2)) and PY2):
```
Eric-Arellano: To confirm, we ignore a check for 3.0-3.2 because Pex doesn't support these versions, right? Consider adding a brief comment.

```
NB: this block is modified to only check if PY2, rather than <3.3, because PEX does not support 3.0-3.2.
```

pex/platforms.py
```diff
@@ -5,14 +5,82 @@
...
+
+def _gen_all_compatible_versions(version):
+  # We select major and minor here in the context of implementation version strings.
+  # These are typically two digit characters; eg "27" or "36", but they can be three digit
+  # characters in the case of pypy; eg "271". In the typical case the 1st digit represents the
+  # python major version and the 2nd digit it's minor version. In the pypy case, the 1st digit still
+  # represents the (hosting) python major version, the 2nd the pypy major version and the 3rd the
+  # pypy minor version. In both cases the last digit is the minor version and the python in question
+  # guarantees backwards compatibility of minor version bumps within the major version as per
+  # semver.
```
Eric-Arellano: The fact that cp37 supports `py3{0-6}` was surprising to me. I recommending link to PEP 425.

```suggestion
  # semver and PEP 425 (https://www.python.org/dev/peps/pep-0425/#id1).
```

pex/platforms.py
```diff
@@ -5,14 +5,82 @@
...
+  #
+  # Concrete examples of what we want to return in each case:
+  # 1. typical case of cpython "36": ["36", "35", "34", "33", "32", "31", "30"]
+  # 2. pypy case of "271": ["271", "270"].
```
Eric-Arellano: Might want to add a test for PyPi to `test_platform.py`, since the format is different than CPython.

pex/platforms.py
```diff
@@ -5,14 +5,82 @@
...
+    for supported in all_supported:
+      yield supported
+      python_tag, abi_tag, platform_tag = supported
+      if platform_tag.startswith('linux') and force_manylinux:
+        yield python_tag, abi_tag, platform_tag.replace('linux', 'manylinux1')
+
+  return list(OrderedSet(iter_all_supported()))
+
+
+def _gen_all_abis(impl, version):
```
Eric-Arellano: Note Python 3 does not have `mu` anymore, thanks to flexible Unicode string storage https://www.python.org/dev/peps/pep-0393.

I think I get why this is okay to keep though. The idea here is to generate all possible ABI tags, then rely on `_get_supported` to filter out the ABI tags we don't need? If this is the case, it is not immediately obvious and a comment could help here.

pex/pep425tags.py
jsirois: That may be so. As per the header comment I copied from our vendored setuptools.

pex/pep425tags.py
jsirois: Nack. The file on the lhs is older than on the rhs. My modification was only to inline setuptools's vendored six.PY2 use here.

pex/platforms.py
jsirois: That test already exists and was failing which is what prompted me to the path of learning that led to the comment above.
https://github.com/pantsbuild/pex/blob/b0fe036ab9762b6ee1ac4ed3e85ab49fedd91ef1/tests/test_platform.py#L87-L93

pex/platforms.py
jsirois: And, fwiw, this code was just moved from pep425tags.py when I updated that file. We had - imo wrongly - edited that file instead of just copying it. We now just copy it with the mknimsl noted modifications.

jsirois:
I filed #696 to track getting rid of these very confusing copies of doubly indirect and double modified copies of pip's glibc.py and pep425tags.py files.

jsirois:
Thanks for the reviews and please accept my apolgies for lack of reviewer notes. I'm adding them below for posterity.

1st, the actual changes to the pep425tags.py  and glibc.py files copied from out vendored setuptools:

```diff
$ diff pex/vendor/_vendored/setuptools/setuptools/pep425tags.py pex/pep425tags.py
0a1,6
> # NB: Copied from our vendored setuptools' version at:
> #   pex/vendor/_vendored/setuptools/setuptools/pep425tags.py
> # Modifications are marked with `# NB: Modified from ...`
> # TODO(John Sirois): Remove this file as part of https://github.com/pantsbuild/pex/issues/696
> # -------------------------------------------------------------------------------------------
> 
3c9,16
< """Generate and work with PEP 425 Compatibility Tags."""
---
> """Generate and work with PEP 425 Compatibility Tags.
> 
> NB: Several functions here have their code grabbed by `pex.interpreter._generate_identity_source`
> to use in runtime identification of python interpreter platform details. These functions are marked
> here and should maintain reliance on only stdlib support. Even then, any new imports added should
> be checked against those setup in `pex.interpreter.ID_PY_TMPL`.
> """
> 
12d24
< import warnings
15,17c27
< from .extern import six
< 
< from . import glibc
---
> from pex import glibc  # NB: Modified from `from . import glibc`
21a32,33
> # NB: Used in `pex.interpreter.ID_PY_TMPL` and should only rely on stdlib. If imports change,
> # please consult `pex.interpreter.ID_PY_TMPL` and adjust stdlib imports there as needed.
26c38
<         warnings.warn("{}".format(e), RuntimeWarning)
---
>         log.warn(str(e))  # NB: Modified from `warnings.warn("{}".format(e), RuntimeWarning)`
29a42,43
> # NB: Used in `pex.interpreter.ID_PY_TMPL` and should only rely on stdlib. If imports change,
> # please consult `pex.interpreter.ID_PY_TMPL` and adjust stdlib imports there as needed.
42a57,58
> # NB: Used in `pex.interpreter.ID_PY_TMPL` and should only rely on stdlib. If imports change,
> # please consult `pex.interpreter.ID_PY_TMPL` and adjust stdlib imports there as needed.
50a67,68
> # NB: Used in `pex.interpreter.ID_PY_TMPL` and should only rely on stdlib. If imports change,
> # please consult `pex.interpreter.ID_PY_TMPL` and adjust stdlib imports there as needed.
62,68c80,81
< def get_impl_tag():
<     """
<     Returns the Tag for this specific implementation.
<     """
<     return "{}{}".format(get_abbr_impl(), get_impl_ver())
< 
< 
---
> # NB: Used in `pex.interpreter.ID_PY_TMPL` and should only rely on stdlib. If imports change,
> # please consult `pex.interpreter.ID_PY_TMPL` and adjust stdlib imports there as needed.
80a94,95
> # NB: Used in `pex.interpreter.ID_PY_TMPL` and should only rely on stdlib. If imports change,
> # please consult `pex.interpreter.ID_PY_TMPL` and adjust stdlib imports there as needed.
98,103c113,119
<         if get_flag('Py_UNICODE_SIZE',
<                     lambda: sys.maxunicode == 0x10ffff,
<                     expected=4,
<                     warn=(impl == 'cp' and
<                           six.PY2)) \
<                 and six.PY2:
---
> 
>         # NB: Modified from ~ `from .extern import six; six.PY2`
>         PY2 = sys.version_info[0] == 2
>         if (get_flag('Py_UNICODE_SIZE',
>                      lambda: sys.maxunicode == 0x10ffff,
>                      expected=4,
>                      warn=(impl == 'cp' and PY2)) and PY2):
104a121
> 
317,319d333
< 
< 
< implementation_tag = get_impl_tag()
```

```diff
$ diff pex/vendor/_vendored/setuptools/setuptools/glibc.py pex/glibc.py
0a1,6
> # NB: Copied from our vendored setuptools' version at:
> #   pex/vendor/_vendored/setuptools/setuptools/glibc.py
> # Modifications are marked with `# NB: Modified from ...`
> # TODO(John Sirois): Remove this file as part of https://github.com/pantsbuild/pex/issues/696
> # -------------------------------------------------------------------------------------------
> 
7c13
< import warnings
---
> from pex import pex_warnings  # NB: Modified from `import warnings`
45,46c51,53
<         warnings.warn("Expected glibc version with 2 components major.minor,"
<                       " got: %s" % version_str, RuntimeWarning)
---
>         # NB: Modified from `warnings.warn(..., RuntimeError)`
>         pex_warnings.warn("Expected glibc version with 2 components major.minor,"
>                           " got: %s" % version_str)
```

2nd, note that the code added to the top of platforms.py:
1. Moves the `get_supported_for_any_abi` from pep425tags.py to `_get_supported_for_any_abi` unaltered.
2. Moves the manylinux code previously embedded in the pep425tags.py `get_supported` code into `_get_supported` - changing to a post-filter technique for 0 disruption to the original pep425tags.py code.

PR #693, opened 2019-03-29
Correct .travis.yml tox syntax to work for 3.8.1+.

jsirois:
The tox 3.8.1 release requires `-v` and `-e <env>` be passed as
seperate arguments where, before, `-ve <env>` was accepted. Since we
float our tox dependency in CI to match our lack of a bound on
developer tox installs, update to the new CLI syntax since it is
compatible with old tox too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -115,93 +115,93 @@ matrix:
   include:
     - <<: *x-linux-27-shard
       name: TOXENV=style
-      script: tox -ve style
+      script: tox -v -e style
 
     - <<: *x-linux-27-shard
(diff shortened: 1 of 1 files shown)


jsirois:
The context here is that tox just released 3.8.1, we float, and `-ve` no longer works and you must separate the flags. Another approach would be to pin tox old, but it traditionally just works and you only bound low via minversion:
https://github.com/pantsbuild/pex/blob/6e35c5f21ac11edb46a696a39e565fa41b9e42cc/tox.ini#L3

PR #695, opened 2019-03-29
Don't change indentation for copied glibc.py and pep425tags.py files

Eric-Arellano:
Since these files are copied, we want to keep the original styling so that we can more easily copy over updates to the files and have a smaller diff when seeing how we've diverged from the originals.

Prework for https://github.com/pantsbuild/pex/pull/692.

### DIFF
--- a/pex/glibc.py
+++ b/pex/glibc.py
@@ -10,52 +10,52 @@ from pex import pex_warnings
 
 
 def glibc_version_string():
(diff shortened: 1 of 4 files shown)


Eric-Arellano:
Closing in favor of https://github.com/pantsbuild/pex/pull/692, which removes our custom code and fully copies the file correctly.

PR #701, opened 2019-04-03
Fix pex bootstrap interpreter selection.

### LINKED ISSUE #700: If sys.executable is not on PATH a pex will re-exec itself forever.
This can happen via:
```
/not/on/PATH/python a.pex
```

If `a.pex` in this example has compatibility constraints that rule out all other interpreters on `PATH` (or there are no intepreters on `PATH`!, c.f.: #53), then, even if sys.executable would be compatible to execute `a.pex` with, it is not considered useable:
https://github.com/pantsbuild/pex/blob/442ce7952674cb8a367ecded7d138661194d4f12/pex/pex_bootstrapper.py#L45-L50
https://github.com/pantsbuild/pex/blob/442ce7952674cb8a367ecded7d138661194d4f12/pex/interpreter.py#L283-L287

And so the pex_bootstrapper re-execs the pex ad infinitum:
https://github.com/pantsbuild/pex/blob/442ce7952674cb8a367ecded7d138661194d4f12/pex/pex_bootstrapper.py#L122-L137

jsirois:
Previously, `sys.executable` would be ignored if:
1. Selection fell back to `PATH` scanning.
2. The `sys.executable` was not on `PATH`.
3. No interpreters on `PATH` met interpreter contraints for the pex in
   question.

We now always include the current interpreter as a candidate when
falling back to a `PATH` scan.

Fixes #700

### DIFF
--- a/pex/pex_bootstrapper.py
+++ b/pex/pex_bootstrapper.py
@@ -44,7 +44,7 @@ def find_compatible_interpreters(pex_python_path=None, compatibility_constraints
       die('PEX_PYTHON_PATH was defined, but no valid interpreters could be identified. Exiting.')
   else:
     # We may have been invoked with a specific interpreter not on the $PATH, make sure our
-    # sys.executable included as a candidate in this case.
+    # sys.executable is included as a candidate in this case.
     interpreters = OrderedSet([PythonInterpreter.get()])
 
     # Add all qualifying interpreters found in $PATH.
--- a/pex/pex_bootstrapper.py
+++ b/pex/pex_bootstrapper.py
@@ -11,6 +11,7 @@ from pex.common import die
 from pex.executor import Executor
 from pex.interpreter import PythonInterpreter
 from pex.interpreter_constraints import matched_interpreters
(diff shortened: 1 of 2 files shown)


pex/pex_bootstrapper.py
```diff
@@ -42,12 +43,12 @@ def find_compatible_interpreters(pex_python_path, compatibility_constraints):
...
       die('PEX_PYTHON_PATH was defined, but no valid interpreters could be identified. Exiting.')
   else:
-    if not os.getenv('PATH', ''):
-      # no $PATH, use sys.executable
-      interpreters = [PythonInterpreter.get()]
-    else:
-      # get all qualifying interpreters found in $PATH
-      interpreters = PythonInterpreter.all()
+    # We may have been invoked with a specific interpreter not on the $PATH, make sure our
+    # sys.executable included as a candidate in this case.
```
stuhood: is included

jsirois: Fixed.

PR #702, opened 2019-04-03
Prepare the 1.6.6 release.

### LINKED ISSUE #699: Release 1.6.6
On the docket:
+ [x] Release more flexible pex binaries. #654
+ [x] If sys.executable is not on PATH a pex will re-exec itself forever. #700

jsirois:
Fixes #699

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Release Notes
 1.6.6
 -----
 
+This is the first release including only a single PEX pex, which
+supports execution under all interpreters pex supports.
+
 * Fix pex bootstrap interpreter selection. (#701)
   `PR #701 <https://github.com/pantsbuild/pex/pull/701>`_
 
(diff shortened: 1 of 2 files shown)


CHANGES.rst
```diff
@@ -1,6 +1,15 @@
 Release Notes
 =============
 
+1.6.6
+-----
+
```
Eric-Arellano: Consider adding a note explaining what #698 means for the end user. Something like:

```
This is the first release as a single multi-pex, meaning the released Pex may be ran with multiple different Python interpreter versions, rather than needing a distinct release for each interpreter.
```

jsirois: Isn't that redundant on the releases page where there will be a single pex binary available for download which says the same thing by inference?

jsirois: Erm. I guess this is also displayed - wonkily - on pypi where it isn't redundant. Will fix.

jsirois: Fixed.

PR #703, opened 2019-04-04
Fix pex file looses the executable permissions of binary files

fhoering:


tests/test_pex.py
```diff
@@ -236,6 +239,57 @@ def test_pex_run():
...
+    pex_dir = os.path.join(temp_dir, 'pex_dir')
+    safe_mkdir(pex_dir)
+
+    with open(os.path.join(pex_dir, 'exe.py'), 'w') as fp:
+      fp.write(textwrap.dedent('''
+               import subprocess
+               import os
+               import sys
+               import my_package
+               path = os.path.join(os.path.dirname(my_package.__file__),'bin/start.sh')
```
jsirois: comma-space: `s|,'bin|, 'bin/|`

pex/util.py
```diff
@@ -181,13 +181,19 @@ def cache_distribution(cls, zf, source, target_dir):
           with contextlib.closing(zf.open(name)) as zi:
             with safe_open(os.path.join(target_dir_tmp, target_name), 'wb') as fp:
               shutil.copyfileobj(zi, fp)
+              cls.copy_perm(zf.getinfo(zi.name), fp.name)
```
jsirois: Do you mind trying this diff instead with your existing test?:
```diff
$ git diff
diff --git a/pex/util.py b/pex/util.py
index 160ad51..e656eba 100644
--- a/pex/util.py
+++ b/pex/util.py
@@ -176,12 +176,9 @@ class CacheHelper(object):
       target_dir_tmp = target_dir + '.' + uuid.uuid4().hex
       for name in zf.namelist():
         if name.startswith(source) and not name.endswith('/'):
-          # strip off prefix + '/'
-          target_name = os.path.join(dependency_basename, name[len(source) + 1:])
-          with contextlib.closing(zf.open(name)) as zi:
-            with safe_open(os.path.join(target_dir_tmp, target_name), 'wb') as fp:
-              shutil.copyfileobj(zi, fp)
-
+          zf.extract(name, target_dir_tmp)
+      os.rename(os.path.join(target_dir_tmp, source),
+                os.path.join(target_dir_tmp, dependency_basename))
       rename_if_empty(target_dir_tmp, target_dir)
 
     dist = DistributionHelper.distribution_from_path(target_dir)
```

We already have perm-perserving logic built into the zips we open here:
https://github.com/pantsbuild/pex/blob/b2beec7ea48abdaf288ed40ac2b7644afa549c09/pex/common.py#L81-L102

It would be nice to re-use that and also use higher-level zip apis here to extract instead of rolling our own.

tests/test_pex.py
```diff
@@ -236,6 +239,57 @@ def test_pex_run():
...
+      bdist = DistributionHelper.distribution_from_path(installer.bdist())
+      pex_builder.add_dist_location(bdist.location)
+      pex_builder.set_executable(os.path.join(pex_dir, 'exe.py'))
+      pex_builder.freeze()
+
+      app_pex = os.path.join(os.path.join(temp_dir, 'out_pex_dir'), 'app.pex')
+      pex_builder.build(app_pex)
+      std_out, rc = run_simple_pex(app_pex,
+                                   env={'PEX_ROOT': os.path.join(temp_dir, '.pex')})
+      print(std_out)
```
jsirois: Please kill this print now that the test is working.

tests/test_pex.py
fhoering: Done

pex/util.py
fhoering: Tried it. It is also working. Thanks

tests/test_pex.py
fhoering: Done

jsirois:
I fixed a style error and everything is green so I'm going to go ahead and merge. Thanks @fhoering!

fhoering:
Thanks

PR #708, opened 2019-04-17
[fix] python 2 with None value in _pex_info

hbmartin:
A valid key might exit in _pex_info with a value of `None`. Under Python 2 this is guaranteed to crash.
This PR fixes that by checking for a None value to avoid calling `value.encode()`

### DIFF
--- a/pex/pex_info.py
+++ b/pex/pex_info.py
@@ -137,7 +137,7 @@ class PexInfo(object):
     if key not in self._pex_info:
       return None
     value = self._pex_info[key]
-    return value.encode('utf-8') if PY2 else value
+    return value.encode('utf-8') if PY2 and value is not None else value
 
   @property
   def build_properties(self):


Eric-Arellano:
Thanks @hbmartin for filing the bug report at https://github.com/pantsbuild/pants/issues/4150 and contributing this fix!

While this fix is correct, we're going to hold off on merging because it seems that the issue is upstream with Pants and not with Pex itself. We'd like to avoid changing Pex where possible.

jsirois:
Thanks for the effort you invested here @hbmartin but it looks like the issue at https://github.com/pantsbuild/pants/issues/4150 has gone stale and you can definitely create no-entry-point no-script Pex today in Pants. I'm going to close out this PR for those reasons; i.e.: the need for this / cause of this appears to have been fixed over in Pants.

PR #712, opened 2019-04-21
Fix import sorting.

jsirois:
Mechanically generated using `tox -eisort-run,style`. I'm not sure how
this got past CI.


jsirois:
> How did you notice this was wrong to begin with?

I run `tox -eisort-run,style` habitually before committing. That turned up edits to a file I wasn't editing.

PR #713, opened 2019-04-22
Delay importing pkg_resources until sys.path is ready.

georgi-yanchev:
Otherwise the master working set will be built upon the first namespace package, instead of when sys.path is fully ready.

### DIFF
--- a/pex/environment.py
+++ b/pex/environment.py
@@ -368,6 +368,9 @@ class PEXEnvironment(Environment):
         with TRACER.timed('Adding sitedir', V=2):
           site.addsitedir(dist.location)
 
-        self._declare_namespace_packages(dist)
+    # Delay calling 'self._declare_namespace_packages' until 'sys.path' contains all of the
+    # resolved dists.
+    for dist in resolved:
+      self._declare_namespace_packages(dist)
 
     return working_set


jsirois:
Thanks for the contribution! @georgi-yanchev can you detail the problem encountered before this change?

georgi-yanchev:
I have a project that has a few namespace packages (protobuf, matplotlib etc.) in its requirements.

When the the first of them (in a random manner) gets activated `environment._declare_namepspace_packages` will import `pkg_resources`, which in turn will initialize its master working set from `sys.path`.

The problem is that this happens while iterating resolved packages and `sys.path` is not fully prepared: https://github.com/pantsbuild/pex/blob/93000eed2c765c816141b8cf9bd90261d5c4a5ac/pex/environment.py#L371

So any packages activated after that will be missing from the master working set and respectively error upon first use.

This is my call stack (some line numbers might be offset a bit, because I added logging statements):

```
  File "/home/ec2-user/deploy/.bootstrap/pex/pex.py", line 351, in execute
    working_set = self._activate()
  File "/home/ec2-user/deploy/.bootstrap/pex/pex.py", line 93, in _activate
    for dist in env.activate():
  File "/home/ec2-user/deploy/.bootstrap/pex/environment.py", line 261, in activate
    self._working_set = self._activate()
  File "/home/ec2-user/deploy/.bootstrap/pex/environment.py", line 376, in _activate
    self._declare_namespace_packages(dist)
  File "/home/ec2-user/deploy/.bootstrap/pex/environment.py", line 327, in _declare_namespace_packages
    pkg_resources, vendored = _import_pkg_resources()
  File "/home/ec2-user/deploy/.bootstrap/pex/environment.py", line 34, in _import_pkg_resources
    import pkg_resources  # vendor:skip
  File "/home/ec2-user/deploy/.deps/setuptools-41.0.0-py2.py3-none-any.whl/pkg_resources/__init__.py", line 3246, in <module>
    @_call_aside
  File "/home/ec2-user/deploy/.deps/setuptools-41.0.0-py2.py3-none-any.whl/pkg_resources/__init__.py", line 3230, in _call_aside
    f(*args, **kwargs)
  File "/home/ec2-user/deploy/.deps/setuptools-41.0.0-py2.py3-none-any.whl/pkg_resources/__init__.py", line 3259, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "/home/ec2-user/deploy/.deps/setuptools-41.0.0-py2.py3-none-any.whl/pkg_resources/__init__.py", line 579, in _build_master
    ws = cls()
  File "/home/ec2-user/deploy/.deps/setuptools-41.0.0-py2.py3-none-any.whl/pkg_resources/__init__.py", line 566, in __init__
```

jsirois:
Thanks for the explanation. Do you feel up for adding a unit test so this doesn't regress? An example that goes through the hoops you need can be found here:
https://github.com/pantsbuild/pex/blob/45140bf30f7509c1203bdedcdc4f5b2ff772bf72/tests/test_environment.py#L61-L160

In particular, this adds wheels to a pex built from temporary content. You could do similar, creating several wheels that contribute to a namespace package.

georgi-yanchev:
I'm not sure I understand all the details about this test, but my concern is that a test like this would be very flaky if we have a regression. The order in which packages are activated is random, so if the namespace package happens to be the last, then the test will succeed. Eventually it will fail some other time when somebody makes unrelated change.

jsirois:
Sorry to keep pushing a bit on this. Can you provide more details on your failure case? I can't repro and I'd like to to find out what in setuptools is blowing up - it shouldn't need to see the whole final sys.path ideally.

In particular:
1. What line is blowing up in WorkingSet? Can you point to it here?:
https://github.com/pypa/setuptools/blob/7a80e29b31f255a3fff5147e50d0135271b7c101/pkg_resources/__init__.py#L553-L567
2. Can you provide more details about your pex? I tried to repro with `pex protobuf matplotlib setuptools==41.0.0 -orepro.pex && ./repro.pex`. Now that pex just has the details gleaned from your report - it may need more - namely an entrypoint, etc. Is it possible to fill more of that information in? Ideally you could provide a repro case that include the python version being selected both at pex build time and more importantly at pex runtime to help debug if this has anything to do with implicit namespace packages (may not at all).

georgi-yanchev:
To reproduce:

```
root@f1fc4c900189:~# pex -o crash.pex protobuf pysaml2
root@f1fc4c900189:~# echo 'import pkg_resources' > test.py
root@f1fc4c900189:~# echo 'print(pkg_resources.working_set.entry_keys.keys())' >> test.py
root@f1fc4c900189:~# echo 'import saml2' >> test.py
root@f1fc4c900189:~# ./crash.pex < test.py
```

Run the last line a few times until it crashes. Also note that the print statement has inconsistent output between runs.

georgi-yanchev:
A crash looks like this:

```
root@f1fc4c900189:~# ./crash.pex < test.py
Python 3.6.7 (default, Oct 22 2018, 11:32:17) 
[GCC 8.2.0] on linux
(InteractiveConsole)
>>> >>> dict_keys(['/root/crash.pex/.bootstrap', '/root/crash.pex', '/usr/lib/python36.zip', '/usr/lib/python3.6', '/usr/lib/python3.6/lib-dynload', '/root/.pex/install/idna-2.8-py2.py3-none-any.whl.645e82ab4157606d15a8766060dd690deebcc4f2/idna-2.8-py2.py3-none-any.whl', '/root/.pex/install/chardet-3.0.4-py2.py3-none-any.whl.c4332e34e38b781695ce775973fe40663558a897/chardet-3.0.4-py2.py3-none-any.whl', '/root/.pex/install/pytz-2019.1-py2.py3-none-any.whl.47e08203e9ccc6aa20e9930777429002176a0d0d/pytz-2019.1-py2.py3-none-any.whl', '/root/.pex/install/pyOpenSSL-19.0.0-py2.py3-none-any.whl.26fde777e246089ee721e597f21d697218990c6f/pyOpenSSL-19.0.0-py2.py3-none-any.whl', '/root/.pex/install/asn1crypto-0.24.0-py2.py3-none-any.whl.3ba24179c20a532e1adc78925a38dabcf1723d1a/asn1crypto-0.24.0-py2.py3-none-any.whl', '/root/.pex/install/setuptools-41.0.1-py2.py3-none-any.whl.ffa66c57fd90eee9c21efeb1d3d918a7b960fa4f/setuptools-41.0.1-py2.py3-none-any.whl', '/root/.pex/install/certifi-2019.3.9-py2.py3-none-any.whl.ab875f092e0a466c83c28bdfa7386590768a58ab/certifi-2019.3.9-py2.py3-none-any.whl', '/root/.pex/install/defusedxml-0.6.0-py2.py3-none-any.whl.1b790555c79b1025737c4dcc7068562a34b5bd7c/defusedxml-0.6.0-py2.py3-none-any.whl', '/root/.pex/install/urllib3-1.24.2-py2.py3-none-any.whl.65cee9682f93e61161329e3a7cfc995e3cd63282/urllib3-1.24.2-py2.py3-none-any.whl', '/root/.pex/install/protobuf-3.7.1-cp36-cp36m-manylinux1_x86_64.whl.6172f4520af8229633187590076297ca060e28c3/protobuf-3.7.1-cp36-cp36m-manylinux1_x86_64.whl'])
>>> Traceback (most recent call last):
  File "/root/.pex/install/pysaml2-4.7.0-py2.py3-none-any.whl.6f6dab5d3da6e843118fa9bb74f1fed8dad7fa69/pysaml2-4.7.0-py2.py3-none-any.whl/saml2/__init__.py", line 24, in <module>
    import saml2.version
  File "/root/.pex/install/pysaml2-4.7.0-py2.py3-none-any.whl.6f6dab5d3da6e843118fa9bb74f1fed8dad7fa69/pysaml2-4.7.0-py2.py3-none-any.whl/saml2/version.py", line 10, in <module>
    version = _parse_version()
  File "/root/.pex/install/pysaml2-4.7.0-py2.py3-none-any.whl.6f6dab5d3da6e843118fa9bb74f1fed8dad7fa69/pysaml2-4.7.0-py2.py3-none-any.whl/saml2/version.py", line 5, in _parse_version
    data = _pkg_resources.get_distribution('pysaml2')
  File "/root/.pex/install/setuptools-41.0.1-py2.py3-none-any.whl.ffa66c57fd90eee9c21efeb1d3d918a7b960fa4f/setuptools-41.0.1-py2.py3-none-any.whl/pkg_resources/__init__.py", line 481, in get_distribution
    dist = get_provider(dist)
  File "/root/.pex/install/setuptools-41.0.1-py2.py3-none-any.whl.ffa66c57fd90eee9c21efeb1d3d918a7b960fa4f/setuptools-41.0.1-py2.py3-none-any.whl/pkg_resources/__init__.py", line 357, in get_provider
    return working_set.find(moduleOrReq) or require(str(moduleOrReq))[0]
  File "/root/.pex/install/setuptools-41.0.1-py2.py3-none-any.whl.ffa66c57fd90eee9c21efeb1d3d918a7b960fa4f/setuptools-41.0.1-py2.py3-none-any.whl/pkg_resources/__init__.py", line 900, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/root/.pex/install/setuptools-41.0.1-py2.py3-none-any.whl.ffa66c57fd90eee9c21efeb1d3d918a7b960fa4f/setuptools-41.0.1-py2.py3-none-any.whl/pkg_resources/__init__.py", line 786, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'pysaml2' distribution was not found and is required by the application
>>> 
```

jsirois:
@georgi-yanchev - thanks for that repro case. I'll experiment for a bit before I comment further.

jsirois:
This gets things always working with the occasional print proving the fallback is used:
```diff
$ diff /home/jsirois/.pex/install/pysaml2-4.7.0-py2.py3-none-any.whl.7c2349ecb028ae0d8fe8ad943a08b1631b70ac7f/pysaml2-4.7.0-py2.py3-none-any.whl/saml2/version.py.orig /home/jsirois/.pex/install/pysaml2-4.7.0-py2.py3-none-any.whl.7c2349ecb028ae0d8fe8ad943a08b1631b70ac7f/pysaml2-4.7.0-py2.py3-none-any.whl/saml2/version.py
5c5,9
<     data = _pkg_resources.get_distribution('pysaml2')
---
>     try:
>         data = _pkg_resources.get_distribution('pysaml2')
>     except _pkg_resources.DistributionNotFound as e:
>         print('>>> searching more robustly after {}'.format(e))
>         data = _pkg_resources.WorkingSet().find(_pkg_resources.Requirement.parse('pysaml2'))
```

I know this is not a solution to your problem, but the underlying issue of setuptools building a global workingset object on 1st import is highly problematic. This means pex can easily "regress" if it ever imports pkg_resources midway through building up a sys,path. This sort of whac-a-mole is what I'm worried about. Thinking on options...

cosmicexplorer:
This change actually also fixes #729, which describes another deterministic repro case appearing in the wild.

cosmicexplorer:
Ok, I've just merged #730, since the fix in this PR happened to fix the problem in #729 as well. I'm not sure if there's work left here to do, since there is now some deterministic testing added for a closely related error.

georgi-yanchev:
Thanks for merging it. Closing this one.

Issue #714, opened 2019-04-25
Pex crashes because of incorrect source file


georgi-yanchev:
To reproduce:

```
(.venv) root@6a95660cfff6:~/aaa# pex -o crash.pex
(.venv) root@6a95660cfff6:~/aaa# echo 'import traceback' > /tmp/src.py
(.venv) root@6a95660cfff6:~/aaa# echo 'import inspect' >> /tmp/src.py
(.venv) root@6a95660cfff6:~/aaa# echo 'traceback.extract_stack()' >> /tmp/src.py 
(.venv) root@6a95660cfff6:~/aaa# echo 'inspect.stack()' >> /tmp/src.py 
(.venv) root@6a95660cfff6:~/aaa# ./crash.pex /tmp/src.py
Traceback (most recent call last):
  File ".bootstrap/pex/pex.py", line 352, in execute
  File ".bootstrap/pex/pex.py", line 284, in _wrap_coverage
  File ".bootstrap/pex/pex.py", line 315, in _wrap_profiling
  File ".bootstrap/pex/pex.py", line 400, in _execute
  File ".bootstrap/pex/pex.py", line 454, in execute_interpreter
  File ".bootstrap/pex/pex.py", line 490, in execute_content
  File ".bootstrap/pex/compatibility.py", line 81, in exec_function
  File "/tmp/src.py", line 4, in <module>
    inspect.stack()
  File "/usr/lib/python3.6/inspect.py", line 1499, in stack
    return getouterframes(sys._getframe(1), context)
  File "/usr/lib/python3.6/inspect.py", line 1476, in getouterframes
    frameinfo = (frame,) + getframeinfo(frame, context)
  File "/usr/lib/python3.6/inspect.py", line 1450, in getframeinfo
    lines, lnum = findsource(frame)
  File "/usr/lib/python3.6/inspect.py", line 828, in findsource
    if pat.match(lines[lnum]): break
IndexError: list index out of range
```

georgi-yanchev:
Happens because `linecache` module gets confused by overwritten `globals()['__main__']` at: https://github.com/pantsbuild/pex/blob/b2beec7ea48abdaf288ed40ac2b7644afa549c09/pex/pex.py#L488

jsirois:
Fixed by #715

PR #715, opened 2019-04-25
Make a copy of globals() before updating it.

georgi-yanchev:
Addresses: https://github.com/pantsbuild/pex/issues/714


jsirois:
Thanks @georgi-yanchev! I think the copying stands on its own without an added test. Mutating globals is clearly not what these few lines of code were after.

Issue #716, opened 2019-04-29
Make pex output deterministic / reproducible


stuhood:
To have deterministic / reproducible output, pex should produce a byte-for-byte identical output given identical inputs.

There are a few common cases where this breaks down:
1. unstable timestamps included in output archives (zip/tar files, generally)
    * This can be fixed by ensuring that pex uses hardcoded timestamps for the entries that it places in archives.
2. unstable shas/timestamps intentionally included in metadata
    * Fixing things like this might involve adding an option to either disable including this info, or to stabilize it.
3. unstable ordering of hash iteration between machines
    * Harder to hunt down, and harder to defend against. But fixing it involves using order preserving or sorted structures.
4. use of absolute paths, or paths that are host specific.

It is out of scope (for this ticket) to stabilize the input files to pex (ie, adding lockfile support). So in cases where the network is involved, structures should be sorted.

----

It's not clear which combination of these issues might be in play in pex, so it would be good to start by getting a reproducibility test/experiment harness in place that makes it easy to compare two pex outputs and identify the above issues.

jsirois:
This will be an excellent property to have. We can break out issues linked here as we pick off individual bits . There may or may not need to be upstream work to support stability of the archives embedded in built pexes - namely exploded wheels in .deps/ that pex sometimes has to build via `setup.py` execution.

Eric-Arellano:
Some initial thoughts on how to approach this.

* Use TDD. Add failing tests that we skip.
* Use Python stdlib's [`filecmp`](https://docs.python.org/3/library/filecmp.html) and maybe [`difflib`](https://docs.python.org/3/library/difflib.html) modules. Former is like shell's `cmp` program, will compare byte by byte. Latter will help to print diffs for easier to understand tests.
* Acceptance tests:
    * `pex -o 1.pex; pex -o 2.pex; cmp 1.pex 2.pex`
    * `pex requests cryptography -o 1.pex; pex requests cryptography -o 2.pex; cmp 1.pex 2.pex`
         * The actual dependencies used is subject to change. Idea is to test a universal wheel, ABI wheel, and we should also test a wheel that gets built from sdist.
    * `pex -m pydoc -o 1.pex;  pex -m pydoc -o 2.pex; cmp 1.pex 2.pex`
    * Test `-c` scenario. Haven't grokked what `-c` does yet, but seems important.
    * Specify `--python` and `--python-shebang`.
* Over time, identify which unit tests will be meaningful to include.
[...]

PR #717, opened 2019-04-30
Add failing acceptance tests for reproducible Pex builds

Eric-Arellano:
Per https://github.com/pantsbuild/pex/issues/716, we want to make Pex builds reproducible / deterministic.

As the first step in that process, we add failing acceptance tests that will act as the indicator of whether the issue is complete or not. As we get tests passing, such as the base test of no args, we will unskip those tests to avoid any regressions.

Here, we test the major permutations of building a Pex via `-o my-name.pex`, per https://pex.readthedocs.io/en/stable/buildingpex.html#saving-pex-files. We use Python's [`filecmp`](https://docs.python.org/3/library/filecmp.html) for a byte-by-byte comparison.

We must also add a `safe_sleep()` function to ensure Python 2 is not flaky, as `time.sleep()` is not guaranteed to run for the specified time until Python 3.5.

### DIFF
--- a/pex/common.py
+++ b/pex/common.py
@@ -166,8 +166,11 @@ def safe_sleep(seconds):
     time.sleep(seconds)
   else:
     start_time = time.time()
-    while time.time() - start_time < seconds:
-      pass
+    current_time = time.time()
+    while current_time - start_time < seconds:
+      remaining_time = seconds - (current_time - start_time)
+      time.sleep(remaining_time)
+      current_time = time.time()
 
 
 def rename_if_empty(src, dest, allowable_errors=(errno.EEXIST, errno.ENOTEMPTY)):
--- a/pex/common.py
+++ b/pex/common.py
@@ -11,6 +11,7 @@ import shutil
 import stat
 import sys
 import tempfile
+import time
 import threading
 import zipfile
 from collections import defaultdict
@@ -154,6 +155,19 @@ def safe_rmtree(directory):
   """Delete a directory if it's present. If it's not present, no-op."""
   if os.path.exists(directory):
     shutil.rmtree(directory, True)
+
+
+def safe_sleep(seconds):
+  """Ensure that the thread sleeps at a minimum the requested seconds.
+
+  Until Python 3.5, there was no guarantee that time.sleep() would actually sleep the requested
+  time. See https://docs.python.org/3/library/time.html#time.sleep."""
+  if sys.version_info[0:2] >= (3, 5):
+    time.sleep(seconds)
+  else:
+    start_time = time.time()
+    while time.time() - start_time < seconds:
+      pass
 
 
 def rename_if_empty(src, dest, allowable_errors=(errno.EEXIST, errno.ENOTEMPTY)):
--- a/pex/common.py
+++ b/pex/common.py
@@ -165,8 +165,7 @@ def safe_sleep(seconds):
   if sys.version_info[0:2] >= (3, 5):
     time.sleep(seconds)
   else:
-    start_time = time.time()
-    current_time = time.time()
+    start_time = current_time = time.time()
     while current_time - start_time < seconds:
       remaining_time = seconds - (current_time - start_time)
       time.sleep(remaining_time)
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -1375,29 +1375,29 @@ def test_reproducible_build_no_args():
 @pytest.mark.skip("Acceptance test for landing https://github.com/pantsbuild/pex/issues/716.")
 def test_reproducible_build_bdist_requirements():
   # We test both a pure Python wheel (six) and a platform-specific wheel (cryptography).
-  assert_reproducible(['six==1.12.0', 'cryptography==2.6.1'])
+  assert_reproducible_build(['six==1.12.0', 'cryptography==2.6.1'])
 
 
 @pytest.mark.skip("Acceptance test for landing https://github.com/pantsbuild/pex/issues/716.")
 def test_reproducible_build_sdist_requirements():
-  assert_reproducible(['pycparser==2.19', '--no-build'])
+  assert_reproducible_build(['pycparser==2.19', '--no-build'])
 
 
 @pytest.mark.skip("Acceptance test for landing https://github.com/pantsbuild/pex/issues/716.")
 def test_reproducible_m_flag():
-  assert_reproducible(['-m', 'pydoc'])
+  assert_reproducible_build(['-m', 'pydoc'])
 
 
 @pytest.mark.skip("Acceptance test for landing https://github.com/pantsbuild/pex/issues/716.")
 def test_reproducible_c_flag():
-  assert_reproducible(['black==19.3b0', '-c', 'black'])
+  assert_reproducible_build(['black==19.3b0', '-c', 'black'])
 
 
 @pytest.mark.skip("Acceptance test for landing https://github.com/pantsbuild/pex/issues/716.")
 def test_reproducible_python_flag():
-  assert_reproducible(['--python=python2.7'])
+  assert_reproducible_build(['--python=python2.7'])
 
 
 @pytest.mark.skip("Acceptance test for landing https://github.com/pantsbuild/pex/issues/716.")
 def test_reproducible_python_shebang_flag():
-  assert_reproducible(['--python-shebang=/usr/bin/python'])
+  assert_reproducible_build(['--python-shebang=/usr/bin/python'])
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -1346,8 +1346,11 @@ def assert_reproducible_build(args):
   with temporary_dir() as td:
     pex1 = os.path.join(td, '1.pex')
     pex2 = os.path.join(td, '2.pex')
-    run_pex_command(args + ['-o', pex1])
-    run_pex_command(args + ['-o', pex2])
+    # Note that we change the `PYTHONHASHSEED` to ensure that there are no issues resulting
+    # from the random seed, such as data structures, as Tox sets this value by default.
+    # See https://tox.readthedocs.io/en/latest/example/basic.html#special-handling-of-pythonhashseed.
+    run_pex_command(args + ['-o', pex1], env=make_env(PYTHONHASHSEED=111))
+    run_pex_command(args + ['-o', pex2], env=make_env(PYTHONHASHSEED=22222))
     # First explode the PEXes to compare file-by-file for easier debugging.
     with ZipFile(pex1) as zf1, ZipFile(pex2) as zf2:
       unzipped1 = os.path.join(td, "pex1")
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -7,7 +7,6 @@ import os
 import platform
 import subprocess
 import sys
-import time
 from contextlib import contextmanager
 from textwrap import dedent
 from zipfile import ZipFile
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -7,6 +7,7 @@ import os
 import platform
 import subprocess
 import sys
+import time
 from contextlib import contextmanager
 from textwrap import dedent
 from zipfile import ZipFile
@@ -1349,8 +1350,11 @@ def assert_reproducible_build(args):
     # Note that we change the `PYTHONHASHSEED` to ensure that there are no issues resulting
     # from the random seed, such as data structures, as Tox sets this value by default. See
     # https://tox.readthedocs.io/en/latest/example/basic.html#special-handling-of-pythonhashseed.
-    run_pex_command(args + ['-o', pex1], env=make_env(PYTHONHASHSEED=111))
-    run_pex_command(args + ['-o', pex2], env=make_env(PYTHONHASHSEED=22222))
+    def create_pex(path, seed):
+      run_pex_command(args + ['-o', path], env=make_env(PYTHONHASHSEED=seed))
+    create_pex(pex1, seed=111)
+    time.sleep(2)
+    create_pex(pex2, seed=22222)
     # First explode the PEXes to compare file-by-file for easier debugging.
     with ZipFile(pex1) as zf1, ZipFile(pex2) as zf2:
       unzipped1 = os.path.join(td, "pex1")
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -1390,7 +1390,16 @@ def test_reproducible_build_m_flag():
 
 @pytest.mark.skip("Acceptance test for landing https://github.com/pantsbuild/pex/issues/716.")
 def test_reproducible_build_c_flag():
-  assert_reproducible_build(['black==19.3b0', '-c', 'black'])
+  setup_py = dedent("""
+    from setuptools import setup
+
+    setup(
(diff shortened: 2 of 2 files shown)


tests/test_integration.py
```diff
@@ -1338,3 +1339,30 @@ def test_issues_539_abi3_resolution():
...
+  # With pre-built wheels, including
+  # 1) a universal wheel - six
+  # 2) an abi wheel - cryptography
+  assert_reproducible(['six==1.12.0', 'cryptography==2.6.1'])
+
+  # With an sdist
+  assert_reproducible(['pycparser==2.19', '--no-build'])
+
+  assert_reproducible(['-m', 'pydoc'])
+  assert_reproducible(['black==19.3b0', '-c', 'black'])
```
Eric-Arellano: I'd love to use a command that does not involve PyPi, since we already test that above. However, I can't think of a command that involves `-c` and does not first download a dependency. Let me know if you know of anything.

tests/test_integration.py
```diff
@@ -1338,3 +1339,30 @@ def test_issues_539_abi3_resolution():
...
+@pytest.mark.skip("Acceptance test for landing https://github.com/pantsbuild/pex/issues/716.")
+def test_reproducible_build():
+
+  def assert_reproducible(args):
+    with temporary_dir() as td:
+      pex1 = os.path.join(td, '1.pex')
+      pex2 = os.path.join(td, '2.pex')
+      run_pex_command(args + ['-o', pex1])
+      run_pex_command(args + ['-o', pex2])
+      assert filecmp.cmp(pex1, pex2, shallow=False)
```
stuhood: I think that you'll want to maybe have two asserts: one that checks the unzipped pex, then another that checks the zipped pex. I think that the former might have more useful/readable output while being easier to pass, while the latter is harder to pass, and has less useful output.

Eric-Arellano: Good suggestion! Added. Think this is the appropriate level of testing now—original was nearly impossible to debug and then anything more than this would be getting too specific and complicated.

tests/test_integration.py
```diff
@@ -1338,3 +1340,41 @@ def test_issues_539_abi3_resolution():
...
+
+  def assert_reproducible(args):
+    with temporary_dir() as td:
+      pex1 = os.path.join(td, '1.pex')
+      pex2 = os.path.join(td, '2.pex')
+      run_pex_command(args + ['-o', pex1])
+      run_pex_command(args + ['-o', pex2])
+      # First explode the PEXes to compare file-by-file for easier debugging.
+      with ZipFile(pex1) as zf1, ZipFile(pex2) as zf2:
+        zf1.extractall(path=os.path.join(td, "pex1"))
```
jsirois: Standard zipfile will lose perms on extract, but the overall filecmp on 1365 will catch if perms are different, so all good as is.

tests/test_integration.py
jsirois: Sure, you can build a PEX with just sources and no deps by either pointing at a local setup.py - you spell the requirement as the path of the dir containing setup.py - or use ... `--sources-directory` & `--resources-directory`.

tests/test_integration.py
```diff
@@ -1338,3 +1340,41 @@ def test_issues_539_abi3_resolution():
...
+        for member1, member2 in zip(zf1.namelist(), zf2.namelist()):
+          assert filecmp.cmp(
+            os.path.join(td, "pex1", member1),
+            os.path.join(td, "pex2", member2),
+            shallow=False
+          )
+      # Then compare the original .pex files. This is the assertion we truly care about.
+      assert filecmp.cmp(pex1, pex2, shallow=False)
+
+  assert_reproducible([])
```
jsirois: You might want to isolate these asserts as separate tests for sanity sake.

tests/test_integration.py
```diff
@@ -1338,3 +1340,41 @@ def test_issues_539_abi3_resolution():
...
+
+
+@pytest.mark.skip("Acceptance test for landing https://github.com/pantsbuild/pex/issues/716.")
+def test_reproducible_build():
+
+  def assert_reproducible(args):
+    with temporary_dir() as td:
+      pex1 = os.path.join(td, '1.pex')
+      pex2 = os.path.join(td, '2.pex')
+      run_pex_command(args + ['-o', pex1])
```
jsirois: So, you may miss alot unless you set two different PYTHONHASHSEED. Pytest sets one via ... I'm not sure. But if via env car export, you may get that inherited for both PEX commands which is not what you want to ferret out the datastructure-related issues.

tests/test_integration.py
Eric-Arellano: Good suggestion. Makes it easier to unskip tests, for example.

tests/test_integration.py
Eric-Arellano: Agreed. Good catch. Setting the `env` in `run_pex_command` will temporarily override for the subprocess whatever Tox sets beforehand (which indeed is via an env var https://tox.readthedocs.io/en/latest/example/basic.html#special-handling-of-pythonhashseed).

tests/test_integration.py
```diff
@@ -1338,3 +1340,73 @@ def test_issues_539_abi3_resolution():
...
+  assert_reproducible_build(['pycparser==2.19', '--no-build'])
+
+
+@pytest.mark.skip("Acceptance test for landing https://github.com/pantsbuild/pex/issues/716.")
+def test_reproducible_build_m_flag():
+  assert_reproducible_build(['-m', 'pydoc'])
+
+
+@pytest.mark.skip("Acceptance test for landing https://github.com/pantsbuild/pex/issues/716.")
+def test_reproducible_build_c_flag():
```
Eric-Arellano: @jsirois does this test look right? Do I need `zip_safe=True`?

The PEXes get built without error and only fail because they aren't equal to each other, so I think it's working correctly.

jsirois: The `zip_safe` characterisitc is completely orthoganal to `-c`. If `-c` is a `console_scripts` entry point, thats found via `sys.path` and if its a script, that's found via resource files (so also via `sys.path`). So this looks fine, but it only tests 1/2 of -c. I'd move on though and circle back to test the other half later.

tests/test_integration.py
```diff
@@ -1349,8 +1350,11 @@ def assert_reproducible_build(args):
     # Note that we change the `PYTHONHASHSEED` to ensure that there are no issues resulting
     # from the random seed, such as data structures, as Tox sets this value by default. See
     # https://tox.readthedocs.io/en/latest/example/basic.html#special-handling-of-pythonhashseed.
-    run_pex_command(args + ['-o', pex1], env=make_env(PYTHONHASHSEED=111))
-    run_pex_command(args + ['-o', pex2], env=make_env(PYTHONHASHSEED=22222))
+    def create_pex(path, seed):
+      run_pex_command(args + ['-o', path], env=make_env(PYTHONHASHSEED=seed))
+    create_pex(pex1, seed=111)
+    time.sleep(2)
```
jsirois: This is horrible to have to do. Do you have any understanding why this is needed? I'm just very sensitive to adding sleeps - we've done much too much of that back over in Pants.

Eric-Arellano: You beat me to writing my explanation:

I had to add this line to avoid flaky tests. By removing `.pyc`, the exploded Pex becomes equivalent for all but the `bdist` case! 

However, the final assertion on the `.pex` files flakily fails. With an appropriate `time.sleep()`, it consistently fails. I suspect this has to do with file metadata changing, but haven't investigated yet.

Flagging this because I know `time.sleep()` has caused much pain in Pants CI. 2 seconds consistently failed all tests for me on macOS. Should it 3 in case CI is slower than my mac? Or is there a better alternative?

jsirois: Really the only good alternative is to figure out what timing has to do with anything and coment the sleep purpose if it's needed. Thinking about this, it probably is needed and its needed to ensure "now" is different to get different timestamps in bits that have timestamp dependence.

jsirois: OK - feeling better about this - I think that's it. A comment is enough in that case. This is one of the very rare cases sleeping is ok / needed.

jsirois: Discussed offline, but older time.sleep could / can sleep less than requested.

See:
+ documented too little or too much: https://docs.python.org/2/library/time.html#time.sleep
+ fixed to never be too little: https://github.com/python/cpython/commit/79d68f929d8def878766965f513b628023f809b5#diff-0fbe6dc9a4b14052d6815e3ae9dce53eR353

I'd be interested to know if the failures you see using `time.sleep(1)` are only on python < 3.5. If so, yay - full explanation. Also if so, a test utility to use time.sleep + tracking elapsed time to ensure the requested sleep time is met or exceeded would be great. We could keep the sleep both minimal and fully understood.

Eric-Arellano: No, this is flaking with Python 3.7 :/ confirmed that 1.1 seconds flakes for me with Py37.

To reproduce, apply this diff:
```diff
diff --git a/pex/bin/pex.py b/pex/bin/pex.py
index 7095833..081e6a2 100755
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -291,6 +291,17 @@ def configure_clp_pex_options(parser):
            'No value (alias for prefer, for backwards compatibility). '
            '[Default: %default]')

+  group.add_option(
+      '--reproducible', '--not-reproducible',
+      dest='reproducible',
+      default=True,
+      action='callback',
+      callback=parse_bool,
+      help='Ensure that the generated Pex file is completely reproducible, i.e. if you were '
+           'to run the same command again, the new PEX would be byte-for-byte equivalent '
+           'to the original. This means that the generated Pex will not have .pyc files '
+           'included, which results in a slight startup performance hit.')
+
   parser.add_option_group(group)


@@ -674,7 +685,7 @@ def main(args=None):
     with TRACER.timed('Building pex'):
       pex_builder = build_pex(reqs, options, resolver_options_builder)

-    pex_builder.freeze()
+    pex_builder.freeze(bytecode_compile=not options.reproducible)
     pex = PEX(pex_builder.path(),
               interpreter=pex_builder.interpreter,
               verify_entry_point=options.validate_ep)
@@ -683,7 +694,7 @@ def main(args=None):
       log('Saving PEX file to %s' % options.pex_name, V=options.verbosity)
       tmp_name = options.pex_name + '~'
       safe_delete(tmp_name)
-      pex_builder.build(tmp_name)
+      pex_builder.build(tmp_name, bytecode_compile=not options.reproducible)
       os.rename(tmp_name, options.pex_name)
     else:
       if not _compatible_with_current_platform(options.platforms):
```

Then this diff to unskip the tests:
```diff
diff --git a/tests/test_integration.py b/tests/test_integration.py
index 24a8ea0..7d08f63 100644
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -1373,7 +1373,7 @@ def assert_reproducible_build(args):
     assert filecmp.cmp(pex1, pex2, shallow=False)


-@pytest.mark.skip("Acceptance test for landing https://github.com/pantsbuild/pex/issues/716.")
+# @pytest.mark.skip("Acceptance test for landing https://github.com/pantsbuild/pex/issues/716.")
 def test_reproducible_build_no_args():
   assert_reproducible_build([])

@@ -1384,17 +1384,17 @@ def test_reproducible_build_bdist_requirements():
   assert_reproducible_build(['six==1.12.0', 'cryptography==2.6.1'])


-@pytest.mark.skip("Acceptance test for landing https://github.com/pantsbuild/pex/issues/716.")
+# @pytest.mark.skip("Acceptance test for landing https://github.com/pantsbuild/pex/issues/716.")
 def test_reproducible_build_sdist_requirements():
   assert_reproducible_build(['pycparser==2.19', '--no-build'])


-@pytest.mark.skip("Acceptance test for landing https://github.com/pantsbuild/pex/issues/716.")
+# @pytest.mark.skip("Acceptance test for landing https://github.com/pantsbuild/pex/issues/716.")
 def test_reproducible_build_m_flag():
   assert_reproducible_build(['-m', 'pydoc'])


-@pytest.mark.skip("Acceptance test for landing https://github.com/pantsbuild/pex/issues/716.")
+# @pytest.mark.skip("Acceptance test for landing https://github.com/pantsbuild/pex/issues/716.")
 def test_reproducible_build_c_flag():
   setup_py = dedent("""
     from setuptools import setup
@@ -1408,11 +1408,11 @@ def test_reproducible_build_c_flag():
     assert_reproducible_build([project_dir, '-c', 'my_app'])


-@pytest.mark.skip("Acceptance test for landing https://github.com/pantsbuild/pex/issues/716.")
+# @pytest.mark.skip("Acceptance test for landing https://github.com/pantsbuild/pex/issues/716.")
 def test_reproducible_build_python_flag():
   assert_reproducible_build(['--python=python2.7'])


-@pytest.mark.skip("Acceptance test for landing https://github.com/pantsbuild/pex/issues/716.")
+# @pytest.mark.skip("Acceptance test for landing https://github.com/pantsbuild/pex/issues/716.")
 def test_reproducible_build_python_shebang_flag():
   assert_reproducible_build(['--python-shebang=/usr/bin/python'])
```

Then `tox -v -e py37-integration -- -k test_reproducible_build`.

--

Given this stabilization to `time.sleep()` in Py3, I don't think our own custom function should make a difference? I can try if you think it will.

jsirois: It won't make a difference and we don't understand what's going on here. If you can just call that out in a comment near the sleep and perhaps file a linked issue to get to the bottom of it

jsirois: Aha. RIF strikes again. See '4.4.6 date and time fields:' here https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT

I think this is fully solved. We need a two second sleep. Under python < 3.5 we need a sleep helper function.

jsirois: And to make sense of the two seconds, they allocated 5 bits to the second field, so 32 distinct values, so can only represent every 2 seconds: http://www.vsft.com/hal/dostime.htm

"Neat"!

pex/common.py
```diff
@@ -156,6 +157,19 @@ def safe_rmtree(directory):
...
+  """Ensure that the thread sleeps at a minimum the requested seconds.
+
+  Until Python 3.5, there was no guarantee that time.sleep() would actually sleep the requested
+  time. See https://docs.python.org/3/library/time.html#time.sleep."""
+  if sys.version_info[0:2] >= (3, 5):
+    time.sleep(seconds)
+  else:
+    start_time = time.time()
+    while time.time() - start_time < seconds:
+      pass
```
jsirois: Eek! Busy looping is for a handful of ms or less. This should still use time.sleep in the loop, but probably sleep the current defecit, keeping track of the defecit.

pex/common.py
```diff
@@ -166,8 +166,11 @@ def safe_sleep(seconds):
     time.sleep(seconds)
   else:
     start_time = time.time()
-    while time.time() - start_time < seconds:
-      pass
+    current_time = time.time()
```
jsirois: Excellent. You might just say `current_time = start_time = time.time()` here.

Eric-Arellano:
> Yay!

Thanks for your help John in landing this! Very happy we have confidence in why everything works the way it does and that we are not introducing flakes. Impressive debugging work with the zip spec 😀

PR #720, opened 2019-05-01
Document how Pex developers can run specific tests and run Pex from source

Eric-Arellano:
These newly documented commands are highly useful for Pants developers, and not immediately obvious for people who have no worked with Pex or used tox before.

However, we also want people reading the Tox docs and not solely relying on our guide, so we try to strike a balance between the two.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -146,8 +146,8 @@ running ``tox --listenvs-all``, then invoke like this:
 
     $ tox -e style
 
-All of our tox environments allow passthrough arguments, which can be helpful to run specific
-tests:
+All of our tox test environments allow passthrough arguments, which can be helpful to run
+specific tests:
 
 .. code-block::
 
--- a/README.rst
+++ b/README.rst
@@ -134,17 +134,22 @@ If you don't have tox, you can generate a pex of tox:
 
     $ pex tox -c tox -o ~/bin/tox
 
+Tox provides many useful commands and options, explained at https://testrun.org/tox/en/latest/.
+Below, we provide some of the most commonly used commands used when working on Pex, but the
+docs are worth acquainting yourself with to better understand how Tox works and how to do more
+advanced commmands.
+
 To run a specific test command, look in `tox.ini` for the name and run like this:
 
 .. code-block::
 
-    $ pex tox -v -e style
+    $ tox -e style
 
 Tox allows passthrough arguments, which can be helpful to run specific tests:
 
 .. code-block::
 
-    $ pex tox -v -e py37-integration -- -k test_reproducible_build
+    $ tox -e py37-integration -- -k test_reproducible_build
 
 To run Pex from source, rather than through what is on your PATH, invoke via Python:
 
--- a/README.rst
+++ b/README.rst
@@ -140,7 +140,7 @@ docs are worth acquainting yourself with to better understand how Tox works and 
 advanced commmands.
 
 To run a specific environment, identify the name of the environment you'd like to invoke by
-running ``tox --listenvs-all`, then invoke like this:
+running ``tox --listenvs-all``, then invoke like this:
 
 .. code-block::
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -134,6 +134,23 @@ If you don't have tox, you can generate a pex of tox:
...
+
+.. code-block::
+
+    $ pex tox -v -e style
+
+Tox allows passthrough arguments, which can be helpful to run specific tests:
+
+.. code-block::
+
+    $ pex tox -v -e py37-integration -- -k test_reproducible_build
```
blorente: IIRC, I had to modify this line to be able to pass arbitrary passthrough args (`-k test_...`): https://github.com/pantsbuild/pex/blob/master/tox.ini#L44

In particular, I had to change `{posargs:-vvs}` to just `{posargs}`. I may have been doing something else wrong, just checking that this was not the case anymore.

README.rst
```diff
@@ -134,6 +134,23 @@ If you don't have tox, you can generate a pex of tox:
 
     $ pex tox -c tox -o ~/bin/tox
 
+To run a specific test command, look in `tox.ini` for the name and run like this:
+
+.. code-block::
+
+    $ pex tox -v -e style
```
jsirois: Copypasta here and below: `s/pex //`

README.rst
```diff
@@ -134,6 +134,23 @@ If you don't have tox, you can generate a pex of tox:
 
     $ pex tox -c tox -o ~/bin/tox
 
+To run a specific test command, look in `tox.ini` for the name and run like this:
```
jsirois: There is also `tox -a`.

Eric-Arellano:
> I'm not a big fan of summarizing things canonically explained at https://testrun.org/tox/en/latest/. Tox is super-standard - so pointing to it and expecting a dev to learn it seemed to make sense to me when Wickman instituted it. I'm fine letting these stand but I'd just like to point out that giving folks shortcuts also lets them shortcut and this can be to their own detriment.

Acknowledged. On the other hand, there are a couple commands that we would expect everyone to run, which are 1) pre-commit lint, 2) running a specific test file and/or specific tests, and 3) running the Pex binary from source. I tried with this new version to strike a balance between not expecting devs to reinvent the wheel with how to do these 3 commands, while still encouraging them to read the docs.

For what it's worth, I was very intimidated when first starting working on Pex because I had never used Tox. Even after reading the docs, I felt frustrated with how difficult it was for me to figure out how to do simple commands like run a single test. I found (find) the docs primarily focused on creating the `tox.ini` and virtualenv, rather than on how to consume it. 

I'd like to reduce friction for new devs as much as we can, while balancing your concern about overreach.

README.rst
Eric-Arellano: My search skills are failing me..can't find what this means either on Tox's site or duckduckgo/google search. Mind explaining?

Tox's docs seem to use `-e` everywhere I find.

jsirois: Try running `tox --help`.

Eric-Arellano: Ah yes that would do the trick... Thanks!

README.rst
```diff
@@ -134,6 +134,30 @@ If you don't have tox, you can generate a pex of tox:
...
+advanced commmands.
+
+To run a specific environment, identify the name of the environment you'd like to invoke by
+running ``tox --listenvs-all``, then invoke like this:
+
+.. code-block::
+
+    $ tox -e style
+
+All of our tox environments allow passthrough arguments, which can be helpful to run specific
```
jsirois: This is definitely not true. I think you mean to say "all of our to test environments"

Eric-Arellano: Auditing `tox.ini`, it seemed every environment extends `testenv`. Either way, fixed to qualify this, especially in case anything changes in the future.

Thanks for your help in landing this PR!

PR #722, opened 2019-05-03
Add new `--no-use-system-time` flag to use a deterministic timestamp in built PEX

Eric-Arellano:
### Problem
Part of https://github.com/pantsbuild/pex/issues/716 to make built PEXes be reproducible.

One of the major sources of unreproducibility is timestamps embedded into the zipfile. Python's `zipfile.py` uses the current system time when creating a new zipfile, per https://github.com/python/cpython/blob/45e92fc02d57a7219f4f4922179929f19b3d1c28/Lib/zipfile.py#L509. So, the Pex cannot be reproducible if built after the original run.

### Solution
Introduce a new flag `--use-system-time` / `--no-use-system-time` that will toggle whether we use a deterministic timestamp or keep the default behavior of using the current system time. We will default to `--use-system-time` for now to keep prior behavior, and then in a future release like 1.70 will switch over to default to reproducible builds.

If `--no-use-system-time` is specified, we use midnight at January 1, 1980, which is the start of the MS-DOS time used by Zipfiles. We then use this value to construct a `ZipInfo` with the desired date.

#### Not respecting `SOURCE_DATE_EPOCH`
We do not respect the standard env var `$SOURCE_DATE_EPOCH` often used by tools for deterministic timestamps, as doing so risks reducing the reproducibility of built pexes across different platforms, e.g. if two platforms set the env var differently. Usually, this is supposed to be set for the sake of security to check that a shipped binary was not tampered with, but that is not our primary use case, so we do not respect it both for simplicity of our codebase and to avoid this potential reduction in reproducibility. Refer to https://reproducible-builds.org/docs/source-date-epoch/ for more information.

### Result
4 / 7 of the acceptance tests for reproducibility now pass consistently!

When setting CI to always use a deterministic timestamp, [CI was green](https://travis-ci.org/pantsbuild/pex/builds/527997029) so things will work as expected when we change the default to deterministic timestamps.

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -307,7 +307,7 @@ def configure_clp_pex_options(parser):
       dest='use_system_time',
       # TODO: set this to True. It's only False for now to test what happens in CI if we
       # always use deterministic timestamps.
-      default=False,
+      default=True,
       action='callback',
       callback=parse_bool,
       help='Use the current system time to generate timestamps for the new pex. Otherwise, Pex '
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -302,6 +302,18 @@ def configure_clp_pex_options(parser):
            'likely will not be reproducible, meaning that if you were to run `./pex -o` with the '
            'same inputs then the new pex would not be byte-for-byte identical to the original.')
 
+  group.add_option(
+      '--use-system-time', '--no-use-system-time',
+      dest='use_system_time',
+      # TODO: set this to True. It's only False for now to test what happens in CI if we
+      # always use deterministic timestamps.
+      default=False,
+      action='callback',
+      callback=parse_bool,
+      help='Use the current system time to generate timestamps for the new pex. Otherwise, Pex '
+           'will first try to read the time from the environment variable SOURCE_DATE_EPOCH and '
+           'will fallback to midnight on January 1, 1980. TODO: explain reproducibility.')
+
   parser.add_option_group(group)
 
 
@@ -694,7 +706,11 @@ def main(args=None):
       log('Saving PEX file to %s' % options.pex_name, V=options.verbosity)
       tmp_name = options.pex_name + '~'
       safe_delete(tmp_name)
-      pex_builder.build(tmp_name, bytecode_compile=options.compile)
+      pex_builder.build(
+        tmp_name,
+        bytecode_compile=options.compile,
+        deterministic_timestamp=not options.use_system_time
+      )
       os.rename(tmp_name, options.pex_name)
     else:
       if not _compatible_with_current_platform(options.platforms):
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -307,7 +307,7 @@ def configure_clp_pex_options(parser):
       dest='use_system_time',
       # TODO: set this to True. It's only False for now to test what happens in CI if we
       # always use deterministic timestamps.
-      default=True,
+      default=False,
       action='callback',
       callback=parse_bool,
       help='Use the current system time to generate timestamps for the new pex. Otherwise, Pex '
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -312,7 +312,9 @@ def configure_clp_pex_options(parser):
       callback=parse_bool,
       help='Use the current system time to generate timestamps for the new pex. Otherwise, Pex '
            'will first try to read the time from the environment variable SOURCE_DATE_EPOCH and '
-           'will fallback to midnight on January 1, 1980. TODO: explain reproducibility.')
+           'will fallback to midnight on January 1, 1980. By using system time, the generated pex '
+           'will not be reproducible, meaning that if you were to run `./pex -o` with the '
+           'same inputs then the new pex would not be byte-for-byte identical to the original.')
 
   parser.add_option_group(group)
 
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -309,7 +309,6 @@ def configure_clp_pex_options(parser):
       action='callback',
       callback=parse_bool,
       help='Use the current system time to generate timestamps for the new pex. Otherwise, Pex '
-           'will first try to read the time from the environment variable SOURCE_DATE_EPOCH and '
            'will fallback to midnight on January 1, 1980. By using system time, the generated pex '
            'will not be reproducible, meaning that if you were to run `./pex -o` with the '
            'same inputs then the new pex would not be byte-for-byte identical to the original.')
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -309,7 +309,7 @@ def configure_clp_pex_options(parser):
       action='callback',
       callback=parse_bool,
       help='Use the current system time to generate timestamps for the new pex. Otherwise, Pex '
-           'will fallback to midnight on January 1, 1980. By using system time, the generated pex '
+           'will use midnight on January 1, 1980. By using system time, the generated pex '
            'will not be reproducible, meaning that if you were to run `./pex -o` with the '
            'same inputs then the new pex would not be byte-for-byte identical to the original.')
 
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -305,9 +305,7 @@ def configure_clp_pex_options(parser):
   group.add_option(
       '--use-system-time', '--no-use-system-time',
       dest='use_system_time',
-      # TODO: set this to True. It's only False for now to test what happens in CI if we
-      # always use deterministic timestamps.
-      default=False,
+      default=True,
       action='callback',
       callback=parse_bool,
       help='Use the current system time to generate timestamps for the new pex. Otherwise, Pex '
--- a/pex/common.py
+++ b/pex/common.py
@@ -20,11 +20,10 @@ from uuid import uuid4
 
 
 # We use the start of MS-DOS time, which is what zipfiles use (see section 4.4.6 of
-# https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT). Note that ZipInfo expects a
-# time.struct_time, which we set here.
+# https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT).
 DETERMINISTIC_DATETIME = datetime(
   year=1980, month=1, day=1, hour=0, minute=0, second=0, tzinfo=None
-).timetuple()
+)
 
 
 def die(msg, exit_code=1):
@@ -407,7 +406,7 @@ class Chroot(object):
         zinfo = zf.zip_info_from_file(
             filename=full_path,
             arcname=f,
-            date_time=DETERMINISTIC_DATETIME if deterministic_timestamp else None
+            date_time=DETERMINISTIC_DATETIME.timetuple() if deterministic_timestamp else None
         )
         with open(full_path, 'rb') as open_f:
           data = open_f.read()
--- a/pex/common.py
+++ b/pex/common.py
@@ -175,11 +175,11 @@ class DeterministicTimestampZipFile(PermPreservingZipFile):
 
 
 @contextlib.contextmanager
-def open_zip(path, *args, use_deterministic_timestamp=False, **kwargs):
+def open_zip(path, *args, deterministic_timestamp=False, **kwargs):
   """A contextmanager for zip files. Passes through positional and kwargs to zipfile.ZipFile."""
   zf = (
     DeterministicTimestampZipFile(path, *args, **kwargs)
-    if use_deterministic_timestamp
+    if deterministic_timestamp
     else PermPreservingZipFile(path, *args, **kwargs)
   )
   with contextlib.closing(zf) as zip:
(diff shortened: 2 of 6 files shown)


pex/bin/pex.py
```diff
@@ -302,6 +302,18 @@ def configure_clp_pex_options(parser):
            'likely will not be reproducible, meaning that if you were to run `./pex -o` with the '
            'same inputs then the new pex would not be byte-for-byte identical to the original.')
 
+  group.add_option(
+      '--use-system-time', '--no-use-system-time',
```
Eric-Arellano: Open to suggestions for a better name.

The main use case I want to optimize for is when deterministic time will be the default. For example, atm `--use-deterministic-time` would probably be a good name, but once we make the switch then `--no-use-deterministic-time` wouldn't be very descriptive for what the behavior would end up being.

illicitonion: In https://github.com/pantsbuild/pants/pull/6094 we made a flag `-jar-creation-time` which, if not set, defaulted to the current time. Not sure whether this is a good or bad default for pex, but it's some prior art :)

stuhood: @illicitonion : I think that we agreed on the other issue that we want this enabled by default, so the default would need to be stable.

@Eric-Arellano : This name looks good I think?

pex/common.py
```diff
@@ -15,9 +15,32 @@
...
 import zipfile
 from collections import defaultdict
+from datetime import datetime
 from uuid import uuid4
 
 
+def deterministic_datetime():
+  """Return a deterministic UTC datetime object that does not depend on the system time.
+
+  First try to read from the standard $SOURCE_DATE_EPOCH, then default to midnight January 1,
```
stuhood: IMO, ever successfully reading from `SOURCE_DATE_EPOCH` would introduce a bug. So I think that it would be good to ignore any environment variables for these purposes.

pex/common.py
```diff
@@ -15,9 +15,32 @@
...
+  1980, which is the start of time for MS-DOS time. Per section 4.4.6 of the zip spec at
+  https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT, Zipfiles use MS-DOS time. So, we
+  use this default to ensure no issues with Zipfiles.
+
+  Even though we use MS-DOS to inform the default value, note that we still return a normal UTC
+  datetime object for flexibility with how the result is used.
+
+  For more information on $SOURCE_DATE_EPOCH, refer to
+  https://reproducible-builds.org/docs/source-date-epoch/."""
+  return (
```
stuhood: This method is going to be called a lot: it might be good to just make it a constant instead? Or have a private method that is called to initialize a public constant at import time?

pex/common.py
```diff
@@ -363,7 +386,45 @@ def __str__(self):
...
       for f in sorted(self.files()):
-        zf.write(os.path.join(self.chroot, f), arcname=f, compress_type=zipfile.ZIP_DEFLATED)
+        full_path = os.path.join(self.chroot, f)
+        if not deterministic_timestamp:
+          zf.write(
+            full_path,
+            arcname=f,
+            compress_type=zipfile.ZIP_DEFLATED
+          )
+        else:
```
stuhood: This `else` should be broken out as a separate method I think.

pex/pex_builder.py
```diff
@@ -428,11 +428,12 @@ def freeze(self, bytecode_compile=True):
       self._precompile_source()
     self._frozen = True
 
-  def build(self, filename, bytecode_compile=True):
+  def build(self, filename, bytecode_compile=True, deterministic_timestamp=False):
```
stuhood: Should this default to `True`? Or are we assuming that folks using pex as a library would not want something like this to change?

John would likely know better, but: my feeling is that this is a reasonable behaviour to change in a `1.6.x` -> `1.7.x` bump.

pex/common.py
Eric-Arellano: I don’t understand why this would introduce a bug? The reproducible builds project highly encourages tools like ours to let people define the value. If they don’t define it, we just use our default.

pex/common.py
Eric-Arellano: Would the motivation be to save performance? This function appears quite cheap in terms of performance. Both the proposals would increase the complexity and I’m concerned would be premature optimization.

pex/pex_builder.py
Eric-Arellano: The tentative plan is to change the default in 1.70 to be True. For now, we want to keep behavior the same as before during the transition period.

pex/common.py
stuhood: The concern is that if I'm on a linux distribution that sets that variable, I will cache miss on a distribution that doesn't set that variable... which makes the whole thing less deterministic, rather than more. I don't see a good reason to consume it... can you share a link?

pex/common.py
stuhood: The suggestion adds ~1 line afaict:
```
def _deterministic_datetime():
  ..

DETERMINISTIC_DATETIME = _deterministic_datetime()
```

pex/pex_builder.py
stuhood: Got it.

pex/common.py
Eric-Arellano: Okay I agree this should be a constant. I made the change. Thanks!

pex/common.py
Eric-Arellano: https://reproducible-builds.org/docs/source-date-epoch/

> The concern is that if I'm on a linux distribution that sets that variable, I will cache miss on a distribution that doesn't set that variable...

Could you please elaborate on this concern? For example, in Pants it doesn't seem like this would be an issue. Our Ubuntu shards and Centos6 shards both don't set `SOURCE_DATE_EPOCH`, so I don't see any immediate concern for Pants.

More generally, this is typically a distribution-set variable rather than individual. For example, Debian sets the value as the most recent value of its changelog. So, it does not seem likely that users within an org would have an issue of cache misses.

The reason to include support for `SOURCE_DATE_EPOCH` is to be a good citizen with this standard. I don't see why a user would want to use `SOURCE_DATE_EPOCH` over our hardcoded date, but Reproducible Builds makes a big hoopla about it so it seems like a best practice to respect.

pex/common.py
Eric-Arellano: I thought about doing this and played with it just now, and am not convinced it should be split out. The logic is not generalizable to anything else, which imo is the reason to extract a function. 

We also don't explicitly test every detail of the implementation beyond the overall deterministic timestamp unit test, and I don't think we should because every single line comes directly from CPython's source, modulo the `date_time` line, comments, and parsing the data. So, extracting the function for better testing isn't a motivation imo.

I can split it out if you'd like but think it's best to keep the logic closer to where it's used.

pex/common.py
Eric-Arellano: I had to revert making it a constant because it caused the unit test to fail—the update to `os.environ` would not be picked up because once the constant is evaluated at import time, it never gets re-evaluated.

pex/common.py
stuhood: > https://reproducible-builds.org/docs/source-date-epoch/

The motivation in these docs does not actually explain why using a hardcoded date would be less reproducible.

> Could you please elaborate on this concern? For example, in Pants it doesn't seem like this would be an issue. Our Ubuntu shards and Centos6 shards both don't set SOURCE_DATE_EPOCH, so I don't see any immediate concern for Pants.

And OSX and Windows don't. So we'd immediately have an unnecessary mismatch between our two relevant platforms.

pex/common.py
stuhood: I think that the env var is a misfeature anyway.

pex/common.py
stuhood: Fine with me.

pex/common.py
```diff
@@ -15,9 +15,23 @@
...
+from datetime import datetime
 from uuid import uuid4
 
 
+# We hardcode the datetime to January 1, 1980, which is the start of MS-DOS time. Per section 
+# 4.4.6 of the zip spec at https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT, 
+# Zipfiles use MS-DOS time.
+# NB: we do not respect the standard env var $SOURCE_DATE_EPOCH often used by tools for
+# deterministic timestamps, as the intended use case is security by checking if the binary was
+# tampered with; we do not have a compelling use case for respecing this env var, and it only
```
stuhood: s/respecing/respecting/

Re: the reasoning: it's more to do with the existence of the variable triggering a _reduction_ in reproducibility when pex is used on platforms that don't set the variable, or if multiple distributions were to set the env var to different values.

pex/common.py
```diff
@@ -363,7 +379,45 @@ def __str__(self):
...
+          # This code mostly reimplements ZipInfo.from_file(), which is not available in Python
+          # 2.7. See https://github.com/python/cpython/blob/master/Lib/zipfile.py#L495.
+          st = os.stat(full_path)
+          isdir = stat.S_ISDIR(st.st_mode)
+          # Determine arcname, i.e. the archive name.
+          arcname = f
+          arcname = os.path.normpath(os.path.splitdrive(arcname)[1])
+          while arcname[0] in (os.sep, os.altsep):
+            arcname = arcname[1:]
+          if isdir:
```
jsirois: Since `self.files()` cintains no dirs, you could drop all dir related code here.

pex/common.py
```diff
@@ -363,7 +379,45 @@ def __str__(self):
...
+            arcname += '/'
+          # Setup ZipInfo with deterministic datetime.
+          zinfo = zipfile.ZipInfo(
+            filename=arcname,
+            # NB: the constructor expects a six tuple of year-month-day-hour-minute-second.
+            date_time=DETERMINISTIC_DATETIME.timetuple()[:6]
+          )
+          zinfo.external_attr = (st.st_mode & 0xFFFF) << 16
```
jsirois: This seems like it would all more naturally house on PermPreservingZipFile suitably renamed. It already holds our custom zip logic, which is now split here in in that class, and you might take small adavantage of the co-location to re-use this perm magic which is also handled there.

jsirois:
RE: the SOURCE_DATE_EPOCH - this is just a flag in spirit that allows external tools like Pants to control the timestamp. Presumably Pants would pass a single value no matter what platform it was run on. As such I don't really follow the objection from Pants point of view except for the fact that:
1. Its not a PEX_ env var so its sneaky / hard to discover.
2. It does not have a paired pex flag to allow passing the value arguably more explicitly.

All said though I'm happy enough to start with a fixed hardcoded value for simplicity sake. We could always add support later.

Eric-Arellano:
Thanks @jsirois for the thoughtful review. Great suggestions. 

I took them and added a function `zip_info_from_file()` to `PermPreservingZipFile`. Notably, I do not have the function always use a deterministic timestamp, but only parametrize it so that we can support both cases of system time and deterministic.

This allowed me to refactor `zip()` to always use `writestr()`, which imo leads to much more consistent and simpler code.

pex/common.py
```diff
@@ -363,7 +400,17 @@ def __str__(self):
...
+  def zip(self, filename, mode='w', deterministic_timestamp=False):
     with open_zip(filename, mode) as zf:
       for f in sorted(self.files()):
-        zf.write(os.path.join(self.chroot, f), arcname=f, compress_type=zipfile.ZIP_DEFLATED)
+        full_path = os.path.join(self.chroot, f)
+        zinfo = zf.zip_info_from_file(
+            filename=full_path,
+            arcname=f,
+            date_time=(
+              None if not deterministic_timestamp else DETERMINISTIC_DATETIME.timetuple()[:6]
```
jsirois: This is the last wonky bit I can see. The `zip_info_from_file` already knows [:6] magic, on the other hand - the extracted constant is not in final form. I think probably having `zip_info_from_file` take a `time.struct_time` and storing what is currently `DETERMINISTIC_DATETIME.timetuple()` as `DETERMINISTIC_DATETIME` would clean this up.

pex/common.py
```diff
@@ -15,9 +15,18 @@
...
+from datetime import datetime
 from uuid import uuid4
 
 
+# We use the start of MS-DOS time, which is what zipfiles use (see section 4.4.6 of
+# https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT). ZipInfo expects a 6-tuple of
+# year-month-day-hour-minute-second, which we use here.
+DETERMINISTIC_DATETIME = datetime(
+  year=1980, month=1, day=1, hour=0, minute=0, second=0, tzinfo=None
+).timetuple()[:6]
```
jsirois: I think you're missing an opportunity here to just know about [:6] down on line 113. In other words, always take a `time.struct_time` in `zip_info_from_file` for the `date_time` parameter. If `None`, use the constant defined here (but unsliced). Then slice the `date_time` just once!

Eric-Arellano: Ah okay _now_ I get what you were suggesting all along. Thanks! Good recommendation.

pex/common.py
```diff
@@ -20,11 +20,11 @@
 
 
 # We use the start of MS-DOS time, which is what zipfiles use (see section 4.4.6 of
-# https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT). ZipInfo expects a 6-tuple of
-# year-month-day-hour-minute-second, which we use here.
+# https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT). Note that ZipInfo expects a
+# time.struct_time, which we set here.
```
jsirois: But now this is a lie. `PermPreservingZipFile.zip_info_from_file` expects a `time.struct_time`. ZipInfo expects a 6-tuple of year-month-day-hour-minute-second.

PR #723, opened 2019-05-04
Fix PexInfo requirements using a non-deterministic data structure

Eric-Arellano:
## Problem
Even though `PexInfo` stores its requirements as an `OrderedSet`, when getting converted into a `list` in the `dump()` function, the requirements were losing their deterministic order for an unclear reason.

Specifically, the acceptance test diverged with `PEX-INFO`, with both files having different orders for their requirements. 

```
# pex1
"requirements": ["six==1.12.0", "cryptography==2.6.1", "cffi!=1.11.3,>=1.8", "asn1crypto>=0.21.0", "pycparser"]

# pex2
["six==1.12.0", "cryptography==2.6.1", "asn1crypto>=0.21.0", "cffi!=1.11.3,>=1.8", "pycparser"]
```

## Solution
Sort the requirements when being dumped to ensure we always have a consistent result.

We also make `interpreter_constraints`'s output always be sorted, as it is internally stored as a set so by definition is not deterministic in ordering.

### DIFF
--- a/pex/pex_info.py
+++ b/pex/pex_info.py
@@ -117,7 +117,7 @@ class PexInfo(object):
     raise ValueError('Malformed PEX requirement: %r' % (requirement_tuple,))
 
   def __init__(self, info=None):
-    """Construct a new PexInfo.  This should not be used directly."""
+    """Construct a new PexInfo. This should not be used directly."""
 
     if info is not None and not isinstance(info, dict):
       raise ValueError('PexInfo can only be seeded with a dict, got: '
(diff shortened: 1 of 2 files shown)


Eric-Arellano:
The meaningful commit to review is https://github.com/pantsbuild/pex/pull/723/commits/3834a426da10f57c238baedc7477659db1f0c941

I do not know why `requirements` is not deterministic when cast into a `list`, as it's an `OrderedSet`. I tried experimenting with this and could not figure it out. All I know is that this fixes the test and that the fix to `interpreter_constraints` does make logical sense.

PR #730, opened 2019-05-13
add delayed pkg_resources import fix from #713, with an integration test

cosmicexplorer:
This is a fix to the way we load pkg_resources, along with an integration test which fails on master since the setuptools vendoring in #624, but is fixed by the change in #713. This can be merged before that one to fix the failures induced in several 3rdparty packages, while #713 can be left to understanding and testing the separate but related problem that this change also happens to fix.

### DIFF
--- a/pex/environment.py
+++ b/pex/environment.py
@@ -368,6 +368,9 @@ class PEXEnvironment(Environment):
         with TRACER.timed('Adding sitedir', V=2):
           site.addsitedir(dist.location)
 
-        self._declare_namespace_packages(dist)
+    # Delay calling 'self._declare_namespace_packages' until 'sys.path' contains all of the
+    # resolved dists.
+    for dist in resolved:
+      self._declare_namespace_packages(dist)
 
     return working_set
--- a/pex/environment.py
+++ b/pex/environment.py
@@ -368,8 +368,9 @@ class PEXEnvironment(Environment):
         with TRACER.timed('Adding sitedir', V=2):
           site.addsitedir(dist.location)
 
-    # Delay calling 'self._declare_namespace_packages' until 'sys.path' contains all of the
-    # resolved dists.
+    # NB: we use this loop, rather than putting the below line in the prior loop, to ensure that
+    # 'sys.path' contains all of the resolved dists before calling
+    # 'self._declare_namespace_packages'.
     for dist in resolved:
       self._declare_namespace_packages(dist)
 
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -1315,6 +1315,33 @@ def test_no_emit_warnings_verbose_override():
   assert stderr
 
 
+def test_undeclared_setuptools_import_on_pex_path():
+  with temporary_dir() as td:
+    setuptools_pex = os.path.join(td, 'setuptools.pex')
+    run_pex_command(['setuptools==40.6.3', '-o', setuptools_pex]).assert_success()
+    bigquery_pex = os.path.join(td, 'bigquery.pex')
+    run_pex_command(['google-cloud-bigquery==1.10.0', '-o', bigquery_pex]).assert_success()
+
+    src_dir = os.path.join(td, 'src')
+    os.mkdir(src_dir)
+
+    src_file = os.path.join(src_dir, 'execute_import.py')
+    with open(src_file, 'w') as fp:
+      fp.write(dedent("""\
+        from google.cloud import bigquery
+
+        print('bigquery version: {}'.format(bigquery.__version__))
+        """))
+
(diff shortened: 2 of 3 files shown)


tests/test_integration.py
```diff
@@ -1315,6 +1315,33 @@ def test_no_emit_warnings_verbose_override():
   assert stderr
 
 
+def test_undeclared_setuptools_import_on_pex_path():
+  with temporary_dir() as td:
+    setuptools_pex = os.path.join(td, 'setuptools.pex')
+    run_pex_command(['setuptools==40.6.3', '-o', setuptools_pex]).assert_success()
```
Eric-Arellano: Is there a particular reason for this version? Does it need to be different than what Pex vendors here?

https://github.com/pantsbuild/pex/blob/3a7e4d701f8ace8bd070e53f03cc694e885e52c5/pex/vendor/__init__.py#L74-L80.

If so, I recommend adding a brief `NB` about this.

tests/test_integration.py
```diff
@@ -1315,6 +1315,33 @@ def test_no_emit_warnings_verbose_override():
...
+        print('bigquery version: {}'.format(bigquery.__version__))
+        """))
+
+    res = run_pex_command([
+      '--pex-path={}'.format(':'.join([setuptools_pex, bigquery_pex])),
+      '-D', src_dir,
+      '--entry-point', 'execute_import',
+    ])
+    res.assert_success()
+    assert res.output == 'bigquery version: 1.10.0\n'
```
Eric-Arellano:
```suggestion
    assert res.output.strip() == 'bigquery version: 1.10.0'
```

tests/test_integration.py
```diff
@@ -1315,6 +1315,33 @@ def test_no_emit_warnings_verbose_override():
   assert stderr
 
 
+def test_undeclared_setuptools_import_on_pex_path():
```
Eric-Arellano: Consider adding a note about what this test means. For example, link to the relevant issue.

tests/test_integration.py
cosmicexplorer: I've only been able to repro this issue with setuptools being provided in this exact way. The version of setuptools doesn't seem to matter, but I wanted to pin it so that we don't have setuptools eventually fixing the bug and silently break the test.

Eric-Arellano: That'd be helpful to document, e.g.

```
NB: the specific setuptools does not necessarily matter. We only pin the version to avoid a future version of setuptools potentially fixing this issue and then us no longer checking that Pex is behaving properly for older setuptools versions.
```

cosmicexplorer:
Given that this is fixing a live breakage and has a reliable test case, I'm thinking of adding the fix from #713 to this PR and merging it and leaving #713 to be figuring out how to simulate the separate problem described in https://github.com/pantsbuild/pex/pull/713#issuecomment-485571771.

tests/test_integration.py
cosmicexplorer: Done!

tests/test_integration.py
cosmicexplorer: Done!

tests/test_integration.py
cosmicexplorer: Done!

pex/environment.py
```diff
@@ -368,6 +368,9 @@ def _activate(self):
         with TRACER.timed('Adding sitedir', V=2):
           site.addsitedir(dist.location)
 
-        self._declare_namespace_packages(dist)
+    # Delay calling 'self._declare_namespace_packages' until 'sys.path' contains all of the
+    # resolved dists.
```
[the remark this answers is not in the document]
cosmicexplorer: Done!

pex/environment.py
```diff
@@ -368,6 +368,10 @@ def _activate(self):
         with TRACER.timed('Adding sitedir', V=2):
           site.addsitedir(dist.location)
 
-        self._declare_namespace_packages(dist)
+    # NB: we use this loop, rather than putting the below line in the prior loop, to ensure that
+    # 'sys.path' contains all of the resolved dists before calling
```
jsirois: This comment misses the real point - the issue has nothing to do with namespace packages per se for example as pointed out in #713. The issue is if any pex bootstrap code whatsoever imports `pkg_resources` or some sub-package prior to completeing sys.path modification then any pex user code or requirements that use the pkg_resources global working set will potentially break. This is whac-a-mole. I'll dig a bit this morning to see if there is a more robust solution.

jsirois: Not alot more robust, but I think a bit better containment and, hopefully, a bit clearer explanation of the case we're tiptoeing around here #732.

PR #732, opened 2019-05-16
Tighten up namespace declaration logic.

jsirois:
Re-arrange namespace declaration logic slightly and provide a more
detailed explanation of the case we're being careful about.

### DIFF
--- a/pex/environment.py
+++ b/pex/environment.py
@@ -307,31 +307,40 @@ class PEXEnvironment(Environment):
     return resolveds
 
   @staticmethod
-  def _declare_namespace_packages(dist):
-    if not dist.has_metadata('namespace_packages.txt'):
+  def _declare_namespace_packages(resolved_dists):
+    namespace_package_dists = [dist for dist in resolved_dists
+                               if dist.has_metadata('namespace_packages.txt')]
+    if not namespace_package_dists:
       return  # Nothing to do here.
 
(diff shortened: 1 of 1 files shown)


pex/environment.py
```diff
@@ -307,31 +307,40 @@ def _resolve(self, working_set, reqs):
...
+    #    above.
     # 2. Declare namespace packages at runtime via the canonical:
     #    `__import__('pkg_resources').declare_namespace(__name__)`
     #
-    # As such, we assume the best and try to import pkg_resources from the distribution they depend
-    # on, and then fall back to our vendored version only if not present. This is safe, since we'll
-    # only introduce our shaded version when no other standard version is present and even then tear
-    # it all down when we hand off from the bootstrap to user code.
+    # As such, we assume the best and try to import `pkg_resources` from the distribution they
+    # depend on (1), and then fall back to our vendored version only if not present. This is safe,
```
Eric-Arellano: The `(1)` took me a second to figure out what it meant. I recommend using `*` instead, although fine to keep as is.

jsirois: Heh, that's where I started and it looked weird to me so I went htis route (I really wanted dagger). I'll re-structure to use neither but treat both branches - we supply setuptools for bad actors and we use their setuptools - as equals / bullet points.

PR #734, opened 2019-06-13
WIP: ensure that sdist resolution is reproducible

Eric-Arellano:
Run `tox -e py37-integration -- -k test_reproducible_build_sdist` to reproduce the now failing test.

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -7,6 +7,7 @@ import errno
 import filecmp
 import functools
 import glob
+import itertools
 import json
 import multiprocessing
 import os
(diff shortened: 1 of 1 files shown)


tests/test_integration.py
```diff
@@ -1752,30 +1753,49 @@ def assert_reproducible_build(args):
         # from the random seed, such as data structures, as Tox sets this value by default. See
         # https://tox.readthedocs.io/en/latest/example/basic.html#special-handling-of-pythonhashseed.
         def create_pex(path, seed):
-            result = run_pex_command(args + ["-o", path], env=make_env(PYTHONHASHSEED=seed))
+            result = run_pex_command(
+                args + ["-o", path, "--disable-cache"], env=make_env(PYTHONHASHSEED=seed)
```
jsirois, line 1757: The only issue here is `--disable-cache` which causes a new tmp dir to be used for each PEX build. Since part of a PEX build is a `pip install` of all downloaded and built wheels, this leaves a direct_url.json file in the dist-info of each installed wheel chroot with a varing local file:// url. See: https://packaging.python.org/specifications/direct-url/#specification

The upshot is that the built PEX is stable when the PEX_ROOT is stable which will be the case that Pants cares mainly about. It still may make sense to delete that file though and its record in RECORD but I need to put some more thought into that.

PR #741, opened 2019-06-24
Fixup pex re-exec during bootstrap.

### LINKED ISSUE #709: Pex should not re-exec when the current interpreter satifies constraints
Currently Pex can re-exec against a new interpreter even when the current interpreter is on the PEX_PYTHON_PATH or PATH and matches any compatibility constraints. This should be avoided both for efficiency and to allow explicit control of the interpreter used to execute the pex.

### LINKED ISSUE #710: Pex should not lose PEX_PYTHON or PEX_PYTHON_PATH when re-exec-ing
Currently Pex loses PEX_PYTHON and PEX_PYTHON_PATH constraints on re-exec. The bug stems from #441 as commented there (https://github.com/pantsbuild/pex/pull/441/files#r277150236) in the following code:
https://github.com/pantsbuild/pex/blob/28869f6af835a39eae96453c0e5ab3dc01f7b00e/pex/pex_bootstrapper.py#L89-L136

In particular note lines 135 and 136. We re-exec vie `execve`, passing an explicit environment. That environment dict is obtained from `ENV.copy()` which does not see the `os.environ` mutation on the prior line:
https://github.com/pantsbuild/pex/blob/ffd56ce35cec240469567e2a51710d62baa21f08/pex/variables.py#L67-L76

jsirois:
Ensure `PEX_PYTHON` and `PEX_PYTHON_PATH` are actually scrubbed from the
environment of the pex process when re-exec'd fixing #710. Also
uniformize re-exec interpreter selection to ensure that the current
interpreter is preferred when it meets any constraints fixing #709.

Fixes #709
Fixes #710

### DIFF
--- a/pex/pex_bootstrapper.py
+++ b/pex/pex_bootstrapper.py
@@ -116,12 +116,12 @@ def maybe_reexec_pex(compatibility_constraints=None):
 
   current_interpreter = PythonInterpreter.get()
 
-  # NB: Used only for debugging and tests.
-  pex_exec_chain = []
+  # NB: Used only for tests.
   if '_PEX_EXEC_CHAIN' in os.environ:
-    pex_exec_chain.extend(os.environ['_PEX_EXEC_CHAIN'].split(os.pathsep))
-  pex_exec_chain.append(current_interpreter.binary)
-  os.environ['_PEX_EXEC_CHAIN'] = os.pathsep.join(pex_exec_chain)
+    flag_or_chain = os.environ.pop('_PEX_EXEC_CHAIN')
+    pex_exec_chain = [] if flag_or_chain == '1' else flag_or_chain.split(os.pathsep)
+    pex_exec_chain.append(current_interpreter.binary)
+    os.environ['_PEX_EXEC_CHAIN'] = os.pathsep.join(pex_exec_chain)
 
   current_interpreter_blessed_env_var = '_PEX_SHOULD_EXIT_BOOTSTRAP_REEXEC'
   if os.environ.pop(current_interpreter_blessed_env_var, None):
--- a/pex/pex_bootstrapper.py
+++ b/pex/pex_bootstrapper.py
@@ -20,12 +20,10 @@ __all__ = ('bootstrap_pex',)
 
 def _find_pex_python(pex_python):
   def try_create(try_path):
-    if os.path.exists(try_path):
-      try:
-        return PythonInterpreter.from_binary(try_path)
-      except Executor.ExecutionError:
-        pass
-    return None
+    try:
+      return PythonInterpreter.from_binary(try_path)
+    except Executor.ExecutionError:
+      return None
 
   interpreter = try_create(pex_python)
   if interpreter:
--- a/pex/pex_bootstrapper.py
+++ b/pex/pex_bootstrapper.py
@@ -18,38 +18,45 @@ from pex.variables import ENV
 __all__ = ('bootstrap_pex',)
 
 
-def find_in_path(target_interpreter):
-  if os.path.exists(target_interpreter):
-    return target_interpreter
-
-  for directory in os.getenv('PATH', '').split(os.pathsep):
-    try_path = os.path.join(directory, target_interpreter)
+def _find_pex_python(pex_python):
+  def try_create(try_path):
     if os.path.exists(try_path):
-      return try_path
-
-
-def find_compatible_interpreters(pex_python_path=None, compatibility_constraints=None):
-  """Find all compatible interpreters on the system within the supplied constraints and use
-     PEX_PYTHON_PATH if it is set. If not, fall back to interpreters on $PATH.
-  """
-  if pex_python_path:
-    interpreters = []
-    for binary in pex_python_path.split(os.pathsep):
       try:
-        interpreters.append(PythonInterpreter.from_binary(binary))
+        return PythonInterpreter.from_binary(try_path)
       except Executor.ExecutionError:
-        print("Python interpreter %s in PEX_PYTHON_PATH failed to load properly." % binary,
-          file=sys.stderr)
-    if not interpreters:
-      die('PEX_PYTHON_PATH was defined, but no valid interpreters could be identified. Exiting.')
+        pass
+
+  interpreter = try_create(pex_python)
+  if interpreter:
+    # If the target interpreter specified in PEX_PYTHON is an existing absolute path - use it.
+    yield interpreter
   else:
-    # We may have been invoked with a specific interpreter not on the $PATH, make sure our
-    # sys.executable is included as a candidate in this case.
-    interpreters = OrderedSet([PythonInterpreter.get()])
+    # Otherwise scan the PATH for matches:
+    for directory in os.getenv('PATH', '').split(os.pathsep):
+      try_path = os.path.join(directory, pex_python)
+      interpreter = try_create(try_path)
+      if interpreter:
+        yield interpreter
 
-    # Add all qualifying interpreters found in $PATH.
-    interpreters.update(PythonInterpreter.all())
 
+def find_compatible_interpreters(path=None, compatibility_constraints=None):
+  """Find all compatible interpreters on the system within the supplied constraints and use
+     path if it is set. If not, fall back to interpreters on $PATH.
+  """
+  interpreters = OrderedSet()
+  paths = None
+  if path:
+    paths = path.split(os.pathsep)
+  else:
+    # We may have been invoked with a specific interpreter, make sure our sys.executable is included
+    # as a candidate in this case.
+    interpreters.add(PythonInterpreter.get())
+  interpreters.update(PythonInterpreter.all(paths=paths))
+  return _filter_compatible_interpreters(interpreters,
+                                         compatibility_constraints=compatibility_constraints)
+
+
+def _filter_compatible_interpreters(interpreters, compatibility_constraints=None):
   return list(
     matched_interpreters(interpreters, compatibility_constraints)
     if compatibility_constraints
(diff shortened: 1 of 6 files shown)


jsirois:
NB: Integration tests are forthcoming, but the production code changes are reviewable.

jsirois:
Now fully reviewable.

pex/pex_bootstrapper.py
```diff
@@ -118,6 +118,13 @@ def maybe_reexec_pex(pex_info):
 
   current_interpreter = PythonInterpreter.get()
 
+  # NB: Used only for debugging and tests.
```
jsirois: I'm not happy with this, but its the best idea I could come up with.

pex/pex_bootstrapper.py
```diff
@@ -18,75 +18,86 @@
...
-  for directory in os.getenv('PATH', '').split(os.pathsep):
-    try_path = os.path.join(directory, target_interpreter)
+def _find_pex_python(pex_python):
+  def try_create(try_path):
     if os.path.exists(try_path):
-      return try_path
+      try:
+        return PythonInterpreter.from_binary(try_path)
+      except Executor.ExecutionError:
+        pass
```
Eric-Arellano: More Pythonic to return `None`, which ends up happening implicitly right now. By doing so, the return type of `try_create` becomes very clearly `Optional[PythonInterpreter]`.

pex/pex_bootstrapper.py
```diff
@@ -104,36 +115,51 @@ def maybe_reexec_pex(compatibility_constraints):
   :param compatibility_constraints: list of requirements-style strings that constrain the
```
Eric-Arellano: Outdated docstring

Eric-Arellano: Also why make this change? It doesn't appear that we use `pex_info` for anything other than `pex_info.interpreter_constraints`. By requiring `pex_info`, we make this function more complex and harder to test.

pex/pex_bootstrapper.py
```diff
@@ -104,36 +115,51 @@ def maybe_reexec_pex(compatibility_constraints):
...
-               'PEX_PYTHON_PATH="%s", COMPATIBILITY_CONSTRAINTS="%s"'
-               % (cmdline, sys.executable, ENV.PEX_PYTHON, ENV.PEX_PYTHON_PATH,
-                  compatibility_constraints))
-    ENV.delete('PEX_PYTHON')
-    ENV.delete('PEX_PYTHON_PATH')
-    os.environ['SHOULD_EXIT_BOOTSTRAP_REEXEC'] = '1'
-    os.execve(target, cmdline, ENV.copy())
+    elif ENV.PEX_PYTHON_PATH or compatibility_constraints:
+      TRACER.log('Using PEX_PYTHON_PATH={} constrained by {}'
+                 .format(ENV.PEX_PYTHON, compatibility_constraints), V=3)
```
Eric-Arellano: Should be `ENV.PEX_PYTHON_PATH`.

Eric-Arellano: Also this logging statement could be confusing if `PEX_PYTHON_PATH` is `None`. The message could be based against a check:

```python
if ENV.PEX_PYTHON_PATH is not None:
   log('Using PEX_PYTHON_PATH={}, constrained by {}')
else:
  log('Using your $PATH, constrained by {}')
```

pex/variables.py
```diff
@@ -75,9 +75,6 @@ def __init__(self, environ=None, rc=None, use_defaults=True):
   def copy(self):
     return self._environ.copy()
 
-  def delete(self, variable):
```
Eric-Arellano: 👍

pex/pex_bootstrapper.py
jsirois: I still think it shouldn't exist so I'd like to keep only. We do already log all this! See:
+ re-exec:
  + https://github.com/pantsbuild/pex/pull/741/files#diff-670bc388096d1ef0c7994c4d37976392R163
+ actual used:
  + https://github.com/pantsbuild/pex/pull/741/files#diff-670bc388096d1ef0c7994c4d37976392R149
  + https://github.com/pantsbuild/pex/pull/741/files#diff-670bc388096d1ef0c7994c4d37976392R157

And I refuse to scrape logs in tests - an even worse solution imo!

tests/test_integration.py
```diff
@@ -1462,3 +1466,83 @@ def test_reproducible_build_python_flag():
...
+    test_pex = os.path.join(td, 'test.pex')
+
+    args = ['-o', test_pex]
+    if interpreter_constraints:
+      args.extend('--interpreter-constraint={}'.format(ic) for ic in interpreter_constraints)
+
+    env = os.environ.copy()
+    PATH = env.get('PATH').split(os.pathsep)
+
+    def add_to_path(path):
```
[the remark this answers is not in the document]
jsirois: Fixed

tests/test_integration.py
```diff
@@ -1462,3 +1466,83 @@ def test_reproducible_build_python_flag():
...
+
+def test_pex_reexec_constraints_dont_match_current_pex_python_path():
+  py36_interpreter = ensure_python_interpreter(PY36)
+  py27_interpreter = ensure_python_interpreter(PY27)
+  _assert_exec_chain(exec_chain=[py36_interpreter],
+                     pex_python_path=[py27_interpreter, py36_interpreter],
+                     interpreter_constraints=['=={}'.format(PY36)])
+
+
+def test_pex_reexec_constraints_dont_match_current_pex_python_path_min():
```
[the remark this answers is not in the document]
jsirois: Fixed.

pex/pex_bootstrapper.py
jsirois: Fixed.

pex/pex_bootstrapper.py
jsirois: Reverted and docstring updated. The pexrc verbage was always irrelevant and overly constraining.

pex/pex_bootstrapper.py
jsirois: Done.

pex/pex_bootstrapper.py
```diff
@@ -18,122 +18,161 @@
...
-    try_path = os.path.join(directory, target_interpreter)
+def _find_pex_python(pex_python):
+  def try_create(try_path):
     if os.path.exists(try_path):
-      return try_path
+      try:
+        return PythonInterpreter.from_binary(try_path)
+      except Executor.ExecutionError:
+        pass
+    return None
```
Eric-Arellano: Ah did no realize this `pass` was originally nested in the `if`. My bad. This is getting a little nit picky, but I think this would be the most explicit:

```python
if not os.path.exists(try_path):
  return None
try:
  returnPythonInterpreter.from_binary(try_path)
except Executor.ExecutionError:
  return None
```

jsirois: Erm, OK. This is truly in the eye of the beholder. I killed the path existence check though and that swings this back around to reasonable to me. Done-ish.

pex/pex_bootstrapper.py
```diff
@@ -116,12 +116,12 @@ def maybe_reexec_pex(compatibility_constraints=None):
 
   current_interpreter = PythonInterpreter.get()
 
-  # NB: Used only for debugging and tests.
-  pex_exec_chain = []
+  # NB: Used only for tests.
   if '_PEX_EXEC_CHAIN' in os.environ:
```
jsirois: I feel slightly less dirty with this tweak. Now user code will never see `_PEX_EXEC_CHAIN` in the environment in production.

pex/pex_bootstrapper.py
```diff
@@ -18,122 +18,159 @@
...
+  if os.environ.pop(current_interpreter_blessed_env_var, None):
     # We've already been here and selected an interpreter. Continue to execution.
     return
 
-  target = None
-  with TRACER.timed('Selecting runtime interpreter based on pexrc', V=3):
+  with TRACER.timed('Selecting runtime interpreter', V=3):
     if ENV.PEX_PYTHON and not ENV.PEX_PYTHON_PATH:
       # preserve PEX_PYTHON re-exec for backwards compatibility
       # TODO: Kill this off completely in favor of PEX_PYTHON_PATH
```
CMLivingston: This something that should probably happen soon. Not sure if we have a deprecation policy around this but it is redundant with PPP and  more complexity to support than necessary.

pex/pex_bootstrapper.py
CMLivingston: Was this used somehow before? Confused about the "now" verbage.

pex/pex_bootstrapper.py
```diff
@@ -18,122 +18,159 @@
...
+    # TODO: Allow the selection strategy to be parameterized:
+    #   https://github.com/pantsbuild/pex/issues/430
+    return min(candidate_interpreters)
 
-def maybe_reexec_pex(compatibility_constraints):
-  """
-  Handle environment overrides for the Python interpreter to use when executing this pex.
+
+def maybe_reexec_pex(compatibility_constraints=None):
+  """Handle environment overrides for the Python interpreter to use when executing this pex.
```
CMLivingston: Is this a style change or does it change any presentations from help/pydocs, etc? Just curious.

jsirois: Just a style change - I wanted to make it clear None was an acceptable input. The general implication of `def maybe_reexec_pex(compatibility_constraints)` is that `compatibility_constraints` are required.

pex/pex_bootstrapper.py
jsirois: I think the bit needed to kick off deprecation / removal is expanding "it's redundant" in an issue. FWICT it is not simply redundant. Today I can `PEX_PYTHON=python2 ...` to ensure a pex runs under some python2 interpreter on any machine. I can't do that with PEX_PYTHON_PATH since its entries are full paths or path prefixes, both of which are machine specific.

Issue #293, opened 2016-07-21
OSError: : [Errno 2] No such file or directory


objmagic:
Hi, 

Pex is used to generate Python executable files in [twitter/heron](https://github.com/twitter/heron). However, we sometimes can see the following error message that makes build fail:

```
    sys.exit(main())
  File "/home/travis/.cache/bazel/_bazel_travis/030a6cc9b27b34f2fe0e137b942f59bf/heron/bazel-out/local_linux-fastbuild/bin/third_party/pex/_pex.runfiles/__main__/third_party/pex/_pex.py", line 159, in main
    resolver_options_builder, interpreter=interpreter)
  File "/home/travis/.cache/bazel/_bazel_travis/030a6cc9b27b34f2fe0e137b942f59bf/heron/bazel-out/local_linux-fastbuild/bin/third_party/pex/_pex.runfiles/__main__/third_party/pex/pex/bin/pex.py", line 472, in build_pex
    resolveds = resolver.resolve(resolvables)
  File "/home/travis/.cache/bazel/_bazel_travis/030a6cc9b27b34f2fe0e137b942f59bf/heron/bazel-out/local_linux-fastbuild/bin/third_party/pex/_pex.runfiles/__main__/third_party/pex/pex/resolver.py", line 191, in resolve
    dist = self.build(package, resolvable.options)
  File "/home/travis/.cache/bazel/_bazel_travis/030a6cc9b27b34f2fe0e137b942f59bf/heron/bazel-out/local_linux-fastbuild/bin/third_party/pex/_pex.runfiles/__main__/third_party/pex/pex/resolver.py", line 254, in build
    os.rename(target + '~', target)
OSError: [Errno 2] No such file or directory
```

This error happens intermittently. I thought #271 fixed the problem and upgrade the Pex, but seems that the problem is still there.

Here is a sample CI log: https://travis-ci.org/twitter/heron/builds/146219942

/cc: @kwlzn

kwlzn:
which version of pex? can you provide a repro?

afaict, this wouldn't be related to #271 but more likely related to bazel's use of pex.

objmagic:
[...]

Issue #155, opened 2015-08-25
PEX_ROOT doesn't seem to be honoured for all directories.


KatieLucas-Grapeshot:
After setting PEX_ROOT, pex still attempts to write to ~/.pex -- notably trying to create ~/.pex/interpreters.

This is a problem because on our build system the build process's users $HOME is non-writeable.

I suspect that the default values for the flags in pex.py lines 162 and 269 are at fault here; expanding to the $HOME regardless of the PEX_ROOT setting.

(For the time being have worked round this by setting flags explicitly.)

Thanks,
  Katie

jsirois:
Sorry for the huge delay, but this issue has been revived through a reference.

How do you set `PEX_ROOT`? NB: A very confusing aspect of pex is its build time vs runtime behavior.

At build time, if you set PEX_ROOT as an env var or pass the --pex-root flag to the cli, you only set the PEX_ROOT for the build process and not for the runtime of the resulting pex when executed. The `pex --help` hints at this:
```
$ pex --help
...
  --pex-root=PEX_ROOT   Specify the pex root used in this invocation of pex.
                        [Default: ~/.pex]
...
```

So, to set PEX_ROOT at _runtime_ you need to do one of:
1. Ensure the desired `PEX_ROOT` environment variable is exported.
2. Set `PEX_ROOT` inline if using a shell like bash; ie `PEX_ROOT=x ./my.pex ...`
3. Deploy appropriate pexrc files on target machines. These can be in 3 locations:
    1. sibling to the pex file being run in `.pexrc`
    2. in `~/.pexrc`
    3. in `/etc/pexrc`
    Line in pexrc are env-var lines like `PEX_ROOT=y` and they are read and applied with the precedence above, a sibling `.pexrc` having the highest precedence.

We currently offer no option to embed the PEX_ROOT setting in the pex at build time since pexes are generally non-machine specific - they can run anywhere; so the PEX_ROOT embedded in the pex if we supported this might not exist or be appropriate on some target hosts.

jsirois:
I'm going to close this as answered, but if the above explanation is unclear or unsatisfactory for your usecase, please re-open.

PR #747, opened 2019-07-25
Have bootstrap_pex_env pull in contents of pexrc files

rtkjliviero:
Added optional arg to bootstrap_pex_env which will cause pexrc configuration files from default locations to be pulled in.

This PR relates to the following issues:
1. https://github.com/pantsbuild/pex/issues/155 (I'm having the same problem symptoms with the same basic cause)
1. https://github.com/pantsbuild/pex/issues/293 (sort of related)

In my current use-case, I am activating one of several .pex files programmatically, and importing certain classes from those files.  This works well, but since I'm using the Python API I can't reasonably pass environment variables to the command.  `os.environ` seems to be ignored by `bootstrap_pex_env()`; having that dict be optionally respected would be another possible solution.

As an aside...while I think this is a good idea, right now I'm open to building my .pex files _with_ the env vars I require, if that is currently supported by the tooling. I have tried `pex --pex-root '/tmp/.pex/' -r <(pip freeze) -o some_output.pex`, but that still results in `~` being used as PEX_ROOT.


rtkjliviero:
Also allowing the passing in of a dict or JSON string, to override vars for a single invocation.

jsirois:
> As an aside...while I think this is a good idea, right now I'm open to building my .pex files with the env vars I require, if that is currently supported by the tooling. I have tried pex --pex-root '/tmp/.pex/' -r <(pip freeze) -o some_output.pex, but that still results in ~ being used as PEX_ROOT.

Pex does not in fact support embedding things like machine-specific paths in its metadata (PEX_INFO) at build time. You can only set PEX_ROOT at runtime via env var or `.pexrc`, '~/.pexrc` or `/etc/pexrc` containing a `PEX_ROOT=x` line.

pex/pex_bootstrapper.py
```diff
@@ -199,9 +213,13 @@ def is_compressed(entry_point):
...
+def bootstrap_pex_env(entry_point, include_pexrc=False, extra_vars=None):
+  """
+  Bootstrap the current runtime environment using a given pex. Optionally, pexrc files from
+  'standard' locations can be included - see pex.variables.DEFAULT_PEXRC_LOCATIONS - or
+  runtime-specific overrides (dict or json)
+  """
+  pex_info = _bootstrap(entry_point, include_pexrc=include_pexrc, extra_vars=extra_vars)
 
-  from .environment import PEXEnvironment
+  from pex.environment import PEXEnvironment
```
jsirois: Please do this for all lazy imports in this file or None (ie: revert). I'd prefer the import style remain consistent in the file; otherwise a reasonable question as a reader is "Why is this different? What import mechanic subtlety am I missing?!".

pex/pex_bootstrapper.py
```diff
@@ -199,9 +213,13 @@ def is_compressed(entry_point):
   return os.path.exists(entry_point) and not os.path.exists(os.path.join(entry_point, PexInfo.PATH))
 
 
-def bootstrap_pex_env(entry_point):
-  """Bootstrap the current runtime environment using a given pex."""
-  pex_info = _bootstrap(entry_point)
+def bootstrap_pex_env(entry_point, include_pexrc=False, extra_vars=None):
+  """
```
jsirois: Please follow pep 8 here:
```python
"""Bootstrap the current runtime environment using a given pex.

Optionally, pexrc files from...
"""
```

pex/pex_bootstrapper.py
```diff
@@ -173,9 +174,22 @@ def maybe_reexec_pex(compatibility_constraints=None):
...
   pex_info = PexInfo.from_pex(entry_point)
+
+  # Pull in the default ENV, which is an instance of pex.variables.Variables, which reads
+  # from the default pexrc locations.
+  if include_pexrc:
+    pex_info.update(PexInfo.from_env())
+
+  # Override env settings using extra passed-in vars
+  if extra_vars:
+    if isinstance(extra_vars, dict):
```
jsirois: How about just accepting `**extra_vars` and pushing json back to the user that needs it. It's terrifying that we have a consumer of this API at all, and as a relunctant / conservative library maintainer I'd much rather keep API surface area minimal.

rtkjliviero: I'm not totally sure what you mean here, would you mind clarifying?

Are you saying this should only allow properly formatted json here, and require the user to convert from a dict if necessary?

pex/pex_bootstrapper.py
rtkjliviero: Nice catch, that was left in accidentally.  Completely agree re: consistency.

pex/variables.py
```diff
@@ -14,6 +14,11 @@
 
 __all__ = ('ENV', 'Variables')
 
+DEFAULT_PEXRC_LOCATIONS = [
```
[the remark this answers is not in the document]
rtkjliviero: Sure.  My thinking was that having the default locations front-and-centre like this makes the file a little more self-documenting - but I'll defer to your preference.

pex/pex_bootstrapper.py
rtkjliviero: 👍

pex/pex_bootstrapper.py
rtkjliviero: @jsirois, if you could clarify I'm more than happy to update this PR!

rtkjliviero:
Added some unit tests, as per
> I'd like to block on unit test being added that help enforce the idea this is depended upon functionality.

I still need to add unit tests for considering pexrc files, but that requires some input from the maintainer.  I'd like to put back my previous changes (DEFAULT_PEXRC_LOCATIONS, [7451a62](https://github.com/pantsbuild/pex/pull/747/commits/7451a62fc3dc420c90a416841d7d591e7d8f2855#diff-e7c82c87884d034d408b533c4e083f27R17)), so that we can just `mock.patch` that object and avoid polluting the test by pulling in any pexrc files that might already be on the file system.  If that is acceptable to @jsirois, I'll go ahead!

rtkjliviero:
@jsirois, any input on the above?

pex/pex_bootstrapper.py
jsirois: I just mean support the minimal api, which is allowing the user to pass environment variables via **extra_vars. If code that call this wants to support a json encoding of the env vars that code can decode the json to a dict to pass in as **extra_vars.

Concreteley, just:
```python
def _bootstrap(entry_point, include_pexrc=False, **extra_vars):
  from .pex_info import PexInfo
  pex_info = PexInfo.from_pex(entry_point)
  env = dict(PEX_IGNORE_RCFILES=str(not include_pexrc), **extra_vars)
  pex_info.update(PexInfo.from_env(env))
```

Calling code in an application can do this if it want to unburdening Pex from supporting json in this API in perpetuity:
```python
_bootstrap(entry_point, **json.loads(extra_vars_json_string))
```

And, of course, where this is really important is in the public API function, so the **extra_vars style should propagate up to `bootstrap_pex_env`.

jsirois:
@rtkjliviero thanks for working through things here. Your work is not in vain and got me thinking hard about how to solve this problem more generally in a way that does not require coding against internal PEX APIs.

I have a branch that should obviate the need to use the PEX bootstrap API like this. It allows you to place a PEX on the sys.path, say via `PYTHONPATH`, and activate it with a single `import __pex__` early in your code or else directly using `__pex__` as a psuedo package like so: `from __pex__ import thing.inside.the.pex`. I think this should remove the need the need for things like Lambdex and other tools that work around PEX files not working like traditional Python zip archive sys.path entries. I'm going to close this PR since Pex has tried really hard to lock down its API to just the CLI and this was the last compelling reason this couldn't be done. I'll add you to that PR review.

Issue #749, opened 2019-07-29
pkg_resources imported too early


jhamet93:
I ran across two cases recently which resulted in pkg_resources being imported too early (resulting in the global working set being set before sys.path is done being finalized). This issue is also relevant to #713, #730, #732 .

1.  In setuptools, If a package sets `namespace_packages` to be an empty list in the setup.py, `has_metadata` will return True in the executed code below in environment.py.
```
def _declare_namespace_packages(resolved_dists):
    namespace_package_dists = [dist for dist in resolved_dists if dist.has_metadata('namespace_packages.txt')]
    if not namespace_package_dists:
      return  # Nothing to do here.
    pkg_resources, vendored = _import_pkg_resources()
    ...
```
However, since the list is empty, it is not actually a namespace package. This results in pkg_resources to be imported too early. I discovered this issue when using the package, `autopep8` in which the dependency, `pycodestyle`, was causing this.(https://github.com/PyCQA/pycodestyle/blob/2.5.0/setup.py). Therefore, to be defensive, should the code also ensure there's a value within the metadata ? 

2.  It seems the current logic is only cognizant of making sure not to import pkg_resources too early according to the executed pex.  Therefore, when setting PEX_PATH to merge in other pexes into the current environment, pkg_resources can be imported too early if there is a namespace package before the last merged pex. Should the code be modified to declare all the namespace packages only after all the pexes have been merged ? 

Both issues can be reproduced identically.
```
pex setuptools autopep8 -o environment.pex
pex six -o missing_dependency.pex
PEX_PATH=$(pwd)/missing_dependency.pex: ./environment.pex
```
```
>>> import pkg_resources
>>> pkg_resources.get_distribution('six')
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/Users/jhamet/.pex/install/setuptools-41.0.1-py2.py3-none-any.whl.ffa66c57fd90eee9c21efeb1d3d918a7b960fa4f/setuptools-41.0.1-py2.py3-none-any.whl/pkg_resources/__init__.py", line 481, in get_distribution
    dist = get_provider(dist)
  File "/Users/jhamet/.pex/install/setuptools-41.0.1-py2.py3-none-any.whl.ffa66c57fd90eee9c21efeb1d3d918a7b960fa4f/setuptools-41.0.1-py2.py3-none-any.whl/pkg_resources/__init__.py", line 357, in get_provider
[...]

PR #750, opened 2019-07-31
Fix pkg resource early import

jhamet93:
As referenced in #749 , I ran into a situation where `pkg_resources` was imported too early before `sys.path` was finalized, occurring in the situation where a pex in my `PEX_PATH` needed to invoke: `pkg_resources.get_distribution`.  This was failing because a namespace package caused `pkg_resources` to be imported before the other pexes were merged into the environment 

Therefore, wait until all deps from all pexes (i.e. pexes in `PEX_PATH`) have been resolved before attempting to declare the namespace packages. This ensures `sys.path` is finalized before `pkg_resources` is possibly imported.

### DIFF
--- a/pex/environment.py
+++ b/pex/environment.py
@@ -10,6 +10,7 @@ import site
 import sys
 import uuid
 import zipfile
+from collections import OrderedDict
 
 from pex import pex_builder, pex_warnings
 from pex.bootstrap import Bootstrap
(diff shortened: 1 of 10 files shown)


jsirois:
The CI error is reproducible locally. I'm taking a look...

jsirois:
@jhamet93 this one was not trivial to explain. I took the liberty of pushing an addtional commit in ef0a13eb11a1b43cd1bf7d7b19b52f89cb2cac08 with the fix for CI + cleanups.

jsirois:
@jhamet93 could you please review the changes I made on top of yours? They are these: https://github.com/pantsbuild/pex/pull/750/files/62af44be964f6f916810a3ac4622241da1b34ff0..0489225aaa65cd41250b708e9210dd47a7e20695

Any comments would be great and if all looks good a LGTM would be good.

Thanks for your contribution!

jhamet93:
LGTM.

Thanks for the help finishing this one!

PR #752, opened 2019-08-07
Fix `sys.path` scrubbing of pex extras modules.

### LINKED ISSUE #745: /usr/lib64/python2.7 from system python gets into sys.path at runtime
(I filed this before but realised I didn't have all the info)

This is with rhel7. I've also tried on a clean vm.

Building a pex with poetry, which requires subprocess32:
```
$ pex -c poetry poetry==0.12.17 --interpreter-constraint ">=2.7,<2.8" -o poetry
$ ./poetry show
                                          
[AttributeError]           
'module' object has no attribute 'run'

$ PEX_INTERPRETER=1 ./poetry
Python 2.7.5 (default, Sep 12 2018, 05:31:16) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-36)] on linux2
(InteractiveConsole)
>>> import subprocess32
>>> subprocess32.__file__
'/usr/lib64/python2.7/site-packages/subprocess32.pyc'
>>> import sys
>>> sys.path
['/home/daniel.j.murray/git/mas/poetry', '/usr/lib64/python27.zip', '/usr/lib64/python2.7', '/usr/lib64/python2.7/plat-linux2' ... <snip>
```
(The error is because I have a super old subprocess32 installed as part of the OS and it doesn't seem to have a run() method)

This only happens when running the pex, if I use pex (which is in a virtual environment) to start an interpreter the paths work ok.

If anyone could assist it would be greatly appreciated :)

jsirois:
Previously we would fail to scrub any extras modules imported by the pex
bootstrapping code. Add failing unit and integration tests and fix this
with a missing scrub step to check modules (previously only packages
were scrubbed).

Fixes #745

### DIFF
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -174,7 +174,13 @@ class PEX(object):  # noqa: T000
     new_modules = {}
 
     for module_name, module in modules.items():
-      # builtins can stay
+      # Tainted modules should be dropped.
+      module_file = getattr(module, '__file__', os.devnull)
+      if cls._tainted_path(module_file, site_libs):
+        TRACER.log('Dropping %s' % (module_name,), V=3)
+        continue
+
+      # Untainted non-packages (builtin modules) need no further special handling and can stay.
       if not hasattr(module, '__path__'):
         new_modules[module_name] = module
         continue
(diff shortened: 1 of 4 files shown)


jsirois:
@danmur I'm not sure if you have time to review, but the integration test added should make some sense and replicate your root poetry issue. See here if so: https://github.com/pantsbuild/pex/pull/752/files#diff-7eb47fc18d4a6196f0940784484a5bf3R1609

jsirois:
Thanks @TansyArron. I'm going to merge and circle back if there is further feedback. This is a small fix with the complexity all in the tests so I'm going to proceed with the 1.6.9 release to get this and #750 out there.

Issue #690, opened 2019-03-26
Release 1.6.5


jsirois:
On the docket:
+ [x] pex does not resolve abi3 wheels correctly #539

PR #754, opened 2019-08-09
imp is deperecated, using importlib.machinery instead.

asherf:
https://docs.python.org/3/library/imp.html#imp.get_suffixes

```
In [10]: import imp
src/python/toolchain/service/buildstats/api/manage.py:1: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
  #! /usr/bin/env ./python3

In [11]: import importlib

In [12]: imp.get_suffixes()
Out[12]:
[('.cpython-37m-darwin.so', 'rb', 3),
 ('.abi3.so', 'rb', 3),
 ('.so', 'rb', 3),
 ('.py', 'r', 1),
 ('.pyc', 'rb', 2)]

In [13]: importlib.machinery.all_suffixes()
Out[13]: ['.py', '.pyc', '.cpython-37m-darwin.so', '.abi3.so', '.so']

```

Screenshot 2019-08-09 14 35 10

### DIFF
--- a/pex/pep425tags.py
+++ b/pex/pep425tags.py
@@ -271,10 +271,16 @@ def get_supported(versions=None, noarch=False, platform=None,
         abis[0:0] = [abi]
 
     abi3s = set()
-    import imp
-    for suffix in imp.get_suffixes():
-        if suffix[0].startswith('.abi'):
-            abi3s.add(suffix[0].split('.', 2)[1])
+    if sys.version_info[0] == 2:
+        import imp
+        suffixes = [sfx[0] for sfx in imp.get_suffixes()]
+    else:
+        import importlib
+        suffixes = importlib.machinery.all_suffixes()
+
+    for suffix in suffixes:
+        if suffix.startswith('.abi'):
+            abi3s.add(suffix.split('.', 2)[1])
 
     abis.extend(sorted(list(abi3s)))
 
--- a/tests/test_platform.py
+++ b/tests/test_platform.py
@@ -102,6 +102,7 @@ def test_pypy_abi_prefix():
 # NB: Having to patch here is a symptom of https://github.com/pantsbuild/pex/issues/694
 # Kill when the Platform API is fixed to not need to consult the local interpreter.
 @patch('imp.get_suffixes', lambda: [('.abi3.so', 'rb', 3)])
+@patch(' importlib.machinery.all_suffixes', lambda: ['.abi3.so'])
 def test_platform_supported_tags_abi3():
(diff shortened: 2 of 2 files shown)


pex/pep425tags.py
```diff
@@ -271,10 +271,10 @@ def get_supported(versions=None, noarch=False, platform=None,
         abis[0:0] = [abi]
 
     abi3s = set()
-    import imp
-    for suffix in imp.get_suffixes():
-        if suffix[0].startswith('.abi'):
-            abi3s.add(suffix[0].split('.', 2)[1])
+    import importlib
```
asherf: @jsirois any reason this is a local import ?
can we move it to the top of the file?

pex/pep425tags.py
```diff
@@ -271,10 +271,16 @@ def get_supported(versions=None, noarch=False, platform=None,
         abis[0:0] = [abi]
 
     abi3s = set()
-    import imp
-    for suffix in imp.get_suffixes():
-        if suffix[0].startswith('.abi'):
-            abi3s.add(suffix[0].split('.', 2)[1])
+    if sys.version_info[0] == 2:
```
jsirois: Please see the header on this file - it's complicated:
https://github.com/pantsbuild/pex/blob/5dfc8c2bb717ead39a897c82320d9d9e4c8bdb84/pex/pep425tags.py#L1-L15

At the very least, the modified code here should be marked.

Better might be to update our vendored version of setuptools and implement #690.
That, though won't work since master of setuptools still uses `imp` unconditionally: https://github.com/pypa/setuptools/blob/5b128e8c82f4886b4ff110219de80a61bf8aa3b0/setuptools/pep425tags.py#L257-L260.
That said, pip, which it copies from, does do a conditional import now: https://github.com/pypa/pip/blob/6f0e873182e45d2f7a4b61b9800599723d2636f1/src/pip/_internal/utils/compat.py#L205-L214
Unfortunately though, pip's pep425tags.py is no longer self-contained.

tests/test_platform.py
```diff
@@ -102,6 +102,7 @@ def test_pypy_abi_prefix():
 # NB: Having to patch here is a symptom of https://github.com/pantsbuild/pex/issues/694
 # Kill when the Platform API is fixed to not need to consult the local interpreter.
 @patch('imp.get_suffixes', lambda: [('.abi3.so', 'rb', 3)])
+@patch(' importlib.machinery.all_suffixes', lambda: ['.abi3.so'])
```
jsirois: Please strip the leading embedded space in the patch here and below.

PR #762, opened 2019-08-23
Avoid reordering of equivalent packages from multiple fetchers

clintharrison:
`Iterator.iter()` iterates over the set of links returned from `Crawler.crawl()`. This means the hash seed randomization potentially changes the order, when there are two packages equivalent in both version, package type (wheel, etc), platform tags, and they're both remote.

Ideally, these would not be any different, so it's not really a correctness issue, but I am using pex's resolver internals to store the actual URLs used, and they flap over time between different-but-equivalent ones.

(And in case it's relevant: to be honest I tried running the integration tests locally, but they were failing before this change, and it didn't look related. Are they known to fail on macOS (10.14)?)

### DIFF
--- a/pex/sorter.py
+++ b/pex/sorter.py
@@ -37,7 +37,9 @@ class Sorter(object):
         package.version,  # highest version
         cls.package_type_precedence(package, precedence=precedence),  # type preference
         cls.package_platform_tag_precedence(package),  # platform preference
-        package.local)  # prefer not fetching over the wire
+        package.local,  # prefer not fetching over the wire
+        package.url,  # otherwise equivalent files should deterministically sort
+    )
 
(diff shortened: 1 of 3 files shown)


jsirois:
Thanks @clintharrison!

jsirois:
@clintharrison re:
> (And in case it's relevant: to be honest I tried running the integration tests locally, but they were failing before this change, and it didn't look related. Are they known to fail on macOS (10.14)?)

CI currently runs against macos-10.13-xcode9.4.1 so we don't have the 10.14 case covered in an automated way. I don't use OSX so I can't comment either. If you want to post a seperate issue with failures or else do that over in the pantsbuild slack channel #pex, we could dig more. This _should_ work on OSX.

PR #769, opened 2019-09-30
PythonInterpreter: support python binary names with single letter suffixes

murph:
On gentoo linux, my python binaries have the 'm' suffix, and then are symlinked from the non-suffixed versions:

usr/bin/python3.5 -> python3.5m
/usr/bin/python3.5m
/usr/bin/python3.6 -> python3.6m
/usr/bin/python3.6m
etc.

PEP-3149 indicates this is okay, so pex should probably support it.

I updated a test, but already have a bunch of tox failures on master so it's hard to know if it passes:
"15 failed, 61 passed, 3 skipped, 1 xfailed in 198.12 seconds"

I'll dig into that in another branch hopefully =)

### DIFF
--- a/pex/interpreter.py
+++ b/pex/interpreter.py
@@ -267,6 +267,11 @@ class PythonInterpreter(object):
 
     re.compile(r'python[23]$'),
     re.compile(r'python[23].[0-9]$'),
+
+    # Some distributions include a suffix on the in the interpreter name, similar to PEP-3149
+    # E.g. Gentoo has /usr/bin/python3.6m to indicate it was built with pymalloc
+    re.compile(r'python[23].[0-9][a-z]$'),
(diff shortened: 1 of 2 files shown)


pex/interpreter.py
```diff
@@ -267,6 +267,10 @@ class PythonInterpreter(object):
 
     re.compile(r'python[23]$'),
     re.compile(r'python[23].[0-9]$'),
+
+    # PEP-3149: Python implementations MAY include additional flags in the file name tag
+    re.compile(r'python[23].[0-9][a-z]$'),
```
jsirois: You've found cases of this in the wild and that may be more important to mention since, IIUC, that pep refers to so names and not python interpreter names.

Issue #768, opened 2019-09-16
Support of intel packages


jdlesage:
I would like to package in a pex, the intel packages of the classic python numerical packages (numpy, scipy, etc.)
Pex seems not able to find the package that provides numpy because the name differs.
Here a simple reproducer:

_System description_
pex version: 1.6.11
python version: 3.6.3
OS: centos7

**Using classic numpy, it works**

`pex numpy -o numpy.pex`
`./numpy.pex -c "import numpy; print(numpy.__file__)"`

/home/jd.lesage/.pex/install/numpy-1.17.2-cp36-cp36m-manylinux1_x86_64.whl.0833169d405facf4fccb13d37dcbf723d0827c1e/numpy-1.17.2-cp36-cp36m-manylinux1_x86_64.whl/numpy/__init__.py

**Using intel numpy, import fails**
`pex intel_numpy -o intel_numpy.pex`
`./intel_numpy.pex -c "import numpy; print(numpy.__file__)"`

Traceback (most recent call last):
  File ".bootstrap/pex/pex.py", line 397, in execute
  File ".bootstrap/pex/pex.py", line 329, in _wrap_coverage
  File ".bootstrap/pex/pex.py", line 360, in _wrap_profiling
  File ".bootstrap/pex/pex.py", line 445, in _execute
  File ".bootstrap/pex/pex.py", line 484, in execute_interpreter
  File ".bootstrap/pex/pex.py", line 535, in execute_content
  File ".bootstrap/pex/compatibility.py", line 81, in exec_function
  File "-c <cmd>", line 1, in <module>
ModuleNotFoundError: No module named 'numpy'

jsirois:
Alrighty - intel-numpy is an exceedingly odd package:
```
$ unzip -l intel_numpy-1.21.5-cp39-cp39-manylinux2014_x86_64.whl | grep numpy/__init__.py
    16819  2023-11-30 14:03   intel_numpy-1.21.5.data/data/lib/python3.9/site-packages/numpy/__init__.py
   133925  2023-11-30 14:03   intel_numpy-1.21.5.data/data/lib/python3.9/site-packages/numpy/__init__.pyi
```

It stuffs the code in data directories instead of at the top-level. This will never work with a traditional PEX, but things do work in `--venv` mode though:
```
$ pex intel-numpy --venv -o intel_numpy.pex && ./intel_numpy.pex -c "import numpy; print(numpy.__file__)"
[...]

Issue #763, opened 2019-08-28
Pex fails with tensorflow==2.0.0rc0  (probably because of manylinux2010)


zachmoshe:
```
<<start in a new virtualenv>>
pip install -U pip
# got me pip-19.2.3
pip install pex
# got me pex-1.6.10
pip install tensorflow==2.0.0rc0

pip freeze
# shows:
# absl-py==0.8.0
# astor==0.8.0
# gast==0.2.2
# google-pasta==0.1.7
# grpcio==1.23.0
# h5py==2.9.0
# Keras-Applications==1.0.8
# Keras-Preprocessing==1.1.0
# Markdown==3.1.1
# numpy==1.17.1
# opt-einsum==3.0.1
# pex==1.6.10
# protobuf==3.9.1
# six==1.12.0
# tb-nightly==1.15.0a20190806
# tensorflow==2.0.0rc0
# termcolor==1.1.0
# tf-estimator-nightly==1.14.0.dev2019080601
# Werkzeug==0.15.5
# wrapt==1.11.2

pex -r <(pip freeze) -o tmp.pex
# fails on:
# Could not satisfy all requirements for tensorflow==2.0.0rc0:
#    tensorflow==2.0.0rc0
```

This only happens when I try `tensorflow==2.0.0rc0`, for example `tensorflow-2.0.0-b1` works fine. It also doesn't happen on macos.
I've noticed that the `-b1` version detects the distribution as `manylinux1` while `rc0` calls it `manylinux2010` - could that be the difference? 

Any solution or a workaround?

Thanks!

dcherniv:
@zachmoshe the only workaround is to repackage tensorflow as manylinux1.

zachmoshe:
Thanks @fhoering, #778 helped me! I see it's still open, so I assume you're planning some more long-term support.

For future readers: I installed pex directly from git (got version 1.6.12 rather than 1.6.10 from apt-get), and it solved the problem.

Issue #761, opened 2019-08-21
pex does not resolve the same version as pip for llvmlite 0.29.0


tibonihoo:
When trying to use the documented pattern of collecting wheels with pip and then using the wheelhouse with `--no-index` for pex generation of a project, I came across a blocking issue due to a difference in which pex and pip resolved the package name for the 0.29.0 version of llvmlite.

Please find below the sequence of commands that "minimally" reproduce the issue (along with showing the versions of the various tools involved):

```
~/shared/Development $mkdir pex-repro
~/shared/Development $cd pex-repro/
~/shared/Development/pex-repro $python3.6 -m venv venv-linux
~/shared/Development/pex-repro $source venv-linux/bin/activate
(venv-linux) ~/shared/Development/pex-repro $python -m pip install --upgrade pip
(...)
Successfully installed pip-19.2.2
(venv-linux) ~/shared/Development/pex-repro $python -m pip install --upgrade pex
(...)
Successfully installed pex-1.6.10
(venv-linux) ~/shared/Development/pex-repro $python -m pip freeze
pex==1.6.10
(venv-linux) ~/shared/Development/pex-repro $python -m pip wheel -w wheelhouseDir "llvmlite>=0.29.0,<0.30.0"
(...)
  Saved ./wheelhouseDir/llvmlite-0.29.0-1-cp36-cp36m-manylinux1_x86_64.whl
Skipping llvmlite, due to already being wheel.
(venv-linux) ~/shared/Development/pex-repro $python -m pex -f wheelhouseDir --no-index --disable-cache "llvmlite>=0.29.0,<0.30.0" -o test.pex
Could not satisfy all requirements for llvmlite<0.30.0,>=0.29.0:
    llvmlite<0.30.0,>=0.29.0
```

For reference and comparison, using pex directly with the same requirement and allowing it to resolve the reference from the package index directly works:
```
(venv-linux) ~/shared/Development/pex-repro $python -m pex -f wheelhouseDir -i https://pypi.org/simple/ -vv --disable-cache "llvmlite>=0.29.0,<0.30.0
" -o test.pex
pex: Warning, using a UrllibContext which is known to be flaky.
pex: Please build pex with the requests module for more reliable downloads.
[...]

Issue #93, opened 2015-04-17
Implement VCS Resolvables


wickman:
ResolvableRepository is included in pex/resolvable.py but is unimplemented.

Eric-Arellano:
Now supported because Pex 2.x vendors Pip and uses Pip for all resolution.

PR #770, opened 2019-09-30
VCS Resolvable

adeandrade:
We rely on the `pip` package to export the repo into a temporary location and let `ResolvableDirectory` do the rest of the work.


adeandrade:
As described in: https://github.com/pantsbuild/pex/issues/93

adeandrade:
There's a problem with `setuptools` in the way it is reading a package definition. It does not return the URL with the package.

bendidi:
hello, thanks for your work, any news on when this would land?
would be glad to help if you give me direction to complete your work :)

jsirois:
@adeandrade vendoring pip is pretty heavyweight just to get vcs resolvables. It's not out of the question as a move for pex to make, but if we make it it would seem to make sense to let pip take over resolution completely - folks expect pex resolves to work just like pip resolves and sometimes they do not. This would fix manylinux2010+ support and other similar issues. See:
+ #771
+ #763
+ #768
+ #761

Pip is a large package though, adding ~5MB IIRC to pex which currently sits at 3MB + due to vendoring setuptools.

jsirois:
Thanks for helping spur movement here @adeandrade. Now that Pex 2+ uses pip internally, VCS requirements are fully supported. I'm going to close this PR as a result.

PR #775, opened 2019-10-17
Add an --intransitive option.

benjyw:
If specified, only the explicitly provided requirements are
resolved.  Their transitive requirements are not.

Since pex checks that all requirements are satisfied, this means
that a complete transitive set must be provided explicitly in
order to use this flag.

This is useful if you have a lockfile from some other source
and want pex to just apply it.

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -121,6 +121,12 @@ def process_precedence(option, option_str, option_value, parser, builder):
   elif option_str in ('--no-manylinux', '--no-use-manylinux'):
     setattr(parser.values, option.dest, False)
     builder.no_use_manylinux()
+  elif option_str == '--transitive':
+    setattr(parser.values, option.dest, True)
+    builder.transitive()
+  elif option_str in ('--no-transitive', '--intransitive'):
+    setattr(parser.values, option.dest, False)
+    builder.intransitive()
   else:
     raise OptionValueError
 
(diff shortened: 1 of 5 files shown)


pex/bin/pex.py
```diff
@@ -121,6 +121,12 @@ def process_precedence(option, option_str, option_value, parser, builder):
   elif option_str in ('--no-manylinux', '--no-use-manylinux'):
     setattr(parser.values, option.dest, False)
     builder.no_use_manylinux()
+  elif option_str == '--transitive':
```
jsirois: Adding this to `process_precedence` seems like a stretch. The precedence setting in PEX deals with which to prefer, wheel or egg or sdist. The binary or not and manylinux options are fairly related but transitivity really isn't. Breaking out a separate callback for `process_intransitive` seems a bit more logical.

pex/requirements.py
```diff
@@ -33,6 +33,9 @@ def __init__(self, filename):
 
 # Process lines in the requirements.txt format as defined here:
 # https://pip.pypa.io/en/latest/reference/pip_install.html#requirements-file-format
+# Note that the lines can be not just pip-style requirement specifiers as detailed
+# at the link above, but any string representation of a Resolvable that pex understands,
+# including package URLs and setup.py directories.
```
jsirois: Your comment is true, so lgtm, but, as of today, Resolvable.get is still a strict subset. You can pip install an url pointing at a package as well as a local directory containing a setup.py.

benjyw: Interesting. That isn't mentioned in that pip documentation though.

jsirois: It sortof is. The archive utl/path one is explicit. The local project path less so, since they only talk about the editable (prefix with -e) case. You can install a local project without -e and it just never updates after that if edited.

---
Each line of the requirements file indicates something to be installed, and like arguments to pip install, the following forms are supported:
```
[[--option]...]
<requirement specifier> [; markers] [[--option]...]
<archive url/path>
[-e] <local project path>
[-e] <vcs project url>
```
For details on requirement specifiers, see Requirement Specifiers.

PR #776, opened 2019-10-17
Allow positional args to be read from a file.

benjyw:
Expands any positional args of the form @path/to/file to the contents
of that file - one line per arg.

This is useful to control the length of the command line when
providing a large number of requirements (e.g., many URLs from
a lockfile).

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -667,6 +667,29 @@ def _compatible_with_current_platform(platforms):
   )
 
 
+def expand_reqs(reqs):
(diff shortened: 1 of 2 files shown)


benjyw:
Hmm, you are correct. The comment at the top of `requirements_from_lines` says 

```# Process lines in the requirements.txt format as defined here:
# https://pip.pypa.io/en/latest/reference/pip_install.html#requirements-file-format```

But it actually does accept the other formats as well. I guess I should just update that comment then.

benjyw:
Updated comment in https://github.com/pantsbuild/pex/pull/775 and closing this.

PR #778, opened 2019-10-18
Add manylinux2010 to list of supported platforms

fhoering:
Proposal to to fix https://github.com/pantsbuild/pex/issues/763 as we are also blocked by this issue.

I investigated this problem and in reality it is complicated.

Pex has forked https://github.com/pantsbuild/pex/blob/master/pex/pep425tags.py from setuptools which handles the supported tags. The last pep425tags version from setuptools also handles manylinux2010 & manylinux2014 support but there is currently a [PR](https://github.com/pypa/setuptools/pull/1829 ) to replace this by pypa/packaging.tags. So I suppose the real fix would be waiting this to be merged, update vendor/setuptools and then reapply the changes. This looks like a lot of work.

In the meantime (waiting for a proper fix) I suggest just adding manylinux2010 to the list of supported platforms. In reality I think having people publish linux_x86_64 and manylinux1 wheels at the same time already doesn't work very well. So this fix should not be much worse.

### DIFF
--- a/pex/platforms.py
+++ b/pex/platforms.py
@@ -54,6 +54,7 @@ def _get_supported(version=None, platform=None, impl=None, abi=None, force_manyl
       python_tag, abi_tag, platform_tag = supported
       if platform_tag.startswith('linux') and force_manylinux:
         yield python_tag, abi_tag, platform_tag.replace('linux', 'manylinux1')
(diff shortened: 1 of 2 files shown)


fhoering:
Tried with the new version. It correctly pulls manylinux2010 wheels.

Issue #746, opened 2019-07-23
Problem running PEX files with restricted permissions


jfinkhaeuser:
When running PEX files, `PexInfo` determines the `install_cache` path based on the `pex_root` path, which in turn can be set in the `PEX-INFO` file - but isn't.

That means it defaults to `~/.pex` for writing the install cache - which fails when that path isn't writeable for the user. That can happen - is common practice, even - when securing a system against unwanted file system access. Writing to the home directory, for example, can modify environment variables in unwanted ways.

There's a number of things that could be improved in the approach of defaulting to `~/.pex`, but different requirements here would conflict with each other - so I'll leave that discussion to other people. The main point, though, is that it would be *very, very useful* if one could set the `pex_root` value that goes into the `PEX-INFO` file, so that you can effectively choose other paths for the `install_cache`. If additionally, `install_cache` could be overridden as well, that'd be excellent.

jsirois:
> If additionally, install_cache could be overridden as well, that'd be excellent.

You should be able to do this. See `pex --help-variables`, but the `PEX_ROOT` environment variable should do what you want. Additionally, pex scans for a `.pexrc` file where you can define this variable as well. See:
https://github.com/pantsbuild/pex/blob/f5e00ad07a4943b101b4d17f94a1146856cfb585/pex/variables.py#L38-L59

Perhaps these two mechanisms solve your issue. Sealing pex root into the pex seems less than ideal since it's a machine specific setting and the pex is machine specific.

I'm going to close this as answered, but please report back with more questions or a success report.

jfinkhaeuser:
Thanks, but this is precisely *not* what I need, going by the documentation that is. I do not care about setting the pex root during the build - during build, I have all the access to the home directory that I require. I need to control the install cache during execution, because that's when my environment is restricted.

I would like to urge you to re-open the issue.

jfinkhaeuser:
I should add, I also tried it out just to confirm: it's not what I need.

jfinkhaeuser:
To be clear: as a matter of fact, I do not need to set the pex root as such: all I need to be able to do is control the directory where the distribution gets unpacked to during execution. Unfortunately, that's tied to the pex root - I'm not currently sure why that would have to be, or should be the case.

I'd also be happy enough if the default was a globally writeable directory, such as `/tmp` - it's the `~/.pex` default that causes issues in restricted runtime environments.

jfinkhaeuser:
[...]

PR #780, opened 2019-10-20
Adding `--runtime-pex-root` option.

yelly:
Adding the option to set the PEX root in in built PEXes.

The option already existed for use from the PexInfo API but was not
exposed to the CLI.

Work towards #746
Work towards #926

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -612,6 +612,9 @@ def build_pex(reqs, options):
   elif options.script:
     pex_builder.set_script(options.script)
 
+  if options.pex_root:
+    pex_builder.set_pex_root(options.pex_root)
+
   if options.python_shebang:
     pex_builder.set_shebang(options.python_shebang)
 
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -619,9 +619,6 @@ def build_pex(reqs, options):
   elif options.script:
     pex_builder.set_script(options.script)
 
-  if options.pex_root:
-    pex_builder.set_pex_root(options.pex_root)
-
   if options.python_shebang:
     pex_builder.set_shebang(options.python_shebang)
 
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -298,6 +298,13 @@ def configure_clp_pex_options(parser):
            'will not be reproducible, meaning that if you were to run `./pex -o` with the '
            'same inputs then the new pex would not be byte-for-byte identical to the original.')
 
+  group.add_option(
+      '--pex-root',
+      dest='pex_root',
+      default=None,
+      help='Specify the pex root to be used in the generated .pex file. [Default: ~/.pex]'
+  )
+
   parser.add_option_group(group)
 
 
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -299,8 +299,8 @@ def configure_clp_pex_options(parser):
            'same inputs then the new pex would not be byte-for-byte identical to the original.')
 
   group.add_option(
-      '--pex-root',
-      dest='pex_root',
+      '--runtime=pex-root',
+      dest='runtime_pex_root',
       default=None,
       help='Specify the pex root to be used in the generated .pex file. [Default: ~/.pex]'
   )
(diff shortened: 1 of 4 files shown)


jsirois:
@yelly thanks for the contribution. Can you provide some motivation for the feature after reading the problems with it below? Perhaps you don't need the feature at all, or, if you do, you can propose fixes for the difficulties I describe.

---
On the `pex/bin/pex.py`, that use of `--pex-root` is explicitly different:
```
$ pex --version
pex 1.6.12
$ pex --help
...
  --pex-root=PEX_ROOT   Specify the pex root used in this invocation of pex.
                        [Default: ~/.pex]
...
```
So to make --pex-root also apply to the generated pex at least requires an update to to the option helps string. That said...

You can already specify `--pex-root` using a `PEXBuilder` via `PexInfo`:
https://github.com/pantsbuild/pex/blob/60fd51395076e8b52d8a403dbf0add466a0a3051/pex/pex_builder.py#L66-L85
https://github.com/pantsbuild/pex/blob/60fd51395076e8b52d8a403dbf0add466a0a3051/pex/pex_info.py#L287-L289

But that is a chancy thing to do since since that allows embedding a pex root path in a pex that then gets shipped to a machine without that path or permissions to create it. Say the pex contains cross-platform python code and is built and tested by an engineer on OSX and CI on linux - both with /var/cache/pex but deployed to an linux machine without that directory. Currently the pex will error when executed unless the bad path is corrected by exporting PEX_ROOT to a valid path via one of the means descibed below.

You can control a pex's `--pex-root` via the `PEX_ROOT` envirnment variable:
```
$ pex --help-variables
...
PEX_ROOT: Directory

    The directory location for PEX to cache any dependencies and code.
    PEX must write not-zip-safe eggs and all wheels to disk in order
    to activate them.  Default: ~/.pex
...
```

To control the default for a machine, user or pex binary you can deploy `/etc/pexrc`, `~/.pexrc` and `.pexrc` as a sibling to the pex executable respectively:
https://github.com/pantsbuild/pex/blob/e5317cbc94a49d5129144806ce4117ccd926fb6a/pex/variables.py#L38-L59
```
PEX_ROOT=/varc/cache/pex
```

yelly:
1. I agree that this change also requires a change in the usage, I will update the PR with a suggestion.
2. Of course this can be set directly in code, but it is useful to have this option directly on the command line.
3. My use case is setting a pex root that is relative to the pex file (for example '.pex'). This is useful for cases where you want the executable to run "locally" without polluting the file system it's running on or making any assumptions about it. This use case also answers your concerns regarding CI. I may add that the current situation (defaulting to '~/.pex') may also fail when run by a user with no home directory.
4. All the other solutions offered would work but obviously complicate deployment, which is exactly the problem the pex format is trying to solve.
5. Overall, the mindset that led me to suggesting this change is that I intuitively expect the executable created by `pex [options] -o app.pex` to behave exactly the same as `pex [options]`, and this change brings pex closer to that behaviour. Assuming I am not the only one that has this approach to the `-o` flag, it might be wise to at least print a warning if it is used in conjunction with `--pex-root` in the case this change is not accepted.

jsirois:
The default of ~/.pex not necessarily being writable is a good point. I think my only hard requirement then is to have the pex CLI tool `--pex-root` be separate from the `--pex-root` assigned to pexes it generates. Perhaps the CLI pex root can only be set via PEX_ROOT / pexrc and the `--pex-root` CLI flag changes meaning to setting the resulting pex's pex root.

yelly:
@jsirois,  I've updated the PR to reflect your suggestion that `--pex-root` only apply to generated `.pex` files.

yelly:
Working on this, will have a working PR soon.

yelly:
@jsirois, I've updated the PR so that `--pex-root` only applies to generated pexes, as you described.

yelly:
@jsirois, should i be resolving the merge conflicts? Will this get merged?

jsirois:
@yelly thank you very much for pushing this forward. I've since filed #926 which will go a bit further, but I'd like to get your contribution in. I'm going to fix things up a bit to be inline with that change presently so we can get this submitted.

jsirois:
Thanks for your perserverance @yelly! This feature along with fixed --cache-dir support in PEXed Pex will ship shortly in 2.1.7.

yelly:
Awesome, thank you!

Issue #782, opened 2019-10-22
Interpreter discovery and pyenv don't interact well


benjyw:
At pex runtime, interpreter discovery looks on the PATH for any binary with a name that looks like it might be a python interpreter, and executes it to get its version (and verify that it is in fact a python interpreter). 

Unfortunately, pyenv can create a large number of such files in `~/.pyenv/shims`. E.g.,

```
> ls /Users/benjyw/.pyenv/shims/ | grep -E "python\d?(.\d)?m?$"
python
python2
python2.7
python3
python3.5
python3.5m
python3.6
python3.6m
python3.7
python3.7m
>
```
Note that these aren't symlinks, but wrapper scripts, so pex's realpath deduping doesn't help.

So, pex will execute every one of these, plus whatever other python interpreters it finds, which adds up to non trivial overhead (~1 second in the above example) on every run of the pex. Note that not all of these are actually valid python interpreters - some are shim stubs that fail on purpose.

benjyw:
I don't have an obvious solution here. Some heuristics to discuss:
1. Checking the interpreter pex is running under first. It may be likely to be the right one?
2. Stopping as soon as we find a valid interpreter, instead of first executing all of them and then comparing. This won't help in the worst case but probably will the average case.  This will effectively implement 1.
3. An on-disk interpreter cache, in some well-known location, to speed up subsequent runs.  This won't help much in container scenarios.
4. Be smarter about matching names to versions. E.g., if the user specified Python>=3.7, we can eliminate many versions just by naming convention.

jsirois:
It would definitely be good to implement 1 directly, this is already nodded towards in some of the machinations in `maybe_reexec_pex`, but the treatment does not include the cases where PEX_PYTHON or PEX_PYTHON_PATH are set:
https://github.com/pantsbuild/pex/blob/e5317cbc94a49d5129144806ce4117ccd926fb6a/pex/pex_bootstrapper.py#L165-L168
Presumably this catches the majority of cases - the current interpreter is fine.

Instead of 2 a multiprocessing approach could probe all interpreters in parallel. Perhaps too complex.

Item 4 makes sense, but I think 1 and maybe 2 might do it.

jsirois:
I took a whack at point 1 in #783.

jsirois:
[...]

Issue #694, opened 2019-03-29
pex multiplatform resolution is broken for non-host-available platforms


jsirois:
In particular `Platform.get_supported` delegates much of it's work to pep425tags / glibc which both consult a local interpreter to answer questions like:
1. Does the interpreter support manylinux?
2. Does the interpreter have abi3 support?

Concretely, if a user runs pex on OSX and supplies a `--platform=linux-x86_64-cp-37-m`, there is no way for pex to determine the answers to those questions for the python 3.7 the pex will eventually run on. The problem is worse for non-extended platforms; eg: `--platform=linux-x86_64`. In those cases we don't know if the eventual linux runtime interpreter will be `d`, `m`, `mu`, `dmu`, etc... and currently accept wheels for any of these!

We need ways to do one or all of:
1. Let the user specify the answers to the 1st two questions - perhaps with new pex API/CLI options.
2. Fail if building a pex encounters situations where these unanswered questions are important - namely pex resolves platform-specific wheels.

jsirois:
Fixing #539 brough this issue to light. #658 also touches on this area of broken multiplatform resolution.

Issue #658, opened 2019-01-30
When using resolve_multi, not all combinations of `interpreters` and `platforms` work.


jsirois:
There is at least one bug / ambiguity in behavior. If a user passes 1 interpreter (say the current interpreter executing pex) and one or more platforms that do not match the current interpreter version but do match the os arch, then the wrong interpreter will be passed to resolve and used to silently build the wrong wheel:
https://github.com/pantsbuild/pex/blob/8fd5e1b989224046cd0fb038cfebc11b4c10a8c9/pex/resolver.py#L505-L568

Either the build should fail or else an attempt to find a correct local interpreter should proceed and then, if not found, the build should fail; otherwise _that interpreter_ should be passed to resolve.

PR #783, opened 2019-10-25
Bias selecting the current interpreter.

jsirois:
Preiously we made some effort towards this, but were not thorough.
Fix all interpreter selection logic to be both lazy and bias towards
re-using the current interpreter, only falling back to picking the
minimum version compatible interpreter from all possible compatibile
interpreters as the very last step.

Addresses item 1 in #782.

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -18,7 +18,7 @@ from pex.fetcher import Fetcher, PyPIFetcher
 from pex.interpreter import PythonInterpreter
 from pex.interpreter_constraints import validate_constraints
 from pex.pex import PEX
-from pex.pex_bootstrapper import find_compatible_interpreters
+from pex.pex_bootstrapper import iter_compatible_interpreters
 from pex.pex_builder import PEXBuilder
 from pex.platforms import Platform
 from pex.requirements import requirements_from_file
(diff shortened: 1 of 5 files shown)


pex/bin/pex.py
```diff
@@ -574,7 +574,7 @@ def to_python_interpreter(full_path_or_basename):
       pex_python_path = rc_variables.get('PEX_PYTHON_PATH', '')
     else:
       pex_python_path = ""
-    interpreters = find_compatible_interpreters(pex_python_path, constraints)
+    interpreters = list(iter_compatible_interpreters(pex_python_path, constraints))
```
benjyw: Why create a list here? Wouldn't it cause the same performance problem again (albeit only a build time)? Could we do something like this?

```
try:
  interpreter = next(interpreters)
except StopIteration:
  die('Could not find compatible interpreter', CANNOT_SETUP_INTERPRETER)
```

Or is it specifically important to choose the `min` interpreter here?

jsirois: I was just gunning for changing ~no logic and, yeah, it was only at build time so even less of a concern.

benjyw: Thing is, pex build time can be build tool user runtime. For example, when iterating on some tests and rebuilding the test sources chroot, this would run every time I think?

jsirois: It would. Keep in mind two things though:
1. The interpreters list is used below to build a multi-interpreter / multi-platform pex.
2. There are only actually more than one interpreter if either more than one --python are passed (ie: `--python=python2.7 --python=python3`) or one or more `--interpreter-constraint` is passed.

Due to item 1, we must materialize the full list. The 1st half of item 2 is a feature, the second half of item 2 may be a bug in the ballpark of #658 and #694.

Issue #785, opened 2019-10-31
minimum_sys_modules keeps tainted non-package modules


zhanwu:
For some packages (e.g. Jinja2), both tainted packages and non-package modules can be imported. However `minimum_sys_modules` would only drop the tainted packages but kept non-package modules.

When `Jinja2` is imported, there are both `Jinja2` and `Jinja2.utils` in sys.modules; minimum_sys_modules() would drop `Jinja2` but keep `Jinja2.utils` (because it doesn't have `__path__`), and this cause user code to unable to load another version of `Jinja2` correctly:  
https://github.com/pantsbuild/pex/blob/master/pex/pex.py#L184-L203

zhanwu:
Had been fixed by:
https://github.com/pantsbuild/pex/issues/745

PR #786, opened 2019-10-31
Remove tainted non-package modules

zhanwu:
Description of the issue:
https://github.com/pantsbuild/pex/issues/785

Don't include the sub-modules of the main modules which are tainted packages.

### DIFF
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -200,16 +200,18 @@ class PEX(object):  # noqa: T000
       # The package still contains untainted path elements, so it can stay.
       if module.__path__:
         new_modules[module_name] = module
-     else:
+      else:
         tainted_package_names += [module_name]
 
     def is_tainted_sub_modules(sub_module_name):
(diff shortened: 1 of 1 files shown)


jsirois:
Thanks for digging @zhanwu! I'll give this a proper review later today, but a test case would be really good to have. I think https://github.com/pantsbuild/pex/blob/master/tests/test_environment.py has all the test infrastaucture / examples of its use that you'd need, but I'll also try to provide better suggestions on that as well when I review.

zhanwu:
Sounds good, I will spend more time this weekend adding test

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

zhanwu:
While I was checking the test cases, I realized I developed my change on a previous version of pex and applied it to the latest code.
The issue actually had already been fixed in latest code:
https://github.com/pantsbuild/pex/commit/0f9bf622e1565dcdd27c3b79b788571c9608c272

I am closing this PR.

PR #787, opened 2019-11-07
Add a --dehydrated option to dump requirements to PEX-INFO, then resolve them at bootstrap time.

cosmicexplorer:
### TODO
- [ ] Fix this PR to perform the resolve process by installing into PEX_ROOT upon the first run, instead of during pex bootstrap.
- [ ] Validate internally that this solution results in the expected latency improvement for redeploying large ML applications within our datacenter.
  - Consider parallelizing the bootstrap resolve and investigating other caching solutions if not.
- [ ] Add testing.
- [ ] Wait for review (or, if this change isn't compatible with the pex project, we can just externalize the "hydration" process using something like virtualenv internally and close this PR).

### Problem

Redeploying pex files full of many extremely large 3rdparty requirements (tensorflow, etc) into our datacenter at Twitter currently takes a very long time, since we upload them all at once into an internal artifact resolution utility, and then pull down the entire pex file before executing it. This slowness to redeploy then also affects multiple of our internal python development workflows and tooling for machine learning (including a Jupyter wrapper developed by @kwlzn, who was not consulted before making this PR) which depend on executing a pex file within the datacenter -- in that case, modifying any python source files in our monorepo currently requires waiting several minutes for changes to be usable within that Jupyter notebook.

As far as we are aware, other users of pex who package large machine learning applications also suffer from this issue and do not have an easy workaround.

### Solution

We would like to be able to ship around "dehydrated" pex files without 3rdparty requirements embedded in the pex, and resolve ("hydrate") them before executing the pex. This removes one half of the current process of synchronously waiting to upload and download 3rdparty requirements, and moves the remaining download part off the critical path of the entire redeploy process. Because the requirements to hydrate were already resolved when building the pex, we know all the exact versions of all the transitive dependencies to resolve at bootstrap time.

There are many ways we could potentially make the bootstrap resolve process faster -- this PR just uses a `CachingResolver`, with the idea that the machines that execute our pex files will eventually have most of the large distributions cached and won't need to redownload them on every redeploy (or that we can provision machines to have these requirements already contained within their local pex cache).

An alternative implementation of "hydration" that was considered was to use a virtualenv to hydrate requirements before running the pex with `PEX_INHERIT_PATH=fallback`, but it would be extremely helpful for us to avoid having to maintain separate tooling with virtualenv, and it would be really nice if pex could do that itself at bootstrap.

### Implementation

- Add `dehydrated_requirements` field to `PexInfo`.
- Add `--dehydrated` command line option, which resolves all requirements, then dumps the transitive requirements into `dehydrated_requirements` in PEX-INFO.
- Add `PEX_BOOTSTRAP_HYDRATION_INDICES` to `variables.py`, which determines which PyPI-compatible package indices to resolve against at bootstrap.
- At bootstrap time when executing the pex, check for `dehydrated_requirements` in PEX-INFO and perform a resolve against `PEX_BOOTSTRAP_HYDRATION_INDICES`. The locations of the downloaded distributions are then appended to `scrubbed_sys_path` when `sys.path` is patched.

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -660,7 +660,7 @@ def build_pex(args, options, resolver_option_builder):
             V=options.verbosity)
 
         if options.dehydrated:
-          pex_info.add_dehydrated_requirement(resolved_dist.requirement)
+          pex_info.add_dehydrated_requirement(resolved_dist.distribution.as_requirement())
         else:
           pex_builder.add_distribution(resolved_dist.distribution)
           pex_builder.add_requirement(resolved_dist.requirement)
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -260,6 +260,17 @@ def configure_clp_pex_resolution(parser, builder):
     callback_args=(builder,),
     help='Whether to transitively resolve requirements. Default: True')
 
+  group.add_option(
+    '--dehydrated',
+    dest='dehydrated',
+    default=False,
+    action='callback',
+    callback=parse_bool,
+    help='[EXPERIMENTAL] Whether to avoid adding requirements to the generated pex file and '
+         'instead add an entry to PEX-INFO containing the transitive dependencies of all '
+         'requirements.'
+  )
+
   # Set the pex tool to fetch from PyPI by default if nothing is specified.
   parser.set_default('repos', [PyPIFetcher()])
   parser.add_option_group(group)
@@ -647,8 +658,12 @@ def build_pex(args, options, resolver_option_builder):
       for resolved_dist in resolveds:
         log('  %s -> %s' % (resolved_dist.requirement, resolved_dist.distribution),
             V=options.verbosity)
-        pex_builder.add_distribution(resolved_dist.distribution)
-        pex_builder.add_requirement(resolved_dist.requirement)
+
+        if options.dehydrated:
+          pex_info.add_dehydrated_requirement(resolved_dist.requirement)
+        else:
+          pex_builder.add_distribution(resolved_dist.distribution)
+          pex_builder.add_requirement(resolved_dist.requirement)
     except Unsatisfiable as e:
       die(e)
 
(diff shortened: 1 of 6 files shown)


jsirois:
1st - this should definitely should have started as an issue - no code. This changes a fundamental bit of the informal pex contract. You no-longer ship an app, you resolve it at runtime and allow in all manner of non-reproduceabilities that need to be carefully excluded or contended with. So, please do that, but comments that can go on that issue include:

Doesn't use of PEX_PATH already solve this? Build 2 pexes, one with the highly variable bits - maybe just sources and resources, and one with the more stable bits - large dependencies. Then on the deploy side, of the 2 pexes that now compose the app, only deploy the ondes that have changed. To make this work well, we may need to enhance the build-side of pex; i.e.: have the pex tool / API understand how to resolve from other pexes as 1st-class indexes / find-links repos; but, crucially, this leaves the runtime unchanged. The same guarantees apply.

cosmicexplorer:
Created #789 to discuss this before attempting to move forward with this PR at all.

pex/bin/pex.py
```diff
@@ -647,8 +658,12 @@ def walk_and_do(fn, src_dir):
       for resolved_dist in resolveds:
         log('  %s -> %s' % (resolved_dist.requirement, resolved_dist.distribution),
             V=options.verbosity)
-        pex_builder.add_distribution(resolved_dist.distribution)
-        pex_builder.add_requirement(resolved_dist.requirement)
+
+        if options.dehydrated:
+          pex_info.add_dehydrated_requirement(resolved_dist.distribution.as_requirement())
```
jsirois: This is incorrect. `Distribution.as_requirement()` loses environment markers, you must use `resolved_dist.requirement`.

jsirois: Why is a new field in PEX-INFO needed at all. Can't the existing requirements be used and just addition of distributions be skipped? IE: On pex boot, if len(requirements) > 0 and len(distributions) == 0, we're in rehydration mode. Further, this would imply that re-hydration logic belongs in PEXEnvironment, where distributions are normally found inside the PEX and added to the sys.path. Inventing a new way to add these to the sys.path below worryingly expands the testing parameter space for no good reason I can see.

cosmicexplorer:
There are some useful comments here and in #789. Given that we haven't figured out a way to address the above comments yet, I'm closing this for now.

PR #798, opened 2019-11-20
Fix `iter_compatible_interpreters` path biasing.

jsirois:
Biasing interpreter selection towards the current interpreter was
implemented in #783 but not test was added. Add a test that fails due to
a bug in path trimming logic and fix the bug.


jsirois:
N.B.: Discovered trying to upgrade Pants to Pex 2.0.0.

PR #795, opened 2019-11-15
Fix #794 issue by add missing return statement in __str__

Djailla:
Fixing issue #794, every build failed using --index-url option

PR #799, opened 2019-11-20
Add a test of pypi index rendering.

jsirois:
Previously rendering was broken but fixed in #795. Add a test of that
fix.


jsirois:
Thanks for the original fix @Djailla. This adds a test to make sure we don't regress. Not a wonderful test, but it feels like forward progress.

PR #801, opened 2019-11-21
Fix current platform handling.

jsirois:
Three aspects of current platform handling are fixed:

1. Ensure that `Platform.supported_tags` returns the full range of tags
when the platform is current. This uses the special no-arg form of
`pep425tags.get_supported` which consults the local interpreter to
return the broadest possible set of tags. Also ensure that we always
pass all compatible versions when the platform is not current to
generate the broadest possible set of tags for the foreign platform.

2. Ensure that the resolver interprets 'current' to mean any local
interpreter instead of just the current runtime interpreter. This is
faithful to prior behavior where a PEX built for the 'current' platform
would always work on the current machine.

3. Fix the PEX cli when run with a fully foreign set of `--platform`s.
Previously the attempt to warn the user would error on an invalid log
call that was missing a `V` argument.

Tests of all of the above are added and platform parsing is pushed up to
the CLI and resolver API layers for cleaner code under those layers and
earlier failures for invalid platform specifiers.

### DIFF
--- a/pex/platforms.py
+++ b/pex/platforms.py
@@ -4,8 +4,8 @@
 from __future__ import absolute_import
 
 from collections import namedtuple
+from textwrap import dedent
 
-from pex.orderedset import OrderedSet
 from pex.pep425tags import get_abbr_impl, get_abi_tag, get_impl_ver, get_platform, get_supported
 
 
@@ -39,39 +39,29 @@ def _gen_all_compatible_versions(version):
   return list(iter_compatible_versions())
 
 
-def _get_supported(version=None, platform=None, impl=None, abi=None, force_manylinux=False):
-  versions = _gen_all_compatible_versions(version) if version is not None else None
-  all_supported = get_supported(
-    versions=versions,
-    platform=platform,
-    impl=impl,
-    abi=abi
-  )
-
-  def iter_all_supported():
-    for supported in all_supported:
-      yield supported
-      python_tag, abi_tag, platform_tag = supported
-      if platform_tag.startswith('linux') and force_manylinux:
-        yield python_tag, abi_tag, platform_tag.replace('linux', 'manylinux1')
-
-  return list(OrderedSet(iter_all_supported()))
-
-
 class Platform(namedtuple('Platform', ['platform', 'impl', 'version', 'abi'])):
   """Represents a target platform and it's extended interpreter compatibility
   tags (e.g. implementation, version and ABI)."""
 
+  class InvalidPlatformError(Exception):
+    """Indicates an invalid platform string."""
+
   SEP = '-'
 
-  def __new__(cls, platform, impl=None, version=None, abi=None):
+  def __new__(cls, platform, impl, version, abi):
+    if not all((platform, impl, version, abi)):
+      raise cls.InvalidPlatformError(
+        'Platform specifiers cannot have blank fields. Given platform={platform!r}, impl={impl!r}, '
+        'version={version!r}, abi={abi!r}'.format(
+          platform=platform, impl=impl, version=version, abi=abi
+        )
+      )
     platform = platform.replace('-', '_').replace('.', '_')
-    if all((impl, version, abi)):
-      abi = cls._maybe_prefix_abi(impl, version, abi)
+    abi = cls._maybe_prefix_abi(impl, version, abi)
     return super(cls, Platform).__new__(cls, platform, impl, version, abi)
 
   def __str__(self):
-    return self.SEP.join(self) if all(self) else self.platform
+    return self.SEP.join(self)
 
   @classmethod
   def current(cls):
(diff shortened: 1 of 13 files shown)


jsirois:
N.B.: These issues were discovered trying to upgrade Pants to Pex 2.0.0.

tests/test_resolver.py
```diff
@@ -139,6 +151,70 @@ def test_resolve_extra_wheel():
...
+  current_python = PythonInterpreter.get()
+
+  resolved_other = resolve_current(interpreters=[other_python])
+  resolved_current = resolve_current()
+
+  assert 1 == len(resolved_other)
+  assert 1 == len(resolved_current)
+  assert resolved_other != resolved_current
+  assert resolved_current == resolve_current(interpreters=[current_python])
+  assert resolved_current == resolve_current(interpreters=[current_python, current_python])
```
Eric-Arellano: I don't understand this line. When would you ever have two instances of the same interpreter?

tests/test_resolver.py
```diff
@@ -139,6 +151,70 @@ def test_resolve_extra_wheel():
...
+
+  # Here we have 2 local interpreters satisfying current but with different platforms and thus
+  # different dists for 2 total dists.
```
Eric-Arellano: Thank you for the explanation!

tests/test_resolver.py
```diff
@@ -139,6 +151,70 @@ def test_resolve_extra_wheel():
...
+  assert 2 == len(resolve_current(interpreters=[current_python, other_python]))
+
+
+@pytest.mark.skipif(PY_VER < (3, 5) or IS_PYPY,
+                    reason="The p537 distribution only builds for CPython 3.5+")
+def test_resolve_current_and_foreign_platforms(p537_resolve_cache):
+  foreign_platform = 'macosx-10.13-x86_64-cp-37-m' if IS_LINUX else 'manylinux1_x86_64-cp-37-m'
```
Eric-Arellano: Check on my understanding: the `cp-37-m` piece is irrelevant here, even if running the test with a CP37 interpreter, correct? The reason for that is that the OS difference is enough for them to be different.

tests/test_platform.py
```diff
@@ -99,67 +103,58 @@ def test_pypy_abi_prefix():
...
-  ]
-  assert expected_tags == tags
-
-
-# NB: Having to patch here is a symptom of https://github.com/pantsbuild/pex/issues/694
-# Kill when the Platform API is fixed to not need to consult the local interpreter.
+  assert_tags(
+    'linux-x86_64-cp-37-m',
+    [
+      ('cp37', 'cp37m', 'linux_x86_64'),
```
Eric-Arellano: Should `cp38` be added now that it's officially released?

tests/test_pex_binary.py
```diff
@@ -167,3 +175,45 @@ def test_clp_prereleases_resolver():
     pex_builder = build_pex(reqs, options)
     assert pex_builder is not None
     assert len(pex_builder.info.distributions) == 3, 'Should have resolved deps'
+
+
+def test_build_pex():
+  with temporary_dir() as sandbox:
+    pex_path = os.path.join(sandbox, 'pex')
+    results = run_pex_command(['ansicolors==1.1.8', '--output-file', pex_path])
```
Eric-Arellano: Huh...all this time I thought `1.0.2` was the latest. Going to see if this newer version (still 2 years old) fixes the regex DeprecationWarning Pants CI is filled with.

pex/platforms.py
```diff
@@ -87,15 +77,33 @@ def create(cls, platform):
...
+        + <python impl abbr>: cp
+        + <python version>: 37
+        + <abi>: cp37m
+        
+        The abbreviated platform string is:
+          linux-x86_64-cp-37-m
+          
+        These fields stem from wheel name conventions as outlined in
+        https://www.python.org/dev/peps/pep-0427#file-name-convention and influenced by
+        https://www.python.org/dev/peps/pep-0425.
```
Eric-Arellano: Fantastic explanation. Thank you.

tests/test_resolver.py
jsirois: You wouldn't, but primarily this is test of de-duping and at the unit level, interpreters are handed in as a list, so an API user could hand in dups. More towards reality, given an interpreter constraint (that whole concept has been broken from the get-go really), you might collect a python 2.7.15 - say from pyenv, and a 2.7.17, say from the system. Since wheels / platform specifiers resolve only down to major/minor, these two different interpreters would produce the same wheels in which case we'd want to dedup.

tests/test_resolver.py
jsirois: It's relevant in that we need the test to work and cp-37-m is one of the pre-built wheels available. We need to have pre-built wheels available since wheel building is turned off & impossible when platforms are foreign: https://pypi.org/project/p537/#files

tests/test_platform.py
jsirois: Perhaps, but then the platform would need to change. The algorithm / correct answer is a countdown from 3.minor to 3.2 for abi3 wheels. Counting up to some arbitrary future minor cweiling is not a thing.

I'll defer 3.8 support to another PR, there is more that needs to be done in CI and tox.

PR #813, opened 2019-11-26
Implicitly trust explicitly requested repos.

### LINKED ISSUE #812: Pex should trust any host passed via `--index` or `--find-links`.
The underlying pip calls don't assume this and error out, e.g.:
```
E   	WARNING: The repository located at www.antlr3.org is not a trusted or secure host and is being ignored. If this repository is available via HTTPS we recommend you use HTTPS instead, otherwise you may silence this warning and allow it anyway with '--trusted-host www.antlr3.org'.
E   	ERROR: Could not find a version that satisfies the requirement antlr_python_runtime==3.1.3 (from versions: none)
E   	ERROR: No matching distribution found for antlr_python_runtime==3.1.3
...
E   	Exception caught: (pex.resolver.Unsatisfiable)
...
E   	  File "/b/f/w/pants/backend/python/subsystems/pex_build_util.py", line 278, in _resolve_multi
E   	    allow_prereleases=python_setup.resolver_allow_prereleases)
E   	  File "/b/f/w/pex_root/install/pex-2.0.2-py2.py3-none-any.whl.b89b057c31d08e6bebdb4531720b75237c56588e/pex-2.0.2-py2.py3-none-any.whl/pex/resolver.py", line 195, in resolve
E   	    raise Unsatisfiable(str(e))
E   	
E   	Exception message: Executing /usr/bin/python2.7 -s -m pip --disable-pip-version-check --isolated -q --cache-dir /home/nobody/.cache/pants/python_cache/requirements/CPython-2.7.12 download --dest /tmp/tmpj2dt894x/resolved --index-url https://pypi.org/simple/ --find-links http://www.antlr3.org/download/Python/ antlr_python_runtime==3.1.3 failed with 1
```

jsirois:
Previously we did not do this and could thus confusingly fail resolves
against explicitly requested insecure repos.

Fixes #812

### DIFF
--- a/pex/pip.py
+++ b/pex/pip.py
@@ -8,6 +8,7 @@ import os
 from collections import deque
 
 from pex import third_party
+from pex.compatibility import urlparse
 from pex.interpreter import PythonInterpreter
 from pex.platforms import Platform
 from pex.variables import ENV
@@ -46,6 +47,16 @@ def execute_pip_isolated(args, cache=None, interpreter=None):
 
 
 def _calculate_package_index_options(indexes=None, find_links=None):
+  trusted_hosts = []
+
+  def maybe_trust_insecure_host(url):
+    url_info = urlparse.urlparse(url)
+    if 'http' == url_info.scheme:
+      # Implicitly trust explicitly asked for http indexes and find_links repos instead of requiring
+      # seperate trust configuration.
+      trusted_hosts.append(url_info.netloc)
+    return url
+
   # N.B.: We interpret None to mean accept pip index defaults, [] to mean turn off all index use.
   if indexes is not None:
     if len(indexes) == 0:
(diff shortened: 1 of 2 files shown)


pex/pip.py
```diff
@@ -46,23 +47,37 @@ def execute_pip_isolated(args, cache=None, interpreter=None):
...
 
 def _calculate_package_index_options(indexes=None, find_links=None):
+  trusted_hosts = []
+
+  def maybe_trust_insecure_host(url):
+    url_info = urlparse.urlparse(url)
+    if 'http' == url_info.scheme:
+      # Implicitly trust explicitly asked for http indexes and find_links repos instead of requiring
+      # seperate trust configuration.
+      trusted_hosts.append(url_info.netloc)
```
Eric-Arellano: Do you think we should log when this happens? I'm not sure what `-v` level that would be, but maybe something people should know about and encourage them to see if there's an `https` version.

jsirois: In order for this to happen, they said something like `pex --index=http://my.index ...`. If they really mean to use the insecure index - say its internal - they probably would appreciate not being nagged. I think we'd need to support some way of saying "no, really, trust me" in that case.

PR #815, opened 2019-11-26
Use the resolve cache to skip installs.

### LINKED ISSUE #809: A cache should always be used by `pex.resolver.resolve`.
Even if no cache is requested, the `resolve` function executes `pip download` and `pip wheel` in serial, both of which can hit the network. The `pip wheel` call can safey use one-shot cache to take advantage of work just done by `pip download`.

### LINKED ISSUE #811: Slow builds in 2.0 - how does --python interact with --platform?
I want to build a pex that supports Linux armv7l and x86_64, on Python versions 3.5 through 3.7.

Before pex 2.0, I used

`--platform linux_armv7l --platform linux_x86_64 --python python3.5 --python python3.6 --python python3.7 --python-shebang "/usr/bin/env python3"`

And that produced a pex file meeting those requirements in about ten seconds.

As of 2.0 I'm not sure what I need to do. pex seems to resolve requirements for every combination of platform and python, so if I specify all six relevant `--platform` options (`--platform linux_armv7l-cp-35-m`, `--platform linux_x86_64-cp-36-m`, etc) plus all three `--python` options it takes **200s**.

For my use case I can simplify and specify just `--platform any-py-<35/36/37>-none` and the `--python`s. That takes **90s**.

Ultimately the dependencies are in this case resolved to py3-none-any wheels, so I figure I can probably get away with specifying `--platform any-py-35-none` and the `--python`s. That takes **60s**.

If I omit `--python` entirely with the single `--platform` arg, it takes about **15s** - close to the pre-2.0 time. For every `--python` I add, including the default interpreter, it adds another ~15s.

In all these cases the actual resolved and bundled dependencies are identical. What effect does `--python` actually have on the pex over and above the platform specifier? Is there anything I can do to get the build time down while ensuring that I'm producing a file that will work where it needs to?

jsirois:
Previously the cache was only used by pip for http queries. Now we also
use the cache to skill wheel installs in their isolated chroots.

Fixes #809
Fixes #811

### DIFF
--- a/pex/resolver.py
+++ b/pex/resolver.py
@@ -6,13 +6,13 @@ from __future__ import absolute_import
 import functools
 import json
 import os
-import shutil
 import subprocess
 from collections import defaultdict, namedtuple
 from textwrap import dedent
+from uuid import uuid4
 
 from pex import third_party
-from pex.common import safe_mkdir, safe_mkdtemp
+from pex.common import safe_mkdtemp
 from pex.interpreter import PythonInterpreter
 from pex.orderedset import OrderedSet
 from pex.pip import PipError, build_wheels, download_distributions, install_wheel
@@ -172,27 +172,29 @@ def resolve(requirements=None,
     return []
 
   workspace = safe_mkdtemp()
+  cache = cache or workspace
   resolved_dists = os.path.join(workspace, 'resolved')
   built_wheels = os.path.join(workspace, 'wheels')
-  installed_chroots = os.path.join(workspace, 'chroots')
+  installed_chroots = os.path.join(cache, 'chroots')
 
   # 1. Resolve
-  try:
-    download_distributions(target=resolved_dists,
-                           requirements=requirements,
-                           requirement_files=requirement_files,
-                           constraint_files=constraint_files,
-                           allow_prereleases=allow_prereleases,
-                           transitive=transitive,
-                           interpreter=interpreter,
-                           platform=parsed_platform(platform),
-                           indexes=indexes,
-                           find_links=find_links,
-                           cache=cache,
-                           build=build,
-                           use_wheel=use_wheel)
-  except PipError as e:
-    raise Unsatisfiable(str(e))
+  with TRACER.timed('Resolving and downloading distributions'):
+    try:
+      download_distributions(target=resolved_dists,
+                             requirements=requirements,
+                             requirement_files=requirement_files,
+                             constraint_files=constraint_files,
+                             allow_prereleases=allow_prereleases,
+                             transitive=transitive,
+                             interpreter=interpreter,
+                             platform=parsed_platform(platform),
+                             indexes=indexes,
+                             find_links=find_links,
+                             cache=cache,
+                             build=build,
+                             use_wheel=use_wheel)
+    except PipError as e:
+      raise Unsatisfiable(str(e))
 
   # 2. Build
   to_build = []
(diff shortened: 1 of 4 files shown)


pex/resolver.py
```diff
@@ -205,57 +207,64 @@ def resolve(requirements=None,
...
-  markers_by_req_key = _calculate_dependency_markers(resolved_distributions,
-                                                     interpreter=interpreter)
+  with TRACER.timed('Installing distributions:\n  {}'.format('\n  '.join(to_install))):
+    for wheel_file_path in to_install:
+      wheel_file = os.path.basename(wheel_file_path)
+      chroot = os.path.join(installed_chroots, wheel_file)
+      if os.path.exists(chroot):
+        TRACER.log('Using cached installation of {} at {}'.format(wheel_file, chroot))
+      else:
+        TRACER.log('Installing {} in {}'.format(wheel_file_path, chroot))
```
Eric-Arellano: Nit: `in` makes it sound like you are installing it in the chroot directory. Really, the `chroot` variable here is the specific file location, not the parent directory. A simple fix would be to change the proposition from `in` to `at`.

Better, I think, is to rename `chroot` to something like `chrooted_wheel_path` or something that more clearly indicates this is the chrooted file location, not the general chrooted directory location.

pex/resolver.py
```diff
@@ -205,57 +207,64 @@ def resolve(requirements=None,
...
+        tmp_chroot = '{}.{}'.format(chroot, uuid4().hex)
+        try:
+          install_wheel(wheel=wheel_file_path,
+                        target=tmp_chroot,
+                        compile=compile,
+                        overwrite=True,
+                        cache=cache,
+                        interpreter=interpreter)
+        except PipError as e:
+          raise Untranslateable('Failed to install {}:\n\t{}'.format(wheel_file_path, str(e)))
```
Eric-Arellano: The `str` has no effect. `{}` automatically calls `str()`.

```
In [3]: try:
   ...:     raise ValueError("doom")
   ...: except ValueError as e:
   ...:     print("{}".format(e))
   ...:
   ...:
doom
```

```
In [4]: try:
   ...:     raise ValueError("doom")
   ...: except ValueError as e:
   ...:     print("{}".format(str(e)))
   ...:
   ...:
doom
```

Also, consider using `!r` for the exception. I still believe that more information is useful, but not going to fight very hard for that one because I think you have a point that there's a downside.

pex/resolver.py
```diff
@@ -205,57 +207,64 @@ def resolve(requirements=None,
...
+        os.rename(tmp_chroot, chroot)
+
+      environment = Environment(search_path=[chroot])
+      for dist_project_name in environment:
+        resolved_distributions.extend(environment[dist_project_name])
```
Eric-Arellano: Should this go into the below `TRACER.timed` block? It seems this is only setup for the environment markers code to work properly.

pex/resolver.py
jsirois: Will do on the str - I still completely disagree with the repr thing.

pex/resolver.py
jsirois: > Nit: `in` makes it sound like you are installing it in the chroot directory.

I am actually. The chroot directory is like a mini-`site-packages` that will contain exactly one distribution after this operation, the wheel file we install there (and which is exploded + other things by pip in the process). To be super clear hopefully:

```
$ rm -rf ~/.pex
$ python -mpex requests -- -c ''
$ tree -d -L 2 ~/.pex/build/chroots/
/home/jsirois/.pex/build/chroots/
├── certifi-2019.9.11-py2.py3-none-any.whl
│   ├── certifi
│   └── certifi-2019.9.11.dist-info
├── chardet-3.0.4-py2.py3-none-any.whl
│   ├── bin
│   ├── chardet
│   └── chardet-3.0.4.dist-info
├── idna-2.8-py2.py3-none-any.whl
│   ├── idna
│   └── idna-2.8.dist-info
├── requests-2.22.0-py2.py3-none-any.whl
│   ├── requests
│   └── requests-2.22.0.dist-info
└── urllib3-1.25.7-py2.py3-none-any.whl
    ├── urllib3
    └── urllib3-1.25.7.dist-info

16 directories
```

In that example, taking one `chroot` of `~/.pex/build/chroots/requests-2.22.0-py2.py3-none-any.whl`, you'll see that installed in it is exactly one distribution, `requests-2.22.0`.

jsirois: Perhaps it's confusing that the chroot dir basename is the zipped wheel basename? That is a hack that allows PEXEnvironment to activate the correct set of wheels at runtime as explained in the comment up above.

pex/resolver.py
jsirois: Not easily since its part of this loop. Also, its used to get the `pkg_resources.Distribution` object which is part of the return type of the function: `pex.resolver.ResolvedDistribution.distribution`

pex/resolver.py
Eric-Arellano: What confused me is thinking of `chroot` as one singular thing and not realizing you are referring to each wheel as having its own chroot. I was conceptualizing `chroot` to be the parent folder `/home/jsirois/.pex/build/chroots/`. 

Renaming `chroot` to `wheel_chroot` would have helped to figure that out.

PR #821, opened 2019-12-06
Unify PEX buildtime and runtime wheel caches.

### LINKED ISSUE #820: Unify resolve and runtime wheel caches.
The discussion in #789 and continued in the [associated design doc](https://docs.google.com/document/d/1B_g0Ofs8aQsJtrePPR1PCtSAKgBG1o59AhS_NwfFnbI/edit?usp=sharing) led to some spelunking and the realization that Pex currently has two distinct caches for wheels. One for the buildtime resolve of the wheel and one for the runtime extraction of the wheel from zipped pex files. By unifying these caches, we save disk space on machines where we both build and run wheels.

More interesting, apropos to #789, we gain the ability to populate the runtime wheel cache from a local build operation due to unification. For zipped pexes, this allows those pexes to fully resolve their requirements from the resolve-populated cache and thus not actually ever need any of those requirement wheels embedded in the zipped pex.

jsirois:
Previously these caches were seperate. Downloaded wheels and sdists were
cached in `~/.pex/build` and wheels unzipped from zipped pexes at
runtime were cached to `~/.pex/install`.

Now the caches are unified by the resolver such that any wheel installs
performed by it can be seen by zipped PEXes on the same machine when
they go to potentially unzip wheel distributions stored within at PEX
boot time.

N.B.: The cache is not unified in the other direction. If a zipped PEX
is executed on the same machine a PEX build resolve later happens on,
any intersecting wheels will be re-downloaded, built and installed by
the resolve finally unifying the caches from that point forward.

Fixes #820

### DIFF
--- a/pex/common.py
+++ b/pex/common.py
@@ -15,6 +15,7 @@ import threading
 import time
 import zipfile
 from collections import defaultdict
+from contextlib import contextmanager
 from datetime import datetime
 from uuid import uuid4
 
@@ -217,19 +218,84 @@ def safe_sleep(seconds):
       current_time = time.time()
 
 
-def rename_if_empty(src, dest, allowable_errors=(errno.EEXIST, errno.ENOTEMPTY)):
-  """Rename `src` to `dest` using `os.rename()`.
+class AtomicDirectory(object):
+  def __init__(self, target_dir):
+    self._target_dir = target_dir
+    self._work_dir = '{}.{}'.format(target_dir, uuid4().hex)
 
-  If an `OSError` with errno in `allowable_errors` is encountered during the rename, the `dest`
-  dir is left unchanged and the `src` directory will simply be removed.
+  @property
+  def work_dir(self):
+    return self._work_dir
+
+  @property
+  def target_dir(self):
+    return self._target_dir
+
+  @property
+  def is_finalized(self):
+    return os.path.exists(self._target_dir)
+
+  def finalize(self, source=None):
+    """Rename `work_dir` to `target_dir` using `os.rename()`.
+
+    :param str source: An optional source offset into the `work_dir`` to use for the atomic
+                       update of `target_dir`. By default the whole `work_dir` is used.
+
+    If a race is lost and `target_dir` already exists, the `target_dir` dir is left unchanged and
+    the `work_dir` directory will simply be removed.
+    """
+    if self.is_finalized:
+      return
+
+    source = os.path.join(self._work_dir, source) if source else self._work_dir
+    try:
+      # Perform an atomic rename.
+      #
+      # Per the docs: https://docs.python.org/2.7/library/os.html#os.rename
+      #
+      #   The operation may fail on some Unix flavors if src and dst are on different filesystems.
+      #   If successful, the renaming will be an atomic operation (this is a POSIX requirement).
+      #
+      # We have satisfied the single filesystem constraint by arranging the `work_dir` to be a
+      # sibling of the `target_dir`.
+      os.rename(source, self._target_dir)
+    except OSError as e:
+      if e.errno not in (errno.EEXIST, errno.ENOTEMPTY):
+        raise e
+    finally:
+      self.cleanup()
+
+  def cleanup(self):
+    safe_rmtree(self._work_dir)
+
+
+@contextmanager
+def atomic_directory(target_dir, source=None):
+  """A context manager that yields a new empty work directory path it will move to `target_dir`.
+
+  :param str target_dir: The target directory to atomically update.
+  :param str source: An optional source offset into the work directory to use for the atomic update
+                     of the target directory. By default the whole work directory is used.
+
+  If the `target_dir` already exists the enclosed block will be yielded `None` to signal there is
+  no work to do.
+
+  If the enclosed block fails the `target_dir` will be undisturbed.
+
+  The new work directory will be cleaned up regardless of whether or not the enclosed block
+  succeeds.
   """
+  atomic_dir = AtomicDirectory(target_dir=target_dir)
+  if atomic_dir.is_finalized:
+    yield None
+    return
+
+  safe_mkdir(atomic_dir.work_dir)
   try:
-    os.rename(src, dest)
-  except OSError as e:
-    if e.errno in allowable_errors:
-      safe_rmtree(src)
-    else:
-      raise
+    yield atomic_dir.work_dir
+    atomic_dir.finalize(source=source)
+  finally:
+    atomic_dir.cleanup()
 
 
 def chmod_plus_x(path):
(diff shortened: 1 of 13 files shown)


jsirois:
Reviewers, feel free to ignore the second commit unless it's of particular interest to you or otherwise request it be reverted. It is fairly unrelated to the main change except that it spells out the motivating case and that may be worth preserving. I'm at your command here.

pex/common.py
```diff
@@ -217,19 +218,84 @@ def safe_sleep(seconds):
...
-    os.rename(src, dest)
-  except OSError as e:
-    if e.errno in allowable_errors:
-      safe_rmtree(src)
-    else:
-      raise
+    yield atomic_dir.work_dir
+    atomic_dir.finalize(source=source)
+  finally:
+    atomic_dir.cleanup()
```
cosmicexplorer: Calling `.cleanup()` in here and in `.finalize()` bugs me a little, but I don't think I see a better way to write it.

pex/pex_builder.py
```diff
@@ -325,8 +302,10 @@ def add_distribution(self, dist, dist_name=None):
 
     if os.path.isdir(dist.location):
       dist_hash = self._add_dist_dir(dist.location, dist_name)
+    elif dist.location.endswith('.whl'):
+      dist_hash = self._add_dist_wheel_file(dist.location, dist_name)
     else:
-      dist_hash = self._add_dist_zip(dist.location, dist_name)
+      raise self.InvalidDistribution('Unsupported distribution type: {}'.format(dist))
```
cosmicexplorer:
```suggestion
      raise self.InvalidDistribution('Unsupported distribution type: {}. pex can accept dist dirs and wheels.'.format(dist))
```

pex/common.py
```diff
@@ -217,19 +218,84 @@ def safe_sleep(seconds):
...
+      # Per the docs: https://docs.python.org/2.7/library/os.html#os.rename
+      #
+      #   The operation may fail on some Unix flavors if src and dst are on different filesystems.
+      #   If successful, the renaming will be an atomic operation (this is a POSIX requirement).
+      #
+      # We have satisfied the single filesystem constraint by arranging the `work_dir` to be a
+      # sibling of the `target_dir`.
+      os.rename(source, self._target_dir)
+    except OSError as e:
+      if e.errno not in (errno.EEXIST, errno.ENOTEMPTY):
```
cosmicexplorer: This is super readable!

pex/environment.py
```diff
@@ -111,21 +109,14 @@ def _force_local(cls, pex_file, pex_info):
...
-    if not os.path.exists(explode_dir):
-      explode_tmp = explode_dir + '.' + uuid.uuid4().hex
-      with TRACER.timed('Unzipping %s' % pex_file):
-        try:
-          safe_mkdir(explode_tmp)
+    with atomic_directory(explode_dir) as explode_tmp:
+      if explode_tmp:
+        with TRACER.timed('Unzipping %s' % pex_file):
           with open_zip(pex_file) as pex_zip:
             pex_files = (x for x in pex_zip.namelist()
```
cosmicexplorer: Is this logic likely to ever be used elsewhere? Could it perhaps be moved into a separate method named something like `_iter_non_cache_bootstrap_entries(pex_zip)`? "No" is a great answer -- just curious. This code is pretty readable though.

pex/resolver.py
```diff
@@ -256,6 +231,63 @@ def install_chroot(self):
...
+    # pex: PYTHONPATH contains:
+    # pex:     /tmp/pex.pex
+    # pex:   * /usr/lib/python38.zip
+    # pex:     /usr/lib/python3.8
+    # pex:     /usr/lib/python3.8/lib-dynload
+    # pex:     /home/jsirois/.pex/installed_wheels/2a594cef34d2e9109bad847358d57ac4615f81f4/pex-2.0.2-py2.py3-none-any.whl  # noqa
+    # pex:   * /tmp/pex.pex/.bootstrap
+    # pex:   * - paths that do not exist or will be imported via zipimport
+    # pex.pex 2.0.2
+    #
```
cosmicexplorer: This is really useful and effective for understanding the change!

tests/test_unified_install_cache.py
```diff
@@ -0,0 +1,208 @@
...
+from pex.pex_info import PexInfo
+from pex.testing import run_pex_command
+
+
+@pytest.fixture(scope='module')
+def pex_project_dir():
+  return subprocess.check_output(['git', 'rev-parse', '--show-toplevel']).decode('utf-8').strip()
+
+
+def test_issues_789_demo(pex_project_dir):
```
cosmicexplorer: I really appreciate the way the pex codebase maps github issues to resolve to specific tests and comments in the code!

cosmicexplorer:
>  It is fairly unrelated to the main change except that it spells out the motivating case and that may be worth preserving.

I would not have a problem with it being reverted! I think it may be worth preserving if it doesn't seem too brittle, but I also plan to hit the ground running with this concept immediately in pants, so we may not need to keep around a reference for the technique. It may be helpful for other build tools wishing to employ the "dehydrated" pex technique however, at least in giving them a place to start from.

pex/environment.py
```diff
@@ -162,34 +153,14 @@ def _update_module_paths(cls, pex_file):
...
-        dist = DistributionHelper.distribution_from_path(os.path.join(pex, internal_dist_path))
-        with TRACER.timed('Caching %s' % dist):
-          newly_cached_distributions.append(
-            CacheHelper.cache_distribution(zf, internal_dist_path, cached_location))
-
-    return existing_cached_distributions, newly_cached_distributions
+  def _write_zipped_internal_cache(cls, zf, pex_info):
+    cached_distributions = []
+    for distribution_name, dist_digest in pex_info.distributions.items():
+      internal_dist_path = '/'.join([pex_info.internal_cache, distribution_name])
```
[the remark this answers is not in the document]
jsirois: Yes. See section 4.4.17 in the zip spec: https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT
```
    4.4.17 file name: (Variable)

       4.4.17.1 The name of the file, with optional relative path.
       The path stored MUST NOT contain a drive or
       device letter, or a leading slash.  All slashes
       MUST be forward slashes '/' as opposed to
       backwards slashes '\' for compatibility with Amiga
       and UNIX file systems etc.  If input came from standard
       input, there is no file name field.  
```

pex/environment.py
jsirois: Not likely, no.

pex/common.py
jsirois: Agreed on both counts. I also had noodles on that.

pex/environment.py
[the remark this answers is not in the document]
cosmicexplorer: Thank you!

PR #825, opened 2019-12-10
Cleanup extras_require.

jsirois:
We no longer use requests or CacheControl as optional enhancements, our
vendored pip does and it vendors those in.


jsirois:
A sanity check on the distribution we publish led to noticing this cruft.

Issue #806, opened 2019-11-22
Officially support python 3.8


jsirois:
We should update CI / tox / and our project metadata to reflect / test our support for python 3.8.

PR #826, opened 2019-12-12
Use flit to distribute pex.

### LINKED ISSUE #805: Move pex to a pyproject.toml build.
We should dogfood PEP-517 / PEP-518 support inherited through pip, set a good example and I think this can also allow removal of the trickery used currently to make the setup.py self-hosting using `pex.third_party`, although that in itself is a dogfood demonstration of the vendoring we'll lose.

### LINKED ISSUE #807: Switch distribution publishing to `twine`.
We currently use the `upload` command which is deprcated. Use the reccomended `twine` package / cli to do this instead.

jsirois:
This gets rid of the tricky setup.py setuptools and wheel bootstrapping
we were doing using pex vendored versions and gives us non-deprecated
publishing support. In the course of adding a pyproject.toml for flit,
also update pex distribution metadata to officially support Python 3.8
with 1 test tweak. CI for Python 3.8 will come in a follow-up change.

Fixes #805
Fixes #807

Work towards #806


setup.cfg
```diff
@@ -1,2 +0,0 @@
-[wheel]
-universal = 1
```
Eric-Arellano: Are we still marking this somehow? I don't see it in `pyproject.toml`.

jsirois: Its an implication of `requires-python = ">=2.7,<3.9,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*"`
If you run `flit build` you'll see a universal wheel gets plopped to dist/. There is no tox target for this, just `tox -e publish`, but that's safe to run since pypi does not allow over-writes.

pex/__init__.py
```diff
@@ -1,2 +1,10 @@
 # Copyright 2014 Pants project contributors (see CONTRIBUTORS.md).
 # Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+"""The PEX packaging toolchain."""  # N.B.: Flit uses this as our distribution description.
+
+from __future__ import absolute_import
+
+from pex.version import __version__ as __pex_version__
+
+__version__ = __pex_version__  # N.B.: Flit uses this as out distribution version.
```
codealchemy: s/out/our ?

PR #827, opened 2019-12-12
Kill `pkg_resources` finders monkey-patching.

jsirois:
We no longer used any of this in production except for one `PEXBuilder`
API which is converted to use `pex.pip` to install zipped wheels when
encountered so that they can be found by standard `pkg_resources`
finders.

Tests that use zipped wheels are similarly updated.

### DIFF
--- a/pex/pex_builder.py
+++ b/pex/pex_builder.py
@@ -227,12 +227,12 @@ class PEXBuilder(object):
       return
 
     # check if 'script' is an ordinary script
-    dist, _, _ = get_script_from_distributions(script, self._distributions)
-    if dist:
+    dist_script = get_script_from_distributions(script, self._distributions)
+    if dist_script:
       if self._pex_info.entry_point:
         raise self.InvalidExecutableSpecification('Cannot set both entry point and script of PEX!')
       self._pex_info.script = script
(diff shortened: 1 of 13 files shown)


pex/pex_builder.py
```diff
@@ -324,11 +324,18 @@ def add_dist_location(self, dist, name=None):
...
-    self.add_distribution(bdist, dist_name=name)
-    self.add_requirement(bdist.as_requirement())
+    dist_path = dist
+    if os.path.isfile(dist_path) and dist_path.endswith('.whl'):
+      dist_path = os.path.join(safe_mkdtemp(), os.path.basename(dist))
+      spawn_install_wheel(
+        wheel=dist,
+        install_dir=dist_path,
+        target=DistributionTarget.for_interpreter(self.interpreter)
+      ).wait()
```
Eric-Arellano: This part I'm unclear on. Is this an optimization or necessary to get things working?

jsirois: Its needed to kick our custom zipped-wheel finder to the curb. It's not an optimization since all wheels added to a pex are installed, not zipped (this goes back to a few commits after wheel support was added way back). Really, the zipped wheel finder support was only ~needed in tests and now that's killed too.

PR #828, opened 2019-12-12
Fix pex resolution to respect --ignore-errors.

jsirois:
Previously, pip might warn in red that versions were mismatched but
we would get no failure until pex runtime resolution. Improve pex
runtime resolution error messages and implement pex buildtime
resolution sanity checks when --no-ignore-errors.

At buildtime we now get:
```
$ python -mpex wheel pantsbuild.pants==1.24.0.dev2
Failed to resolve compatible distributions:
1: pantsbuild.pants==1.24.0.dev2 requires wheel==0.31.1 but wheel==0.33.6 was resolved
```

And the corresponding runtime error is now:
```
$ PEX_IGNORE_ERRORS=false python -mpex wheel pantsbuild.pants==1.24.0.dev2 --ignore-errors
Failed to execute PEX file. Needed linux_x86_64-cp-38-cp38 compatible dependencies for:
 1: wheel==0.31.1
    Required by:
      pantsbuild.pants==1.24.0.dev2
    But this pex only contains:
      wheel-0.33.6-py2.py3-none-any.whl
```

### DIFF
--- a/pex/package.py
+++ b/pex/package.py
@@ -15,12 +15,18 @@ def distribution_compatible(dist, supported_tags=None):
     by the platform in question; defaults to the current interpreter's supported tags.
   :returns: True if the distribution is compatible, False if it is unrecognized or incompatible.
   """
-  if supported_tags is None:
-    supported_tags = get_supported()
 
   filename, ext = os.path.splitext(os.path.basename(dist.location))
   if ext.lower() != '.whl':
-    return False
+    # This supports resolving pex's own vendored distributions which are vendored in directory
+    # directory with the project name (`pip/` for pip) and not the corresponding wheel name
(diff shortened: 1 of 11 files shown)


pex/package.py
```diff
@@ -15,12 +15,18 @@ def distribution_compatible(dist, supported_tags=None):
...
   """
-  if supported_tags is None:
-    supported_tags = get_supported()
 
   filename, ext = os.path.splitext(os.path.basename(dist.location))
   if ext.lower() != '.whl':
-    return False
+    # This supports resolving pex's own vendored distributions which are vendored in directory
+    # directory with the project name (`pip/` for pip) and not the corresponding wheel name
+    # (`pip-19.3.1-py2.py3-none-any.whl/` for pip). Pes only vendors universal wheels for all
```
Eric-Arellano: s/Pes/Pex

jsirois: Fixed.

Issue #823, opened 2019-12-10
Pex does not download foreign abi3 wheels correctly


jsirois:
A similar looking issue was encountered in #539 and fixed in #692, although that was for the Pex runtime resolve case; i.e.: resolving wheels stored inside the PEX at runtime. With the switch from the pex resolver to pip, we have a similar issue for downloads of foreign platforms. In particular, passing pip an explicit foreign platform, say: `--platform macosx_10_11_x86_64 --implementation cp --python-version 36 --abi cp36m`, we get the following calculation of supported tags:
```
$ PYTHONPATH=$PWD/pex/vendor/_vendored/pip python -mpip -vvv --disable-pip-version-check --isolated --exists-action i --no-cache-dir download --dest /tmp/tmp7991z44v/resolved_dists/macosx_10_11_x86_64-cp-36-cp36m --no-index --find-links /home/jsirois/dev/pantsbuild/pants/dist/deploy/wheels/pantsbuild.pants/879df20df1df1b9cf16e29cb4a2836f4831f23d9/1.24.0.dev1+git879df20d --find-links /home/jsirois/dev/pantsbuild/pants/dist/deploy/wheels/3rdparty/879df20df1df1b9cf16e29cb4a2836f4831f23d9/1.24.0.dev1+git879df20d --platform macosx_10_11_x86_64 --implementation cp --python-version 36 --abi cp36m --only-binary :all: cryptography>=1.9
>>> calculating supported for versions=36, noarch=False, platform=macosx_10_11_x86_64, impl=cp, abi=cp36m:...

	('cp36', 'cp36m', 'macosx_10_11_x86_64')
	('cp36', 'cp36m', 'macosx_10_11_intel')
	('cp36', 'cp36m', 'macosx_10_11_fat64')
	('cp36', 'cp36m', 'macosx_10_11_fat32')
	('cp36', 'cp36m', 'macosx_10_11_universal')
	('cp36', 'cp36m', 'macosx_10_10_x86_64')
	('cp36', 'cp36m', 'macosx_10_10_intel')
	('cp36', 'cp36m', 'macosx_10_10_fat64')
	('cp36', 'cp36m', 'macosx_10_10_fat32')
	('cp36', 'cp36m', 'macosx_10_10_universal')
	('cp36', 'cp36m', 'macosx_10_9_x86_64')
	('cp36', 'cp36m', 'macosx_10_9_intel')
	('cp36', 'cp36m', 'macosx_10_9_fat64')
[...]

PR #831, opened 2019-12-26
Use pypa/packaging.

### LINKED ISSUE #696: Eliminate copies of glibc.py and pep425tags.py
Currently we maintain minimally modified versions of our vendored setuptools own copies of pip's glibc.py and pep425tags.py files. We do this in order to support zero-dependency interpreter identification in PythonIdentity:
https://github.com/pantsbuild/pex/blob/b0fe036ab9762b6ee1ac4ed3e85ab49fedd91ef1/pex/interpreter.py#L16-L68

Ideally, we'd run interpreter identification using our vendored setuptools using something like `pex.third_party.expose('setuptools')`:
https://github.com/pantsbuild/pex/blob/b0fe036ab9762b6ee1ac4ed3e85ab49fedd91ef1/pex/third_party/__init__.py#L410-L426

jsirois:
Vendoring pypa/packaging allows us to eliminate our copies of
`pep425tags.py` and `glibc.py` from pip as well as eliminate custom
environment marker setup code. We accomplish the latter by re-working
the responsibility split between `PythonIdentity` and `Platform`.

`Platform` now just handles parsing extended platform specifiers and
all capability testing is moved into `PythonIdentity` which consults
`packaging` at runtime to determine the compatible tag set and
appropriate environment markers.

Work towards #823
Fixes #696


jsirois:
Reviewers, note that 64c0ad8 is the vendoring commit. The only manual edits there are two small ones to `pex/vendor/__{init,main}__,py`.

pex/environment.py
```diff
@@ -216,7 +211,20 @@ def _update_candidate_distributions(self, distribution_iter):
           self.add(dist)
 
   def can_add(self, dist):
-    return distribution_compatible(dist, self._supported_tags)
+    filename, ext = os.path.splitext(os.path.basename(dist.location))
+    if ext.lower() != '.whl':
```
Eric-Arellano: Could we tighten this check at all, e.g. `ext.lower() != .whl and os.path.isdir(..)`? Right now, something like `foo.egg` would return `True`.

Eric-Arellano: In fact, consider raising an exception if the `extension != whl and os.path.isfile()`. Perhaps that's overly defensive, but would be nice to fail eagerly if we make programming errors in the future that try passing a non-wheel name to `can_add`.

jsirois: I'll think on this one. The code was just moved, no logic changes. Although the logic you suggest is correct, it's exactly the sort of thing that led to zipped wheel finders hanging around so long (eliminated in #827). When you read a check like that it makes you think it can actually happen but it can't due to a larger systemic invariant (enforced by collusion between resolver and PEXBuilderat build time).

pex/interpreter.py
```diff
@@ -5,66 +5,21 @@
...
-    get_config_var,
-    get_abbr_impl,
-    get_abi_tag,
-    get_impl_version_info,
-    get_impl_ver
-  )
-
-  return ID_PY_TMPL.replace(b'__CODE__',
-                            b'\n\n'.join(getsource(func).encode('utf-8') for func in encodables))
+def encode_identity():
```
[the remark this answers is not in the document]
jsirois: @benjyw I missed this but will circle back before release.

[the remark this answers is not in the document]
jsirois: FIxed in #835 commit 5838c0f6.

PR #832, opened 2019-12-26
Upgrade vendored setuptools to 42.0.2.

jsirois:
The changelogs from 41.6.0 are here:
  https://setuptools.readthedocs.io/en/latest/history.html#v42-0-2

Of most note are the changes to setuptools to use pip when present
to implement `setup_requires`. The way we invoke pip in pex/pip.py
is compatible with this new feature; so we get more successful wheel
builds as a result.

See: https://github.com/pypa/setuptools/pull/1830


jsirois:
Reviewers, note that the only manual change here was the single line edit to `pex/vendor/__init__.py`.

PR #833, opened 2019-12-26
De-vendor pex just once per version.

jsirois:
Previously we would de-vendor pex once per CLI invocation in a new
temporary directory. Now we de-vendor once per pex version under the
pex root (~/.pex/isolated/...) by leveraging existing hashing and atomic
directory creation tools.


jsirois:
> Do you know if this results in a performance bump and by how much?

This results in a nominal performance decrease but it's in the wash - O(10ms) slower on non-spinning disk. The speed win will come when the content hash of the pex distribution is stored in a metadata file in the release.

PR #842, opened 2020-01-06
Parallelize interpreter discovery.

jsirois:
This speeds up both pex builds and pex executions on machines with
multiple interpreters and `--interpreter-constraint`s in-play.

Work towards #782.

### DIFF
--- a/pex/interpreter.py
+++ b/pex/interpreter.py
@@ -8,12 +8,14 @@ from __future__ import absolute_import
 import json
 import os
 import re
+import subprocess
 import sys
 from textwrap import dedent
 
 from pex import third_party
 from pex.compatibility import string
 from pex.executor import Executor
+from pex.jobs import Job, SpawnedJob, execute_parallel
 from pex.third_party.packaging import markers, tags
 from pex.third_party.pkg_resources import Distribution, Requirement
 from pex.tracer import TRACER
@@ -203,7 +205,7 @@ class PythonIdentity(object):
 
 
 class PythonInterpreter(object):
-  REGEXEN = (
+  _REGEXEN = (
     re.compile(r'jython$'),
 
     # NB: OSX ships python binaries named Python so we allow for capital-P.
@@ -220,7 +222,11 @@ class PythonInterpreter(object):
     re.compile(r'pypy-1.[0-9]$'),
   )
 
-  CACHE = {}  # memoize executable => PythonInterpreter
+  _PYTHON_INTERPRETER_BY_NORMALIZED_PATH = {}
+
+  @staticmethod
+  def _normalize_path(path):
+    return os.path.realpath(path)
 
   class Error(Exception): pass
   class IdentificationError(Error): pass
@@ -229,6 +235,10 @@ class PythonInterpreter(object):
   @classmethod
   def get(cls):
     return cls.from_binary(sys.executable)
+
+  @staticmethod
+  def _paths(paths=None):
+    return paths or os.getenv('PATH', '').split(os.pathsep)
 
   @classmethod
   def iter(cls, paths=None):
@@ -240,18 +250,11 @@ class PythonInterpreter(object):
     :param paths: The paths to look for python interpreters; by default the `PATH`.
     :type paths: list str
     """
-    if paths is None:
-      paths = os.getenv('PATH', '').split(os.pathsep)
-    for interpreter in cls._filter(cls._find(paths)):
-      yield interpreter
+    return cls._filter(cls._find(cls._paths(paths=paths)))
 
   @classmethod
   def all(cls, paths=None):
     return list(cls.iter(paths=paths))
-
-  @classmethod
-  def _from_binary_internal(cls):
-    return cls(sys.executable, PythonIdentity.get())
 
   @classmethod
   def _create_isolated_cmd(cls, binary, args=None, pythonpath=None, env=None):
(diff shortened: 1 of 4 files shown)


jsirois:
For example, building the pex pex on a machine with 14 interpreter PATH entries:

Pex 2.1.0:
```
time ~/downloads/pex-2.1.0 --interpreter-constraint="==2.7.*" --interpreter-constraint=">=3.5,<3.8" --no-compile --no-use-system-time --python-shebang="/usr/bin/env python" -o dist/pex -c pex .

real	1m15.685s
user	1m35.309s
sys	0m38.770s
```

HEAD:
```
$ time python -mpex --interpreter-constraint="==2.7.*" --interpreter-constraint=">=3.5,<3.8" --no-compile --no-use-system-time --python-shebang="/usr/bin/env python" -o dist/pex -c pex .

real	0m44.046s
user	0m53.727s
sys	0m17.989s
```

pex/interpreter.py
```diff
@@ -456,3 +477,41 @@ def __lt__(self, other):
 
   def __repr__(self):
     return '%s(%r, %r)' % (self.__class__.__name__, self._binary, self._identity)
+
+
+def spawn_python_job(args, env=None, interpreter=None, expose=None, **subprocess_kwargs):
```
jsirois: This was a move from `pex.jobs` below to avoid a circular dependency.

jsirois:
> Not for this change, but is there value in killing the other outstanding processes at that point?

You read that right. If we know there will only ever be one run of pex on a machine, then definitely yes - but given #856 and the unlikelihood of 1 run of pex ever the extra work will ~likely not got to waste and even if it does go to waste, it will only do so once. I think thats probably enough reason not to attempt the optimization, but making `execute_parallel` cancellable is a good idea generally and may come up again.

pex/jobs.py
```diff
@@ -62,55 +59,41 @@ def kill(self):
       if e.errno != errno.ESRCH:
         raise e
 
-  def _check_returncode(self):
+  def _check_returncode(self, stderr=None):
     if self._process.returncode != 0:
-      raise self.Error('Executing {} failed with {}'
-                       .format(' '.join(self._command), self._process.returncode))
+      msg = 'Executing {} failed with {}'.format(' '.join(self._command), self._process.returncode)
+      if stderr:
```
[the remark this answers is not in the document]
jsirois: As noted below we don't always provide stderr.

Agreed though that stderr being None vs `b''` is extra information, but only information about the code calling the API, not the subprocess execution itself. For code that wires up `stderr=subprocess.PIPE` we'll always at least get `b''`, for code that doesn't we'll get `None`. It seems to me in either case, outputting:
```
...

STDERR:
```
is not helpful debugging the subprocess execution that took place.

pex/jobs.py
```diff
@@ -62,55 +59,41 @@ def kill(self):
       if e.errno != errno.ESRCH:
         raise e
 
-  def _check_returncode(self):
+  def _check_returncode(self, stderr=None):
```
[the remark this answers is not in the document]
jsirois: The other call site is on line 38 above where this default is used.

pex/jobs.py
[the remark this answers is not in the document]
benjyw: Fair point!

pex/interpreter.py
```diff
@@ -441,18 +458,59 @@ def open_process(self, args=None, pythonpath=None, env=None, **kwargs):
...
-    return (self._binary, self._identity) == (other._binary, other._identity)
+    if type(other) is not type(self):
+      return NotImplemented
+    return self._tup() == other._tup()
 
   def __lt__(self, other):
-    if not isinstance(other, PythonInterpreter):
-      return False
+    if type(other) is not type(self):
+      return NotImplemented
```
Eric-Arellano: Great. Thank you for fixing this.

pex/jobs.py
```diff
@@ -230,14 +216,23 @@ def spawn_jobs():
 
     try:
       if isinstance(item, Exception):
-        error = item
+        if raise_type:
+          error = item
+        else:
+          # Otherwise, if we were given no `raise_type`, we skip over the input that raised.
```
Eric-Arellano: Check on understanding: how come we don't want to log when this case happens? If I'm reading correctly, this case will never result in the log line 236 triggering. 

If we're worried about too wordy of logging, consider setting it as debug.

pex/interpreter.py
```diff
@@ -320,67 +329,73 @@ def expand_path(cls, path):
   def from_env(cls, hashbang):
     """Resolve a PythonInterpreter as /usr/bin/env would.
 
-       :param hashbang: A string, e.g. "python3.3" representing some binary on the $PATH.
+    :param hashbang: A string, e.g. "python3.3" representing some binary on the $PATH.
+    :return: the first matching interpreter found or `None`.
```
Eric-Arellano: > the first matching interpreter

I want to ensure that interpreter discovery is deterministic, but I'm having a hard time reasoning about how we'd verify that. Could you please speak to your expectations around determinism of interpreter resolution when there are multiple matching candidates?

jsirois: Everything funnels to `_identify_interpreters` which deterministically iterates interpreters based on the order of an explicit `paths` sequence or else the sequence defined by the `PATH` env var. If that paths input is stable, the resulting iteration order of interpreters is too.

pex/jobs.py
jsirois: Just an oversight. Logging a spawn failure (OSError, IOError) seems just as valuable as logging Job.Error non-zero exits below. I'll fix this.

jsirois: Ah, right. I was probably being lazy as well. There is unuseful info here since we don't have the input that led to the exception, unlike the the case below. That said, That can be fixed too.

PR #843, opened 2020-01-06
Fix `tox -epackage` to create pex supporting 3.8.

jsirois:
This was missed in #826 but noticed creating #842 where pex.pex startup
was slower than expected due to not using the current (python 3.8)
interpreter and instead proceeding to find all interpreters and then
select the min.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -164,7 +164,7 @@ skip_install = true
 commands =
   python -m pex -v \
     --cache-dir {envtmpdir}/buildcache \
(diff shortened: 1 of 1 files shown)


jsirois:
> Looks good, other than a typo in title. s/-epackage/-e package.

Not a typo, I try to abide by the linux kernel commit formatting conventions which calls for a 50 column summary sentence and then 72 column body. `tox -epackage` is a valid command line and in fact the way I spell my tox invocations (no space).

PR #852, opened 2020-01-21
Handle EPERM when Linking

bradhowes:
Fallback to copy if there is an EPERM or EXDEV error when calling os.link()


jsirois:
What is the case where you hit EPERM when creating a link but don't hit EPERM when performing a copy instead?

jsirois:
OK - the CI failure is unrelated to your work. I'll get in a fix for that 1st though and then merge this change.

bradhowes:
Thanks for the doc to protected hardlinks. Our failure is exaclty that.

PR #856, opened 2020-01-22
Introduce an interpreter cache.

### LINKED ISSUE #782: Interpreter discovery and pyenv don't interact well

jsirois:
This speeds up interpreter discovery by a factor of ~10 on cache hits.
    
Fixes #782

### DIFF
--- a/pex/interpreter.py
+++ b/pex/interpreter.py
@@ -5,20 +5,25 @@
 
 from __future__ import absolute_import
 
+import hashlib
 import json
 import os
+import platform
 import re
 import subprocess
 import sys
 from textwrap import dedent
 
 from pex import third_party
+from pex.common import safe_rmtree
 from pex.compatibility import string
 from pex.executor import Executor
 from pex.jobs import Job, SpawnedJob, execute_parallel
 from pex.third_party.packaging import markers, tags
 from pex.third_party.pkg_resources import Distribution, Requirement
 from pex.tracer import TRACER
+from pex.util import CacheHelper
+from pex.variables import ENV
 
 
 class PythonIdentity(object):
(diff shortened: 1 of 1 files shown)


jsirois:
N.B.: This is based off #842 and will be rebased once that is submitted. The reviewable commit here is 00cf371.

pex/interpreter.py
```diff
@@ -290,32 +293,67 @@ def _execute(cls, binary, args=None, pythonpath=None, env=None, stdin_payload=No
...
       return cls(binary, PythonIdentity.decode(identity))
 
-    return SpawnedJob.stdout(job, result_func=create_interpreter)
+    # Part of the PythonInterpreter data are environment markers that depend on the current OS
+    # release. That data can change when the OS is upgraded but (some of) the installed interpreters
+    # remain the same. As such, include the OS in the hash structure for cached interpreters.
+    os_digest = hashlib.sha1()
+    for os_identifier in platform.release(), platform.version():
+      os_digest.update(os_identifier.encode('utf-8'))
+    os_hash = os_digest.hexdigest()
```
benjyw: Nice! This might save us from awkward, hard-to-find bugs later on, I would guess.

jsirois:
After ferreting out the unrelated pyenv issues here, I'm going to break off a seperate PR to address those then circle back here and rebase.

jsirois:
This is now ready for review. Tests in test_pex_bootstrapper should fail, but that is fixed by #860.

jsirois:
> Pex is also not taking the approach of Pants of trying to act like Python 2 is really Python 3 

... yeah, exactly. That certainly helps avoid alot of problems.

PR #863, opened 2020-01-24
Ensure Pex PEX contraints match pex wheel / sdist.

jsirois:
Previously we had to keep these values in sync and failed to do so often
enough leading to released Pex PEXes that would fail to run under
interpreters they should run under.

### DIFF
--- a/scripts/package.py
+++ b/scripts/package.py
@@ -1,0 +1,49 @@
+#!/usr/bin/env python3
+
+import subprocess
+import sys
+from pathlib import Path
+
+import pytoml as toml
+
+PROJECT_METADATA = Path('pyproject.toml')
+
+
+def python_requires() -> str:
+  project_metadata = toml.loads(PROJECT_METADATA.read_text())
+  return project_metadata['tool']['flit']['metadata']['requires-python'].strip()
+
+
+def build_pex_pex() -> None:
+  # NB: We do not include the subprocess extra (which would be spelled: `.[subprocess]`) since we
+  # would then produce a pex that would not be consumable by all python interpreters otherwise
+  # meeting `python_requires`; ie: we'd need to then come up with a deploy environment / deploy
+  # tooling, that built subprocess32 for linux cp27m, cp27mu, pypy, ... etc. Even with all the work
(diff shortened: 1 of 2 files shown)


jsirois:
For example:
```
$ toml get pyproject.toml tool.flit.metadata.requires-python
">=2.7,<3.9,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*"
```

VS:
```
$ tox -epackage
$ unzip -qc dist/pex PEX-INFO | jq .interpreter_constraints
[
  ">=2.7,<3.9,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*"
]
```

tox.ini
```diff
@@ -160,18 +160,11 @@ commands =
 
 [testenv:package]
 skip_install = true
-# TODO(John Sirois): Include subprocess32 when #415, #658 and #694 are resolved.
+basepython = python3
```
jsirois: NB: We now can actually include it and things work properly when the pex is run with pypy or python2.7 or python3; however the overly verbose comment in `build_pex_pex` trumps.

scripts/package.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env python3
```
Eric-Arellano: Thanks for making this a Python script

scripts/package.py
```diff
@@ -0,0 +1,49 @@
...
+    '-c', 'pex',
+    pex_requirement
+  ]
+  subprocess.run(args, check=True)
+
+
+if __name__ == '__main__':
+  if not PROJECT_METADATA.is_file():
+    print('This script must be run from the root of the Pex repo.', file=sys.stderr)
+    sys.exit(1)
```
Eric-Arellano: Nit: I think it's more idiomatic to `raise SystemExit("This script must be run from the root of the Pex repo.")`. Should do the same thing.

jsirois: I have literally never seen that.

jsirois: FWIW, github search shows 1 million hits for it vs 77 million for sys.exit.

PR #873, opened 2020-02-06
Improve hermeticity of vendoring.

benjyw:
- Force specific versions of setuptools and wheel to be used
  when running pip during vendoring.

- Force the vendor-check test to run on python 3.8.

These address issues in our tests, where re-vendoring pip caused
slight differences in its metadata, causing the `git diff` check to
fail.

### DIFF
--- a/pex/vendor/__main__.py
+++ b/pex/vendor/__main__.py
@@ -189,8 +189,22 @@ class VendorizeError(Exception):
 
 def vendorize(root_dir, vendor_specs, prefix):
   for vendor_spec in vendor_specs:
-    cmd = ['pip', 'install', '--upgrade', '--no-compile', '--target', vendor_spec.target_dir,
-           vendor_spec.requirement]
+    # NB: We set --no-build-isolation to prevent pip from installing the requirements listed in
+    # its [build-system] config in its pyproject.toml.
+    #
+    # Those requirements are (currently) the unpinned ["setuptools", "wheel"], which will cause pip
+    # to use the latest version of those packages.  This is a hermeticity problem: re-vendoring a
+    # package at a later time may yield a different result.  At the very least the result will
+    # differ in the version embedded in the WHEEL metadata file, which causes issues with our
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -136,12 +136,18 @@ deps =
   ansicolors==1.1.8
   pip==19.3.1
   redbaron==0.9.2
+  setuptools==42.0.2
+  wheel==0.33.6
```
Eric-Arellano: To confirm, this is where we pin the version of `setuptools` and `wheel` that we're going to ship with the Pex binary?

benjyw: No, this is where we pin the versions used when tox runs `pex.vendor`. The versions we actually vendor are in `__init__.py`.

Eric-Arellano: Might be worth adding a comment that these need to be in sync.

jsirois: They don't need to be in sync, they just need to be pinned versions that work with the version of pip used above which itself does not need to match our vendored pip version as it currently does not. That said, see above - I think we can eliminate this change by pruning more metadata files post install.

pex/vendor/__main__.py
```diff
@@ -189,8 +189,22 @@ class VendorizeError(Exception):
 
 def vendorize(root_dir, vendor_specs, prefix):
   for vendor_spec in vendor_specs:
-    cmd = ['pip', 'install', '--upgrade', '--no-compile', '--target', vendor_spec.target_dir,
-           vendor_spec.requirement]
+    # NB: We set --no-build-isolation to prevent pip from installing the requirements listed in
```
jsirois: I'm going to see if this can be reverted bythe expedient of removing the install metadata we don't need (`entry_points.txt`, `INSTALLER`,  and `WHEEL`). That should allow reverting these changes as well as the patch to pip that was needed in https://github.com/pantsbuild/pip/commit/e2553a16ba0e0272a915ceff848a29d27a088e83

PR #874, opened 2020-02-07
Check expected pex invocation failure reason in tests.

benjyw:
In some tests we just checked that expected invocation failures
happened. These tests would continue to pass if the failure reason
changed from the one we expected.

This change makes the tests check the invocation's stderr for the
expected failure output.

It adds a TODO to verify whether one of those tests is intended to
fail for the reason checked (a wheel compatibility issue).

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -6,6 +6,7 @@ import functools
 import glob
 import json
 import os
+import re
 import subprocess
 import sys
 from contextlib import contextmanager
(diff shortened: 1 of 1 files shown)


benjyw:
Note - the tests are now expected to fail. Will add a commit that fixes or bypasses them before sending for review.

tests/test_integration.py
```diff
@@ -1170,8 +1174,13 @@ def test_setup_python_multiple_direct_markers():
 
     py2_only_program = [pex, '-c', 'import subprocess32']
 
-    with pytest.raises(subprocess.CalledProcessError):
-      subprocess.check_call(py2_only_program, env=make_env(PATH=os.path.dirname(py36_interpreter)))
+    with pytest.raises(subprocess.CalledProcessError) as err:
+      subprocess.check_output(py2_only_program, stderr=subprocess.STDOUT,
+                              env=make_env(PATH=os.path.dirname(py36_interpreter)))
+    # TODO: Is the intention of this test that this should fail on platform incompatibility?
```
jsirois: It is not, this exposes a bug: #899. Thank you!

There is also a very similar bug described in #898 found as a result.

PR #877, opened 2020-02-07
Fix the resolve check in the presence of platform constraints.

benjyw:
If a requirement is resolved vacuously due to a platform constraint
(e.g., 'pathlib2; python_version < "3"' when we're on Python 3) then
it is not an error to not have a resolved dist.

### DIFF
--- a/pex/resolver.py
+++ b/pex/resolver.py
@@ -602,9 +602,11 @@ class ResolveRequest(object):
           )
         else:
           resolved_dist = resolved_requirement_dist.distribution
-          # If a requirement is resolved vacuously due to a platform constraint (e.g., 'pathlib2; python_version < "3"'
-          # when we're on Python 3) then resolved_dist will be None, and this is not an error.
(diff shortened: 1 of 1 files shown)


pex/resolver.py
```diff
@@ -601,7 +602,10 @@ def _check_resolve(self, resolved_distributions):
           )
         else:
           resolved_dist = resolved_requirement_dist.distribution
-          if resolved_dist not in requirement:
+          # If a requirement is resolved vacuously due to a platform constraint (e.g., 'pathlib2; python_version < "3"'
+          # when we're on Python 3) then resolved_dist will be None, and this is not an error.
```
Eric-Arellano: I think you might have meant to use a comma:
```suggestion
          # when we're on Python 3, then resolved_dist will be None, and this is not an error.
```

benjyw: No, this closes the paren opened before "e.g.,"

PR #881, opened 2020-02-12
Use pkg_resources when isolating a pex code chroot.

benjyw:
Previously we used filesystem operations, which assumes
that the pex code is available on the filesystem. This is
not the case, e.g., when bootstrapping a zipped pex file.

This change replaces the os.walk-based implementation
with one that uses equivalent pkg_resources operations.

### DIFF
--- a/pex/third_party/__init__.py
+++ b/pex/third_party/__init__.py
@@ -7,6 +7,7 @@ import contextlib
 import importlib
 import os
 import re
+import shutil
 import sys
 import zipfile
 from collections import OrderedDict, namedtuple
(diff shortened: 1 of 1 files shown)


pex/third_party/__init__.py
```diff
@@ -338,24 +339,31 @@ def isolated():
...
     from pex.util import CacheHelper
     from pex.variables import ENV
+    from pex.third_party.pkg_resources import resource_isdir, resource_listdir, resource_stream
+
+    module = 'pex'
+    def recursive_copy(srcdir, dstdir):
+      os.mkdir(dstdir)
+      for entry_name in resource_listdir(module, srcdir):
+        # NB: Resource path components are always separated by /, on all systems.
+        src_entry = '%s/%s' % (srcdir, entry_name) if srcdir else entry_name
```
Eric-Arellano: Nit: this can use `.format()`. Pex uses both styles, but I think John has recently been converting things to `.format()` because it's easier to read.

PR #885, opened 2020-02-12
Add a clear_cache() method to PythonInterpreter.

benjyw:
Useful in the Pants interpreter selection tests.

### DIFF
--- a/pex/interpreter.py
+++ b/pex/interpreter.py
@@ -319,6 +319,11 @@ class PythonInterpreter(object):
   INTERP_INFO_FILE = 'INTERP-INFO'
 
   @classmethod
+  def clear_cache(cls):
+    interpreter_cache_dir = os.path.join(ENV.PEX_ROOT, 'interpreters')
+    safe_rmtree(interpreter_cache_dir)
+
+  @classmethod
   def _spawn_from_binary_external(cls, binary):
     def create_interpreter(stdout):
       identity = stdout.decode('utf-8').strip()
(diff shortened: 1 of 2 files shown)


pex/interpreter.py
```diff
@@ -405,7 +410,7 @@ def hashbang_matches(fn):
   def _spawn_from_binary(cls, binary):
     normalized_binary = cls._normalize_path(binary)
 
-    # N.B.: The CACHE is written as the last step in PythonInterpreter instance initialization.
+    # N.B.: The cache is written as the last step in PythonInterpreter instance initialization.
     cached_interpreter = cls._PYTHON_INTERPRETER_BY_NORMALIZED_PATH.get(normalized_binary)
```
jsirois: The Pants test currently clears the in-process `CACHE` (now `_PYTHON_INTERPRETER_BY_NORMALIZED_PATH`). This PR only clears the new on-disk cache but leaves the in-process cache in place. If this change actually works over in the Pants test, that implies clearing of the in-process cache was not needed. If that's true, then I think this method could be dropped and the Pants test could just use an ephemeral PEX_ROOT since that is a public API unlike `os.path.join(ENV.PEX_ROOT, 'interpreters')` which this PR properly tries to hide since its an internal API.

jsirois: ... the context being this new `clear_cache` method is undocumented and unused in the pex codebase making it ripe for reaping on any surface read.

benjyw: Ah, I hadn't yet actually modified Pants to call this method, but I made a Pants change in my branch, https://github.com/benjyw/pants/commit/78f9f4c909706ca32bec19940f33a6c8019d7d2d) that mirrors this change and can now be edited appropriately to call this method. I'm pretty sure this was necessary to get tests passing.


And now that I think of it, https://github.com/benjyw/pants/commit/4064806551cc86b450058136f8c39999a7bce72a  can now also be modified appropriately.

jsirois: I fixed benjyw/pants@4064806 and benjyw/pants@78f9f4c can be dropped. There turned out there was no need to cache clear at all since the interpreters propped up by the test are ephemeral tmpfile fake interpreters. Each one has a unique path and each one is eliminated in test tear-down. With your fix to check for the interpreter binary existing in #886, the cache is self-cleared. I'll circle back and revert this as a result since Pants didn't end up using it.

Issue #672, opened 2019-02-14
`PythonInterpreter.from_binary` has a leaky API


jsirois:
Right now, when handed a binary that does not exist, pex.executor.Executor.ExecutableNotFound is raised:
https://github.com/pantsbuild/pex/blob/5f19c916fad3631806a31e025a90333273c90e3d/pex/interpreter.py#L293-L302

This both leaks the internal Executor API used to carry out the method and is not documented.

benjyw:
We can close this now, I think?

jsirois:
Yup. Thanks for that fix.

PR #886, opened 2020-02-12
Error eagerly if an interpreter binary doesn't exist.

benjyw:
The exception class already existed, it just wasn't used.

Due to caching, the interpreter may or may not fail
in find_binary(), so the calling code might fail at some
later stage.

This change causes us to fail eagerly and consistently if
the binary doesn't exist on disk.


benjyw:
Fixes #672

jsirois:
LGTM

Issue #892, opened 2020-02-19
Pex resolver API is not completely hermetic to PYTHONPATH.


jsirois:
The `pex.resolver` module delegates work to pip by running a pip PEX in a subprocess. If the interpreter being resolved for is not the current interpreter and the current interpreter has portions of its stdlib exported on the `PYTHONPATH` and those portions of the stdlib are incompatible with the interpreter being resolved for, the bootstrap of the pip PEX will fail before it can scrub away the `PYTHONPATH`.

A minimal repro:
```
$ python3.8 <<EOF
> import os
> import sys
> 
> 
> # This puts python3.8 stdlib on PYTHONPATH.
> os.environ['PYTHONPATH'] = os.pathsep.join(sys.path)
> 
> 
> from pex import resolver
> from pex.interpreter import PythonInterpreter
> 
> 
> def find_python36():
>   for interp in PythonInterpreter.iter():
>     if interp.version[:2] == (3, 6):
>       return interp
>   raise RuntimeError('This test requires python3.6 be on the PATH.')
> 
> 
> result = resolver.resolve(requirements=['packaging==19.2'], interpreter=find_python36())
> print('Resolved: {}'.format(result))
> EOF
Fatal Python error: Py_Initialize: can't initialize sys standard streams
Traceback (most recent call last):
  File "/usr/lib/python3.8/abc.py", line 64, in <module>
ModuleNotFoundError: No module named '_abc'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.8/io.py", line 52, in <module>
  File "/usr/lib/python3.8/abc.py", line 68, in <module>
  File "/usr/lib/python3.8/_py_abc.py", line 35
    def __new__(mcls, name, bases, namespace, /, **kwargs):
                                              ^
SyntaxError: invalid syntax
Traceback (most recent call last):
  File "<stdin>", line 20, in <module>
  File "/home/jsirois/dev/pantsbuild/pex/pex/resolver.py", line 701, in resolve
    return list(resolve_request.resolve_distributions(ignore_errors=ignore_errors))
  File "/home/jsirois/dev/pantsbuild/pex/pex/resolver.py", line 498, in resolve_distributions
    for resolve_result in self._run_parallel(inputs=to_resolve,
  File "/home/jsirois/dev/pantsbuild/pex/pex/resolver.py", line 366, in _run_parallel
    for result in execute_parallel(
[...]

PR #894, opened 2020-02-19
Fixup resolve debug rendering.

jsirois:


jsirois:
Found debugging #892

jsirois:
The test failure here is a known OSX-only python 3.8 only failure on master tracked by #896 . I'll fix that and circle back.

PR #895, opened 2020-02-19
Fix Pex resolver API PYTHONPATH hermeticity.

jsirois:
As described in #892, when the Pex resolver API is used in-process, a
sufficiently antagonistic PYTHONPATH could lead to a failure to launch
the pip PEX we delegate resolving to. Fix this by scrubbing PYTHONPATH
from the pip PEX environment pre-launch.

Fixes #892

PR #897, opened 2020-02-19
Convert `bdist_pex` tests to explicit cmdclass.

### LINKED ISSUE #896: Master CI is broken for OSX python3.8 `tests/test_bdist_pex.py`
In short the tests in `tests/test_bdist_pex.py` try to run `python setup.py bdist_pex ...` but either the Pex codebase is not on the sys.path / PYTHONPATH or else it is but the distutils `bdist_pex` `Command` provded by Pex is not being registered / discovered.

Errors look like:
```
tests/test_bdist_pex.py::test_entry_points_dict usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
   or: setup.py --help [cmd1 cmd2 ...]
   or: setup.py --help-commands
   or: setup.py cmd --help
error: invalid command 'bdist_pex'
```

jsirois:
Previously the test relied upon the tox venv interpreter (which
implicitly builds and installs the pex wheel) being the interpreter that
executed the `python setup.py bdist_pex ...` command. This assumption
could be violated when running under `pyenv` which OSX CI shards do.
Instead, explicitly add pex to the `PYTHONPATH` of the interpreter and
let distutils know about `bdist_pex` via `--command-packages`.

Fixes #896

### DIFF
--- a/pex/interpreter.py
+++ b/pex/interpreter.py
@@ -549,7 +549,14 @@ class PythonInterpreter(object):
     )
 
 
-def spawn_python_job(args, env=None, interpreter=None, expose=None, **subprocess_kwargs):
+def spawn_python_job(
+  args,
+  env=None,
+  interpreter=None,
+  expose=None,
+  pythonpath=None,
+  **subprocess_kwargs
+):
   """Spawns a python job.
 
   :param args: The arguments to pass to the python interpreter.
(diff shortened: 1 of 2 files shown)


jsirois:
I'm going to submit to unblock landing #894 and #895.

PR #900, opened 2020-02-20
readme: adding a TOC

jhermann:
Preview at https://github.com/jhermann/pex/tree/patch-1#readme

BTW, the GH "Website" setting should point to https://pex.readthedocs.io, see the top of the preview page.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,5 +1,6 @@
+***
 PEX
-===
+***
 .. image:: https://travis-ci.org/pantsbuild/pex.svg?branch=master
     :target: https://travis-ci.org/pantsbuild/pex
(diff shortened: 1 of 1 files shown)


jsirois:
> BTW, the GH "Website" setting should point to https://pex.readthedocs.io, see the top of the preview page.

Ideally, yes. Unfortunately the current maintainers of PEX do not have permissions to push docs there and so they are very stale. The API section refers to long dead modules. As such, actively pointing there in a ~high profile place doesn't seem like a great idea until / if the staleness gets sorted.

jsirois:
Thanks!

PR #901, opened 2020-02-20
Show pex-root from env as default in help output

jhermann:
If PEX_ROOT is set, its value is shown in the --pex-root help string

```console
$ export PEX_ROOT="$HOME/.cache/pex"
$ pex --help 2>&1 | grep .-root
  --pex-root=PEX_ROOT   Specify the pex root used in this invocation of pex. [Default: ~/.cache/pex]
```

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -483,7 +483,8 @@ def configure_clp():
       '--pex-root',
       dest='pex_root',
       default=ENV.PEX_ROOT,
-      help='Specify the pex root used in this invocation of pex. [Default: ~/.pex]'
+      help='Specify the pex root used in this invocation of pex. [Default: {}]'.format(
+          (ENV.PEX_ROOT or '~/.pex').replace(os.path.expanduser('~'), '~')),
   )
 
   parser.add_option(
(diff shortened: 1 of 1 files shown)


pex/bin/pex.py
```diff
@@ -483,7 +483,8 @@ def configure_clp():
       '--pex-root',
       dest='pex_root',
       default=ENV.PEX_ROOT,
-      help='Specify the pex root used in this invocation of pex. [Default: ~/.pex]'
+      help='Specify the pex root used in this invocation of pex. [Default: {}]'.format(
```
jsirois: If you read through `Variables.PEX_ROOT`:
https://github.com/pantsbuild/pex/blob/de5097f7bb9eaba14a8998c24f6fecc7bd90af70/pex/variables.py#L265-L272
And `Variables._get_path`:
https://github.com/pantsbuild/pex/blob/de5097f7bb9eaba14a8998c24f6fecc7bd90af70/pex/variables.py#L96-L99
You'll see this can all just be replace by ENV.PEX_ROOT and that's reassuring since the default above is just that.

As such this can all be DRYed by just replacing `~/.pex` in the original help string with `%default`.

jhermann: Fixed, please squash the PR.

jsirois:
Thanks Jürgen.

PR #908, opened 2020-02-25
Silence pip warnings about Python 2.7.

jsirois:
Pex will want to warn about deprecated support for Python 2.7 some day,
but it is not Pips job to do so on our behalf. Our support cycles are
independent since we vendor in a given version of Pip for ~as long as
we want / need.

### DIFF
--- a/pex/pip.py
+++ b/pex/pip.py
@@ -55,7 +55,7 @@ class Pip(object):
 
   def _spawn_pip_isolated(self, args, cache=None, interpreter=None):
     pip_args = [
-      # We vendor the version of pip we wan't so pip should never check for updates.
+      # We vendor the version of pip we want so pip should never check for updates.
       '--disable-pip-version-check',
 
       # Don't read pip configuration files like `~/.config/pip/pip.conf`.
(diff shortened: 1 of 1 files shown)


jsirois:
N.B.: This option only became available with the recent re-vendor of Pip up to 20.0+.

pex/pip.py
```diff
@@ -54,7 +54,16 @@ def __init__(self, pip_pex_path):
     self._pip_pex_path = pip_pex_path
 
   def _spawn_pip_isolated(self, args, cache=None, interpreter=None):
-    pip_args = ['--disable-pip-version-check', '--isolated', '--exists-action', 'i']
+    pip_args = [
```
jsirois: This PR jumps the gun a bit, #907 eliminates `--exists-action i`. I'll re-base this PR against master once that PR lands.

jsirois:
OK - rebased against #907 on master and ready for review.

pex/pip.py
```diff
@@ -54,7 +54,16 @@ def __init__(self, pip_pex_path):
     self._pip_pex_path = pip_pex_path
 
   def _spawn_pip_isolated(self, args, cache=None, interpreter=None):
-    pip_args = ['--disable-pip-version-check', '--isolated']
+    pip_args = [
+      # We vendor the version of pip we wan't so pip should never check for updates.
```
Eric-Arellano: Typo 
```suggestion
      # We vendor the version of pip we want so pip should never check for updates.
```

PR #924, opened 2020-03-19
Pin our tox version in CI for stability.

jsirois:


jsirois:
#923 is getting burned by a lack of this pinning due to tox-dev/tox#1538.

PR #928, opened 2020-03-23
Improve interpreter not found error messages.

### LINKED ISSUE #922: Improve error message when interpreters cannot be found
Current:

```
pex --interpreter-constraint='CPython>=3.8.*'
Could not find compatible interpreter
```

This is a fine message for trivial constraints, especially when using Pex directly through the CLI. But, this is hard to interpret when there are complex constraints and when Pex is used programmatically, such as Pants passing often quite complex constraints to Pex.

An ideal message will
a) output all requested constraints
b) identify which interpreters were recognized by Pex
c) offer suggestions to resolve the issue

See https://github.com/pantsbuild/pants/pull/7628 for possible inspiration.

jsirois:
Centralize messaging in a new exception type and incorporate this in the
CLI and at PEX runtime. API users can use the new structured exception
type as they wish.

Fixes #922

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -15,7 +15,7 @@ from textwrap import TextWrapper
 
 from pex.common import die, safe_delete, safe_mkdtemp
 from pex.interpreter import PythonInterpreter
-from pex.interpreter_constraints import UnsatisfiableInterpreterConstraints, validate_constraints
+from pex.interpreter_constraints import UnsatisfiableInterpreterConstraintsError, validate_constraints
 from pex.jobs import DEFAULT_MAX_JOBS
 from pex.pex import PEX
 from pex.pex_bootstrapper import iter_compatible_interpreters
@@ -530,7 +530,7 @@ def build_pex(reqs, options):
         pex_python_path = None
       try:
         interpreters = list(iter_compatible_interpreters(pex_python_path, constraints))
-      except UnsatisfiableInterpreterConstraints as e:
+      except UnsatisfiableInterpreterConstraintsError as e:
         die(e.create_message('Could not find a compatible interpreter.'), CANNOT_SETUP_INTERPRETER)
 
   try:
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -15,7 +15,7 @@ from textwrap import TextWrapper
 
 from pex.common import die, safe_delete, safe_mkdtemp
 from pex.interpreter import PythonInterpreter
-from pex.interpreter_constraints import validate_constraints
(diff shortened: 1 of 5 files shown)


jsirois:
Eric suggested a footer of:
```
Ensure that the targeted Python version is installed and discoverable, or tweak your constraints to work with one of the discovered interpreters.
```

I feel like this crosses the line from empathy to coddling but am happy to re-consider. The messaging currently looks like so.

At buildtime:
```
$ python -m pex --interpreter-constraint='>=3.9'
Could not find a compatible interpreter.

   /usr/bin/python3.8 CPython==3.8.2
   /opt/pypy/bin/pypy PyPy==2.7.13
   /usr/bin/python2.7 CPython==2.7.17
   /usr/bin/python3.5 CPython==3.5.9
  /usr/bin/python3.6m CPython==3.6.10
   /usr/bin/python3.7 CPython==3.7.7

None were compatible with the requested constraints:
  >=3.9
```

And at runtime:
```
Failed to find a compatible PEX_PYTHON=/home/jsirois/dev/pantsbuild/jsirois-pex/.pyenv_test/versions/2.7.15/bin/python2.7.

	/home/jsirois/dev/pantsbuild/jsirois-pex/.pyenv_test/versions/2.7.15/bin/python2.7 CPython==2.7.15

None were compatible with the requested constraints:
	>3,<3.8
```

jsirois:
> One of the things I think every great error message includes is a translation of "okay, how do I actually fix this?" 

I'd agree if there was tool specific info to give, Here there is not, just the standard can of worms. I'm claiming the footer you suggest is mainly empty of content since the verbage above notes the interpreter constraints the user supplied, meaning the second phrase is purely repetitive. Likewise we've listed the paths we were able to test, and saying "discoverable" isn't helpful to the newby to a command line and its redundant to a non newbie. For the newbie, to actually be helpful you'd need to run down the possible issues, which include:
1. The interpreter is in the list, how super confsuing!: Permissions are likely the issue and that requires a book to explain usefully.
2. The interpreter is not in the list but it is installed!: Its common on images to have broken installations of python. Toolchain's own internal images have officially installed pythons that are broken.

We agree on providing next steps. We don't agree when to draw the line.

Eric-Arellano:
Well reasoned. I agree with your judgment here to leave off the footer.

Eric-Arellano:
Well reasoned. I agree with your judgment here to leave off the footer.

benjyw:
+1 to not having the footer. Our users are engineers, they understand the environment in which this is running better than we do. Laying out the problem in excruciating detail, as this change does, is the useful bit.

PR #929, opened 2020-03-23
Sanitize PEX_ROOT handling.

### LINKED ISSUE #746: Problem running PEX files with restricted permissions

### LINKED ISSUE #816: Prevent create files on PEX ROOT as root when pex file is run with sudo
Run a pex with sudo the first time:

```
sudo my-pex.pex
```
It creates files under in my HOME `~/.pex/install/` with root as owner, 

```
$ ls -la ~/.pex/install/
total 0
drwxr-xr-x 1 root   root   3818 Nov 28 17:03 .
drwxrwxr-x 1 user user   14 Nov 28 17:03 ..
drwxr-xr-x 1 root   root     76 Nov 28 17:03 attrs-19.3.0-py2.py3-none-any.whl.e91da08a7091c46772d87f0c961a63ac85d4a810
drwxr-xr-x 1 root   root     82 Nov 28 17:03 requests-2.22.0-py2.py3-none-any.whl.9523c112783d316b50f5bfd5d4bcad40736336aa
.........
```

Later when the same pex is run without sudo it produces `PermissionError`

```
  File "/usr/bin/my-pex.pex/.bootstrap/pex/pex.py", line 395, in execute
  File "/usr/bin/my-pex.pex/.bootstrap/pex/pex.py", line 93, in _activate
  File "/usr/bin/my-pex.pex/.bootstrap/pex/environment.py", line 249, in activate
  File "/usr/bin/my-pex.pex/.bootstrap/pex/environment.py", line 357, in _activate
  File "/usr/bin/my-pex.pex/.bootstrap/pex/environment.py", line 238, in _update_candidate_distributions
  File "/usr/bin/my-pex.pex/.bootstrap/pex/environment.py", line 203, in _load_internal_cache
  File "/usr/bin/my-pex.pex/.bootstrap/pex/environment.py", line 190, in _write_zipped_internal_cache
  File "/usr/bin/my-pex.pex/.bootstrap/pex/util.py", line 166, in cache_distribution
  File "/usr/lib/python3.7/zipfile.py", line 1599, in extract
    return self._extract_member(member, path, pwd)
  File "/usr/bin/my-pex.pex/.bootstrap/pex/common.py", line 121, in _extract_member
  File "/usr/lib/python3.7/zipfile.py", line 1662, in _extract_member
    os.makedirs(upperdirs)
  File "/usr/lib/python3.7/os.py", line 211, in makedirs
    makedirs(head, exist_ok=exist_ok)
  File "/usr/lib/python3.7/os.py", line 211, in makedirs
    makedirs(head, exist_ok=exist_ok)
  File "/usr/lib/python3.7/os.py", line 221, in makedirs
    mkdir(name, mode)
PermissionError: [Errno 13] Permission denied: '/home/user/.pex/install/pyrsistent-0.15.6-cp37-cp37m-linux_x86_64.whl.5e52dba7dab3ea7ab9594ad73326d31baf150835.a8407da79a3145ad8d2ed53f8f654852'
```
Maybe this files can be created with the real user ownership

### LINKED ISSUE #926: The pex cache (PEX_ROOT) is not always controllable.
In general, there is cruft here, including the confusing existence of `--cache-dir` _and_ `--pex-root` at build-time, in addition to `PEX_ROOT` at both build-time and run-time, and, finally `PexInfo.pex_root` at run-time. The behavior of PEX with combinations of these settings is not currently well defined, but there is also at least one concrete case where the pex root cannot be controlled, in a Pex PEX. This case is of importance for Pants v2 which uses PEXed Pex to build PEXes.

On PEX startup, all `PEX_` env vars are stripped before handing off control to the underlying application:
https://github.com/pantsbuild/pex/blob/73ce2f4b4fdc1fe3ad76779c06f0b9a25f443384/pex/pex.py#L416-L419
https://github.com/pantsbuild/pex/blob/73ce2f4b4fdc1fe3ad76779c06f0b9a25f443384/pex/pex.py#L45-L54

When that application is itself Pex, this takes away `PEX_ROOT` as a means to control the pex root for PEXed Pex build-time. Whether or not some affordance should be made for passing `PEX_` env vars to Pex when PEXed is tracked in #925. As a result, any direct uses of `PEX_ROOT` instead of `--pex-root` or `--cache-dir` will use the default `PEX_ROOT` of `~/.pex`. We have three of these:
1. https://github.com/pantsbuild/pex/blob/128c03de866c3f0d667a97524aade5b05faad8a5/pex/interpreter.py#L337
2. https://github.com/pantsbuild/pex/blob/bf66b72a9aa467b1bfd6ae6f07d1b679733aaaa5/pex/third_party/__init__.py#L362
3. https://github.com/pantsbuild/pex/blob/91159b34a8c88f8788df41aba2d09cd2938a67f8/pex/pip.py#L56-L82

Item 3 is easy to fix to respect `cache` if passed - this appears to be a bug. Items 1 & 2 are a bit harder to fix:
1. PythonInterpreter is used at both build-time and run-time and even in the public API.
2. The `isolated` API is used by on-demand `pex.third_party` imports.

Fixing item 3 may be enough to solve the 1 known controllability issue, but it would be good to solve for all 3 cases by addressing #925 eventually.

jsirois:
In the past, portions of the pex cache could be controlled individually
but this was no longer the case in practice nor was it desirable. Unify
all cache handling under the PEX_ROOT and deprecate `--cache-dir` to
codify the new single pex cache as the PEX_ROOT.

Also add best-effort support for always finding a a viable PEX_ROOT even
when the configured value is not writeable. In conjunction with
`--runtime-pex-root` this provides for both a Pex CLI and created PEXes
that should always work wrt having a viable PEX_ROOT but still display a
warning if a suboptimal alternate had to be used.

Fixes #746
Fixes #816
Fixes #926

### DIFF
--- a/pex/variables.py
+++ b/pex/variables.py
@@ -277,7 +277,7 @@ class Variables(object):
       return None
 
     if not can_write_dir(pex_root):
-      tmp_root = safe_mkdtemp()
+      tmp_root = os.path.realpath(safe_mkdtemp())
       pex_warnings.warn('PEX_ROOT is configured as {pex_root} but that path is un-writeable, '
                         'falling back to a temporary PEX_ROOT of {tmp_root} which will hurt '
                         'performance.'.format(pex_root=pex_root, tmp_root=tmp_root))
--- a/pex/variables.py
+++ b/pex/variables.py
@@ -273,7 +273,7 @@ class Variables(object):
     pex_root = self._get_path('PEX_ROOT', default=os.path.expanduser('~/.pex'))
 
     if pex_root is None:
-      # PEX_ROOT is no set and we're running in stripped_defaults mode.
+      # PEX_ROOT is not set and we're running in stripped_defaults mode.
       return None
 
     if not can_write_dir(pex_root):
--- a/pex/variables.py
+++ b/pex/variables.py
@@ -10,7 +10,8 @@ import os
 import sys
 from contextlib import contextmanager
 
-from pex.common import die
+from pex import pex_warnings
+from pex.common import can_write_dir, die, safe_mkdtemp
 
 __all__ = ('ENV', 'Variables')
 
@@ -269,7 +270,20 @@ class Variables(object):
     The directory location for PEX to cache any dependencies and code.  PEX must write
     not-zip-safe eggs and all wheels to disk in order to activate them.  Default: ~/.pex
     """
-    return self._get_path('PEX_ROOT', default=os.path.expanduser('~/.pex'))
+    pex_root = self._get_path('PEX_ROOT', default=os.path.expanduser('~/.pex'))
+
+    if pex_root is None:
+      # PEX_ROOT is no set and we're running in stripped_defaults mode.
+      return None
+
+    if not can_write_dir(pex_root):
+      tmp_root = safe_mkdtemp()
+      pex_warnings.warn('PEX_ROOT is configured as {pex_root} but that path is un-writeable, '
+                        'falling back to a temporary PEX_ROOT of {tmp_root} which will hurt '
+                        'performance.'.format(pex_root=pex_root, tmp_root=tmp_root))
+      pex_root = self._environ['PEX_ROOT'] = tmp_root
+
+    return pex_root
 
   @property
   def PEX_PATH(self):
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -13,6 +13,7 @@ import sys
 from optparse import OptionGroup, OptionParser, OptionValueError
 from textwrap import TextWrapper
 
+from pex import pex_warnings
 from pex.common import die, safe_delete, safe_mkdtemp
 from pex.interpreter import PythonInterpreter
 from pex.interpreter_constraints import (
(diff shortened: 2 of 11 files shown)


pex/variables.py
```diff
@@ -269,7 +270,20 @@ def PEX_ROOT(self):
...
+
+    if pex_root is None:
+      # PEX_ROOT is not set and we're running in stripped_defaults mode.
+      return None
+
+    if not can_write_dir(pex_root):
+      tmp_root = os.path.realpath(safe_mkdtemp())
+      pex_warnings.warn('PEX_ROOT is configured as {pex_root} but that path is un-writeable, '
+                        'falling back to a temporary PEX_ROOT of {tmp_root} which will hurt '
+                        'performance.'.format(pex_root=pex_root, tmp_root=tmp_root))
```
Eric-Arellano: Very nice. This is both a helpful mechanism and a thoughtful warning message.

PR #932, opened 2020-03-24
Fix `PEX_*` env stripping and allow turning off.

### LINKED ISSUE #180: Preserve PEX_ env variables after calling pex
https://github.com/pantsbuild/pex/blob/master/pex/pex.py#L48

Currently we delete all env variables from os.environ that start with PEX_.  If pex is called from another process as a library this will override the parent scripts env variables.  If the env variables are reset back to their original state at the end of execution the behavior would be less confusing when using pex as a library.

For example in pants if we try to override the PEX_ROOT every time we call pex to build a target PEX_ROOT is reset.

### LINKED ISSUE #925: Consider passing through `PEX_` env vars to PEXed Pex.
On PEX startup, all `PEX_` env vars are stripped before handing off control to the underlying application:
https://github.com/pantsbuild/pex/blob/73ce2f4b4fdc1fe3ad76779c06f0b9a25f443384/pex/pex.py#L416-L419
https://github.com/pantsbuild/pex/blob/73ce2f4b4fdc1fe3ad76779c06f0b9a25f443384/pex/pex.py#L45-L54

When that application is Pex itself (Someone uses the PEXed Pex we release for example), this leads to surprising behavior where Pex does not respond to the same environment variable control it does respond to when installed as a wheel.

It seems like this should be fixed by either passing through the `PEX_` env vars in this special case (which is likely hard to detect) - or else we should revisit stripping `PEX_` env vars wholesale.

jsirois:
Previously, `PEX_*` environment variable stripping incorrectly stripped
`PEX_*` environment variables in host process when the `PEX.run` API was
used. This is fixed and an option to not strip pex environment variables
at all is added to `PexInfo` / the Pex CLI. This new option is used to
publish a Pex PEX that supports control via environment variables.

Fixes #180
Fixes #925
Work towards #926

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -302,8 +302,16 @@ def configure_clp_pex_options(parser):
       '--runtime-pex-root',
       dest='runtime_pex_root',
       default=None,
-      help='Specify the pex root to be used in the generated .pex file. [Default: ~/.pex]'
-  )
+      help='Specify the pex root to be used in the generated .pex file. [Default: ~/.pex]')
+
+  group.add_option(
+      '--strip-pex-env', '--no-strip-pex-env',
+      dest='strip_pex_env',
+      default=True,
+      action='callback',
+      callback=parse_bool,
+      help='Strip all `PEX_*` environment variables used to control the PEX runtime before handing '
+           'off control to the PEX entrypoint. [Default: %default]')
 
   parser.add_option_group(group)
 
@@ -576,6 +584,7 @@ def build_pex(reqs, options):
   pex_info.emit_warnings = options.emit_warnings
   pex_info.inherit_path = options.inherit_path
   pex_info.pex_root = options.runtime_pex_root
+  pex_info.strip_pex_env = options.strip_pex_env
   if options.interpreter_constraint:
     for ic in options.interpreter_constraint:
       pex_builder.add_interpreter_constraint(ic)
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -310,8 +310,10 @@ def configure_clp_pex_options(parser):
       default=True,
       action='callback',
       callback=parse_bool,
-      help='Strip all `PEX_*` environment variables used to control the PEX runtime before handing '
-           'off control to the PEX entrypoint. [Default: %default]')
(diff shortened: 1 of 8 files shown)


jsirois:
@wisechengyi this will allow Twitter and Pants should they decide to do so, to mark their Pants PEXes as --no-strip-pex-env to allow control of PEX subprocesses they execute via `PEX_*` env vars. I think this is relevant to work you were trying to do during the Pants Pex 2.x upgrade recently.

pex/bin/pex.py
```diff
@@ -302,8 +302,16 @@ def configure_clp_pex_options(parser):
...
+      help='Specify the pex root to be used in the generated .pex file. [Default: ~/.pex]')
+
+  group.add_option(
+      '--strip-pex-env', '--no-strip-pex-env',
+      dest='strip_pex_env',
+      default=True,
+      action='callback',
+      callback=parse_bool,
+      help='Strip all `PEX_*` environment variables used to control the PEX runtime before handing '
+           'off control to the PEX entrypoint. [Default: %default]')
```
Eric-Arellano: It'd be helpful to put the description you wrote elsewhere here: 

> You might want to set this to `False` if this PEX executes other PEXes or the Pex CLI itself and you want the executed PEX to be controlled via PEX environment variables.

Until reading that explanation, it was not obvious to me why I would want this to be False.

jsirois: It is one of those things you don't really want to make obvious, because if you need it you know you need it - aka Pants advanced concept, but fair enough, will fix.

Eric-Arellano: Saying exactly that would help, i.e. saying it's "advanced".

> This is an advanced feature only possibly needed when using PEX to executing other PEXes or the PEX CLI itself.

jsirois:
Hrm, note to self - don't use GitHub app.

Eric-Arellano:
> don't use GitHub app.

How come? I've fount it's not as great as I first thought. I often fall back to the mobile website. I'm hoping it will keep getting better.

jsirois:
> How come?

You don't get a chance to edit the commit message like you do on the web, desktop or mobile: 1c6ae63b

I have never liked the concept of a merge button anyhow.

Issue #930, opened 2020-03-23
Running a trivial pex file could be faster


illicitonion:
Currently, the overhead for a no-deps pex on my machine is about 440ms:

```
$ mkdir -p trivial && echo 'print("Hello")' > trivial/trivial.py && python3.6 -m pex -D trivial -m trivial -o trivial.pex && multitime -n 20 python3.6 trivial/trivial.py >/dev/null && multitime -n 20 python3.6 ./trivial.pex >/dev/null
===> multitime results
1: python3.6 trivial/trivial.py
            Mean        Std.Dev.    Min         Median      Max
real        0.035       0.005       0.029       0.035       0.054       
user        0.025       0.003       0.020       0.025       0.037       
sys         0.008       0.001       0.006       0.008       0.012       
===> multitime results
1: python3.6 ./trivial.pex
            Mean        Std.Dev.    Min         Median      Max
real        0.474       0.020       0.441       0.471       0.528       
user        0.418       0.017       0.391       0.415       0.466       
sys         0.052       0.003       0.047       0.052       0.058       
```

I'm going to spend a bit of time trying to get this to <100ms.

illicitonion:
Next thing I'm looking at:
Importing `packaging.requirements` is non-trivially expensive - it looks like it takes about 75ms:
```
$ multitime -n 20 python3.6 -c 'import packaging'
===> multitime results
1: python3.6 -c "import packaging"
            Mean        Std.Dev.    Min         Median      Max
real        0.035       0.002       0.030       0.036       0.038       
user        0.024       0.001       0.021       0.024       0.025       
sys         0.008       0.001       0.006       0.008       0.009       

$ multitime -n 20 python3.6 -c 'import packaging.requirements'
===> multitime results
1: python3.6 -c "import packaging.requirements"
            Mean        Std.Dev.    Min         Median      Max
real        0.108       0.006       0.095       0.110       0.115       
user        0.091       0.005       0.080       0.093       0.097       
sys         0.013       0.001       0.012       0.013       0.015       
```

Because of transitive vendoring, we import it twice when running a pexfile, once from our vendored `packaging` and again from `pkg_resources`'s vendored `packaging`.

The reason it's so slow is that it imports the `pyparsing` module, which itself takes about 50ms to import each time. We have four copies of it checked in:
```
pex$ find . -name pyparsing.py
./pex/vendor/_vendored/packaging/pyparsing.py
[...]

PR #936, opened 2020-03-27
Use one version of transitively vendored deps

illicitonion:
This is kind of hacky. It probably shouldn't be a hard-coded list in
`__main__.py`.

It's also not guaranteed to work - it relies on us picking good versions
to vendor which are compatibile with all of the transitive Avendored
versions.

We also don't vendor all of the transitive deps, because the rest don't
seem to be used by our codepaths. This is a bit fragile.

Finally, the last rewrite, `pex.third_party.six` ->
`pex.vendor._vendored.six.six` shouldn't be necessary. If the approach
seems reasonable, I'll dig into why it's needed to work, and resolve
that.

Performance difference:
Before:
```
pex$ mkdir -p trivial && echo 'print("Hello")' > trivial/trivial.py && python3.6 -m pex -D trivial -m trivial -o trivial.pex && multitime -n 50 python3.6 ./trivial.pex >/dev/null
===> multitime results
1: python3.6 ./trivial.pex
real        0.466       0.024       0.434       0.466       0.594    
user        0.409       0.022       0.380       0.408       0.527          
sys         0.054       0.002       0.049       0.054       0.063 
```

After:
```
pex$ mkdir -p trivial && echo 'print("Hello")' > trivial/trivial.py && python3.6 -m pex -D trivial -m trivial -o trivial.pex && multitime -n 50 python3.6 ./trivial.pex >/dev/null
===> multitime results
1: python3.6 ./trivial.pex
real        0.405       0.025       0.368       0.402       0.519    
user        0.355       0.023       0.325       0.352       0.457    
sys         0.046       0.003       0.041       0.046       0.057  
```

Relates to #930


pex/vendor/__main__.py
```diff
@@ -105,14 +105,33 @@ def _modify__import__calls(self, red_baron):  # noqa: We want __import__ as part
...
+  def _revendor(self, value):
+    """Re-vendors transitively vendored imports to our vendored versions.
+
+    Some of our transitively vendored imports are incredibly expensive to import. Rather than
+    importing multiple versions of them, we canonicalize all of these imports to the version we
+    ourselves vendor, so that we only have to process the imports one time.
+    """
+    return (
+      value
+        .replace("pex.third_party.setuptools.extern", "pex.third_party")
```
jsirois: It would be really great if `_revendor` or some higher level that passed it in this configuration data also nuked the revendered packages. That would reduce deployed Pex PEX and Pex .whl sizes and perhaps be less confusing. Even better would be nuking and dropping in a small auto-generated README explaining the package was nuked and replaced with `VendorSpec.pinned('six', '1.14.0')` for example.

illicitonion:
> Stale?

Yeah realistically if I revive this at some point it'll be effectively from scratch :) Closing for now

PR #948, opened 2020-04-06
Support for copying requirements from an existing pex.

benjyw:
This can greatly speed up repeatedly creating a pex when
no requirements have changed. A build tool (such as Pants)
can create a "requirements pex" that contains just a static
set of requirements, and build a final pex on top of that,
without having to re-run pip to resolve requirements.

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -497,6 +497,16 @@ def configure_clp():
            'times.')
 
   parser.add_option(
+    '--requirements-pex',
+    dest='requirements_pexes',
+    metavar='FILE',
+    default=[],
+    type=str,
+    action='append',
+    help='Add requirements from the given .pex file or unzipped pex directory.  This option can '
+         'be used multiple times.')
+
+  parser.add_option(
       '-v',
       dest='verbosity',
       default=0,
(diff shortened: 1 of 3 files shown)


pex/pex_builder.py
```diff
@@ -229,6 +229,35 @@ def add_interpreter_constraint(self, ic):
...
 
+  def add_from_requirements_pex(self, pex):
+    """Add requirements from an existing pex.
+
+    :param pex: The path to an existing .pex file or unzipped pex directory.
+    """
+    self._ensure_unfrozen('Adding from pex')
+    pex_info = PexInfo.from_pex(pex)
+
+    def add(location, dname, expected_dhash):
```
Eric-Arellano, line 240: Nit: it took me a minute to figure out whether `dname` was `dependency_name` or `distribution_name` or something else. Spelling it out as `distribution_name` will save future readers this time.

pex/bin/pex.py
```diff
@@ -496,6 +496,16 @@ def configure_clp():
...
            'times.')
 
+  parser.add_option(
+    '--requirements-pex',
+    dest='requirements_pexes',
+    metavar='FILE',
+    default=[],
+    type=str,
+    action='append',
+    help='Add requirements from the given .pex file or unzipped pex directory.  This option can '
```
jsirois, line 506: Even if this happens to work on an unzipped PEX, there is no way to create one from the CLI so this seems like a confusing bit to document in the context of a CLI option as opposed to the PEXBuilder API where it does make sense.

benjyw: Well, you can unzip a `.pex` file to get an unzipped pex directory... But I guess Pants now uses zipped `.pex` files exclusively, so maybe there's no need to document this.

Issue #964, opened 2020-05-09
Can't run PEX file when a dependency's wheel includes a build tag


luislew:
We recently encountered a runtime issue with a successfully built PEX file:
```
(nwpy-ml-clustering_linux-x86-64_cp3.7) app@8bceeea547ee:/srv/nerdwallet/nwpy-ml-clustering$ ./build/nwpy-ml-clustering.pex
Traceback (most recent call last):
  File "/srv/nerdwallet/nwpy-ml-clustering/build/nwpy-ml-clustering.pex/.bootstrap/pex/pex.py", line 391, in execute
  File "/srv/nerdwallet/nwpy-ml-clustering/build/nwpy-ml-clustering.pex/.bootstrap/pex/pex.py", line 89, in _activate
  File "/srv/nerdwallet/nwpy-ml-clustering/build/nwpy-ml-clustering.pex/.bootstrap/pex/environment.py", line 232, in activate
  File "/srv/nerdwallet/nwpy-ml-clustering/build/nwpy-ml-clustering.pex/.bootstrap/pex/environment.py", line 375, in _activate
  File "/srv/nerdwallet/nwpy-ml-clustering/build/nwpy-ml-clustering.pex/.bootstrap/pex/environment.py", line 209, in _update_candidate_distributions
  File "/srv/nerdwallet/nwpy-ml-clustering/build/nwpy-ml-clustering.pex/.bootstrap/pex/environment.py", line 227, in can_add
  File "/srv/nerdwallet/nwpy-ml-clustering/build/nwpy-ml-clustering.pex/.bootstrap/pex/vendor/_vendored/packaging/packaging/tags.py", line 75, in parse_tag
ValueError: too many values to unpack (expected 3)
```

We tracked this down to the following installed wheel: `'llvmlite-0.29.0-1-cp37-cp37m-manylinux1_x86_64.whl'`. The current wheel filename parsing code in `environment.py` assumes that there is no build tag: https://github.com/pantsbuild/pex/blob/916e61e04634c60d09ee25859c76ba4f27282e51/pex/environment.py#L223

As a result, it includes the build tag with `wheel_tags` (e.g. `1-cp37-cp37m-manylinux1_x86_64`) and `parse_tag` fails.

Per PEP 427, wheel filenames may include an optional build tag: https://www.python.org/dev/peps/pep-0427/#file-name-convention

---

[...]

Issue #950, opened 2020-04-07
Release 2.1.10


benjyw:
On the docket:
+ [x] Improve Pex packaging. (#961)
+ [x] Make the interpreter cache deterministic. (#960)
+ [x] Fix deprecation warning for `rU` mode (#956)
+ [x] Fix runtime resolve error message generation. (#955)
+ [x] Kill dead code. (#954)
+ [x] Many Pex tests fail under Python 2.7 in CI #967
+ [x] Add a `--local` mode for packaging the Pex PEX. #971
+ [x] Split Pex resolve API. (#970)
+ [x] Can't run PEX file when a dependency's wheel includes a build tag #964
+ [ ] Expose network configuration in pex options. #803

benjyw:
On the docket:
+ [ ] Top level environment markers are lost in built PEXes. #899
+ [ ] Requirement resolution errors should provide enough information to resolve conflicts. #943

PR #965, opened 2020-05-09
Fix wheel filename parsing in PEXEnvironment.can_add

luislew:
PEX issue: https://github.com/pantsbuild/pex/issues/964

This PR updates `PEXEnvironment.can_add` to split the whole wheel filename in order to gather up the trailing three tags. The previous implementation was overly naive and failed when a wheel filename included a (valid) build tag.

### DIFF
--- a/pex/environment.py
+++ b/pex/environment.py
@@ -11,8 +11,6 @@ import sys
 import zipfile
 from collections import OrderedDict, defaultdict
 
-from setuptools import wheel
-
 from pex import dist_metadata, pex_builder, pex_warnings
 from pex.bootstrap import Bootstrap
 from pex.common import atomic_directory, die, open_zip
@@ -23,6 +21,11 @@ from pex.third_party.packaging import tags
 from pex.third_party.pkg_resources import DistributionNotFound, Environment, Requirement, WorkingSet
 from pex.tracer import TRACER
 from pex.util import CacheHelper, DistributionHelper
+
+if "__PEX_UNVENDORED__" in __import__("os").environ:
+  from setuptools import wheel  # vendor:skip
+else:
+  from pex.third_party.setuptools import wheel
 
 
 def _import_pkg_resources():
--- a/pex/environment.py
+++ b/pex/environment.py
@@ -19,7 +19,7 @@ from pex.orderedset import OrderedSet
 from pex.platforms import Platform
 from pex.third_party.packaging import tags
 from pex.third_party.pkg_resources import DistributionNotFound, Environment, Requirement, WorkingSet
-from pex.third_party.setuptools.wheel import Wheel
+from pex.third_party.setuptools import wheel
 from pex.tracer import TRACER
 from pex.util import CacheHelper, DistributionHelper
 
@@ -221,7 +221,7 @@ class PEXEnvironment(Environment):
       return True
 
     try:
-      wheel = Wheel(filename)
+      wheel = wheel.Wheel(filename)
       wheel_tags = '-'.join([wheel.py_version, wheel.abi, wheel.platform])
     except ValueError:
       return False
--- a/pex/environment.py
+++ b/pex/environment.py
@@ -22,11 +22,6 @@ from pex.third_party.pkg_resources import DistributionNotFound, Environment, Req
 from pex.tracer import TRACER
 from pex.util import CacheHelper, DistributionHelper
 
-if "__PEX_UNVENDORED__" in __import__("os").environ:
-  from setuptools import wheel  # vendor:skip
-else:
-  from pex.third_party.setuptools import wheel
-
 
 def _import_pkg_resources():
   try:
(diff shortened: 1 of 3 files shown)


pex/environment.py
```diff
@@ -22,6 +22,11 @@
 from pex.tracer import TRACER
 from pex.util import CacheHelper, DistributionHelper
 
+if "__PEX_UNVENDORED__" in __import__("os").environ:
```
jsirois, line 25: The PEXEnvironment class is only used at runtime and at runtime a PEX only has access to the Pex code needed to bootstrap the runtime (scrub the sys.path). The bootstrap code is contained in `.bootstrap/` in the PEX and that's controlled here:
https://github.com/pantsbuild/pex/blob/a7d98116daedcacda4716ebc9ca831303f24b71f/pex/pex_builder.py#L436-L444

So you'd need to add `'wheel'` to the `root_module_names` list.

That said, I don't think this is the way to go. See below...

pex/environment.py
```diff
@@ -220,7 +225,8 @@ def can_add(self, dist):
       return True
 
     try:
-      wheel_name_, wheel_raw_version_, wheel_tags = filename.split('-', 2)
+      whl = wheel.Wheel(filename)
```
jsirois, line 228: Unfortunately, this reads in and hashes the full wheel file: https://github.com/pypa/wheel/blob/a51977075740fda01b2c0e983a79bfe753567219/src/wheel/wheelfile.py#L38-L79
We definitely don't want to do this for many wheels used in the ML world for example since those wheels are routinely massive. Although you could jump through some hoops here and set the mode to 'w' but never actually write, etc - that's fragile and getting a bit ugly.

How do you feel about just copying the proper regex from that code and using it here? Its a bit unsatisfying, but we pick up no new runtime dependency and we do less work. Its nt fragile since this is all spec'd in [PEP 427](https://www.python.org/dev/peps/pep-0427/#file-name-convention).

luislew: Yeah, we have lots of build code which uses `Wheel(filename)`, so I thought it was an easy solution here, but as I learned that's **not** the simple path. Even a regex is pretty unnecessary when all we need are the trailing 3 parts of the wheel filename.

I'm simplified the solution to (essentially):
```
wheel_tags = '-'.join(filename.split('-')[-3])
```

tests/test_environment.py
```diff
@@ -348,3 +349,12 @@ def test_present_non_empty_namespace_packages_metadata_does_warn():
 
 def test_present_but_empty_namespace_packages_metadata_does_not_warn():
   assert_namespace_packages_warning('pycodestyle', '2.5.0', expected_warning=False)
+
+
+def test_can_add_handles_wheel_with_build_tag():
+  # The actual PEXEnviroment is irrelevant; all we need is an instance
+  interpreter = PythonInterpreter.get()
```
jsirois, line 356: You don't need the interpreter, just `PexInfo.default()` will do. That said, if you did realy set up a controlled interpreter you could test `can_add` returns `True` which would be a bit better test (tests both that we don't raise like before, and that we parse out the right tags and match as well). For that, you can use the fixture higher in this file:
https://github.com/pantsbuild/pex/blob/85f156548e3ac0d4d9c03246815a4dd0a67a31ed/tests/test_environment.py#L175-L178

And:
```python
def test_can_add_handles_wheel_with_build_tag(python_35_interpreter):
  pex_environment = PEXEnvironment(
    pex="",
    pex_info=PexInfo.default(python_35_interpreter),
    interpreter=python_35_interpreter
  )
  wheel_with_build_tag = 'llvmlite-0.29.0-1-cp35-cp35m-manylinux1_x86_64.whl'
  assert pex_enviroment.can_add(Distribution(wheel_with_build_tag))
```

luislew:
> Thanks a bunch! This will go out this weekend in 2.1.10 tracked in #950.

Awesome, thanks!

pex/environment.py
```diff
@@ -219,11 +219,10 @@ def can_add(self, dist):
...
       return True
 
-    try:
-      wheel_name_, wheel_raw_version_, wheel_tags = filename.split('-', 2)
-    except ValueError:
-      return False
-
+    # Wheel filename format: https://www.python.org/dev/peps/pep-0427/#file-name-convention
+    # `{distribution}-{version}(-{build tag})?-{python tag}-{abi tag}-{platform tag}.whl`
+    wheel_parts = filename[:-4].split('-')
```
jsirois, line 224: This is the red in CI - filename from above already has `'.whl'` stripped.

luislew: 🤦 that's why! i was testing in console and missed that

jsirois: I helped out with the last road bump here.

jsirois:
The remaining error looks unrelated to this work. I'll try to fix that independently - master has outstanding errors of the same form in the same OSX shards - and then circle back and rebase this change with that fix to get green.

luislew:
> The remaining error looks unrelated to this work. I'll try to fix that independently - master has outstanding errors of the same form in the same OSX shards - and then circle back and rebase this change with that fix to get green.

🙌🎉

jsirois:
The CI issue is not fully figured out, but the root cause is and that can be side-stepped. Tracking in #967.

jsirois:
CI is green despite the Github UI still showing pending: https://travis-ci.com/github/pantsbuild/pex/builds/165260311

Going to merge.

Issue #969, opened 2020-05-11
Merge resolve and resolve_multi


jsirois:
These functions are both part of the informal public API. We should deprecate passing a single interpreter and a single platform and just support passing optional lists for each as done in `resolve_multi`. We should also probably pick the name `'resolve'` for the merged function, essentially deleting the current `resolve` and replacing it with the current `resolve_multi` renamed to `resolve`.

Resolving this issue is probably a good time to take up #584 and set up proper deprecation, then removal.
1. [ ] Create a deprecation policy doc.
2. [ ] Implement a deprecation system - emit pex_warning messages as needed.
3. [ ] Change `resolve` to support `None`, singles, or lists for `interpreter` and `platform` and deprecate both calling `resolve` with singles and calling `resolve_multi`.
4. [ ] When the deprecation cycle is up remove `resolve` and rename `resolve_multi` to `resolve`.

This deprecation should probably use version 3.0.0 as the boundary to conform to Semver.

jsirois:
Additionally it would be good to circle back to the exposed datatypes and exception types and thoroughly document them and check their args as part of this resolve API cleanup and the API formalization umbrelled by #584.

jsirois:
Since Pex has moved away from suporting any API besides the CLI, this can now be done without all the above prep work needed to support breaking CLI API changes.

PR #968, opened 2020-05-10
Constrain the virtualenv version used by tox.

jsirois:
Pin virtualenv just under 20.0.5 which is broken for Pex tests when run
under Python 2.7.

Along the way, tidy up tox tool installation by doing it in its own
venv.

Also document, update and align all Python versions used in the various
shards while killing use of pyenv shims.

Finally, kill no longer needed (and unwittingly unused) OSX specific env
vars intended to help pyenv compile ssl libs.

Fixes #967

Issue #967, opened 2020-05-10
Many Pex tests fail under Python 2.7 in CI


jsirois:
I was finally able to reproduce this locally by re-creating my tox venv which, as a side effect, upgraded virtualenv. The latter is the problem.

The errors look like:
```
$ tox -epy27 -- -vvsk test_entry_points_dict
.package create: /home/jsirois/dev/pantsbuild/jsirois-pex/.tox/.package
.package installdeps: flit_core >=2,<3
py27 create: /home/jsirois/dev/pantsbuild/jsirois-pex/.tox/py27
py27 installdeps: pytest==4.6.6, mock==3.0.5
py27 inst: /home/jsirois/dev/pantsbuild/jsirois-pex/.tox/.tmp/package/1/pex-2.1.9.tar.gz
py27 installed: DEPRECATION: Python 2.7 reached the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 is no longer maintained. A future version of pip will drop support for Python 2.7. More details about Python 2 support in pip, can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support,atomicwrites==1.4.0,attrs==19.3.0,configparser==4.0.2,contextlib2==0.6.0.post1,funcsigs==1.0.2,importlib-metadata==1.6.0,mock==3.0.5,more-itertools==5.0.0,packaging==20.3,pathlib2==2.3.5,pex==2.1.9,pluggy==0.13.1,py==1.8.1,pyparsing==2.4.7,pytest==4.6.6,scandir==1.10.0,six==1.14.0,wcwidth==0.1.9,zipp==1.2.0
py27 run-test-pre: PYTHONHASHSEED='1467207441'
py27 run-test: commands[0] | pytest --ignore=tests/test_integration.py -vvsk test_entry_points_dict
==================================================================================================================================== test session starts =====================================================================================================================================
platform linux2 -- Python 2.7.18, pytest-4.6.6, py-1.8.1, pluggy-0.13.1 -- /home/jsirois/dev/pantsbuild/jsirois-pex/.tox/py27/bin/python
cachedir: .tox/py27/.pytest_cache
rootdir: /home/jsirois/dev/pantsbuild/jsirois-pex
collected 179 items / 178 deselected / 1 selected                                                                                                                                                                                                                                            

tests/test_bdist_pex.py::test_entry_points_dict running bdist_pex
Writing my_app to /tmp/tmpsgDVu1/my_app-0.0.0.pex
[...]

PR #970, opened 2020-05-11
Split Pex resolve API.

jsirois:
This change introduces download and install APIs that the existing Pex
resolve / resolve_multi APIs are composed from.

The new APIs will be used by Pants to split traditional resolution
(download in Pip terms) from compilation and install of wheels from the
downloaded distributions.

In the future the APIs could be used to support Pex CLI subcommands that
distinguish these two phases of creating a PEX allowing resolution on
one machine and building on another; thus adding a new way to handle
creation of multi-platform pexes without resorting to pre-built wheel
repositories.

### DIFF
--- a/pex/resolver.py
+++ b/pex/resolver.py
@@ -185,15 +185,13 @@ class DownloadRequest(namedtuple('DownloadRequest', [
           for target in self.targets:
             yield BuildRequest.create(target=target, source_path=local_project)
 
-  def download_distributions(self, workspace=None, max_parallel_jobs=None):
+  def download_distributions(self, dest=None, max_parallel_jobs=None):
     if not self.requirements and not self.requirement_files:
       # Nothing to resolve.
       return None
 
-    workspace = workspace or safe_mkdtemp()
-
-    resolved_dists_dir = os.path.join(workspace, 'resolved_dists')
-    spawn_download = functools.partial(self._spawn_download, resolved_dists_dir)
+    dest = dest or safe_mkdtemp()
+    spawn_download = functools.partial(self._spawn_download, dest)
     with TRACER.timed('Resolving for:\n  {}'.format('\n  '.join(map(str, self.targets)))):
       return list(
         execute_parallel(
@@ -837,7 +835,7 @@ def resolve_multi(requirements=None,
     build=build,
     use_wheel=use_wheel,
     manylinux=manylinux,
-    workspace=workspace,
+    dest=workspace,
     max_parallel_jobs=max_parallel_jobs
   )
 
@@ -879,7 +877,7 @@ def _download_internal(requirements=None,
                        build=True,
                        use_wheel=True,
                        manylinux=None,
-                       workspace=None,
+                       dest=None,
                        max_parallel_jobs=None):
 
   parsed_platforms = [parsed_platform(platform) for platform in platforms] if platforms else []
@@ -920,9 +918,9 @@ def _download_internal(requirements=None,
 
   local_projects = list(download_request.iter_local_projects())
 
-  workspace = workspace or safe_mkdtemp()
+  dest = dest or safe_mkdtemp()
   download_results = download_request.download_distributions(
-    workspace=workspace,
+    dest=dest,
     max_parallel_jobs=max_parallel_jobs
   )
   return local_projects, download_results
@@ -953,6 +951,7 @@ def download(requirements=None,
              build=True,
              use_wheel=True,
              manylinux=None,
+             dest=None,
              max_parallel_jobs=None):
   """Downloads all distributions needed to meet requirements for multiple distribution targets.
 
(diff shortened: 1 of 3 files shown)


jsirois:
CI should fail on the OSX Python 2.7 shards as described in #967 and fixed by #968.

pex/resolver.py
```diff
@@ -151,7 +156,75 @@ def parsed_platform(platform=None):
...
+      transitive=self.transitive,
+      target=target,
+      indexes=self.indexes,
+      find_links=self.find_links,
+      cache=self.cache,
+      build=self.build,
+      manylinux=self.manylinux,
+      use_wheel=self.use_wheel
+    )
+    return SpawnedJob.wait(job=download_job, result=DownloadResult(target, download_dir))
```
jsirois, line 224: Moved from https://github.com/pantsbuild/pex/pull/970/files#diff-82dd0097879461fc5b03ef52278ba3f8L373-L413

pex/resolver.py
```diff
@@ -785,6 +792,96 @@ def resolve_multi(requirements=None,
...
+  #
+  # As a complicating factor, the runtime activation scheme relies on PEP 425 tags; i.e.: wheel
+  # names. Some requirements are only available or applicable in source form - either via sdist,
+  # VCS URL or local projects. As such we need to insert a `pip wheel` step to generate wheels for
+  # all requirements resolved in source form via `pip download` / inspection of requirements to
+  # discover those that are local directories (local setup.py or pyproject.toml python projects).
+  #
+  # Finally, we must calculate the pinned requirement corresponding to each distribution we
+  # resolved along with any environment markers that control which runtime environments the
+  # requirement should be activated in.
```
jsirois, line 822: Moved from https://github.com/pantsbuild/pex/pull/970/files#diff-82dd0097879461fc5b03ef52278ba3f8L468-L494

jsirois:
Reviewers - apologies for the large diff. There actually isn't a whole lot going on here besides re-names and moves and big new doc comments for the download and install APIs. Lots of fluff. I pointed out the two biggest ~pure move fluffs.

cosmicexplorer:
> In the future the APIs could be used to support Pex CLI subcommands that distinguish these two phases of creating a PEX allowing resolution on one machine and building on another; thus adding a new way to handle creation of multi-platform pexes without resorting to pre-built wheel repositories.

Will review the current PR! I'd also like to note along these lines that pypa/pip#53 is introducing the concept of splitting resolve and download (which I'll note was specifically inspired by pypa/pip#7819 :D) and that it would likely make sense for pex to be able to take advantage of that (because exposing that functionality via pex as a library alone would be a super cool contribution).

pex/resolver.py
```diff
@@ -35,15 +36,19 @@ class Unsatisfiable(Exception):
...
 
   def __new__(cls, target, requirement, distribution):
     assert isinstance(target, DistributionTarget)
     assert isinstance(requirement, Requirement)
     assert isinstance(distribution, Distribution)
-    return super(ResolvedDistribution, cls).__new__(cls, target, requirement, distribution)
+    return super(InstalledDistribution, cls).__new__(cls, target, requirement, distribution)
+
+
+# A type alias to preserve API compatibility for resolve and resolve_multi.
```
stuhood, line 50: "...until version X"?

pex/resolver.py
```diff
@@ -151,7 +156,73 @@ def parsed_platform(platform=None):
   return Platform.create(platform) if platform and platform != 'current' else None
 
 
-class ResolveResult(namedtuple('ResolveResult', ['target', 'download_dir'])):
+class DownloadRequest(namedtuple('DownloadRequest', [
+  'targets',
```
stuhood, line 160: I guess that in this context `target` is a DistributionTarget? Unsure if worth spelling out, or less ambiguous in this codebase.

pex/resolver.py
```diff
@@ -806,19 +901,191 @@ def iter_targets():
...
+
+  dest = dest or safe_mkdtemp()
+  download_results = download_request.download_distributions(
+    dest=dest,
+    max_parallel_jobs=max_parallel_jobs
+  )
+  return local_projects, download_results
+
+
+class LocalDistribution(namedtuple('LocalDistribution', ['target', 'path', 'fingerprint'])):
```
stuhood, line 929: The `Local` here is "has been downloaded", rather than "compatible with the local machine"?

pex/resolver.py
jsirois: Definitely unambiguous in this codebase - there are no other targets.

pex/resolver.py
jsirois: Unfortunately not yet. The TODO here anticipates this though:
https://github.com/pantsbuild/pex/pull/970/files#diff-82dd0097879461fc5b03ef52278ba3f8R708-R709

pex/resolver.py
jsirois: Correct. It's only compatible with `target` which may or may not be local.

jsirois:
Rebased to pick up CI fix in #968.

jsirois:
There is more to follow up on here in the context of the API work in #969 including more clearly delineating API boundary types and documenting exception types, but I'm going to push forward with the merge to unblock some Pants-side work consuming this. As always, feedback post-merge is welcome.

PR #971, opened 2020-05-11
Add a `--local` mode for packaging the Pex PEX.

jsirois:
This significantly speeds up `--server` launch for local integration
testing when there are many local interpreters that match Pex
requires-python.

### DIFF
--- a/scripts/package.py
+++ b/scripts/package.py
@@ -22,7 +22,7 @@ def python_requires() -> str:
   return project_metadata['tool']['flit']['metadata']['requires-python'].strip()
 
 
-def build_pex_pex(output_file: PurePath, verbosity: int = 0) -> None:
+def build_pex_pex(output_file: PurePath, local: bool = False, verbosity: int = 0) -> None:
   # NB: We do not include the subprocess extra (which would be spelled: `.[subprocess]`) since we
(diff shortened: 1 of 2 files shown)


scripts/package.py
```diff
@@ -146,6 +152,12 @@ def main(*additional_dist_formats: Format, verbosity: int = 0, serve: bool = Fal
     action='append',
     help='Package Pex in additional formats.'
   )
+  parser.add_argument(
+    '--local',
+    default=False,
+    action='store_true',
+    help='Build Pex PEX with just a single local interpreter.'
```
stuhood, line 159: Is it clear in this context which one you'll get?

jsirois: Yes, unconstrained `pex ...` uses the interpreter it was invoked with which is sys.executable above. The other factor is Pex is multipython and not native, so ~any interpreter will do.

To be clear - this is working around the fact --interpreter-constraint has been broken since inception for Pex builtime since its always been used to pick the list of interpreters used instead of sanity check them. It is fully coherent at runtime in contrast.

This just stops bleeding on a machine like mine where I have ~15 pythons installed which brings everything (including the cursor) to a halt for many minutes.

PR #973, opened 2020-05-11
Restore handling for bad wheel filenames to `.can_add()`

luislew:
My previous PR (#965) unintentionally removed handling for invalid wheel names (e.g. `"foo.whl"` or `"foo-bar.whl"`). This PR restores the previous functionality, namely returning `False` for (a subset of) bad wheel filenames.

I made the conservative choice to only reject filenames with < 3 wheel components (matching the previous behavior), but there's a valid argument to made for rejecting wheels with < 5 components, per [PEP 427](https://www.python.org/dev/peps/pep-0427/#file-name-convention).

---

Previous implementation: https://github.com/pantsbuild/pex/blob/2f048d8df0dbcebdc24d2355bf58130baa35de0f/pex/environment.py#L221-L225

### DIFF
--- a/pex/environment.py
+++ b/pex/environment.py
@@ -221,7 +221,11 @@ class PEXEnvironment(Environment):
 
     # Wheel filename format: https://www.python.org/dev/peps/pep-0427/#file-name-convention
     # `{distribution}-{version}(-{build tag})?-{python tag}-{abi tag}-{platform tag}.whl`
-    wheel_tags = '-'.join(filename.split('-')[-3:])  # `{python tag}-{abi tag}-{platform tag}`
+    wheel_components = filename.split('-')
+    if len(wheel_components) < 3:
+      return False
+
+    wheel_tags = '-'.join(wheel_components[-3:])  # `{python tag}-{abi tag}-{platform tag}`
     if self._supported_tags.isdisjoint(tags.parse_tag(wheel_tags)):
       return False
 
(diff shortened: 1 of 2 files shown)


luislew:
@jsirois Quick addendum to my previous PR... I don't know if this is truly necessary, but it's the safer implementation and ensures that my change doesn't break anything for projects that include  malformed wheels.

pex/environment.py
```diff
@@ -221,7 +221,11 @@ def can_add(self, dist):
 
     # Wheel filename format: https://www.python.org/dev/peps/pep-0427/#file-name-convention
     # `{distribution}-{version}(-{build tag})?-{python tag}-{abi tag}-{platform tag}.whl`
-    wheel_tags = '-'.join(filename.split('-')[-3:])  # `{python tag}-{abi tag}-{platform tag}`
+    wheel_components = filename.split('-')
+    if len(wheel_components) < 3:
+      return False
+
+    wheel_tags = '-'.join(wheel_components[-3:])  # `{python tag}-{abi tag}-{platform tag}`
     if self._supported_tags.isdisjoint(tags.parse_tag(wheel_tags)):
```
jsirois, line 229: Would have gone boom! here in `tags.parse_tags`:
https://github.com/pantsbuild/pex/blob/2a8b950284d4a8be5eb2bcf9dcb63e2bc5f60fe4/pex/vendor/_vendored/packaging/packaging/tags.py#L73-L75

PR #974, opened 2020-05-11
Expose Pip network configuration.

### LINKED ISSUE #94: Remove PEX_HTTP_RETRIES and push into a flag for the pex tool
PEX_\* environment variables are reserved for the PEX runtime.  Everything else should ideally be a flag to the pex tool or used via the API directly.

### LINKED ISSUE #803: Expose network configuration in pex options.
In particular, these pip options should be plumbed through from the pex CLI & the resolver API:
```
  -H, --header <key:val>      HTTP header to include in all requests. This option can be used multiple times. Conflicts with --extra-index-url.
  --proxy <proxy>             Specify a proxy in the form [user:passwd@]proxy.server:port.
  --retries <retries>         Maximum number of retries each connection should attempt (default 5 times).
  --timeout <sec>             Set the socket timeout (default 15 seconds).
  --trusted-host <hostname>   Mark this host as trusted, even though it does not have valid or any HTTPS.
  --cert <path>               Path to alternate CA bundle.
  --client-cert <path>        Path to SSL client certificate, a single file containing the private key and the certificate in PEM format.
```

jsirois:
Also add back the `--cache-ttl` option from Pex 1.x and plumb the header
needed to make it so.

Also remove long-ignored `PEX_HTTP_RETRIES` and `PEX_HTTP_TIMEOUT` env
var support.

Fixes #94
Fixes #803

### DIFF
--- a/pex/network_configuration.py
+++ b/pex/network_configuration.py
@@ -16,7 +16,7 @@ class NetworkConfiguration(namedtuple('NetworkConfiguration', [
   'cert',
   'client_cert'
 ])):
-  """Configurtion for network requests made by the resolver."""
+  """Configuration for network requests made by the resolver."""
 
   @classmethod
   def create(
--- a/pex/network_configuration.py
+++ b/pex/network_configuration.py
@@ -1,0 +1,58 @@
+# coding=utf-8
+# Copyright 2020 Pants project contributors (see CONTRIBUTORS.md).
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+from __future__ import absolute_import
+
+from collections import namedtuple
+
+
+class NetworkConfiguration(namedtuple('NetworkConfiguration', [
+  'cache_ttl',
+  'retries',
+  'timeout',
+  'headers',
+  'proxy',
+  'cert',
+  'client_cert'
+])):
+  """Configurtion for network requests made by the resolver."""
+
+  @classmethod
+  def create(
+      cls,
+      cache_ttl=3600,
+      retries=5,
+      timeout=15,
+      headers=None,
+      proxy=None,
+      cert=None,
+      client_cert=None
+  ):
+    """
+    Creates a new network configuration accepting defaults for any values not explicitly provided.
+
+    :param int cache_ttl: The maximum time an item is retained in the HTTP cache in seconds.
+    :param int retries: The maximum number of retries each connection should attempt.
+    :param timeout: The socket timeout in seconds.
+    :param headers: A list of additional HTTP headers to send in the form NAME:VALUE.
+    :type headers: list of str
+    :param str proxy: A proxy to use in the form [user:passwd@]proxy.server:port.
+    :param str cert: The path to an alternate CA bundle.
+    :param str client_cert: The path to an SSL client certificate which should be a single file
+                            containing the private key and the certificate in PEM format.
+    :return: The resulting network configuration.
+    :rtype: :class:`NetworkConfiguration`
+    """
+    assert cache_ttl >= 0, 'The cache_ttl parameter should be >= 0; given: {}'.format(cache_ttl)
+    assert retries >= 0, 'The retries parameter should be >= 0; given: {}'.format(retries)
+    assert timeout >= 0, 'The timeout parameter should be > 0; given: {}'.format(timeout)
+    return cls(
+      cache_ttl=cache_ttl,
+      retries=retries,
+      timeout=timeout,
+      headers=tuple(headers or ()),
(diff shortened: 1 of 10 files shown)


jsirois:
Reviewers, there are 2 commits:
1. Re-vendoring Pip to pull in @wisechengyi's patch in https://github.com/pantsbuild/pip/pull/7
2. Exposing network configuration options

This finally gets Pex 2.x performing better than Pex 1.x in warm cache cases. An example against Pex itself:

1.6.12:
```
$ multitime -n 10 ./pex-1.6.12.pex pex==2.1.9 --cache-ttl=3600 -opex.pex
===> multitime results
1: ./pex-1.6.12.pex pex==2.1.9 --cache-ttl=3600 -opex.pex
real        1.698       0.050       1.644       1.688       1.825       
user        1.308       0.032       1.243       1.328       1.338       
sys         0.157       0.025       0.127       0.154       0.202     
```

2.1.9:
```
$ multitime -n 10 ./pex-2.1.9.pex pex==2.1.9 -opex.pex
===> multitime results
1: ./pex-2.1.9.pex pex==2.1.9 -opex.pex
real        1.634       0.013       1.616       1.631       1.666       
user        1.307       0.021       1.271       1.302       1.339       
sys         0.176       0.016       0.146       0.177       0.209```
```

HEAD:
```
$ multitime -n 10 ./dist/pex pex==2.1.9 --cache-ttl=3600 -o pex.pex
===> multitime results
1: ./dist/pex pex==2.1.9 --cache-ttl=3600 -o pex.pex
real        1.457       0.011       1.443       1.454       1.483       
user        1.265       0.015       1.229       1.268       1.283       
sys         0.190       0.017       0.167       0.191       0.219
```

pex/network_configuration.py
```diff
@@ -0,0 +1,58 @@
...
+class NetworkConfiguration(namedtuple('NetworkConfiguration', [
+  'cache_ttl',
+  'retries',
+  'timeout',
+  'headers',
+  'proxy',
+  'cert',
+  'client_cert'
+])):
+  """Configurtion for network requests made by the resolver."""
```
Eric-Arellano, line 19:
```suggestion
  """Configuration for network requests made by the resolver."""
```

pex/network_configuration.py
```diff
@@ -0,0 +1,58 @@
...
+    :param str proxy: A proxy to use in the form [user:passwd@]proxy.server:port.
+    :param str cert: The path to an alternate CA bundle.
+    :param str client_cert: The path to an SSL client certificate which should be a single file
+                            containing the private key and the certificate in PEM format.
+    :return: The resulting network configuration.
+    :rtype: :class:`NetworkConfiguration`
+    """
+    assert cache_ttl >= 0, 'The cache_ttl parameter should be >= 0; given: {}'.format(cache_ttl)
+    assert retries >= 0, 'The retries parameter should be >= 0; given: {}'.format(retries)
+    assert timeout >= 0, 'The timeout parameter should be > 0; given: {}'.format(timeout)
```
Eric-Arellano, line 49: Safer to use `raise AssertionError`, as `assert` is not run when byte code is optimized (https://bandit.readthedocs.io/en/latest/plugins/b101_assert_used.html).

pex/variables.py
```diff
@@ -339,28 +339,6 @@ def PEX_VERBOSE(self):
...
-  @property
-  def PEX_HTTP_TIMEOUT(self):
-    """Integer
-
-    When built with the `requests` HTTP library, a timeout to apply for HTTP connect attempts and
-    read attempts. See the `requests` docs (https://2.python-requests.org/en/master/user/quickstart/#timeouts)
-    for more information.
-    Default: 15.
-    """
-    return self._get_int('PEX_HTTP_TIMEOUT', default=15)
```
Eric-Arellano, line 362: Might want to make this a dedicated PR so that Pex's changelog can say "Remove unused `PEX_HTTP_RETRIES` and `PEX_HTTP_TIMEOUT` flags." Otherwise, this is easy to miss.

pex/network_configuration.py
jsirois: I'm not too worried about that. I would like to beef up the API types though to do argument validation more thoroughly though at which point these will switch from asserts to at least ValueErrors. This is tracked in #969 with a note just added.

pex/variables.py
jsirois: I'd like to keep this inline since nothing breaks - these are just silent env var non-reads that have been happening since October. I will call it out in the release header though since they're both gone and were always a bad idea for the reasons Wickman notes in #94.

Issue #978, opened 2020-05-15
wheel cache should use relative path symlinks to allow for chroot environments


tdyas:
When running `pex` within a chroot environment, the symlinks generated by https://github.com/pantsbuild/pex/blob/e2d6bab09f8e9245f608402a904421268223ee58/pex/resolver.py#L421 use the absolute path within the chroot in the path for the symlinks' targets. These symlinks are invalid paths to any program that accesses the wheel cache outside of the chroot. (This was seen within a remote execution environment with Buildfarm.)

These would be better as relative paths within the wheel cache directory structure (probably use `os.path.relpath`).

PR #979, opened 2020-05-15
use relative paths within wheel cache

tdyas:
When running pex within a chroot environment, the symlinks generated by `pex/pex/resolver.py` use the absolute path within the chroot in the path for the symlinks' targets. These symlinks are invalid paths to any program that accesses the wheel cache outside of the chroot. (This was seen within a remote execution environment with Buildfarm.)

Switch to using relative paths for the symlinks.

Fixes https://github.com/pantsbuild/pex/issues/978.

### DIFF
--- a/pex/resolver.py
+++ b/pex/resolver.py
@@ -418,6 +418,9 @@ class InstallResult(namedtuple('InstallResult', ['request', 'installation_root',
     runtime_key_dir = os.path.join(self.installation_root, wheel_dir_hash)
     with atomic_directory(runtime_key_dir) as work_dir:
       if work_dir:
+        # Note: Create a relative path symlink between the two directories so that the PEX_ROOT can be
+        # used within a chroot environment where the prefix of the path may change between programs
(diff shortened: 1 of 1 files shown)


pex/resolver.py
```diff
@@ -418,7 +418,10 @@ def finalize_install(self, install_requests):
     runtime_key_dir = os.path.join(self.installation_root, wheel_dir_hash)
     with atomic_directory(runtime_key_dir) as work_dir:
       if work_dir:
-        os.symlink(self.install_chroot, os.path.join(work_dir, self.request.wheel_file))
+        source_path = os.path.join(work_dir, self.request.wheel_file)
```
Eric-Arellano, line 421: Maybe add a brief comment?

# NB: It's important to use relative paths here because ..

tdyas: Added.

Eric-Arellano:
Style check failed. Iirc you can run `tox -e format-run`.

PR #986, opened 2020-06-04
Fix typos

asherf:


pex/resolver.py
```diff
@@ -28,7 +28,7 @@
 from pex.util import CacheHelper
 
 
-class Untranslateable(Exception):
+class Untranslatable(Exception):
```
stuhood, line 31: Unclear whether this is a public API... but I don't see any usages of the old name in `pants`, so I'm fine with just renaming.

pex/resolver.py
```diff
@@ -698,7 +698,7 @@ def resolve(requirements=None,
   :type indexes: list of str
   :keyword find_links: A list or URLs, paths to local html files or directory paths. If URLs or
     local html file paths, these are parsed for links to distributions. If a local directory path,
-    its listing is used to discover distributons.
+    its listing is used to discover distributors.
```
stuhood, line 701: Should be `distributions`

pex/resolver.py
```diff
@@ -787,7 +787,7 @@ def resolve_multi(requirements=None,
...
```
stuhood, line 790: ditto

pex/resolver.py
```diff
@@ -1004,7 +1004,7 @@ def download(requirements=None,
...
```
stuhood, line 1007: ditto

pex/resolver.py
```diff
@@ -1081,7 +1081,7 @@ def install(local_distributions,
...
```
stuhood, line 1084: ditto

PR #989, opened 2020-06-19
A PEX_EXTRA_SYS_PATH runtime variable.

benjyw:
If set, we add these path elements to PEX's runtime sys.path.
This allows code in the pex to interact with code outside the
pex (for example, running pytest in a pex on code outside it).

### DIFF
--- a/pex/variables.py
+++ b/pex/variables.py
@@ -201,6 +201,8 @@ class Variables(object):
     Using this option is generally not advised, but can help in situations when certain dependencies
     do not conform to standard packaging practices and thus cannot be bundled into PEX files.
 
+    See also PEX_EXTRA_SYS_PATH for how to *add* to the sys.path.
+
     Default: false.
     """
     return self._get_string('PEX_INHERIT_PATH', default='false')
@@ -285,6 +287,11 @@ class Variables(object):
     interact with code outside it.
 
     Ex: "/path/to/lib1:/path/to/lib2"
+
+    This is distinct from PEX_INHERIT_PATH, which controls how the interpreter's
+    existing sys.path (which you may not have control over) is scrubbed.
+
+    See also PEX_PATH for how to merge packages from other pexes into the current environment.
     """
     return self._get_string('PEX_EXTRA_SYS_PATH', default=None)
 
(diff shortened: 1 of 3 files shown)


benjyw:
There is already a PEX_PYTHON_PATH (which is paths to python interpreters) so I wanted to be very unconfusing.

benjyw:
Added a test

tests/test_pex.py
```diff
@@ -253,6 +253,23 @@ def test_pex_run():
...
+        'import sys; sys.stdout.write(":".join(sys.path)); sys.exit(0)'
+      )
+      rc = PEX(pex.path()).run(stdin=None, stdout=fake_stdout, stderr=None, env={
+        'PEX_EXTRA_SYS_PATH': 'extra/syspath/entry1:extra/syspath/entry2'})
+      assert rc == 0
+
+      fake_stdout.seek(0)
+      syspath = fake_stdout.read().split(b':')
+      assert b'extra/syspath/entry1' in syspath
+      assert b'extra/syspath/entry2' in syspath
```
Eric-Arellano, line 270: Looks good! Could you please add a base case where you do not set the env var to be extra defensive? They can “assert foo not in” for example

benjyw: That seems excessive. What would I test for? That a random string isn't in the sys.path? How would I pick that string?

Eric-Arellano: Eh, fair. Looks good

pex/variables.py
```diff
@@ -275,6 +275,19 @@ def PEX_PYTHON_PATH(self):
     """
     return self._get_string('PEX_PYTHON_PATH', default=None)
 
+  @property
+  def PEX_EXTRA_SYS_PATH(self):
+    """String
+
+    A colon-separated string containing paths to add to the runtime sys.path.
```
stuhood, line 282: It would be good to add docs both here and to `PEX_INHERIT_PATH` that explain how these two options interact.

As far as I can tell, the interaction is fine? "inherit" talks about what to do with the existing path defined by the interpreter, which this allows items to be added after that logic has run.

You might also want to describe when you should use this vs `PEX_PATH`, which also allows the PYTHONPATH to be expanded at runtime (but with pexes rather than loose files). Related: should we be using PEX_PATH in some way here rather than loose files?

benjyw:
@stuhood what do you mean by "aligning the docs and naming"?

Eric-Arellano:
Reminder that there are no up-to-date Pex docs because we can’t access the site. Benjy and I have been wanting to make a readme.io-based version of Pex docs, built from the ground up.

pex/variables.py
benjyw: In the pytest runner use case, we don't want to build a sources pex, for performance reasons.

benjyw:
Going to merge since only changes from previous green commit were docs, and I checked lint.

PR #1000, opened 2020-07-10
Add support for resolving --platform locally.

### LINKED ISSUE #996: Add support for satisfying `--platform=X` with a local interpreter iff one is found.
Currently, specifying a `--platform` that doesn't match the interpreter Pex is running under forces resolves to act as if `--no-build` had been specified and thus only fetch pre-built wheels:
https://github.com/pantsbuild/pex/blob/451977efdf987dd299a1b4798ac2ee298cd6d61b/pex/pip.py#L208-L220

Using Pex from other (multiplatform) programs could be made simpler in many cases if, in addition to the current interpreter platform check, Pex could be asked to (or by default) check for any local interpreter that matched specified `--platform`s, using those interpreters when found to build wheels when only sdists were available.

The motivating use case is https://github.com/pantsbuild/pants/issues/10286

jsirois:
Introduce new APIs to get a PythonInterpreter's associated platforms
directly and use this to implement resolution of specified platforms to
local interpreters when possible if --resolve-local-platforms is
specified.

Fixes #996.

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -445,9 +445,7 @@ def configure_clp_pex_environment(parser):
            'https://www.python.org/dev/peps/pep-0427#file-name-convention and influenced by '
            'https://www.python.org/dev/peps/pep-0425. For the current interpreter at {} the full '
            'platform string is {}. To find out more, try `{} --platform explain`.'
-           .format(current_interpreter.binary,
-                   Platform.of_interpreter(current_interpreter),
-                   sys.argv[0]))
+           .format(current_interpreter.binary, current_interpreter.platform, sys.argv[0]))
 
   group.add_option(
       '--resolve-local-platforms',
@@ -621,8 +619,8 @@ def build_pex(reqs, options, cache=None):
 
   pex_python_path = None  # Defaults to $PATH
   if options.rc_file or not ENV.PEX_IGNORE_RCFILES:
-    rc_variables = Variables.from_rc(rc=options.rc_file)
-    pex_python_path = rc_variables.get('PEX_PYTHON_PATH', None)
+    rc_variables = Variables(rc=options.rc_file)
+    pex_python_path = rc_variables.PEX_PYTHON_PATH
 
   # NB: options.python and interpreter constraints cannot be used together.
   if options.python:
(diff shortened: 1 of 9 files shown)


pex/bin/pex.py
```diff
@@ -444,6 +445,17 @@ def configure_clp_pex_environment(parser):
            'string representing the python version (e.g. "27", "36"). ABI is the ABI tag '
            '(e.g. "cp36m", "cp27mu", "abi3", "none"). Default: current platform.')
 
+  group.add_option(
+      '--resolve-local-platforms',
+      dest='resolve_local_platforms',
+      default=False,
```
Eric-Arellano, line 451: Is the default `False` for backward-compatibility or because you think it's the more sound default?

pex/bin/pex.py
```diff
@@ -444,6 +445,17 @@ def configure_clp_pex_environment(parser):
...
+      action='callback',
+      callback=parse_bool,
+      help='When --platforms are specified, attempt to resolve a local interpreter that matches '
+           'each platform specified. If found, use the interpreter to resolve distributions; if '
+           'not, resolve for the platform only allowing matching binary distributions and failing '
```
Eric-Arellano, line 456: bdist = built distribution, not binary.

```suggestion
           'not, resolve for the platform only allowing matching built distributions (bdists) and failing '
```

Eric-Arellano: There are a couple other references in this PR that would need updating.

pex/bin/pex.py
jsirois: Both with the added reason that its slow in a cold cache case. On my machine with ~10 interpreters to search, building the cold interpreter cache takes ~1 second. After that these checks are ~20ms though.

pex/bin/pex.py
jsirois: Well, you're running into reality here. Pip, which is what we delegate all this to uses binary == built:
https://github.com/pantsbuild/pex/blob/dd0cf31f791cfce77ef29ddc0750985f80135850/pex/pip.py#L219-L223

Eric-Arellano: Hm, that's confusing. Distutils calls it a built distribution: https://docs.python.org/3/distutils/builtdist.html, but that's fine to keep using this if you prefer.

jsirois: Yeah, I think we had talked about distinction without a difference. I don't think its long term useful to be hung up on this one. Unless there is a large change, bdist == binary == {wheel}. Once egg was part of the set, but not really any longer.

PR #1001, opened 2020-07-10
Improve and fix --platform help.

### LINKED ISSUE #997: Feature Request : Syntax for --platform to refer to the current platform
I'm working on a project where I'm using an OS X development machine to cross build PEX files that need to run both on OS X and on Linux. To do this, I am using `--platform` to define a new platform list for the build that contains both the OS X platform string and the target platform:

`--platform macosx_10.15-x86_64-cp-37-cp37m  --platform manylinux2010-x86_64-cp-37-cp37m`

This works, but it requires explicitly specifying (and knowing) the current platform string. What would be useful to avoid this is a syntax for the `--platform` option that uses the normal defaulting logic to compute the current platform string and adds that to the platform list.

`--platform host-platform --platform manylinux2010-x86_64-cp-37-cp37m`

Alternatively, a programmatic way to query for the current platform string would also work:

`--platform $(find-python-platform-string) --platform manylinux2010-x86_64-cp-37-cp37m`

jsirois:
Fixes #997.

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -436,13 +436,13 @@ def configure_clp_pex_environment(parser):
       action='callback',
       callback=process_platform,
       help='The platform for which to build the PEX. This option can be passed multiple times '
-           'to create a multi-platform pex. To use wheels for specific interpreter/platform tags'
-           ', you can append them to the platform with hyphens like: PLATFORM-IMPL-PYVER-ABI '
-           '(e.g. "linux_x86_64-cp-27-cp27mu", "macosx_10.12_x86_64-cp-36-cp36m") PLATFORM is '
-           'the host platform e.g. "linux-x86_64", "macosx-10.12-x86_64", etc". IMPL is the '
-           'python implementation abbreviation (e.g. "cp", "pp", "jp"). PYVER is a two-digit '
-           'string representing the python version (e.g. "27", "36"). ABI is the ABI tag '
-           '(e.g. "cp36m", "cp27mu", "abi3", "none"). Default: current platform.')
+           'to create a multi-platform pex. To use the platform corresponding to the current '
+           'interpreter you can pass `current`. To target any other platform you pass a string '
+           'composed of fields: <platform>-<python impl abbr>-<python version>-<abi>. These fields '
+           'stem from wheel name conventions as outlined in '
+           'https://www.python.org/dev/peps/pep-0427#file-name-convention and influenced by '
+           'https://www.python.org/dev/peps/pep-0425. For the current interpreter at {} the full '
+           'platform string is {}'.format(PythonInterpreter.get().binary, Platform.current()))
 
   parser.add_option_group(group)
 


jsirois:
Help now renders on my machine as:
```
   --platform=PLATFORMS
                        The platform for which to build the PEX. This option
                        can be passed multiple times to create a multi-
                        platform pex. To use the platform corresponding to the
                        current interpreter you can pass `current`. To target
                        any other platform you pass a string composed of
                        fields: <platform>-<python impl abbr>-<python
                        version>-<abi>. These fields stem from wheel name
                        conventions as outlined in
                        https://www.python.org/dev/peps/pep-0427#file-name-
                        convention and influenced by
                        https://www.python.org/dev/peps/pep-0425. For the
                        current interpreter at /usr/bin/python3.8 the full
                        platform string is manylinux2014_x86_64-cp-38-cp38
```

pex/bin/pex.py
```diff
@@ -436,13 +436,13 @@ def configure_clp_pex_environment(parser):
...
-           'to create a multi-platform pex. To use wheels for specific interpreter/platform tags'
-           ', you can append them to the platform with hyphens like: PLATFORM-IMPL-PYVER-ABI '
-           '(e.g. "linux_x86_64-cp-27-cp27mu", "macosx_10.12_x86_64-cp-36-cp36m") PLATFORM is '
-           'the host platform e.g. "linux-x86_64", "macosx-10.12-x86_64", etc". IMPL is the '
-           'python implementation abbreviation (e.g. "cp", "pp", "jp"). PYVER is a two-digit '
-           'string representing the python version (e.g. "27", "36"). ABI is the ABI tag '
-           '(e.g. "cp36m", "cp27mu", "abi3", "none"). Default: current platform.')
+           'to create a multi-platform pex. To use the platform corresponding to the current '
+           'interpreter you can pass `current`. To target any other platform you pass a string '
+           'composed of fields: <platform>-<python impl abbr>-<python version>-<abi>. These fields '
```
Eric-Arellano, line 441: I think the old examples of `"linux_x86_64-cp-27-cp27mu", "macosx_10.12_x86_64-cp-36-cp36m"` were really helpful. I know in the past that you've wanted to avoid explaining it all because we won't capture the nuance of the PEP, which makes sense. 

It's fine to not break apart each part like you previously, but giving two working examples is very helpful imo.

Eric-Arellano:
Adding Benjy and Stu because this topic has confused users quite a bit in the past. Relates to the incomplete docs at https://pants.readme.io/docs/python-platforms.

pex/bin/pex.py
jsirois: I'm potentially sympathetic. Keep in mind:
```
$ python -mpex --platform bob
Usage: __main__.py [-o OUTPUT.PEX] [options] [-- arg1 arg2 ...]

__main__.py builds a PEX (Python Executable) file based on the given specifications: sources, requirements, their dependencies and other options.

__main__.py: error: The --platform option is invalid:
Not a valid platform specifier: bob

Platform strings must be in one of two forms:
1. Canonical: <platform>-<python impl abbr>-<python version>-<abi>
2. Abbreviated: <platform>-<python impl abbr>-<python version>-<abbr abi>

Given a canonical platform string for CPython 3.7.5 running on 64 bit linux of:
  linux-x86_64-cp-37-cp37m

Where the fields above are:
+ <platform>: linux-x86_64 
+ <python impl abbr>: cp
+ <python version>: 37
+ <abi>: cp37m

The abbreviated platform string is:
  linux-x86_64-cp-37-m

These fields stem from wheel name conventions as outlined in
https://www.python.org/dev/peps/pep-0427#file-name-convention and influenced by
https://www.python.org/dev/peps/pep-0425.
```

The formatting of OptParse help string is horrible. The formatting of the error we have complate control over. Maybe two examples there?

Eric-Arellano: Ah, good, glad we still have this. A linux and a macOS example would be great in this error message, indeed.

Maybe one example in the help message? Better UX to get it right on the first attempt than needing the error message. My concern is "Okay, I see this PEP, but I don't really see how the PEP connects to this tool". My hypothesis is that having at least one example would make it easier to know what to scan for in the PEP. And if that fails, you get this nice error message.

jsirois: There is one example on the last line of help. Its the example derived from the interpreter running pex which is an affordance for finding out the correct platform on any machine you're targeting the pex to as a hack. At some point pex will need to grow subcommands.

jsirois: Perhaps you want to see that example in a different position in the help?

jsirois: FWIW - this is a case where you can't ever get it right on the first attempt. You really have to know what you're doing or - now - run `pex --help` on every machine you're targeting to harvest the help example and thus collect up all the platform strings you need.

Eric-Arellano: > There is one example on the last line of help. Its the example derived from the interpreter running pex 

Ah, I misread this. This is really helpful, indeed. Pardon not reading the last line more closely.

jsirois: Ok, great. I'm going to circle back separately to add an extra example or two to the parse error message and likely add verbage like "Use --platform explain to see more examples.".

jsirois: Follow up here: #1002

PR #1006, opened 2020-07-18
Autoformat with Black

Eric-Arellano:
All of the pantsbuild codebases use Black + Isort, except for Pex. This brings us alignment across the organization.


Eric-Arellano:
Much more feasible now that we have little BUILD file boilerplate! Earlier, I'd say it would be overkill, but Pants is starting to get the point where it makes sense even in small repos :)

Related to this, I tried setting up Docformatter and it was too hard to get working without Pants. No way to easily exclude the vendored folder without explicitly enumerating every single file to format. If we had Pants, this would be easy.

PR #1008, opened 2020-07-20
Add `--use-first-matching-interpreter`

Eric-Arellano:
Before:
```
1: python -m pex --disable-cache --interpreter-constraint=CPython>=3.6 cryptography -o t.pex
real        7.541       0.258       7.283       7.378       7.967
user        21.989      0.609       20.890      21.887      23.020
sys         5.979       0.169       5.690       5.918       6.237
```

After:
```
1: python -m pex --disable-cache --use-first-matching-interpreter --interpreter-constraint=CPython>=3.6 cryptography -o t.pex
real        6.217       0.478       5.820       6.118       7.592
user        11.233      0.349       10.593      11.292      11.751
sys         3.538       0.106       3.305       3.542       3.675
```

This is expected to be particularly useful for Pants, where often have loose interpreter constraints like the default of `>=3.6`. The only time we actually need to build for `>=3.6` is when running `binary`. All other times, we only need the first matching interpreter.

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -483,7 +483,7 @@ def configure_clp_pex_environment(parser):
         dest="python_shebang",
         default=None,
         help="The exact shebang (#!...) line to add at the top of the PEX file minus the "
-        "#!.  This overrides the default behavior, which picks an environment python "
+        "#!. This overrides the default behavior, which picks an environment Python "
         "interpreter compatible with the one used to build the PEX file.",
     )
 
@@ -517,6 +517,20 @@ def configure_clp_pex_environment(parser):
         "each platform specified. If found, use the interpreter to resolve distributions; if "
         "not, resolve for the platform only allowing matching binary distributions and failing "
         "if only sdists or non-matching binary distributions can be found.",
+    )
+
+    group.add_option(
+        "--use-first-matching-interpreter",
+        dest="use_first_matching_interpreter",
+        default=False,
+        action="callback",
+        callback=parse_bool,
+        help=(
+            "If multiple interpreters are valid, use the first one. Normally, when multiple "
+            "interpreters match, Pex will resolve requirements for each interpreter; this allows "
+            "the resulting Pex to work with more interpreters, such as different Python versions. "
+            "However, resolving for multiple interpreters results in worse performance."
+        ),
     )
 
     parser.add_option_group(group)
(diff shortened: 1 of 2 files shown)


pex/bin/pex.py
```diff
@@ -519,6 +519,20 @@ def configure_clp_pex_environment(parser):
...
+        "--use-first-matching-interpreter",
+        dest="use_first_matching_interpreter",
+        default=False,
+        action="callback",
+        callback=parse_bool,
+        help=(
+            "If multiple interpreters are valid, use the first one. Normally, when multiple "
+            "interpreters match, Pex will resolve requirements for each interpreter; this allows "
+            "the resulting Pex to work with more interpreters, such as different Python versions. "
+            "However, resolving for multiple interpreters results in worse performance."
```
Eric-Arellano, line 532: I struggled writing this. I wasn't sure how much detail to give about how Pex does interpreter matching, which is via a combination of `--python`, `--interpreter-constraint`, and `--platforms` (but only if `--resolve-local-platforms` is enabled).

I also wasn't sure if we should suggest when you might want this and when you might not care.

pex/bin/pex.py
```diff
@@ -766,6 +780,18 @@ def to_python_interpreter(full_path_or_basename):
             )
 
     interpreter = min(interpreters) if interpreters else None
+    if options.use_first_matching_interpreter and interpreters:
+        if len(interpreters) > 1:
+            unused_interpreters = set(interpreters) - {interpreter}
+            TRACER.log(
```
Eric-Arellano, line 786: FYI this is triggered with `-v`, i.e. level 1.

tests/test_integration.py
```diff
@@ -575,6 +575,33 @@ def test_interpreter_resolution_pex_python_path_precedence_over_pex_python():
...
+                "--interpreter-constraint=>=3.5",
+                "--use-first-matching-interpreter",
+                "-v",
+                "-o",
+                pex_out_path,
+            ], env=env
+        )
+        res.assert_success()
+        assert "using {}".format(py35_path) in res.error.decode()
+        assert "will not be used: {}".format(py36_path) in res.error.decode()
```
Eric-Arellano, line 599: This isn't as comprehensive as I'd like. All this verifies is that we're logging, but doesn't actually verify the behavior is correct.

An alternative idea is to build with a Pex that has distinct wheels for each Python version, e.g. `cp35m` and `cp36m` wheels. Then, try running that Pex with both Py35 and Py36. Py36 should fail. If we wanted to be really robust, we could do the same thing without `--use-first-matching-interpreters` and ensure that both Py35 and Py36 work that time.

pex/bin/pex.py
benjyw: I guess this should clarify what "first one" means. "First one it happens to encounter" or "Lowest version"?

Issue #999, opened 2020-07-10
Release 2.1.16


jsirois:
On the docket:
+ [x] Zipped PEXes don't contain directory entries. #1021 
+ [x] Path scrubbing not working for python 3.8 package installed with homebrew #1025
+ [x] Improve UnsatisfiableInterpreterConstraintsError. #1028

jsirois:
Moved these to #1013:
+ Pex skipping pandas activation #1017
+ Invalid requirement, parse error at "'python_v' #940
+ Decouple PEX runtime interpreter selection from buildtime interpreter selection. #1020

PR #1010, opened 2020-07-23
Prepare 2.1.14

Eric-Arellano:
Closes https://github.com/pantsbuild/pex/pull/1011

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,19 @@
 Release Notes
 =============
 
+2.1.14
+------
+
+This release adds the ``--use-first-matching-interpreter`` flag, which
+can speed up performance when building a Pex at the expense of being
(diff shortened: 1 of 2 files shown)


benjyw:
In pex, we have a "Release X.Y.Z" issue per release. John has already created one for 2.1.14: https://github.com/pantsbuild/pex/issues/999

Those issues haven't been addressed in your release, so what you should do is:
1. Edit https://github.com/pantsbuild/pex/issues/999 to change the version to 2.1.15. 
2. Create a "Release 2.1.14" issue similar to that one, mentioning just your change (and anything else of note that is different from 2.1.13).

benjyw:
And then link the issue and the corresponding PR, naturally.

Eric-Arellano:
PTAL.

PR #1014, opened 2020-07-28
Fix --use-first-matching-interpreter at runtime.

jsirois:
Restrict the runtime interpreter to match the buildtime interpreter
by the crude means of relying on the pythonX.Y shebang.

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -598,7 +598,9 @@ def test_use_first_matching_interpreter():
             # We do not attempt to update the PexInfo, even if `--use-first-matching-interpreter`
             # is used.
             pex_info = PexInfo.from_pex(pex_out_path)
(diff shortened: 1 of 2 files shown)


jsirois:
This should right the ship over in https://github.com/pantsbuild/pants/pull/10442. Unfortunately I won't be able to do the release. Sorry to leave the rest here in your hands. I'm not happy with this solution on the PEX side but there was not enough time to get the Pants side debuggability of Pants side shims up to par.

tests/test_integration.py
```diff
@@ -598,7 +598,9 @@ def run_pex_with_py36(use_first_matching_flag):
             # We do not attempt to update the PexInfo, even if `--use-first-matching-interpreter`
             # is used.
```
Eric-Arellano, line 599: Stale.

PR #1022, opened 2020-08-29
PEX zips now contain directory entries.

### LINKED ISSUE #1021: Zipped PEXes don't contain directory entries.
[PEP-420 implicit namespace packages](https://www.python.org/dev/peps/pep-0420/) require directories with no `__init__.py` children to implicitly declare a namespace package. As such, the zipfile created by the Pex CLI / PEXBuilder is invalid since it only contains file entries and not directory entries. This, for example, has caused Pants (see: https://github.com/pantsbuild/pants/pull/10608) to need to always mark all PEXes `--not-zip-safe` even when they otherwise are to force the PEX code portion to be exploded to disk, which implicitly creates parent directories on the filesystem if needed, satisfying PEP-420.

jsirois:
Previously they did not which led to PEP-420 namespace packages not
working in `--zip-safe` context.

Fixes #1021

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -1130,6 +1130,34 @@ def test_pex_source_bundling():
             assert stdout == b"hello\n"
 
 
+def test_pex_source_bundling_pep420():
+    with temporary_dir() as output_dir:
+        with temporary_dir() as input_dir:
+            with safe_open(os.path.join(input_dir, "a/b/c.py"), "w") as fh:
+                fh.write("GREETING = 'hello'")
+
+            with open(os.path.join(input_dir, "exe.py"), "w") as fh:
+                fh.write(
+                    dedent(
+                        """
+                        from a.b.c import GREETING
+
+                        print(GREETING)
+                        """
+                    )
+                )
+
+            pex_path = os.path.join(output_dir, "pex1.pex")
+            py36 = ensure_python_interpreter(PY36)
+            res = run_pex_command(["-o", pex_path, "-D", input_dir, "-e", "exe"], python=py36)
+            res.assert_success()
+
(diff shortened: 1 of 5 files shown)


tests/test_integration.py
```diff
@@ -1130,6 +1130,34 @@ def test_pex_source_bundling():
             assert stdout == b"hello\n"
 
 
+def test_pex_source_bundling_pep420():
+    with temporary_dir() as output_dir:
+        with temporary_dir() as input_dir:
+            with safe_open(os.path.join(input_dir, "a/b/c.py"), "w") as fh:
```
Eric-Arellano, line 1136: Nit, you can combine these to reduce nesting:

```python
with temporary_dir() as output_dir, temporary_dir() as input_dir():
```

This is fine to keep - only want to make sure you're aware.

tests/test_integration.py
```diff
@@ -1640,11 +1668,11 @@ def create_pex(path, seed):
...
             zf2.extractall(path=unzipped2)
             for member1, member2 in zip(sorted(zf1.namelist()), sorted(zf2.namelist())):
-                assert filecmp.cmp(
-                    os.path.join(unzipped1, member1),
-                    os.path.join(unzipped2, member2),
-                    shallow=False,
-                )
+                member1_path = os.path.join(unzipped1, member1)
+                if os.path.isdir(member1_path):
+                    continue
```
Eric-Arellano, line 1673: What's the reason for skipping? Are the directory entries not byte-for-byte identical?

jsirois: It's because `filecmp.cmp` considers anything not a regular file (symlink, directory, etc.) unequal.

tests/test_integration.py
jsirois: Ack. I'll stick with the copypasta from the test above for now.

PR #1023, opened 2020-08-29
Upgrade Black.

jsirois:
This fixes (arguably) incorrect docstring formatting (indentation).
Other changes are just part of ceding control.

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -525,7 +525,12 @@ def test_interpreter_constraints_honored_without_ppp_or_pp():
         pex_out_path = os.path.join(td, "pex.pex")
         env = make_env(
             PEX_IGNORE_RCFILES="1",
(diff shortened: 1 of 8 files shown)


tests/test_integration.py
```diff
@@ -1103,7 +1112,16 @@ def test_pex_source_bundling():
...
+            res = run_pex_command(
+                [
+                    "-o",
+                    pex_path,
+                    "-D",
+                    input_dir,
+                    "-e",
+                    "exe",
+                ]
+            )
```
Eric-Arellano, line 1124: FYI this is only because of the trailing comma. If you removed it, Black would keep things on the same line.

Not a big deal either way - only wanted to explain the new algorithm.

jsirois: OK, thanks. I'll let this stand.

PR #1024, opened 2020-08-29
Fix UnsatisfiableInterpreterConstraintsError.

jsirois:
Handle the case when there are no candidates and add a test for this.


jsirois:
Discovered working #1022.

jsirois:
> Oh, hm, this should still probably render the constraints, right?

It seemed to me they are irrelevant. If you find 0 interpreters period, constrinats are noise. There is a bigger problem afoot.

PR #1026, opened 2020-08-31
Scrub direct sys.path manipulations by .pth files.

### LINKED ISSUE #1025: Path scrubbing not working for python 3.8 package installed with homebrew
Hi there, I got an issue with a pex file embedding a dependency (protobuf) under OSX. I guess it could be related to the protobuf package and how the package is installed with brew but maybe also in the way the scrubbing is done by PEX.

If I install `protobuf` through `pip3.8 install protobuf` the scrubbing works fine because `protobuf` is inside the path `/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages//google/protobuf/`.

But if you install the `protobuf` brew package through `brew install protobuf`, it is installed under `/usr/local/Cellar/protobuf/` and the python package is registered inside `/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages//homebrew-protobuf.pth` with the following line:
```
import site; site.addsitedir('/usr/local/Cellar/protobuf/3.12.4/libexec/lib/python3.8/site-packages')
```

I put some logs to print `module.__path__` and `site_libs` before `cls._tainted_path(...)` is [called](https://github.com/pantsbuild/pex/blob/master/pex/pex.py#L209). Here is the difference between python3.7 and 3.8.

Python 3.7:
```
Module path: ['/Users/XXX/repack.pex/google', '/usr/local/Cellar/python@3.7/3.7.8_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/google']
Site_libs : {'/usr/local/lib/python3.7/site-packages', '/usr/local/Cellar/python@3.7/3.7.8_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages'}
pex: Scrubbing google.__path__: /usr/local/Cellar/python@3.7/3.7.8_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/google
```

Python 3.8:
```
Module path: ['/Users/XXX/repack.pex/google', '/usr/local/Cellar/protobuf/3.12.4/libexec/lib/python3.8/site-packages/google']
Site_libs : {'/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages', '/usr/local/lib/python3.8/site-packages'}

```

Notice there is no "Scrubbing" for python 3.8.  It is due to the change in where python package is located. Instead of being inside the python 3.8 directory, it is put as a standard `brew` package. The import is indirectly managed inside the `homebrew-protofbu.pth` file (see above for the path and content)

The `cls._tainted_path(...)` is then not able to to identify that the path matches in python 3.8 since they are completely different.

python 3.7 match
Module path : **/usr/local/Cellar/python@3.7/3.7.8_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages**/google
Site_libs : **/usr/local/Cellar/python@3.7/3.7.8_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages**

python 3.8 no match
Module path : /usr/local/Cellar/protobuf/3.12.4/libexec/lib/python3.8/site-packages/google
Site_libs : /usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages

Any solution ?

jsirois:
Previously direct (or indirect) sys.path manipulation via import lines
in .pth files would not be detected and scrubbed. We now detect these
modifications and scrub them as appropriate.

Fixes #1025

### DIFF
--- a/pex/util.py
+++ b/pex/util.py
@@ -5,6 +5,7 @@ from __future__ import absolute_import
 
 import contextlib
 import os
+import sys
 import tempfile
 from hashlib import sha1
 from site import makepath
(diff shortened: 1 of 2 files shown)


pex/util.py
```diff
@@ -198,21 +199,42 @@ def iter_pth_paths(filename):
...
+                # sys.path directly or indirectly as a programmatic way to add sys.path entries
+                # in contrast to the standard .pth mechanism of including fixed paths as
+                # individual lines in the file. Here we capture all such programmatic attempts
+                # to expand the sys.path and report the additions.
+                original_sys_path = sys.path[:]
                 try:
+                    # N.B.: Setting sys.path to empty is ok since all the .pth files we find and
+                    # execute have already been found and executed by our ambient sys.executable
+                    # when it started up before running this PEX file. As such, all symbols imported
+                    # by the .pth files then will still be available now as cached in sys.modules.
```
benjyw, line 217: Technically I think the .pth file could import some file that itself has a conditional import in code? Probably too esoteric to worry about, but might be worth a caveat in the comment if true.

jsirois: I've seen lots of conditional imports in .pth files, but all based on the version of the interpreter and that won't change here. In fact its hard to see what at all could be detected as different by the .pth between its execution 10ms ago duing sys.executable startup and here inside the same sys.executable. I think the only detectable difference is the fact we just blanked out sys.path.

jsirois: Similar - I think in the end unjustified - fears caused me to add the exception logging below though! Do you think that's enough?

benjyw: Yeah, I guess I was thinking of some weird corner case where we execute a conditional import in a function here, and it fails because we blanked out sys.path, but as you say, that would already have been executed during startup. Every line here has already been executed and its imports cached. So I think the logging is more than sufficient.

Issue #1027, opened 2020-09-01
Known interpreter identification issues.


jsirois:
Sometimes Pex may fail to identify and use an interpreter it finds on a system. This can often be due to the vagaries of how various operating systems and their packaging systems configure and install python. This issue tracks these known issues and workarounds.

jsirois:
Ubuntu systems are known to leave out (portions of) the `distutils` stdlib from their "standard" python package. For example, on Ubunutu 20.4 and many older Ubuntu distributions, installing the `python3` package is not enough, you must also install `python3-distlib`. The error seen when only `python3` is installed will look something like the error seen below in the 'Skipped the following broken interpreters:' section:
```
Could not find a compatible interpreter.

Examined the following working interpreters:
1.) /usr/bin/python2.7 CPython==2.7.18

Skipped the following broken interpreters:
1.) /usr/bin/python3.8:
Traceback (most recent call last):
  File "<string>", line 5, in <module>
  File "/workspace/pex/interpreter.py", line 24, in <module>
    from pex.third_party.packaging import markers, tags
  File "/workspace/pex/vendor/_vendored/packaging/packaging/tags.py", line 7, in <module>
    import distutils.util
ModuleNotFoundError: No module named 'distutils.util'

No working interpreter compatible with the requested constraints was found:
  CPython>=3.7
```

cupen:
My envs.
* archlinux
* python3.10 
```
$ wget https://github.com/pantsbuild/pants/releases/download/release_2.10.0/pants.2.10.0.pex
$ ./pants.2.10.0.pex --version
Failed to find compatible interpreter on path /home/cupen/.nimble/bin:/home/cupen/bin:/usr/local/bin:/usr/local/bin:/usr/bin:/usr/local/sbin:/home/cupen/.dotnet/tools:/var/lib/flatpak/exports/bin:/usr/lib/jvm/default/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl:/home/cupen/go/bin/:/home/cupen/.local/bin:/bin:/bin.


Examined the following working interpreters:
1.) /usr/bin/python3.10 CPython==3.10.4
2.)  /usr/bin/python2.7 CPython==2.7.18

Skipped the following broken interpreters:
1.) /usr/bin/python3.7:
InterpreterNotFound('The interpreter path /usr/bin/python3.7 does not exist.')

[...]

PR #1028, opened 2020-09-01
Improve UnsatisfiableInterpreterConstraintsError.

jsirois:
Surface all broken interpreters and a link to known breaks and
workarounds tracked in #1027.

### DIFF
--- a/pex/interpreter_constraints.py
+++ b/pex/interpreter_constraints.py
@@ -3,11 +3,12 @@
 
 # A library of functions for filtering Python interpreters based on compatibility constraints
 
-from __future__ import absolute_import
+from __future__ import absolute_import, print_function
+
+import os
 
 from pex.common import die
 from pex.interpreter import PythonIdentity
-from pex.tracer import TRACER
 
 
 def validate_constraints(constraints):
@@ -24,7 +25,7 @@ def validate_constraints(constraints):
 class UnsatisfiableInterpreterConstraintsError(Exception):
     """Indicates interpreter constraints could not be satisfied."""
 
-    def __init__(self, constraints, candidates):
+    def __init__(self, constraints, candidates, failures):
         """
         :param constraints: The constraints that could not be satisfied.
         :type constraints: iterable of str
@@ -33,6 +34,7 @@ class UnsatisfiableInterpreterConstraintsError(Exception):
         """
         self.constraints = tuple(constraints)
         self.candidates = tuple(candidates)
+        self.failures = tuple(failures)
         super(UnsatisfiableInterpreterConstraintsError, self).__init__(self.create_message())
 
     def create_message(self, preamble=None):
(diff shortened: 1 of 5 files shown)


pex/interpreter_constraints.py
```diff
@@ -46,53 +48,58 @@ def create_message(self, preamble=None):
         """
         preamble = "{}\n\n".format(preamble) if preamble else ""
         if not self.candidates:
+            if self.failures:
+                return (
+                    "{preamble}"
+                    "Interpreters were found but they all appear to be broken:\n  {failures}"
```
Eric-Arellano, line 54: This should probably link to https://github.com/pantsbuild/pex/issues/1027 as well.

pex/jobs.py
```diff
@@ -172,20 +188,108 @@ def _sanitize_max_jobs(max_jobs=None):
...
+                  processing of the item altogether.
+        :raise: To indicate all item processing should be cancelled and the exception raised.
+        """
+
+
+class Raise(ErrorHandler):
+    """Re-raises errors encountered spawning or awaiting the result of a `Job`."""
+
+    def __init__(self, raise_type):
+        """"""
```
Eric-Arellano, line 238: Stale.

pex/jobs.py
```diff
@@ -172,20 +188,108 @@ def _sanitize_max_jobs(max_jobs=None):
         return min(max_jobs, _ABSOLUTE_MAX_JOBS)
 
 
-def execute_parallel(inputs, spawn_func, raise_type=None, max_jobs=None):
+class ErrorHandler(AbstractClass):
```
Eric-Arellano, line 191: Great modeling.

tdyas:
This is good.

I would like to see this link to the pants.readme.io documentation instead of a GitHub issue. The installation issues should live near the installation documentation.

Eric-Arellano:
> link to the pants.readme.io documentation

Those docs are for Pants, not Pex. 

Pex docs are unfortunately extremely stale, as we don't have the keys to the site anymore: https://pex.readthedocs.io/en/stable/.

Benjy and I have discussed a 10% side project for me to add a dedicated Pex site for pants.readme.io. It would help Pex users and, by extension, Pants users.

tdyas:
> Pex docs are unfortunately extremely stale, as we don't have the keys to the site anymore: https://pex.readthedocs.io/en/stable/.

Understood. I was conflating the two sites then. Separately, we should recover access to the pex site. Seems like support is done through GitHub issues: https://docs.readthedocs.io/en/stable/contribute.html#initial-triage

pex/jobs.py
jsirois: Fixed.

pex/interpreter_constraints.py
jsirois: Fixed.

The full matrix is now:
1. None
    ```
    Could not find a compatible interpreter.

    No interpreters could be found on the system.
    ```

2. None - All broken:
    ```
    Could not find a compatible interpreter.

    Interpreters were found but they all appear to be broken:
    1.) /usr/bin/python3.8:
      File "<string>", line 5, in <module>
      File "/workspace/pex/interpreter.py", line 24, in <module>
        from pex.third_party.packaging import markers, tags
      File "/workspace/pex/vendor/_vendored/packaging/packaging/tags.py", line 7, in <module>
        import distutils.util
    ModuleNotFoundError: No module named 'distutils.util'

    (See https://github.com/pantsbuild/pex/issues/1027 for a list of known breaks and workarounds.)
    ```

3. Some but some broken:
    ```
    Could not find a compatible interpreter.

    Examined the following working interpreters:
    1.) /usr/bin/python2.7 CPython==2.7.18

    Skipped the following broken interpreters:
    1.) /usr/bin/python3.8:
      File "<string>", line 5, in <module>
      File "/workspace/pex/interpreter.py", line 24, in <module>
        from pex.third_party.packaging import markers, tags
      File "/workspace/pex/vendor/_vendored/packaging/packaging/tags.py", line 7, in <module>
        import distutils.util
    ModuleNotFoundError: No module named 'distutils.util'

    (See https://github.com/pantsbuild/pex/issues/1027 for a list of known breaks and workarounds.)

    No working interpreter compatible with the requested constraints was found:
      CPython>=3.7
    ```

4. Some and none broken:
    ```
    Could not find a compatible interpreter.

    1.) /usr/bin/python2.7 CPython==2.7.18
    2.) /usr/bin/python3.8 CPython==3.8.2

      CPython>=3.9
    ```

Eric-Arellano: Excellent!

pex/pex_bootstrapper.py
```diff
@@ -92,9 +89,32 @@ def _iter_interpreters():
...
+        raise UnsatisfiableInterpreterConstraintsError(constraints, candidates, failures)
+
+
 def _compatible_interpreters_iter(interpreters_iter, compatibility_constraints=None):
     if compatibility_constraints:
-        for interpreter in matched_interpreters_iter(interpreters_iter, compatibility_constraints):
+        for interpreter in _matched_interpreters_iter(interpreters_iter, compatibility_constraints):
             yield interpreter
     else:
         for interpreter in interpreters_iter:
```
jsirois, line 120: This branch does not handle the switch from `PythonInterpreter.iter` to `PythonInterpreter.iter_candidates` here: https://github.com/pantsbuild/pex/pull/1028/files#diff-670bc388096d1ef0c7994c4d37976392R82 so we can get error tuples mixed into the yielded interpreters when there are no compatibility constraints.

Eric-Arellano: > so we can get error tuples

This was something I was confused by when adding type hints. In addition to the type hints, we may want to rename this to `interpreter_or_failure`:

https://github.com/pantsbuild/pex/blob/aae2a6ad55e148ada43f36c1d486af9b2312af50/pex/pex_bootstrapper.py#L97-L109

jsirois: I think the problem is better fixed by restructuring for locality. The code that calls PythonInterpreter.iter_candidates should deal with the consequences of that directly.

jsirois: I did not rename since we don't use Hungaian elsewhere but this confusing bit of code where typing is path-dependent is now typed in #1046

PR #1032, opened 2020-09-19
Use MyPy via type comments

Eric-Arellano:
Note that we do not actually need to vendor `typing`, even though it's not provided by Python 2. Read the docstring in `pex/typing.py` to see how we use a similar approach to Pip so that the runtime does not use anything from `typing`, but MyPy is still able to work.

Once we require Python 3 to run, we can remove this facade and use the built-in `typing` without conditional imports.


pex/jobs.py
```diff
@@ -172,7 +172,7 @@ def __str__(self):
         return str(self._job)
 
 
-_CPU_COUNT = cpu_count()
+_CPU_COUNT = cpu_count() or 2
```
cosmicexplorer, line 175: Might add a comment here saying "2 was an arbitrary choice, to avoid being blocked by one long-running job", or something.

pex/typing.py
```diff
@@ -0,0 +1,10 @@
+from __future__ import absolute_import
+
+# We first try to re-export from the std lib for Python 3. If that fails, we're on Python 2 so used
```
cosmicexplorer, line 3:
```suggestion
# We first try to re-export from the std lib for Python 3. If that fails, we're on Python 2, so use
```

tox.ini
```diff
@@ -144,24 +144,30 @@ commands =
...
 
+[testenv:typecheck]
+deps =
+    mypy==0.782
+commands =
+    mypy pex/ tests/
+    mypy --py2 pex/ tests/
+
 [testenv:vendor]
+# We use Python 2.7 because we need to vendor `typing` with Python 2.7. The others do not matter,
```
cosmicexplorer, line 155: Maybe mention that if we don't, we get a python 3 dist without any packages.

jsirois:
I'm about to be offline until the 24th, but wanted to make sure you're considering the `.bootstrap/`. I had not realized this would require adding a distribution to Pex. That's not a big deal for the tool, but it could be for the runtime of the PEX files it generates. If you could run an experiment to determine both the runtime performance hit of importing the typing stuff use in `.bootstrap/` code and the size hit (less important), that would be wise. Runtime overhead of the existing `.bootstrap/` is already problematic and has been worked on by DWH.

cosmicexplorer:
As a random thought: since we already perform rewriting (which affects `pex/*` code as well, not just vendored code), is it possible we could use the rewriting tool to remove `typing` imports when we package up the pex code into a PEX file? This is possible precisely *because* we still use type comments instead of type hints. More specifically:

We have so far exactly one `typing` import: https://github.com/pantsbuild/pex/blob/449e1c54d7e8f565c01e0d7229ce17a5c209959f/pex/pex_bootstrapper.py#L16

While we want to have those imports untouched when running `tox -e typecheck{,2}`, we could make sure to remove them specifically by running another `RedBaron()` invocation within `scripts/package.py`.

I'll take a look at this now!!

cosmicexplorer:
Oh...or we could just do what pip does: https://github.com/pantsbuild/pex/blob/2783d304daf43dc1d59753c7e5fc2599b16acd01/pex/vendor/_vendored/pip/pip/_internal/utils/typing.py#L29

The rest of that file explains how this avoids having `typing` at runtime. We may not even need to remove `typing` from the final dist that pex outputs (which would just keep the packaging script much simpler).

cosmicexplorer:
> If you could run an experiment to determine both the runtime performance hit of importing the typing stuff used in .bootstrap/ code and the size hit (less important), that would be wise.

3b3d38b adopts pip's solution with `MYPY_CHECK_RUNNING` to avoid a runtime performance hit (it would still be good to run the performance experiment to check). The remaining issue is the size hit, which should be determinable by gauging the size of `pex/typing.py` and the directory `pex/vendor/_vendored/typing/`.

cosmicexplorer:
Was able to remove the `typing` vendoring entirely after looking more closely at pip's solution!!! Will see if this passes tests.

pex/compatibility.py
```diff
@@ -8,28 +8,35 @@
```
Eric-Arellano, line 8: This file has some new noise because of explicit re-exports. We configured MyPy to require you to explicitly re-export, like we do with Pants. This seems to have been a very useful check in Pants, as it makes it easier to see where imports really are coming from.

pex/compatibility.py
```diff
@@ -100,32 +108,33 @@ def nested(*context_managers):
...
+    from queue import Queue as Queue
 
     # The `os.sched_getaffinity` function appears to be supported on Linux but not OSX.
     if not hasattr(os, "sched_getaffinity"):
-        from os import cpu_count
+        from os import cpu_count as cpu_count
     else:
 
         def cpu_count():
+            # type: () -> Optional[int]
```
Eric-Arellano, line 129: Technically, this could be `-> int`, but MyPy requires that the function signature is identical to the `if not hasattr` branch.

Lmk if I should comment this.

pex/compatibility.py
```diff
@@ -8,28 +8,35 @@
 
 import os
 from abc import ABCMeta
-from io import StringIO
 from sys import version_info as sys_version_info
 
+from pex.typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from typing import Optional
```
Eric-Arellano, line 16: This is the key part of the change, as explained in `pex/typing.py`.

Note that `from typing import TYPE_CHECKING` would fail on Py2 because `typing` is not defined.

pex/interpreter.py
```diff
@@ -259,7 +263,7 @@ class PythonInterpreter(object):
         re.compile(r"pypy-1.[0-9]$"),
     )
 
-    _PYTHON_INTERPRETER_BY_NORMALIZED_PATH = {}
+    _PYTHON_INTERPRETER_BY_NORMALIZED_PATH = {}  # type: Dict
```
Eric-Arellano, line 266: Will add a proper type hint later once this file has more hints.

pex/util.py
```diff
@@ -167,7 +170,7 @@ def cache_distribution(cls, zf, source, target_dir):
                             zf.extract(name, target_dir_tmp)
 
         dist = DistributionHelper.distribution_from_path(target_dir)
-        assert dist is not None, "Failed to cache distribution ".format(source)
+        assert dist is not None, "Failed to cache distribution: {} ".format(source)
```
Eric-Arellano, line 173: Bug

tests/test_environment.py
```diff
@@ -320,7 +320,7 @@ def run(args, **env):
                 "STDOUT:\n"
                 "{}\n"
                 "STDERR:\n"
-                "{}".format(stdout, stdout, stderr)
+                "{}".format(stdout, stderr)
```
Eric-Arellano, line 323: Bug

Eric-Arellano:
Enormous thanks to @cosmicexplorer for pairing on this and for figuring out that we didn't need to vendor `typing`!

.isort.cfg
```diff
@@ -1,5 +0,0 @@
-[settings]
```
Eric-Arellano, line 1: This should have been deleted previously. We now have the settings in `pyproject.toml`.

mypy.ini
```diff
@@ -0,0 +1,53 @@
...
+python_version: 3.5
+
+# TODO: turn on.
+check_untyped_defs = False
+
+no_implicit_optional = True
+
+# TODO: enable once we drop Python 2. Otherwise, we need it because some ignores depend on which
+#  interpreter we use.
+warn_redundant_casts = False
```
cosmicexplorer, line 12: Does that apply to just this warning, or the top 3? What about the next 3?

pex/compatibility.py
```diff
@@ -8,28 +8,35 @@
...
 try:
-    # Python 2.x
-    from ConfigParser import ConfigParser
-except ImportError:
     # Python 3.x
-    from configparser import ConfigParser
+    from configparser import ConfigParser as ConfigParser  # type: ignore[import]
+except ImportError:
+    # Python 2.x
+    from ConfigParser import ConfigParser as ConfigParser  # type: ignore[no-redef]
```
cosmicexplorer, line 24: Note that this reordering was necessary, as the alternative was impossible to fix by adding ignores.

cosmicexplorer: Well, this could probably be switched back now that we've unified the `mypy.ini` for both versions.

pex/compatibility.py
```diff
@@ -85,6 +92,7 @@ def exec_function(ast, globals_map):
 
 
 else:
+    # This will result in `exec_function` being defined at runtime.
```
cosmicexplorer, line 95: I'm concerned that this will mean type check errors wherever it's used. Is that the case? If so I feel like a no-op might make sense. But if it's used in only one or two places then this is fine.

pex/compatibility.py
cosmicexplorer: I think this is fine, that should be expected for anything in an `if` block like this.

pex/typing.py
```diff
@@ -0,0 +1,41 @@
...
+
+To add type comments, use a conditional import like this:
+
+    ```
+    from pex.typing import TYPE_CHECKING
+
+    if TYPE_CHECKING:
+        from typing import Optional, ...
+    ```
+"""
```
cosmicexplorer, line 28: Great great editing of this section!

pex/compatibility.py
cosmicexplorer: I think not redefining the function with a no-op (as you've done here) is probably safer, in case it messes up how the function is defined. Any type ignores can be removed after dropping py2.

cosmicexplorer (approved):
SUPER glad this worked out without the ridiculous vendoring scheme!!!

mypy.ini
Eric-Arellano: Top 3. I'll clarify.

pex/compatibility.py
Eric-Arellano: Didn't realize it was reordered. I'll try to make the diff smaller.

pex/compatibility.py
Eric-Arellano: > I think not redefining the function with a no-op (as you've done here) is probably safer, in case it messes up how the function is defined. Any type ignores can be removed after dropping py2.

Exactly. This code was written in 2013 and is only used in one single place. I think it's unlikely we'll start using it more.

pex/compatibility.py
Eric-Arellano: Cool. And I figure an eager-beaver like me would try deleting it, then MyPy will complain and they'll realize why it's needed hah.

stuhood (approved):
Looks great: thanks!

Is it worth adding a snippet somewhere (in the README.md or contributor guide?) describing how to run mypy locally?

pex/util.py
luislew: i came here to see if this was still an issue, glad to see it's fixed!

PR #1036, opened 2020-09-22
Add type hints to most tests

Eric-Arellano:
Now all tests have full coverage except for `test_environment.py`, `test_pex.py`, and `test_integration.py`, which are saved for a dedicated followup.

Why start with tests? These are the call-sites of much of our production code, and this will help to ensure that when we add types to our production code, the hints reflect reality. The tests also are "roots", which makes it easier to add hints to than adding hints to code with lots of dependees.

### `str` vs. `bytes` vs. `Text` vs. `AnyStr`

Python 2 lets you freely mix unicode and bytes :/ Pex uses this abundantly, in contrast to Pants's (awkward) use of `future` to ensure that Py2 always used the exact same type as Py3.

Originally, this PR added `from __future__ import unicode_literals`, but that was reverted to instead optimize for fewer code changes. PEX's Py2 support works well as is, and it's well tested. It's not worth the energy to do a Pants-style migration that ensures Py2 behaves like Py3.

Instead, these hints try to capture the truth, as follows:
* `bytes` -> when both Py2 and Py3 use `bytes`
* `str` -> when Py2 uses `bytes`, but Py3 uses unicode
* `Text` -> when both use unicode. This only really happens when we explicitly call `.decode('utf-8')`
* `AnyStr` -> when it can be either unicode or bytes; shorthand for `Union`

Once we require Py3 to run, `Text` and `AnyStr` will both be converted into simply `str` (i.e. unicode).

### DIFF
--- a/pex/compatibility.py
+++ b/pex/compatibility.py
@@ -13,7 +13,7 @@ from sys import version_info as sys_version_info
 from pex.typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
-    from typing import Optional
+    from typing import Optional, AnyStr, Text
 
 
 try:
@@ -29,8 +29,6 @@ PY2 = sys_version_info[0] == 2
 PY3 = sys_version_info[0] == 3
 
 string = (str,) if PY3 else (str, unicode)  # type: ignore[name-defined]
-unicode_string = (str,) if PY3 else (unicode,)  # type: ignore[name-defined]
-bytes = (bytes,)  # type: ignore[has-type]
 
 if PY2:
     from collections import Iterable as Iterable
@@ -42,6 +40,7 @@ else:
 if PY2:
 
     def to_bytes(st, encoding="utf-8"):
+        # type: (AnyStr, Text) -> bytes
         if isinstance(st, unicode):
             return st.encode(encoding)
         elif isinstance(st, bytes):
@@ -50,6 +49,7 @@ if PY2:
             raise ValueError("Cannot convert %s to bytes" % type(st))
 
     def to_unicode(st, encoding="utf-8"):
+        # type: (AnyStr, Text) -> Text
         if isinstance(st, unicode):
             return st
         elif isinstance(st, (str, bytes)):
@@ -61,6 +61,7 @@ if PY2:
 else:
 
     def to_bytes(st, encoding="utf-8"):
+        # type: (AnyStr, Text) -> bytes
         if isinstance(st, str):
             return st.encode(encoding)
         elif isinstance(st, bytes):
@@ -69,6 +70,7 @@ else:
             raise ValueError("Cannot convert %s to bytes." % type(st))
 
     def to_unicode(st, encoding="utf-8"):
+        # type: (AnyStr, Text) -> Text
         if isinstance(st, str):
             return st
         elif isinstance(st, bytes):
--- a/pex/pex_info.py
+++ b/pex/pex_info.py
@@ -10,9 +10,14 @@ from pex import pex_warnings
 from pex.common import can_write_dir, open_zip, safe_mkdtemp
 from pex.compatibility import PY2
 from pex.compatibility import string as compatibility_string
+from pex.interpreter import PythonInterpreter
 from pex.orderedset import OrderedSet
+from pex.typing import TYPE_CHECKING
 from pex.variables import ENV
 from pex.version import __version__ as pex_version
+
+if TYPE_CHECKING:
+    from typing import Optional
 
 
 # TODO(wickman) Split this into a PexInfoBuilder/PexInfo to ensure immutability.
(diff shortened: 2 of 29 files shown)


pex/compatibility.py
```diff
@@ -29,8 +29,6 @@
 PY3 = sys_version_info[0] == 3
 
 string = (str,) if PY3 else (str, unicode)  # type: ignore[name-defined]
-unicode_string = (str,) if PY3 else (unicode,)  # type: ignore[name-defined]
-bytes = (bytes,)  # type: ignore[has-type]
```
Eric-Arellano, line 33: This shadowed the builtin `bytes`, which messed up below return types. It was also unused.

pex/compatibility.py
```diff
@@ -29,8 +29,6 @@
 PY3 = sys_version_info[0] == 3
 
 string = (str,) if PY3 else (str, unicode)  # type: ignore[name-defined]
-unicode_string = (str,) if PY3 else (unicode,)  # type: ignore[name-defined]
```
Eric-Arellano, line 32: Only used in tests for this file. Now inlined.

tests/test_inherits_path_option.py
```diff
@@ -26,26 +31,23 @@ def write_and_run_simple_pex(inheriting=False):
         pb.set_executable(os.path.join(td, "exe.py"))
         pb.freeze()
         pb.build(pex_path)
-        yield run_simple_pex(pex_path, env={"PEX_VERBOSE": "1"})[0]
+        stdout, _ = run_simple_pex(pex_path, env={"PEX_VERBOSE": "1"})
+        yield stdout.decode("utf-8")
```
Eric-Arellano, line 35: We didn't call `.decode()` before. Instead, we used `str()` on the object at each callsite. This is more explicit

tests/test_resolver.py
```diff
@@ -35,12 +36,17 @@
     make_source_dir,
 )
 from pex.third_party.pkg_resources import Requirement
+from pex.typing import TYPE_CHECKING
 
+if TYPE_CHECKING:
+    from typing import Any, List, Union
 
-def create_sdist(*args, **kwargs):
```
Eric-Arellano, line 40: Only kwargs were being used.

tests/test_resolver.py
```diff
@@ -177,10 +193,14 @@ def assert_resolve(expected_version, **resolve_kwargs):
 
 
 def _parse_requirement(req):
-    return Requirement.parse(str(req))
+    # type: (Union[str, Requirement]) -> Requirement
+    if isinstance(req, Requirement):
+        req = str(req)
+    return Requirement.parse(req)
```
Eric-Arellano, line 199: This was changed to be more explicit.

tests/test_unified_install_cache.py
```diff
@@ -200,7 +203,7 @@ def run():
             "--always-write-cache",
             "--pex-root",
             ptex_cache,
-            pex_project_dir,
+            pex_project_dir,  # type: ignore[list-item]  # This is unicode in Py2, whereas everthing else is bytes. That's fine.
```
Eric-Arellano, line 206: Gross

tests/test_util.py
```diff
@@ -10,20 +10,25 @@
...
     import mock  # type: ignore[no-redef]
 
+if TYPE_CHECKING:
+    from typing import Any
+
 
 def test_hash():
-    empty_hash = sha1().hexdigest()
+    # type: () -> None
+    empty_hash_digest = sha1().hexdigest()
```
Eric-Arellano, line 27: MyPy didn't like that the variable was being reassigned later with a different type.

pex/pex_info.py
```diff
@@ -51,8 +56,6 @@ class PexInfo(object):
 
     @classmethod
     def make_build_properties(cls, interpreter=None):
-        from .interpreter import PythonInterpreter
```
cosmicexplorer, line 54: This import may have been intentionally lazy (although it's not written down). I might recommend keeping it in this block, although making it absolute instead of relative.

cosmicexplorer (approved):

pex/pex_info.py
Eric-Arellano: I was looking if the lazy import would actually end up mattering in production. It looks like no. It doesn't impact the imports of this file; if you only ever import `pex_info.py`, then you end up transitively importing way more than before.

But the only two places we import this file in production already import a ton of other things, `resolver.py` and `pex/pex.py`. So this shouldn't matter.

cosmicexplorer: In general, I would normally say if there's no comment, feel free to kill it, but I'm concerned about performance regressions since @jsirois isn't around. If he could make sure of that when he gets back that would be great, but otherwise we can 🚢 for now!

cosmicexplorer (approved):

PR #1038, opened 2020-09-23
Expose the ability to add a local wheel through the CLI.

hrfuller:
Pex already has the ability to add distributions from pre downloaded locations. This change will allow users to use wheel resolving and caching strategies outside of the pex, and then build canonically formed PEXes from external requirement sets.

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -637,7 +637,7 @@ def configure_clp():
         default=[],
         type=str,
         action="append",
-        help="Add a wheel distribution (.whl) from a local path into the generated .pex file."
+        help="Add a wheel distribution (.whl file) from a local path (relative or absolute) into the generated .pex file."
         "  This option can be used multiple times.",
     )
 
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -630,6 +630,18 @@ def configure_clp():
     )
 
     parser.add_option(
+        "-w",
+        "--wheel-dist",
+        dest="wheel_dists",
+        metavar="FILE",
+        default=[],
+        type=str,
+        action="append",
+        help="Add a wheel distribution (.whl) from a local path into the generated .pex file."
+        "  This option can be used multiple times.",
+    )
+
+    parser.add_option(
         "-R",
         "--resources-directory",
         dest="resources_directory",
@@ -853,6 +865,9 @@ def build_pex(reqs, options, cache=None):
 
     for requirements_pex in options.requirements_pexes:
         pex_builder.add_from_requirements_pex(requirements_pex)
+
+    for whl_path in options.wheel_dists:
+        pex_builder.add_dist_location(whl_path)
 
     with TRACER.timed("Resolving distributions ({})".format(reqs + options.requirement_files)):
         network_configuration = NetworkConfiguration.create(


pex/bin/pex.py
```diff
@@ -629,6 +629,18 @@ def configure_clp():
...
+    parser.add_option(
+        "-w",
+        "--wheel-dist",
+        dest="wheel_dists",
+        metavar="FILE",
+        default=[],
+        type=str,
+        action="append",
+        help="Add a wheel distribution (.whl) from a local path into the generated .pex file."
+        "  This option can be used multiple times.",
```
Eric-Arellano, line 641: Might be helpful to explain more about the expectation. Can this be a `.whl` _file_, or it should be an expanded folder?

Eric-Arellano: Relative path or absolute path?

Eric-Arellano (approved):
This sounds like a good idea. John gets back tomorrow and will probably have some good thoughts on this.

cosmicexplorer:
Is this not already possible? e.g. with:
```bash
> pex -f "$(dirname my-requirement-2.0.0-py2-py3.whl)" my-requirement==2.0.0
```

Eric-Arellano:
Good point Danny. See the blue tool tip in https://www.pantsbuild.org/docs/python-third-party-dependencies#basic-setup about how to install from local files in general with Pex.

pex/bin/pex.py
hrfuller: I think the fact that the metavar says it should be a file is enough to distinguish that it isn't an expanded folder, and the fact that the .whl suffix is suggested. I'll make a change to indicate it can be relative or absolute, and to canonicalize any relative paths.

hrfuller:
@Eric-Arellano @cosmicexplorer The main difference I see is that the -w option is supposed to be to stuff some .whl into the pex *without* invoking the pex resolver. Passing the requirement specifier and a --find-links still has pex run its resolver for that requirement, which I want to avoid. And the pex will fail to build if any of the `Requires-Dist` in that wheel can't also be found.

pex/bin/pex.py
hrfuller: Actually the `PEXBuilder.add_dist_location` method that does the work here already deals with relative paths. I'll change the comment.

hrfuller:
I see the PEP440 direct references in the tooltip you suggested. That combined with --no-transitive works for me. Thanks!

cosmicexplorer:
Oh -- thanks for pointing that out! Glad that the use case is still already supported ^_^

jsirois:
Thanks to all for realizing this was an existing feature. The feature actually predates using Pip (where Pex gained support for PEP-440). Since Pip supports a non-PEP'd local path just like Pex used to before adopting Pip internally, this always worked without interuption during the Pip transition.

So long ago (two years ago to the day in this example), you could do the simplest thing imaginable:
```
$ ls -l
total 112
-rw-r--r-- 1 jsirois jsirois 69546 Dec 11  2019 isort-4.3.21.tar.gz
-rw-r--r-- 1 jsirois jsirois 43054 Jul 11 08:04 more_itertools-8.4.0-py3-none-any.whl
$ curl -sSL https://github.com/pantsbuild/pex/releases/download/v1.4.7/pex36 > pex1.4.7.pex
$ chmod +x pex1.4.7.pex 
$ cat << EOF | ./pex1.4.7.pex more_itertools-8.4.0-py3-none-any.whl isort-4.3.21.tar.gz
import isort
import more_itertools

print(f"isrt={isort}\nmore_itertools={more_itertools}")
EOF
Python 3.6.11 (default, Aug  7 2020, 14:31:03) 
[GCC 10.1.0] on linux
(InteractiveConsole)
>>> >>> >>> >>> isrt=<module 'isort' from '/tmp/tmpjvlxwkdp/.deps/isort-4.3.21-py2.py3-none-any.whl/isort/__init__.py'>
more_itertools=<module 'more_itertools' from '/tmp/tmpjvlxwkdp/.deps/more_itertools-8.4.0-py3-none-any.whl/more_itertools/__init__.py'>
>>> 
```

Just like you can today:
```
$ pex --version
pex 2.1.16
$ pex --intransitive more_itertools-8.4.0-py3-none-any.whl isort-4.3.21.tar.gz -- -c 'import isort, more_itertools; print(f"{isort=}\n{more_itertools=}")'
isort=<module 'isort' from '/tmp/tmpftai414k/.deps/isort-4.3.21-py2.py3-none-any.whl/isort/__init__.py'>
more_itertools=<module 'more_itertools' from '/tmp/tmpftai414k/.deps/more_itertools-8.4.0-py3-none-any.whl/more_itertools/__init__.py'>
```

hrfuller:
Because you were kind enough to provide that excellent description, I want to put it on the record that PEP440 direct references don't support relative paths like pex. So I just resorted to using the example you gave. This issue https://github.com/pypa/pip/issues/6658 seems to be at the core of it and pip hasn't addressed it yet. Hopefully they will and PEX and pip can standardize on direct references.

PR #1039, opened 2020-09-24
add a -V VERBOSITY option accepting an explicit nonnegative int

### LINKED ISSUE #988: allow -v9 abbrevation on the command line for -vvvvvvvvv
It might be nice to be able to write `-v9` when invoking pex on the command line instead of painstakingly typing out all 9 characters of `-vvvvvvvvv`.

cosmicexplorer:
Closes #988. The option `-V` now more-or-less corresponds to the `PEX_VERBOSE` environment variable when executing a PEX file. Further `-v` arguments will continue to increment the verbosity, while further `-V` arguments will reset it to the specific value indicated.

Feel free to close this if it introduces unnecessary choice to the CLI.

```bash
> pex --help
...
  -v                    Turn on logging verbosity, may be specified multiple
                        times.
  -V VERBOSITY          Set or reset logging verbosity to a specific
                        nonnegative number (>= 0). A -v argument after a -V
                        will increment the verbosity further.
...
> pex -V1
pex: Building pex: 2.8ms
pex:   Resolving distributions ([]): 0.3ms
...
> pex -V0 -v
pex: Building pex: 1.4ms
pex:   Resolving distributions ([]): 0.2ms
...
> pex -v -V0
Python 3.6.11 (default, Sep 19 2020, 01:11:19)
[GCC Apple LLVM 12.0.0 (clang-1200.0.32.2)] on darwin
(InteractiveConsole)
>>>
```

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -50,8 +50,12 @@ def increment_verbosity(option, opt_str, _, parser):
     setattr(parser.values, option.dest, verbosity + 1)
 
 
-def process_disable_cache(option, option_str, option_value, parser):
-    setattr(parser.values, option.dest, None)
+def process_explicit_verbosity(option, _, option_value, parser):
+    if option_value < 0:
+        raise OptionValueError(
+            "Please specify a nonnegative integer; ie: >=0. Given {}".format(option_value)
+        )
+    setattr(parser.values, option.dest, option_value)
 
 
 class PyPiSentinel(object):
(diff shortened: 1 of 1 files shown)


jsirois:
> Feel free to close this if it introduces unnecessary choice to the CLI.

I think it does. Especially since you could achieve this under the existing `-v` option using the existing `increment_verbosity` callback by examining `parser.rargs`. That would get you everything you have here except that `-v9` would need to be spelled `-v 9` (dash-v-space-9). To get `-v9` support too I think you'd need to create hidden (`optparse.SUPPRESS_HELP`) `-0`, `-1`, ... `-9` options.

cosmicexplorer:
Thank you for the extraordinarily thorough response! I will dig into that.

cosmicexplorer:
> a really hard examination of is this really worth its weight?

I don't know that it is! It's extremely easy to automate the 9x<kbd>v</kbd> insertion in emacs, anyway. I think I agree with your appraisement that `-V` is too much choice, and I think the alternative with the `-0`/`-1`/etc options introduces a rather large amount of complexity in CLI parsing (not even in the implementation but rather the complexity in understanding as a user).

I think `-v 9` may be a sweet spot, but I would have to go through adding testing for a bit. As there are a few other things going on in my head right now I may elect to close this given the above information and resume it later if I feel like it might be worth it.

PR #1040, opened 2020-09-25
Add type hints to the remaining tests

Eric-Arellano:
Now that we have full type coverage for the tests, we'll get useful validation when adding hints to our source code that the hints line up.

### DIFF
--- a/pex/common.py
+++ b/pex/common.py
@@ -22,7 +22,7 @@ from uuid import uuid4
 from pex.typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
-    from typing import Any
+    from typing import Any, Iterator, NoReturn, Optional, Tuple, Union
 
 
 # We use the start of MS-DOS time, which is what zipfiles use (see section 4.4.6 of
@@ -33,12 +33,15 @@ DETERMINISTIC_DATETIME = datetime(
 
 
 def die(msg, exit_code=1):
+    # type: (str, int) -> NoReturn
     print(msg, file=sys.stderr)
     sys.exit(exit_code)
 
 
 def safe_copy(source, dest, overwrite=False):
+    # type: (str, str, bool) -> None
     def do_copy():
+        # type: () -> None
         temp_dest = dest + uuid4().hex
         shutil.copy(source, temp_dest)
         os.rename(temp_dest, dest)
@@ -169,6 +172,7 @@ def open_zip(path, *args, **kwargs):
 
 @contextlib.contextmanager
 def temporary_dir(cleanup=True):
+    # type: (bool) -> Iterator[str]
     td = tempfile.mkdtemp()
     try:
         yield td
@@ -193,6 +197,7 @@ def register_rmtree(directory):
 
 
 def safe_mkdir(directory, clean=False):
+    # type: (str, bool) -> None
     """Safely create a directory.
 
     Ensures a directory is present.  If it's not there, it is created.  If it is, it's a no-op. If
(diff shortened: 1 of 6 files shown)


pex/common.py
```diff
@@ -376,25 +386,11 @@ def can_write_dir(path):
     return os.path.isdir(path) and os.access(path, os.R_OK | os.W_OK | os.X_OK)
 
 
-def touch(file, times=None):
```
Eric-Arellano, line 379: We never used the `times` kwarg. It's really tricky to get right with type hints, so I took the liberty to simplify this function.

I figure that we can add back the `times` part if we need it.

pex/testing.py
```diff
@@ -83,31 +100,16 @@ def temporary_content(content_map, interp=None, seed=31337, perms=0o644):
...
-            filename = os.path.join(root, f)
-            rel_filename = os.path.relpath(filename, directory)
-            yield filename, rel_filename
-
-
-def write_zipfile(directory, dest, reverse=False):
-    with open_zip(dest, "w") as zf:
-        for filename, rel_filename in sorted(yield_files(directory), reverse=reverse):
-            zf.write(filename, arcname=rel_filename)
-    return dest
```
Eric-Arellano, line 98: Unused

pex/testing.py
```diff
@@ -135,7 +137,7 @@ def make_project(
         os.path.join(name, "my_module.py"): 'def do_something():\n  print("hello world!")\n',
         os.path.join(name, "package_data/resource1.dat"): 1000,
         os.path.join(name, "package_data/resource2.dat"): 1000,
-    }
+    }  # type: Dict[str, Union[str, int]]
```
Eric-Arellano, line 140: This is the best we can do with type casting. We can't use `cast()` because we can't use values like `Dict` at runtime.

tests/test_integration.py
```diff
@@ -957,6 +996,7 @@ def test_pex_multi_resolve_2():
 
 @contextmanager
 def pex_manylinux_and_tag_selection_context():
+    # type: () -> Iterator[Tuple[Callable[[str, str, str, str, Optional[str]], None], Callable[[str, str, str, str], None]]]
```
Eric-Arellano, line 999: Absolutely disgusting. But we can't use type aliases because those require setting variables at runtime, and that would break Python 2.

tests/test_pex.py
```diff
@@ -34,22 +34,28 @@
...
+if TYPE_CHECKING:
+    from typing import Dict, Iterator, Union
+
 
 def test_pex_uncaught_exceptions():
+    # type: () -> None
     body = "raise Exception('This is an exception')"
     so, rc = run_simple_pex_test(body)
-    assert b"This is an exception" in so, "Standard out was: %s" % so
+    assert b"This is an exception" in so, "Standard out was: %r" % so
```
Eric-Arellano, line 53: MyPy warns that this will cause you to bring `b"abc"`, with the `b` prefix, when using Py3. It wants us to be explicit about that.

tdyas (approved):
lgtm

tests/test_integration.py
jsirois: This is untrue. You can define the inside the `if TYPE_CHECKING` guard, and, if you need to cast to them, pass those type aliases to cast as strings.

pex/testing.py
jsirois: See above, but this is untrue, you just need to pass the type to cast as a string.

Issue #1041, opened 2020-09-25
Use dataclasses (or `typing.NamedTuple`) once Python 3-only


Eric-Arellano:
`collections.namedtuple` does not work with MyPy. We cannot use `typing.NamedTuple` because it's Python 3-only. So, we temporarily use our own classes.

Once we require Python 3 to run Pex, we can go back to either dataclasses or `typing.NamedTuple`.

cosmicexplorer:
I like `@dataclass`es a lot, especially because they already accept a wide range of configuration, which may incorporate even more features in later python versions! This is what cleared my concern entirely about using them in pants vs the homegrown `datatype` -- it's something that won't cause churn with structs that many people will be editing over time! We also have code in pants to draw from to ensure we write pythonic code (or at least, pants-pythonic code).
 
Being able to switch on/off whether fields are frozen is a really nicely done feature in itself. Is there a reason we might want to consider `typing.NamedTuple` instead?

Eric-Arellano:
Superseded by https://github.com/pantsbuild/pex/pull/1234.

PR #1042, opened 2020-09-25
Add type hints to variables.py and platforms.py

Eric-Arellano:
Unfortunately, so long as we require Python 2, we cannot use `namedtuple` robustly with MyPy because it does not support annotating the fields of a named tuple. So, we recreate namedtuple through our own conventional class. This can be fixed once we require Python 3 to run. See https://github.com/pantsbuild/pex/issues/1041.

### DIFF
--- a/pex/dist_metadata.py
+++ b/pex/dist_metadata.py
@@ -7,18 +7,21 @@ from __future__ import absolute_import
 import email
 
 from pex.third_party.packaging.specifiers import SpecifierSet
-from pex.third_party.pkg_resources import DistInfoDistribution
+from pex.third_party.pkg_resources import DistInfoDistribution, Distribution
+from pex.typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from typing import Optional
 
 
 def requires_python(dist):
+    # type: (Distribution) -> Optional[SpecifierSet]
     """Examines dist for `Python-Requires` metadata and returns version constraints if any.
 
     See: https://www.python.org/dev/peps/pep-0345/#requires-python
 
     :param dist: A distribution to check for `Python-Requires` metadata.
-    :type dist: :class:`pkg_resources.Distribution`
     :return: The required python version specifiers.
-    :rtype: :class:`packaging.specifiers.SpecifierSet` or None
     """
     if not dist.has_metadata(DistInfoDistribution.PKG_INFO):
         return None
--- a/pex/pex_warnings.py
+++ b/pex/pex_warnings.py
@@ -6,13 +6,20 @@ from __future__ import absolute_import
 
 import warnings
 
+from pex.typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from pex.pex_info import PexInfo
+    from pex.variables import Variables
+
 
 class PEXWarning(Warning):
     """Indicates a warning from PEX about suspect buildtime or runtime configuration."""
 
 
 def configure_warnings(pex_info, env):
-    if env.PEX_VERBOSE > 0:
+    # type: (PexInfo, Variables) -> None
+    if env.PEX_VERBOSE is not None and env.PEX_VERBOSE > 0:
         emit_warnings = True
     elif env.PEX_EMIT_WARNINGS is not None:
         emit_warnings = env.PEX_EMIT_WARNINGS
@@ -24,4 +31,5 @@ def configure_warnings(pex_info, env):
 
 
 def warn(message):
+    # type: (str) -> None
     warnings.warn(message, category=PEXWarning, stacklevel=2)
--- a/pex/platforms.py
+++ b/pex/platforms.py
@@ -102,22 +102,22 @@ class Platform(namedtuple("Platform", ["platform", "impl", "version", "abi"])):
     @property
     def platform(self):
         # type: () -> str
-        return super(Platform, self).platform  # type: ignore[no-any-return]
+        return cast(str, super(Platform, self).platform)
 
     @property
     def impl(self):
         # type: () -> str
-        return super(Platform, self).impl  # type: ignore[no-any-return]
+        return cast(str, super(Platform, self).impl)
 
     @property
     def version(self):
         # type: () -> str
-        return super(Platform, self).version  # type: ignore[no-any-return]
+        return cast(str, super(Platform, self).version)
 
     @property
     def abi(self):
         # type: () -> str
-        return super(Platform, self).abi  # type: ignore[no-any-return]
+        return cast(str, super(Platform, self).abi)
 
     def __str__(self):
         # type: () -> str
--- a/pex/platforms.py
+++ b/pex/platforms.py
@@ -3,11 +3,16 @@
 
 from __future__ import absolute_import
 
-from collections import namedtuple
 from textwrap import dedent
 
+from pex.typing import TYPE_CHECKING
 
-class Platform(namedtuple("Platform", ["platform", "impl", "version", "abi"])):
+if TYPE_CHECKING:
+    from typing import Any, Iterator, Union, Tuple
+
+
+# TODO(#1041): Use typing.NamedTuple once we require Python 3.
+class Platform(object):
     """Represents a target platform and it's extended interpreter compatibility tags (e.g.
     implementation, version and ABI)."""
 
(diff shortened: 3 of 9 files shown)


pex/dist_metadata.py
```diff
@@ -7,18 +7,21 @@
...
-from pex.third_party.pkg_resources import DistInfoDistribution
+from pex.third_party.pkg_resources import DistInfoDistribution, Distribution
+from pex.typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from typing import Optional
 
 
 def requires_python(dist):
+    # type: (Distribution) -> Optional[SpecifierSet]
```
Eric-Arellano, line 18: Note that MyPy doesn't actually know what these are due to the vendoring. But I believe that it will at least check we are using type A, rather than type Z; it just doesn't understand the methods/fields on the type.

And this helpful for the reader.

pex/pex_warnings.py
```diff
@@ -6,13 +6,20 @@
...
+
 
 class PEXWarning(Warning):
     """Indicates a warning from PEX about suspect buildtime or runtime configuration."""
 
 
 def configure_warnings(pex_info, env):
-    if env.PEX_VERBOSE > 0:
+    # type: (PexInfo, Variables) -> None
+    if env.PEX_VERBOSE is not None and env.PEX_VERBOSE > 0:
```
Eric-Arellano, line 22: This was a bug, although I don't know if we ever hit it. It's possible for `PEX_VERBOSE` to be `None`, such as if you call `ENV.strip()` and the user did not define the value.

pex/platforms.py
```diff
@@ -16,8 +21,25 @@ class InvalidPlatformError(Exception):
 
     SEP = "-"
 
+    __slots__ = ("platform", "impl", "version", "abi")
+
+    def __init__(self, platform, impl, version, abi):
+        super(Platform, self).__init__()
+        if not all((platform, impl, version, abi)):
```
Eric-Arellano, line 28: It'd be better to check if these are `str`, rather than simply being truthy. But I tried to minimize my changes. Lmk if I should add it.

pex/platforms.py
```diff
@@ -62,33 +84,43 @@ def create(cls, platform):
...
+        return "Platform(platform={}, impl={}, version={}, abi={})".format(*self._tup())
+
     def __str__(self):
-        return self.SEP.join(self)
+        # type: () -> str
+        return self.SEP.join(self._tup())
+
+    def __eq__(self, other):
+        # type: (Any) -> bool
+        return self._tup() == other  # type: ignore[no-any-return]
```
Eric-Arellano, line 118: Idk why MyPy doesn't realize this is a `bool`.

pex/variables.py
```diff
@@ -68,40 +74,47 @@ def _get_kv(cls, variable):
             return kv
 
     def __init__(self, environ=None, rc=None, use_defaults=True):
+        # type: (Optional[Dict[str, str]], Optional[str], bool) -> None
         self._use_defaults = use_defaults
-        self._environ = (environ if environ is not None else os.environ).copy()
+        self._environ = environ.copy() if environ is not None else os.environ.copy()
```
Eric-Arellano, line 79: MyPy was getting confused trying to find the common denominator between our `Dict` vs. the `_OSEnviron` type, then complaining that `.copy()` wasn't define on that common interface of `MutableMapping`. This fixes it.

pex/platforms.py
```diff
@@ -3,11 +3,16 @@
...
 
+from pex.typing import TYPE_CHECKING
 
-class Platform(namedtuple("Platform", ["platform", "impl", "version", "abi"])):
+if TYPE_CHECKING:
+    from typing import Any, Iterator, Union, Tuple
+
+
+# TODO(#1041): Use typing.NamedTuple once we require Python 3.
+class Platform(object):
```
cosmicexplorer, line 15: I would try to determine some quick canonical performance test just since I believe there is a lower memory footprint and possibly faster field access (unsure) when using `namedtuple` in python 2.

pex/variables.py
```diff
@@ -68,40 +74,47 @@ def _get_kv(cls, variable):
...
-            if value.lower() in ("0", "false"):
-                return False
-            elif value.lower() in ("1", "true"):
-                return True
-            else:
-                die("Invalid value for %s, must be 0/1/false/true, got %r" % (variable, value))
-        else:
-            return self._defaulted(default)
+        if value is None:
+            return self._defaulted(default)  # type: ignore[return-value]
```
Eric-Arellano, line 98: We have to ignore due to the imprecise `._defaulted` method.

pex/variables.py
```diff
@@ -68,40 +74,47 @@ def _get_kv(cls, variable):
...
 
+    # TODO(#1041): Use a TypeVar once we require Python 3 to make this more precise.
     def _defaulted(self, default):
+        # type: (Optional[Union[str, int, bool]]) -> Optional[Union[str, int, bool]]
         return default if self._use_defaults else None
 
     def _get_bool(self, variable, default=False):
+        # type: (str, Optional[bool]) -> Optional[bool]
         value = self._environ.get(variable)
-        if value is not None:
```
Eric-Arellano, line 86: This method is the same code as before, only refactored by inverting the conditional and early returns.

pex/variables.py
```diff
@@ -237,15 +262,17 @@ def PEX_MODULE(self):
...
     @property
     def PEX_PROFILE(self):
+        # type: () -> Optional[bool]
         """Boolean.
 
         Enable application profiling.  If specified and PEX_PROFILE_FILENAME is not specified, PEX
         will print profiling information to stdout.
         """
-        return self._get_path("PEX_PROFILE", default=None)
+        return self._get_bool("PEX_PROFILE", default=False)
```
Eric-Arellano, line 271: This was confusing to me. I think a bug. We document it as a `bool`, but treat it as a `path`. @jsirois lmk if you know what we intended.

pex/dist_metadata.py
cosmicexplorer: We can possibly create stubs for these if needed?

pex/variables.py
```diff
@@ -21,15 +25,17 @@ class Variables(object):
 
     @classmethod
     def process_pydoc(cls, pydoc):
+        # type: (Optional[str]) -> Tuple[str, str]
         if pydoc is None:
```
cosmicexplorer, line 29: Possibly consider extending this to extract type hints in the future.

cosmicexplorer (approved):

pex/dist_metadata.py
Eric-Arellano: Oh, that'd be interesting as a followup once we have higher coverage. `pkg_resources` is pretty key to Pex. Having type hints working with Pex would be great.

pex/platforms.py
Eric-Arellano: I think field access might be ever so slightly faster.

For memory footprint, indeed. This is why I set `__slots__`. With `__slots__`, it shouldn't be very noticeable. (We decided in Pants that `__slots__` weren't worth the complexity; but I reason that Pex has different goals and constraints, such as minimizing its footprint. And this brings us closer to what `namedtuple` was doing.)

pex/variables.py
Eric-Arellano: Once we require Python 3, that would be super easy to do with `typing.get_type_hints()`, which is how the Pants rule signature parsing code works.

pex/platforms.py
cosmicexplorer: Oh yes! Thank you for pointing that out.

pex/variables.py
cosmicexplorer: Might this be the place for a `cast()`?

pex/platforms.py
cosmicexplorer: A `cast()` might make it more clear what's being done here? This is temporary enough anyway, so not too important.

pex/platforms.py
cosmicexplorer: I think we might want to make `abi` at least into an enum when we can drop py2, although I'm not sure checking for `str` now makes it harder to do later. I think leaving it as you've got it here sounds good.

pex/platforms.py
Eric-Arellano: Oh, we can actually indeed use `cast` here because it's a builtin type `bool`! Yay. 

Normally `cast` doesn't work if it's a complex type like `Optional[bool]` because we would need to import `Optional` at runtime.

pex/variables.py
Eric-Arellano: We can't because it returns `Optional[bool]` :/

cosmicexplorer: That's quite annoying.

jsirois (approved):

pex/platforms.py
```diff
@@ -3,11 +3,16 @@
...
 
+from pex.typing import TYPE_CHECKING, cast
 
-class Platform(namedtuple("Platform", ["platform", "impl", "version", "abi"])):
+if TYPE_CHECKING:
+    from typing import Any, Iterator, Union, Tuple
+
+
+# TODO(#1041): Use typing.NamedTuple once we require Python 3.
+class Platform(object):
```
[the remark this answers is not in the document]
Eric-Arellano: Yes, that sounds great. I would love to avoid this all. Thanks!

pex/platforms.py
```diff
@@ -62,33 +69,56 @@ def create(cls, platform):
...
         abi = cls._maybe_prefix_abi(impl, version, abi)
         return super(Platform, cls).__new__(cls, platform, impl, version, abi)
 
-    @classmethod
-    def from_tags(cls, platform, python, abi):
-        """Creates a platform corresponding to wheel compatibility tags.
+    @property
+    def platform(self):
+        # type: () -> str
+        return super(Platform, self).platform  # type: ignore[no-any-return]
```
jsirois, line 105: You use cast below in `__str__` for str, you probably should for all of these too.

pex/variables.py
jsirois: Noted in #1042, but you can use cast, just pass the types as strings when those types (or aliases) are only available at `TYPE_CHECKING` time.

PR #1044, opened 2020-09-26
Add type hints to `util.py` and `tracer.py`

Eric-Arellano:
Also clean up some prior type hints to use `cast`, type vars, and type aliases.

### DIFF
--- a/pex/tracer.py
+++ b/pex/tracer.py
@@ -8,29 +8,36 @@ import threading
 import time
 from contextlib import contextmanager
 
+from pex.typing import TYPE_CHECKING, cast
 from pex.variables import ENV
 
 __all__ = ("TraceLogger",)
+
+if TYPE_CHECKING:
+    from typing import Any, Callable, IO, Iterator, List, Optional
 
 
 class Trace(object):
     __slots__ = ("msg", "verbosity", "parent", "children", "_clock", "_start", "_stop")
 
     def __init__(self, msg, parent=None, verbosity=1, clock=time):
+        # type: (str, Optional[Trace], int, Any) -> None
         self.msg = msg
         self.verbosity = verbosity
         self.parent = parent
         if parent is not None:
-            parent.children.append(self)
-        self.children = []
+            parent.children.append(self)  # type: ignore[has-type]
+        self.children = cast("List[Trace]", [])
         self._clock = clock
-        self._start = self._clock.time()
-        self._stop = None
+        self._start = cast(float, self._clock.time())
+        self._stop = cast("Optional[float]", None)
 
     def stop(self):
+        # type: () -> None
         self._stop = self._clock.time()
 
     def duration(self):
+        # type: () -> float
         assert self._stop is not None
         return self._stop - self._start
 
@@ -39,6 +46,7 @@ class TraceLogger(object):
     """A multi-threaded tracer."""
 
     def __init__(self, predicate=None, output=sys.stderr, clock=time, prefix=""):
+        # type: (Optional[Callable[[int], bool]], IO, Any, str) -> None
         """If predicate specified, it should take a "verbosity" integer and determine whether or not
         to log, e.g.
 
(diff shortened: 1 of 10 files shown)


pex/tracer.py
```diff
@@ -8,29 +8,36 @@
...
 
     def __init__(self, msg, parent=None, verbosity=1, clock=time):
+        # type: (str, Optional[Trace], int, Any) -> None
         self.msg = msg
         self.verbosity = verbosity
         self.parent = parent
         if parent is not None:
-            parent.children.append(self)
-        self.children = []
+            parent.children.append(self)  # type: ignore[has-type]
```
Eric-Arellano, line 29: I couldn't get MyPy happy with this, even with using `cast(Trace, parent)` and `cast("List[Trace]", parent.children)`. It keeps complaining `Cannot determine type of 'children'  [has-type]`.

pex/util.py
```diff
@@ -9,74 +9,34 @@
...
+if TYPE_CHECKING:
+    if PY2:
+        from hashlib import _hash as _Hash
+    else:
+        from hashlib import _Hash
+    from typing import Any, Callable, IO, Iterable, Iterator, Optional
 
-class DistributionHelper(object):
-    @classmethod
-    def access_zipped_assets(cls, static_module_name, static_path, dir_location=None):
```
Eric-Arellano, line 29: Unused, beyond a test.

pex/util.py
```diff
@@ -86,18 +46,21 @@ def distribution_from_path(cls, path, name=None):
             for dist in find_distributions(path):
                 if dist.project_name == name:
                     return dist
+        return None
```
Eric-Arellano, line 49: MyPy wanted an explicit return.

pex/variables.py
```diff
@@ -86,16 +88,15 @@ def copy(self):
         # type: () -> Dict[str, str]
         return self._environ.copy()
 
-    # TODO(#1041): Use a TypeVar once we require Python 3 to make this more precise.
     def _defaulted(self, default):
-        # type: (Optional[Union[str, int, bool]]) -> Optional[Union[str, int, bool]]
+        # type: (Optional[_T]) -> Optional[_T]
```
Eric-Arellano, line 92: Yay!!

pex/tracer.py
```diff
@@ -8,29 +8,36 @@
...
+        self.children = cast("List[Trace]", [])
```
jsirois, line 30: This is not idiomatic as I read the mypy docs. You want `var = val  # type: X` for typing variables.

pex/tracer.py
jsirois: Perhaps the idiom note below clears this up. If not, perhaps an isinstance check instead of a not None check. Hopefully the former.

pex/util.py
jsirois: Ah, this old thing. It is used but we don't have API markers yet. This commit's ba58a854 review is lost inside Twitter but was used to allow a Pexed up application that knew about Pexing to extract its assets; i.e: client code called (calls) it. This shouldn't be killed until API / deprecations / etc is set up. Perhaps add a note to this effect since you did the right thing given all the discoverable information.

tests/test_integration.py
```diff
@@ -994,20 +994,26 @@ def test_pex_multi_resolve_2():
             ), "{} was not found in wheel".format(dist_substr)
 
 
+if TYPE_CHECKING:
+    TEST_RESOLVE_FN = Callable[[str, str, str, str, Optional[str]], None]
```
jsirois, line 998: All caps type names are jarring. I'm guessing you're trying to distinguish an alias from a "real" type via hungarian - but, by definition this is already all local to the file so I'm not seeing the point.

jsirois (approved):

pex/util.py
Eric-Arellano: Added back. Thanks.

tests/test_integration.py
Eric-Arellano: My thinking was it was a constant, so following the convention. But I like PascalCase more here.

pex/tracer.py
Eric-Arellano: This fixed the type ignore. Thanks!

Issue #1020, opened 2020-08-28
Decouple PEX runtime interpreter selection from buildtime interpreter selection.


jsirois:
The Pex CLI supports various options for constraining the interpreters a given PEX should be built to work with. Today these are:
1. `--python` or `--interpreter-constraint` (these are mutually exclusive)
2. `--platform`

And combinations of these are then possibly modified by `--resolve-local-platforms` and `--use-first-matching-interpreter`.

The result is a PEX file built by one or more local interpreters that can run in turn run on an unrelated set of interpreters. The last statement might be surprising, but is true and exposes the fundamental flaw in trying to use buildtime interpreter selection constraints to constrain runtime interpreter selection. To clarify, consider building a PEX file for Pex itself:
```
$ python -mpex pex -cpex -opex.pex
$ unzip -qc pex.pex PEX-INFO | jq .distributions
{
  "pex-2.1.15-py2.py3-none-any.whl": "bc4c28769a8f357bba9579aa5e24534b622d4ada"
}
$ head -1 pex.pex
#!/usr/bin/env python3.8
```
As we can see, the PEX file is universal since all contained distributions are universal and yet the PEX file will only run on a machine with a `python3.8` binary on the `PATH`. Today this can be hacked around by either specifying an alternative `--python-shebang` (say `#!/usr/bin/env python` or executing the PEX file via a manually selected interpreter (`/my/python pex.pex`).

Ideally, once bootstrapped by a Pex compatible interpreter (Python 2.7 or Python3.5+), the PEX runtime should select a local interpreter compatible with the actual contained distributions. This would allow PEX files to run on all machines they should be able to run on assuming `#!/usr/bin/env python` points to a Pex compatible Python interpreter. 

Switching the PEX runtime to select interpreters based on contained distributions would fix at least 2 bugs:
1. PEX files could ship working on the maximum set of interpreters given only a reasonable `#!/usr/bin/env python` (solves the above example bug in pex.pex).
2. PEX files would fail fast when interpreter constraints on the buildtime machine are met by a narrower set of interpreters than on a runtime machine.

The latter bug today manifests when using, say, `--interpreter-constraints="CPython>=3.6"` which finds just a CPython 3.7 distributuion on the build machine. The resulting PEX can then potentially only run on CPython 3.7 interpreter if platform specific wheels were built but at runtime a CPython 3.6 interpreter will be selected if found even if a local CPython 3.7 is present.

cosmicexplorer:
[...]

PR #1047, opened 2020-09-27
slightly demystify the relationship between platforms and interpreters in the library API and CLI

### LINKED ISSUE #1033: specify (interpreter-constraint) x (platform), and pass --python-version to pip regardless of a local binary
When i run `pex.resolver.resolve_multi()`, it asks for both `interpreters` and `platforms` separately. A few questions:
1. Is it possible to construct a `PythonInterpreter` without actually having a matching interpreter binary on the local disk?
2. If the answer to (1) is "no", is there any way to create a pex file with requirements downloaded for an interpreter I don't have on my local machine?
3. If we *are* restricting `PythonInterpreter` to only local files as per (1), why wouldn't the `platform` argument be something we could infer from the interpreter binary path? It looks like `PythonInterpreter` has this `supported_platforms` property. Is there a reason we wouldn't just use that, instead of having to spread that information over a separate `platforms` kwarg to `resolve_multi()`?

It seems like `(interpreter-constraint) x (platform)` are the axes that pip actually uses to resolve, and I'm wondering why we specify those separately, and why we require an interpreter necessarily on the local disk. One alternative might be:
```python
# Monolithic resolve method, instead of a separate resolve_multi().
def resolve(..., matching_interpreters=None):
    if matching_interpreters is None:
        if interpreter is None and platform is None:
            matching_interpreters = [MatchingInterpreter.from_interpreter(PythonInterpreter.get())]
        else:
            # Produces a list of `MatchingInterpreter`s with all combinations of interpreter and platform.
            matching_interpreters = (InterpreterSet(interpreters=[interpreter])
                                     .for_platforms(platforms=[platform]))
    ...

# Create a declarative interpreter spec matching a python 3.6 interpreter 
# on the local machine  -- this is possible currently by iterating over interpreters and checking
# their compatibility. This is a translation of what is possible today.
# This line *does* touch the filesystem, to iterate over interpreters until a matching one is found.
whatever_the_current_36_is = MatchingInterpreter.from_constraints(['CPython==3.6.*']).

# This line, and all the rest, do *not* touch the filesystem or perform any syscalls.
# If this is in the `matched_interpreters` list when calling resolve(), a pip job will be executed with:
# --python-version 3.6.11 --platform linux_x86_64 --abi cp36m
linux_cp_36_m = MatchingInterpreter('CPython==3.6.11', platform=Platform.create('linux_x86_64-cp-36-m'))

# If we pretend we found a 3.6.11 python interpreter in the first example, the above line would be equivalent
# to the below line, which uses a utility method `.copy()`. This is the direct translation of the current
# method of creating a multiplatform pex, which depends on finding an interpreter on the local disk, and
# then providing a set of other platforms to fetch for (totally separate from that local interpreter version)
# discarding the `.platform` of that local interpreter (necessitating the redundant 'current' shorthand).
linux_cp_36_m = whatever_the_current_36_is.copy(platform=Platform.create('linux_x86_64-cp-36-m'))

# Ask for wheels matching the most recent macosx_10_XY_x86_64 platform,
# because that has the widest compatibility, and many wheels are simply published
# with a platform matching the publisher's laptop, which may be very recent, for no reason.
osx_36_abi3 = MatchingInterpreter('CPython==3.6.11', platform=Platform.most_recent_osx(abi_tag='abi3'))

# Note that *only* == interpreter constraints are allowed when using the normal constructor.
# This is a possible convenience method to match CPython specifically, with shorter strings.
# This should produce the exact same result as the previous line.
 osx_36_abi3 = MatchingInterpreter.CPython('3.6.11', platform=Platform.create('macosx_10_14_x86_64-cp36-abi3'))
```

If this was to be entertained, it seems that it could subsume the functionality of the existing `resolve()` method entirely (translating `interpreters` and `platforms` to a list of `MatchingInterpreter`s), and therefore could avoid breaking any existing users of the pex library API at all (while closing #969).

We would have to figure out how to appropriately error when:
1. the user specifies an interpreter version (or platform) that *doesn't* exist on their machine, and
2. some requirements are platform-specific, and there are no prebuilt wheels for that `(interpreter-version) x (platform)` pair.

However, I believe we already error out today in this case, if we request e.g. `linux_86_64-cp36-abi3` in `platforms`, and there are no prebuilt wheels available for that platform. So we should be able to extend or keep the same error message as before when things don't match up -- we would just be adding a new failure state (which just errors out much earlier today, at interpreter selection time), by allowing the user to resolve for interpreters they don't have on their system.

Regarding the Pex CLI, I believe we could consider an `--explicit-python-version` option corresponding to the first argument of `MatchingInterpreter(...)`, aka:
```bash
> pex --explicit-python-version CPython==3.6.11 ...
```
This flag would be mutually exclusive with both `--python` and `--interpreter-constraints`, but could be applied multiple times to resolve for multiple explicit interpreter versions.

cosmicexplorer:
@jsirois read my mind in https://github.com/pantsbuild/pex/issues/1033#issuecomment-699590960 after I asked why Pex couldn't resolve for an interpreter it doesn't have locally. This was in fact already possible, and he suggested a documentation fix to make it more clear. These changes make it clear that:
- `platform{,s}` takes precedence over `interpreter{,s}` when resolving wheels
- `interpreter{,s}` are used to seed `platform{,s}` if not provided
- matching `interpreter{,s}` must be provided for all `platform{,s}` if any distributions need to be built from source

Closes #1033.

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -500,11 +500,9 @@ def configure_clp_pex_environment(parser):
         callback=process_platform,
         help="The platform for which to build the PEX. This option can be passed multiple times "
         "to create a multi-platform pex. To use the platform corresponding to the current "
-        "interpreter you can pass `current` "
-        "(which is the default value if this option is not provided). "
-        "To target any other platform you pass a string composed of fields: "
-        "<platform>-<python impl abbr>-<python version>-<abi>. "
-        "These fields stem from wheel name conventions as outlined in "
+        "interpreter you can pass `current`. To target any other platform you pass a string "
+        "composed of fields: <platform>-<python impl abbr>-<python version>-<abi>. These fields "
+        "stem from wheel name conventions as outlined in "
         "https://www.python.org/dev/peps/pep-0427#file-name-convention and influenced by "
         "https://www.python.org/dev/peps/pep-0425. For the current interpreter at {} the full "
         "platform string is {}. To find out more, try `{} --platform explain`.".format(
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -500,9 +500,7 @@ def configure_clp_pex_environment(parser):
         callback=process_platform,
         help="The platform for which to build the PEX. This option can be passed multiple times "
         "to create a multi-platform pex. To use the platform corresponding to the current "
-        "interpreter you can pass `current`. If any distributions need to be built, use the "
-        "--python or --interpreter-constraint argument instead, providing the corresponding "
-        "interpreter. To target any other platform you pass a string "
+        "interpreter you can pass `current`. To target any other platform you pass a string "
         "composed of fields: <platform>-<python impl abbr>-<python version>-<abi>. "
         "These fields stem from wheel name conventions as outlined in "
         "https://www.python.org/dev/peps/pep-0427#file-name-convention and influenced by "
@@ -537,7 +535,7 @@ def configure_clp_pex_environment(parser):
             "resolve requirements for each interpreter; this allows the resulting Pex to be "
             "compatible with more interpreters, such as different Python versions. However, "
             "resolving for multiple interpreters will take longer to build, and the resulting PEX "
-            "will be larger."
+            "may be larger."
         ),
     )
 
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -500,9 +500,11 @@ def configure_clp_pex_environment(parser):
         callback=process_platform,
         help="The platform for which to build the PEX. This option can be passed multiple times "
         "to create a multi-platform pex. To use the platform corresponding to the current "
-        "interpreter you can pass `current`. To target any other platform you pass a string "
-        "composed of fields: <platform>-<python impl abbr>-<python version>-<abi>. These fields "
-        "stem from wheel name conventions as outlined in "
+        "interpreter you can pass `current`. If any distributions need to be built, use the "
+        "--python or --interpreter-constraint argument instead, providing the corresponding "
+        "interpreter. To target any other platform you pass a string "
+        "composed of fields: <platform>-<python impl abbr>-<python version>-<abi>. "
+        "These fields stem from wheel name conventions as outlined in "
         "https://www.python.org/dev/peps/pep-0427#file-name-convention and influenced by "
         "https://www.python.org/dev/peps/pep-0425. For the current interpreter at {} the full "
         "platform string is {}. To find out more, try `{} --platform explain`.".format(
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -520,9 +520,9 @@ def configure_clp_pex_environment(parser):
         callback=parse_bool,
         help="When --platforms are specified, attempt to resolve a local interpreter that matches "
         "each platform specified. If found, use the interpreter to resolve distributions; if "
-        "not (or if this option is not specified), resolve for the platform only allowing matching "
-        "binary distributions and failing if only sdists or non-matching binary distributions "
-        "can be found.",
+        "not (or if this option is not specified), resolve for each platform only allowing "
+        "matching binary distributions and failing if only sdists or non-matching binary "
+        "distributions can be found.",
     )
 
     group.add_option(
@@ -536,7 +536,8 @@ def configure_clp_pex_environment(parser):
             "compatible Python version. Normally, when multiple interpreters match, Pex will "
             "resolve requirements for each interpreter; this allows the resulting Pex to be "
             "compatible with more interpreters, such as different Python versions. However, "
-            "resolving for multiple interpreters results in worse performance."
+            "resolving for multiple interpreters will take longer to build, and the resulting PEX "
+            "will be larger."
         ),
     )
 
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -500,9 +500,11 @@ def configure_clp_pex_environment(parser):
         callback=process_platform,
         help="The platform for which to build the PEX. This option can be passed multiple times "
         "to create a multi-platform pex. To use the platform corresponding to the current "
-        "interpreter you can pass `current`. To target any other platform you pass a string "
-        "composed of fields: <platform>-<python impl abbr>-<python version>-<abi>. These fields "
-        "stem from wheel name conventions as outlined in "
+        "interpreter you can pass `current` "
+        "(which is the default value if this option is not provided). "
+        "To target any other platform you pass a string composed of fields: "
+        "<platform>-<python impl abbr>-<python version>-<abi>. "
+        "These fields stem from wheel name conventions as outlined in "
         "https://www.python.org/dev/peps/pep-0427#file-name-convention and influenced by "
         "https://www.python.org/dev/peps/pep-0425. For the current interpreter at {} the full "
         "platform string is {}. To find out more, try `{} --platform explain`.".format(
@@ -518,8 +520,9 @@ def configure_clp_pex_environment(parser):
         callback=parse_bool,
         help="When --platforms are specified, attempt to resolve a local interpreter that matches "
         "each platform specified. If found, use the interpreter to resolve distributions; if "
-        "not, resolve for the platform only allowing matching binary distributions and failing "
-        "if only sdists or non-matching binary distributions can be found.",
+        "not (or if this option is not specified), resolve for the platform only allowing matching "
+        "binary distributions and failing if only sdists or non-matching binary distributions "
+        "can be found.",
     )
 
     group.add_option(
--- a/pex/resolver.py
+++ b/pex/resolver.py
@@ -733,11 +733,13 @@ def resolve(
       prerelease or development versions will override this setting.
     :keyword bool transitive: Whether to resolve transitive dependencies of requirements.
       Defaults to ``True``.
-    :keyword interpreter: The interpreter to use for building distributions and for testing
-      distribution compatibility. Defaults to the current interpreter.
+    :keyword interpreter: If specified, distributions will be resolved for this interpreter, and
+      non-wheel distributions will be built against this interpreter. If both `interpreter` and
+      `platform` are ``None`` (the default), this defaults to the current interpreter.
     :type interpreter: :class:`pex.interpreter.PythonInterpreter`
-    :keyword str platform: The exact target platform to resolve distributions for. If ``None`` or
-      ``'current'``, resolve for distributions appropriate for `interpreter`.
+    :keyword str platform: The exact PEP425-compatible platform string to resolve distributions for,
+      in addition to the platform of the given interpreter, if provided. The current interpreter
+      will be used to build any non-wheel distributions.
     :keyword indexes: A list of urls or paths pointing to PEP 503 compliant repositories to search for
       distributions. Defaults to ``None`` which indicates to use the default pypi index. To turn off
       use of all indexes, pass an empty list.
@@ -826,12 +828,14 @@ def resolve_multi(
       prerelease or development versions will override this setting.
     :keyword bool transitive: Whether to resolve transitive dependencies of requirements.
       Defaults to ``True``.
-    :keyword interpreters: The interpreters to use for building distributions and for testing
-      distribution compatibility. Defaults to the current interpreter.
+    :keyword interpreters: If specified, distributions will be resolved for these interpreters, and
+      non-wheel distributions will be built against each interpreter. If both `interpreters` and
+      `platforms` are ``None`` (the default) or an empty iterable, this defaults to a list
+      containing only the current interpreter.
     :type interpreters: list of :class:`pex.interpreter.PythonInterpreter`
     :keyword platforms: An iterable of PEP425-compatible platform strings to resolve distributions
-      for. If ``None`` (the default) or an empty iterable, use the platforms of the given
-      interpreters.
+      for, in addition to the platforms of any given interpreters. The current interpreter will be
+      used to build any non-wheel distributions.
     :type platforms: list of str
     :keyword indexes: A list of urls or paths pointing to PEP 503 compliant repositories to search for
       distributions. Defaults to ``None`` which indicates to use the default pypi index. To turn off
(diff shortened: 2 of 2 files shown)


pex/resolver.py
```diff
@@ -737,7 +737,9 @@ def resolve(
       distribution compatibility. Defaults to the current interpreter.
     :type interpreter: :class:`pex.interpreter.PythonInterpreter`
     :keyword str platform: The exact target platform to resolve distributions for. If ``None`` or
-      ``'current'``, resolve for distributions appropriate for `interpreter`.
+      ``'current'``, resolve for distributions appropriate for `interpreter` instead. If any
+      distributions need to be built, the corresponding interpreter must be provided in the
```
jsirois, line 741: > ... If any distributions need to be built, the corresponding interpreter must be provided in the

This is not right. The platforms and interpreters argumenst are _completely_ independent. IOW, if breaking the resolve API was deemed OK, we could collapse the two interpreters and platforms arguments to one distribution_targets argument since that's exactly what's done internally:
https://github.com/pantsbuild/pex/blob/e67412c31648f6e0d0f2a68b9d38dffabb258ebc/pex/resolver.py#L958-L975

jsirois (commented):
Thanks. I'm still not sure you and the API are on the same page. More clarification below.

pex/resolver.py
jsirois: If that all makes sense, perhaps you mean to say:
```
If any distributions need to be built, use the interpreters argument instead, providing the corresponding interpreter
```

cosmicexplorer: Ok. I was under the impression that an interpreter of the appropriate version was necessary to build any non-wheel for that interpreter version. However, now, IIUC, the `interpreter` from this section: https://github.com/pantsbuild/pex/blob/e67412c31648f6e0d0f2a68b9d38dffabb258ebc/pex/resolver.py#L515-L523
will just resolve to the current interpreter with `PythonInterpreter.get()` whenever the target was created with `DistributionTarget.from_platform()`.

I see now that we explicitly error in the case that we are forced into building wheels from a foreign platform. I didn't realize how that was handled before: https://github.com/pantsbuild/pex/blob/e67412c31648f6e0d0f2a68b9d38dffabb258ebc/pex/pip.py#L202-L212

So things learned:
1. no special interpreter is needed to build non-wheels!
2. if there are no wheels available for a foreign platform, we will error with `Unsatisfiable`, because our pip invocation will include `--only-binary :all:`.
3. if wheels are turned off and we try to resolve for a foreign platform, we have the very specific error message displayed above.

I will look again and see whether there is any useful change I can make to the docs. Thank you again for your patience.

cosmicexplorer:
Ok, I've made another change to the docstring. In this case I have:
- clearly specified that distributions are resolved for `interpreters` as well as `platforms`, but that the current interpreter is used to build any non-wheel distributions for `platforms`.
- added (correctly) two `:raises ValueError:` clauses in `resolve()`, `resolve_multi()`, and `download()` to clarify what happens when you don't allow building non-wheels and don't allow building wheels.

My final question is: is it possible to successfully build wheels for foreign platforms from pure-python libraries that are only provided as `.tar.gz`? This would occur e.g. if you wanted to build a PEX containing *only* a linux dist of `tensorflow==1.14.0`. It appears pip stops you from doing this too, but it feels artificial. Not sure if I'm missing something here.

pex/bin/pex.py
```diff
@@ -500,9 +500,11 @@ def configure_clp_pex_environment(parser):
         callback=process_platform,
         help="The platform for which to build the PEX. This option can be passed multiple times "
         "to create a multi-platform pex. To use the platform corresponding to the current "
-        "interpreter you can pass `current`. To target any other platform you pass a string "
-        "composed of fields: <platform>-<python impl abbr>-<python version>-<abi>. These fields "
-        "stem from wheel name conventions as outlined in "
+        "interpreter you can pass `current` "
+        "(which is the default value if this option is not provided). "
```
jsirois, line 504: The parenthetical contradicts the default above and its not always true either. Consider the case where 1 interpreter is specified - say via `--python python2.7` and yet the Pex CLI is running under Python 3.6. In that case the resulting PEX file will only be built using a local Python 2.7 interpreter and not additionally for `current` which is a local Python 3.6 interpreter. How about just dropping this addition?

pex/resolver.py
```diff
@@ -733,11 +733,13 @@ def resolve(
...
-    :keyword interpreter: The interpreter to use for building distributions and for testing
-      distribution compatibility. Defaults to the current interpreter.
+    :keyword interpreter: If specified, distributions will be resolved for this interpreter, and
+      non-wheel distributions will be built against this interpreter. If both `interpreter` and
+      `platform` are ``None`` (the default), this defaults to the current interpreter.
     :type interpreter: :class:`pex.interpreter.PythonInterpreter`
-    :keyword str platform: The exact target platform to resolve distributions for. If ``None`` or
-      ``'current'``, resolve for distributions appropriate for `interpreter`.
+    :keyword str platform: The exact PEP425-compatible platform string to resolve distributions for,
+      in addition to the platform of the given interpreter, if provided. The current interpreter
```
jsirois, line 741: > The current interpreter will be used to build any non-wheel distributions.

This is not true. If non-wheel distributions are encountered when resolving for the given platform, resolution will fail.

pex/resolver.py
```diff
@@ -824,12 +828,14 @@ def resolve_multi(
...
-      distribution compatibility. Defaults to the current interpreter.
+    :keyword interpreters: If specified, distributions will be resolved for these interpreters, and
+      non-wheel distributions will be built against each interpreter. If both `interpreters` and
+      `platforms` are ``None`` (the default) or an empty iterable, this defaults to a list
+      containing only the current interpreter.
     :type interpreters: list of :class:`pex.interpreter.PythonInterpreter`
     :keyword platforms: An iterable of PEP425-compatible platform strings to resolve distributions
-      for. If ``None`` (the default) or an empty iterable, use the platforms of the given
-      interpreters.
+      for, in addition to the platforms of any given interpreters. The current interpreter will be
```
jsirois, line 837: Ditto - the last sentence is false.

pex/bin/pex.py
cosmicexplorer: Great!

pex/resolver.py
jsirois: Perhaps the confusion is that the CLI, which uses the API, will turn `--platform ...` arguments into `PythonInterpreters` when -`--resolve-local-platforms`. It then passes interpreters for those platforms it could resolve instead of passing platforms. Does that clear things up? I.E.: `--resolve-local-platforms` is a CLI feature and not a resolve API feature.

cosmicexplorer: What if the `platform` matches the `supported_tags` of the current interpreter? I'm looking at https://github.com/pantsbuild/pex/blob/e67412c31648f6e0d0f2a68b9d38dffabb258ebc/pex/distribution_target.py#L31-L34 and https://github.com/pantsbuild/pex/blob/e67412c31648f6e0d0f2a68b9d38dffabb258ebc/pex/interpreter.py#L61-L72, and it seems like `tags.sys_tags()` can return more than one element in its tuple: https://github.com/pantsbuild/pex/blob/e67412c31648f6e0d0f2a68b9d38dffabb258ebc/pex/vendor/_vendored/packaging/packaging/tags.py#L372-L374

jsirois: This code here:
https://github.com/pantsbuild/pex/blob/cff40893dd1b0ae185201f4871baae87a9fe1335/pex/bin/pex.py#L758-L786

pex/resolver.py
cosmicexplorer: Yes, I missed your addendum earlier but I think it's much better:
> If any distributions need to be built, use the interpreters argument instead, providing the corresponding interpreter.

I'll modify these to use that.

pex/bin/pex.py
cosmicexplorer: Thank you for explaining this.

pex/resolver.py
jsirois: > What if the platform matches the supported_tags of the current interpreter? I'm looking at

We raced each other. I think you're confusing a CLI feature for an API feature. Let me know if I've guessed your confusion or if it persists.

pex/bin/pex.py
cosmicexplorer: After reverting the change you noted, I modified this further to include the distinction you gave above:
> If any distributions need to be built, use the --python or --interpreter-constraint argument instead, providing the corresponding interpreter.

pex/resolver.py
cosmicexplorer: I was referring to the fact that `DistributionTarget.is_foreign` is calculated according to the tags of the current interpreter (`PythonInterpreter.get()`) when its `._interpreter` field is None (when it's created via `.for_platform()`). While it's true that it would be quite unlikely for `.is_foreign` to return False, if the `platform` provided matches the current interpreter's tags, then, it would actually use the current interpreter to build any non-wheels.

If we actually wanted to document this rare edge case, I might think something like:
> If the platform matches the current interpreter, the current interpreter will be used to build any non-wheels.

could be reasonable.

cosmicexplorer: Which isn't exactly rare, now that I think of it -- I believe `platform='current'` would do the trick. I think it would be useful to add the sentence I proposed above, but if that seems like too much information I'm not beholden to it.

cosmicexplorer: Ah! But we have a specific check for that! https://github.com/pantsbuild/pex/blob/e67412c31648f6e0d0f2a68b9d38dffabb258ebc/pex/resolver.py#L972-L974

It seems like we might want to consider expanding that check to cover when the platform *matches* any of the given interpreters, in addition to checking if it's just `None`. I don't know if there's any deduplication later. I'll test out what happens if you give an explicit platform string matching the current interpreter.

jsirois: You're losing me. The block of code you point to just ensures that for `platforms=[None]` (which is how `parsed_platforms` `translates platforms=['current']`) and `interpreters=None` or `interpreters=[]` we use  the current interpreter as the single interpreter to resolve with.

The general search you describe is what the CLI does in the code I pointed to above in pex/bin/pex.py when you sepcify `--resolve-local-platforms`.

jsirois: If you want to document the edge case where `platforms=['X', ...]` and 'X' happens to match the current interpreter accessing the Pex API, that's fine with me. But, to be clear, 'X' could either be 'current' in that case or a full platform string that just happens to match the current interpreter.

cosmicexplorer: Yes, thank you for piecing that together. Let me know if 2fc5078b09e24ff82ac8eacd984e51a082d1a5d7 seems sufficient to document that.

pex/resolver.py
```diff
@@ -978,13 +978,26 @@ def iter_targets():
                 # Build for the specified local interpreters (on the current platform).
                 yield DistributionTarget.for_interpreter(interpreter)
 
+        def check_platform_against_interpreters(platform):
```
jsirois, line 981: Please limit this PR to the description and break this out in a new PR.

cosmicexplorer: You got it!

pex/bin/pex.py
```diff
@@ -500,9 +500,11 @@ def configure_clp_pex_environment(parser):
         callback=process_platform,
         help="The platform for which to build the PEX. This option can be passed multiple times "
         "to create a multi-platform pex. To use the platform corresponding to the current "
-        "interpreter you can pass `current`. To target any other platform you pass a string "
-        "composed of fields: <platform>-<python impl abbr>-<python version>-<abi>. These fields "
-        "stem from wheel name conventions as outlined in "
+        "interpreter you can pass `current`. If any distributions need to be built, use the "
+        "--python or --interpreter-constraint argument instead, providing the corresponding "
```
jsirois, line 504: Ok. I think explaining complex things in certain contexts like CLI help and error messages is a tarpit but I know others don't agree. This now tells half the truth. The whole truth is you can also pass `--resolve-local-platforms`.

pex/bin/pex.py
```diff
@@ -533,7 +536,8 @@ def configure_clp_pex_environment(parser):
             "compatible Python version. Normally, when multiple interpreters match, Pex will "
             "resolve requirements for each interpreter; this allows the resulting Pex to be "
             "compatible with more interpreters, such as different Python versions. However, "
-            "resolving for multiple interpreters results in worse performance."
+            "resolving for multiple interpreters will take longer to build, and the resulting PEX "
+            "will be larger."
```
jsirois, line 540: s/will be/may be/ - the tarpit again. It will only be larger if there are wheels that vary. If its a pure universal wheel resolve for example, this is false as is.

jsirois (approved):

pex/bin/pex.py
cosmicexplorer: I think `--resolve-local-platforms` has exactly the right information and it is directly below `--platforms`. I will revert this change.

pex/bin/pex.py
cosmicexplorer: My only concern with the original wording is that I might be worried upon reading the help for the first time that there's e.g. some weird bug, or some complex emergent effect, that causes multi-interpreter PEX files to run more slowly. I would have left it at "takes longer to build", but there is at least one runtime effect -- namely the file size (which as you pointed out is not guaranteed to differ).

I think that too much information, especially in CLI help, can absolutely be overwhelming (and I know I am prone to causing this more than others). I would be fine reverting this too, although I think this one is much more defensible than the change to `--platform` help that I just reverted.

jsirois: Yeah - this is a smell of the feature. It was added for Pants use but does not make much sense for Pex use. It's in fact still buggy and will only be fixed #1020. That said, Eric and I came to light agreement on killing the feature since Pants no longer uses it.

cosmicexplorer: Ok, that sounds fine then. I'll perhaps leave this change in then just to be done with it, but from what I can tell killing the feature seems quite reasonable. In the future I suspect that Pants could make use of checked-in python scripts like `ipex_launcher.py` to make Pex tools that work with the process execution framework (and avoid churn in the upstream Pex project).

Eric-Arellano: Yep, I'm cool with removing it.

PR #1049, opened 2020-09-29
Do not rely on `sys.executable` being accurate.

### LINKED ISSUE #1009: No such file for interpreter's binary name
With `2.1.12`, @asherf encountered a case where `pex` detected an interpreter with a bad binary name:
```
pex: Building pex
pex: Building pex :: Resolving interpreters
pex: Constraints on interpreters: ['CPython>=3.6'], Matching Interpreter: /Users/asher/.pyenv/versions/3.8.2/bin/python3.8
pex: Constraints on interpreters: ['CPython>=3.6'], Matching Interpreter: /Users/asher/.toolchain/homebrew/Cellar/python@3.8/3.8.4/bin/python3.7
pex: Constraints on interpreters: ['CPython>=3.6'], Matching Interpreter: /usr/local/Cellar/python/3.7.8/Frameworks/Python.framework/Versions/3.7/bin/python3.7
pex: Constraints on interpreters: ['CPython>=3.6'], Matching Interpreter: /Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.7/bin/python3.7
```

The file `/Users/asher/.toolchain/homebrew/Cellar/python@3.8/3.8.4/bin/python3.7` does not exist (`python3.8` does, as might be expected).

This bug was reproduced in the context of a long lived `pex_root`, which @asherf can provide if need be.

jsirois:
Unfortunately, OSX framework Python interpreters have had launch schemes
that could lead to misreported `sys.executable`. We no longer trust
`sys.executable` when caching interpreters or reading them from the
cache.

Fixes #1009

### DIFF
--- a/pex/interpreter.py
+++ b/pex/interpreter.py
@@ -68,11 +68,22 @@ class PythonIdentity(object):
     }
 
     @classmethod
-    def get(cls):
+    def get(cls, binary=None):
+        # N.B.: We should not need to look past `sys.executable` to learn the current interpreter's
+        # executable path, but on OSX there has been a bug where the `sys.executable` reported is
+        # _not_ the path of the current interpreter executable:
+        # https://bugs.python.org/issue22490#msg283859
+        if binary and binary != sys.executable:
+            TRACER.log(
+                "Identifying interpreter found at {} which reports an incorrect sys.executable of "
+                "{}.".format(binary, sys.executable),
+                V=9,
+            )
+
         supported_tags = tuple(tags.sys_tags())
         preferred_tag = supported_tags[0]
         return cls(
-            binary=sys.executable,
+            binary=binary or sys.executable,
             python_tag=preferred_tag.interpreter,
             abi_tag=preferred_tag.abi,
             platform_tag=preferred_tag.platform,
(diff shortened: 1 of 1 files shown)


pex/interpreter.py
```diff
@@ -400,11 +411,20 @@ def _execute(cls, binary, args=None, pythonpath=None, env=None, stdin_payload=No
...
             if not identity:
-                raise cls.IdentificationError("Could not establish identity of %s" % binary)
-            return cls(PythonIdentity.decode(identity))
+                raise cls.IdentificationError("Could not establish identity of {}.".format(binary))
+            interpreter = cls(PythonIdentity.decode(identity))
+            # We should not need to check this since binary == interpreter.binary should always be
+            # true, but historically this could be untrue as noted in `PythonIdentity.get`.
+            if check_binary and not os.path.exists(interpreter.binary):
+                raise cls.InterpreterNotFound(
+                    "Cached interpreter for {} reports a binary of {} which is not found".format(
```
Eric-Arellano, line 423:
```suggestion
                    "Cached interpreter for {} reports a binary of {}, which could not be found".format(
```

Eric-Arellano: Is there any remedy here? Could we instruct the user to purge the cache, for example?

Eric-Arellano (approved):
Ludicrous behavior from macOS. Thanks for digging into this.

pex/interpreter.py
jsirois: The code you commented on above does purge the cache automatically already. This log line at finest log level is just to say FYI for someone debugging things deeply.

Eric-Arellano: I'm not following that. This raises an exception, rather than logging. Where is the cache purge happening?

pex/interpreter.py
```diff
@@ -435,7 +455,7 @@ def create_interpreter(stdout):
         if os.path.isfile(cache_file):
             try:
                 with open(cache_file, "rb") as fp:
-                    return SpawnedJob.completed(create_interpreter(fp.read()))
+                    return SpawnedJob.completed(create_interpreter(fp.read(), check_binary=True))
             except (IOError, OSError, cls.Error, PythonIdentity.Error):
                 safe_rmtree(cache_dir)
```
jsirois, line 460: We purge the cache right here. Note the new `check_binary` argument is only passed here (in the try above).

Issue #1055, opened 2020-10-03
Ensure `--unzip` mode performance does not regress.


jsirois:
As noted in https://github.com/pantsbuild/pex/issues/1054#issuecomment-703150138 Pex `--unzip` mode performance improvements are sensitive to import scope in the `__main__.py` injected in PEXes to bootstrap. We should ensure performance does not regress un-noticed. This probably means adding some form of automated performance test that can fail CI at base and then probably some change to the bootstrap code to be systemically less sensitive to import scope creep slowing the bootstrap down.

jsirois:
See #1037 which tracks the need for a broader class of performance / size tests beyond this narrow `--unzip` bootstrap overhead case.

PR #1056, opened 2020-10-03
Fix `--unzip` performance regression.

### LINKED ISSUE #1054: Pex PEX perf regression
Pex ships a PEX of itself for those who wish to download a Pex binary instead of pip installing it. Pants is an important user of the Pex PEX and f07ae4704df40968918b6b4f80fcfafbee1c4327 (via git bisect) introduced a performance regression that induces ~300ms of startup overhead over a v2.1.16 baseline.

Tested with the following which emulates the simplest Pex PEX use case in Pants for interpreter selection for Pants tools:
```
$ python -mpex . -cpex --unzip -opex.pex
$ ./pex.pex --interpreter-constraint=">=3.6" -- -c 'import sys; print(sys.executable)'
$ multitime -n10 ./pex.pex --interpreter-constraint=">=3.6" -- -c 'import sys; print(sys.executable)'
```

v2.1.16
```
===> multitime results
1: ./pex.pex --interpreter-constraint=>=3.6 -- -c "import sys; print(sys.executable)"
real        0.861       0.007       0.851       0.859       0.874       
user        0.780       0.008       0.770       0.779       0.798       
sys         0.080       0.009       0.067       0.080       0.093 
```

f07ae4704df40968918b6b4f80fcfafbee1c4327
```
===> multitime results
1: ./pex.pex --interpreter-constraint=>=3.6 -- -c "import sys; print(sys.executable)"
real        1.184       0.013       1.170       1.182       1.217       
user        1.076       0.012       1.056       1.080       1.091       
sys         0.101       0.012       0.080       0.099       0.123
```

jsirois:
This is a bandaid fix only. Appropriate comments have been added to
help prevent re-regression with a longer term fix tracked by #1055.

Fixes #1054

### DIFF
--- a/pex/pex_info.py
+++ b/pex/pex_info.py
@@ -10,13 +10,14 @@ from pex import pex_warnings
 from pex.common import can_write_dir, open_zip, safe_mkdtemp
 from pex.compatibility import PY2
 from pex.compatibility import string as compatibility_string
-from pex.interpreter import PythonInterpreter
 from pex.orderedset import OrderedSet
 from pex.typing import TYPE_CHECKING
 from pex.variables import ENV
 from pex.version import __version__ as pex_version
 
 if TYPE_CHECKING:
+    from pex.interpreter import PythonInterpreter
+
     from typing import Optional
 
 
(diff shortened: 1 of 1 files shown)


Eric-Arellano (approved):
Thank you

pex/pex_info.py
```diff
@@ -56,6 +57,15 @@ class PexInfo(object):
 
     @classmethod
     def make_build_properties(cls, interpreter=None):
+        # This lazy import is currently needed for performance reasons. At PEX runtime PexInfo is
+        # read in the bootstrap to see if the PEX should run in `--unzip` mode. If so, it must
+        # re-exec itself to run against its unzipped contents. Since `make_build_properties` is only
+        # used at PEX buildtime and the transitive imports of PythonInterpreter are large and slow,
+        # we avoid this import cost for runtime-only use.
```
cosmicexplorer, line 64: Great great comment!

cosmicexplorer (approved):
Thank you for expanding/clarifying the use case/implementation of regression testing with command lines! That is a great place for us to get started! :D

> This is a bandaid fix only.

This probably shouldn't have confused me, but I interpreted this to mean that there was some larger performance bug in the code that wasn't being fixed here. I was satisfied upon realizing the bug was not covering the `--unzip` case with regression testing.

PR #1061, opened 2020-10-08
Fix Variables typing.

### LINKED ISSUE #1059: Variables.strip_defaults leads to misleading types
Variables currently typecheck, but the types do not reflect reality. This forces marking many properties as `Optional` that can never be in production use. Fix this.

jsirois:
Re-structure `strip_defaults` to allow typing to reflect real usage and
simplify call sites.

Fixes #1059


Eric-Arellano (approved):
Thanks. This looks great. I appreciate restructuring the code rather than using `cast` everywhere.

PR #1065, opened 2020-10-09
Honor `sys.executable` unless macOS Framework.

jsirois:
The prior fix for #1009 undid the fix in #860. Re-enable that fix
for any case that's clearly not in the context of a macOS Framework
Python build. Also add tests for pyenv shims to prevent further
regressions.

### DIFF
--- a/pex/interpreter.py
+++ b/pex/interpreter.py
@@ -69,16 +69,17 @@ class PythonIdentity(object):
 
     @classmethod
     def get(cls, binary=None):
+        # type: (Optional[str]) -> PythonIdentity
+
         # N.B.: We should not need to look past `sys.executable` to learn the current interpreter's
         # executable path, but on OSX there has been a bug where the `sys.executable` reported is
         # _not_ the path of the current interpreter executable:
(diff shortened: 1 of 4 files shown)


pex/interpreter.py
```diff
@@ -69,16 +69,17 @@ class UnknownRequirement(Error):
...
     @classmethod
     def get(cls, binary=None):
+        # type: (Optional[str]) -> PythonIdentity
+
         # N.B.: We should not need to look past `sys.executable` to learn the current interpreter's
         # executable path, but on OSX there has been a bug where the `sys.executable` reported is
         # _not_ the path of the current interpreter executable:
-        # https://bugs.python.org/issue22490#msg283859
-        if binary and binary != sys.executable:
-            TRACER.log(
```
jsirois, line 77: These logs were never viewable since they were emitted durring a subprocess that captures both stdout and stderr as part of the execute_parallel Job / Retain framework. As such, just killed.

Eric-Arellano (approved):
Thank you

PR #1072, opened 2020-10-10
Encapsulate `--inherit-path` handling.

### LINKED ISSUE #921: Pex should validate the PEX_INHERIT_PATH option and fail fast.
Currently `--inherit-path` / `PEX_INHERIT_PATH` is a tristate string and the value of that string is not validated against the valid set of `{'false', 'prefer', 'fallback'}` when passed via the `PEX_INHERIT_PATH` environment variable. We should fix `Variables` to do this validation like `pex.bin.pex` does in its option parsing:
https://github.com/pantsbuild/pex/blob/a07954a1bce1cf0c023c67f36e4e37f078eb23ba/pex/bin/pex.py#L264-L274

Lack of fail-fast parsing tripped up @smezouar in #920.

jsirois:
Factor out an enum-like class that encodes legacy bool behavior
centrally and use this to provide the same validation in
`--inherit-path` and PEX_INHERIT_PATH.

Fixes #921

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -16,6 +16,7 @@ from textwrap import TextWrapper
 
 from pex import pex_warnings
 from pex.common import die, safe_delete, safe_mkdtemp
+from pex.inherit_path import InheritPath
 from pex.interpreter import PythonInterpreter
 from pex.interpreter_constraints import (
     UnsatisfiableInterpreterConstraintsError,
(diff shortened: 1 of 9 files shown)


pex/pex_info.py
```diff
@@ -143,9 +151,7 @@ def __init__(self, info=None):
             raise ValueError(
                 "PexInfo can only be seeded with a dict, got: " "%s of type %s" % (info, type(info))
             )
-        self._pex_info = info or {}
-        if "inherit_path" in self._pex_info:
```
jsirois, line 147: N.B.: This dance was just enforcing coercion of old bool values into new str enum values. This is now centralized in `InheritPath.for_value` so there is no longer a need here.

pex/bin/pex.py
```diff
@@ -926,11 +928,12 @@ def walk_and_do(fn, src_dir):
 
 
 def transform_legacy_arg(arg):
+    # type: (str) -> str
     # inherit-path used to be a boolean arg (so either was absent, or --inherit-path)
     # Now it takes a string argument, so --inherit-path is invalid.
     # Fix up the args we're about to parse to preserve backwards compatibility.
     if arg == "--inherit-path":
-        return "--inherit-path=prefer"
+        return "--inherit-path={}".format(InheritPath.PREFER.value)
```
illicitonion, line 936: In retrospect, would've been nice to drop this legacy behaviour in 2.0. Maybe worth deprecating?

pex/inherit_path.py
```diff
@@ -0,0 +1,42 @@
...
+
+from __future__ import absolute_import
+
+from pex.typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from typing import Tuple, Union
+
+
+class InheritPath(object):
```
illicitonion, line 12: Why not use `enum`?

illicitonion (approved):
There's a lot of stuff mixed in with this change, but it all looks reasonable and plausible :) Thanks!

pex/inherit_path.py
jsirois: Pex still supports Python 2.7.

jsirois:
> There's a lot of stuff mixed in with this change, but it all looks reasonable and plausible :) Thanks!

Its all typing and @Eric-Arellano likes that.

pex/bin/pex.py
jsirois: Agreed, but we still have no deprecation mechanism: #584

Eric-Arellano (approved):

Issue #1012, opened 2020-07-24
Allow constraining the interpreter search path directly.


jsirois:
Right now the Pex API allows this:
https://github.com/pantsbuild/pex/blob/b51b340bea8a585a246997071a815d008986d6e3/pex/interpreter.py#L278-L288

Neither the CLI (buildtime) nor runtime via PEX-INFO / ENV vars does however forcing manipulation of the PATH to acheive the desired effect. This is problematic at buildtime since during wheel building tools may need to be invoked by subprocesses (think gcc). This could also be problematic at runtime for any PEX housing user code that needs to shell out to tools expected to be on the PATH.

See https://github.com/pantsbuild/pants/issues/9760 for the motivating issue.

stuhood:
Echoing a comment from the other ticket: Pants' API for this is: https://github.com/pantsbuild/pants/blob/e328b58af917440dc98c8aa88dbd49a9355c5e8d/src/python/pants/python/python_setup.py#L87-L99, which is implemented here:
https://github.com/pantsbuild/pants/blob/e328b58af917440dc98c8aa88dbd49a9355c5e8d/src/python/pants/python/python_setup.py#L178-L205

jsirois:
OK - Lies above. Both buildtime and runtime support PEX_PYTHON_PATH which is a path-like env var that accepts both directory entries and paths to interpreter binaries:
1. Build time:
https://github.com/pantsbuild/pex/blob/5fd9b1527691f3efb42a469a563a7653a3a1594d/pex/bin/pex.py#L722-L725
2. Run time:
https://github.com/pantsbuild/pex/blob/b51b340bea8a585a246997071a815d008986d6e3/pex/pex_bootstrapper.py#L203-L215

This fully suffices for the Pants motivating use case. It still may make sense to add a CLI flag / PexInfo metadata slot for this to bake in a "corporate" interpreter path without resorting to ~/.pexrc or /etc/pexrc, but that resort has been working? for Twitter at least as an example of corporate.

Eric-Arellano:
Closed by https://github.com/pantsbuild/pex/pull/1077. https://github.com/pantsbuild/pex/issues/1075 tracks removing `PEX_PYTHON_PATH` from build time.

Issue #1075, opened 2020-10-14
`PEX_PYTHON_PATH` impacts interpreter selection for Pex installed from pip, but not the `pex` release file


Eric-Arellano:
Using `pex` installed via `pip install pex`, I get this behavior:

```
pex --interpreter-constraint='CPython>=3.5' -- -c "import sys, os; print(os.path.realpath(sys.executable))"
/Users/eric/.pyenv/versions/3.6.10/bin/python3.6
```

```
PEX_PYTHON_PATH=/Users/eric/.pyenv/versions/3.7.7/bin pex --interpreter-constraint='CPython>=3.5' -- -c "import sys, os; print(os.path.realpath(sys.executable))"
/Users/eric/.pyenv/versions/3.7.7/bin/python3.7
```

But using `pex` installed from the releases page, I get this behavior:

```
./pex --interpreter-constraint='CPython>=3.5' -- -c "import sys, os; print(os.path.realpath(sys.executable))"
/Users/eric/.pyenv/versions/3.6.10/bin/python3.6
```

```
PEX_PYTHON_PATH=/Users/eric/.pyenv/versions/3.7.7/bin ./pex --interpreter-constraint='CPython>=3.5' -- -c "import sys, os; print(os.path.realpath(sys.executable))"
/Users/eric/.pyenv/versions/3.6.10/bin/python3.6
```

Eric-Arellano:
I now understand why setting `PEX_PYTHON_PATH` works when running Pex when installed via `pip`. `Variables(rc=options.rc_file)` ends up reading from `os.environ` still:

https://github.com/pantsbuild/pex/blob/20a67c3e95e4d15b9e200dc15e9b01aad2cbb83e/pex/variables.py#L184-L192

This means that unless the user explicitly has set `PEX_IGNORE_RCFILES`, `PEX_PYTHON_PATH` _is_ already being used, regardless of RC files having been created. I'm not sure we were aware this happens.

--

It's not clear to me yet why the behavior is different when running Pex via the release page.

Eric-Arellano:
> It's not clear to me yet why the behavior is different when running Pex via the release page.

Found it: https://github.com/pantsbuild/pex/pull/741. We strip `PEX_PYTHON_PATH` from `os.environ` when running `./pex`. By the time we're building a new PEX, this env var has been stripped. Atm, the sole way to load `PEX_PYTHON_PATH` when building a new Pex by running `./pex` is to use an RC File, which won't get stripped.

[...]

PR #1077, opened 2020-10-14
Add `--python-path` to change interpreter search paths when building a PEX

Eric-Arellano:
Closes https://github.com/pantsbuild/pex/issues/1012.

Currently, PEX defaults to searching the $PATH when `--interpreter-constraint` and/or `--resolve-local-platforms` are used. (Otherwise, it uses the current interpreter or `--python`).

You can specify `PEX_PYTHON_PATH`, but this has several issues:

* Env vars should only be used for runtime, not build time. We break that separation here.
* This only works if `PEX_IGNORE_RCFILES` is not set to true. Pants must set this option, meaning Pants cannot leverage this mechanism.
* The behavior doesn't work as expected if running Pex as a Pex: https://github.com/pantsbuild/pex/issues/1075

Instead of further using `PEX_PYTHON_PATH`, we add a proper `--python-path` flag. In a followup, we can remove the `PEX_PYTHON_PATH` mechanism.

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -458,10 +458,26 @@ def configure_clp_pex_environment(parser):
         default=[],
         type="str",
         action="append",
-        help="The Python interpreter to use to build the pex.  Either specify an explicit "
-        "path to an interpreter, or specify a binary accessible on $PATH. This option "
-        "can be passed multiple times to create a multi-interpreter compatible pex. "
-        "Default: Use current interpreter.",
+        help=(
+            "The Python interpreter to use to build the PEX (default: current interpreter). This "
+            "cannot be used with `--interpreter-constraint`, which will instead cause PEX to "
+            "search for valid interpreters. Either specify an absolute path to an interpreter, or "
+            "specify a binary accessible on $PATH like `python3.7`. This option can be passed "
+            "multiple times to create a multi-interpreter compatible PEX."
+        ),
+    )
+    group.add_option(
+        "--python-path",
+        dest="python_path",
+        default=None,
+        type="str",
+        help=(
+            "The paths used by PEX to find valid interpreters when `--interpreter-constraint` "
+            "and/or `--resolve-local-platforms` are used (default: the $PATH). Specify a "
+            "colon-separated string containing paths of blessed Python interpreters. Can be "
+            "absolute paths to interpreters or standard $PATH style directory entries that are "
+            "searched for child files that are python binaries. "
+        ),
     )
 
     group.add_option(
@@ -470,18 +486,23 @@ def configure_clp_pex_environment(parser):
         default=[],
         type="str",
         action="append",
-        help="Constrain the selected Python interpreter. Specify with Requirement-style syntax, "
-        'e.g. "CPython>=2.7,<3" (A CPython interpreter with version >=2.7 AND version <3) '
-        'or "PyPy" (A pypy interpreter of any version). This argument may be repeated multiple '
-        "times to OR the constraints.",
+        help=(
+            "Constrain the selected Python interpreter. Specify with Requirement-style syntax, "
+            'e.g. "CPython>=2.7,<3" (A CPython interpreter with version >=2.7 AND version <3) '
+            'or "PyPy" (A pypy interpreter of any version). This argument may be repeated multiple '
+            "times to OR the constraints."
+        ),
     )
 
     group.add_option(
         "--rcfile",
         dest="rc_file",
         default=None,
-        help="An additional path to a pexrc file to read during configuration parsing. "
-        "Used primarily for testing.",
+        help=(
+            "An additional path to a pexrc file to read during configuration parsing, in addition "
+            "to reading `/etc/pexrc` and `~/.pexrc`. If `PEX_IGNORE_RCFILES=true`, then all rc "
+            "files will be ignored."
+        ),
     )
 
     group.add_option(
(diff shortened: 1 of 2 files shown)


pex/bin/pex.py
```diff
@@ -458,10 +458,26 @@ def configure_clp_pex_environment(parser):
         default=[],
         type="str",
         action="append",
-        help="The Python interpreter to use to build the pex.  Either specify an explicit "
-        "path to an interpreter, or specify a binary accessible on $PATH. This option "
-        "can be passed multiple times to create a multi-interpreter compatible pex. "
-        "Default: Use current interpreter.",
+        help=(
```
Eric-Arellano, line 461: Reworded. I wanted to add a little insight into how Pex only sometimes searches for interpreters.

pex/bin/pex.py
```diff
@@ -470,18 +486,23 @@ def configure_clp_pex_environment(parser):
         default=[],
         type="str",
         action="append",
-        help="Constrain the selected Python interpreter. Specify with Requirement-style syntax, "
-        'e.g. "CPython>=2.7,<3" (A CPython interpreter with version >=2.7 AND version <3) '
-        'or "PyPy" (A pypy interpreter of any version). This argument may be repeated multiple '
-        "times to OR the constraints.",
+        help=(
```
Eric-Arellano, line 489: Only reformatted.

pex/bin/pex.py
```diff
@@ -470,18 +486,23 @@ def configure_clp_pex_environment(parser):
...
+        ),
     )
 
     group.add_option(
         "--rcfile",
         dest="rc_file",
         default=None,
-        help="An additional path to a pexrc file to read during configuration parsing. "
-        "Used primarily for testing.",
+        help=(
```
Eric-Arellano, line 501: Reworded to emphasize this appends, rather than replaces. And to mention the env var.

pex/bin/pex.py
```diff
@@ -458,10 +458,26 @@ def configure_clp_pex_environment(parser):
...
+    )
+    group.add_option(
+        "--python-path",
+        dest="python_path",
+        default=None,
+        type="str",
+        help=(
+            "The paths used by PEX to find valid interpreters when `--interpreter-constraint` "
+            "and/or `--resolve-local-platforms` are used (default: the $PATH). Specify a "
+            "colon-separated string containing paths of blessed Python interpreters. Can be "
```
benjyw, line 477: I don't think "blessed" is a useful word here (I know it's in the docstring for `PEX_PYTHON_PATH`). What does it mean? And the wording here is rather strained and redundant in general. 

How about something like: 

```
help=(
    "Colon-separated paths to search for interpreters when `--interpreter-constraint` or `--resolve-local-platforms` are specified (default: $PATH). Each element can be the absolute path of an interpreter binary, or a directory containing interpreter binaries.
)
```

benjyw (approved):

pex/bin/pex.py
```diff
@@ -721,8 +742,9 @@ def _safe_link(src, dst):
 def build_pex(reqs, options, cache=None):
     interpreters = None  # Default to the current interpreter.
 
-    pex_python_path = None  # Defaults to $PATH
-    if options.rc_file or not ENV.PEX_IGNORE_RCFILES:
+    pex_python_path = options.python_path  # If None, this will result in using $PATH.
+    # TODO(#1075): stop looking at PEX_PYTHON_PATH and solely consult the `--python-path` flag.
+    if not pex_python_path and (options.rc_file or not ENV.PEX_IGNORE_RCFILES):
```
benjyw, line 747: Should this be `if pex_python_path is not None`? What do we want the behavior to be if `--python-path` is explicitly set to an empty path?

benjyw (approved):

Eric-Arellano:
Oops...my GitHub UI was messed up and showed this was green, when CI checks were still running. Let's hope this does go green..I'll fix forward otherwise.

jsirois:
Thanks for digging into this issue. This is definitely the way to go.

Eric-Arellano:
Yay! Should we indeed remove `PEX_PYTHON_PATH` from ever being used at build time? Happy to do so, only wanted your ack first.

jsirois:
I'm not sure. The whole PEX_PYTHON, PEX_PYTHON_PATH, .pexrc business was a Twitter attempt in good faith that is currently a mess. There is no urgency I can see at any rate.

Eric-Arellano:
The only urgency imo is that `PEX_PYTHON_PATH` will be used unless you set `PEX_IGNORE_RCFILES=true`. I wouldn't be very concerned if you have to set `--rcfile` for it to be activated. But as found in https://github.com/pantsbuild/pex/issues/1075, you only have to leave off `PEX_IGNORE_RCFILES` for this behavior. That's the default, which isn't ideal.

jsirois:
I think no-one except Twitter and Pants use PEX_PYTHON_PATH and IIUC Pants will now be dropping use - at least at buildtime. Twitter uses it in .pexrc files and so they (should) know what they're in for, especially since they wrote the feature.

Eric-Arellano:
> IIUC Pants will now be dropping use - at least at buildtime.

This is correct. Pants will only use `PEX_PYTHON_PATH` at runtime if the Pex was created with `internal_only=False`. We currently never have Pants run an `internal_only=False` Pex, so in practice, we'll never end up setting it.

PR #1083, opened 2020-10-19
Switch from optparse to argparse.

benjyw:
The impetus is to allow us to use a sphinx plugin to generate CLI documentation.

This is also an opportunity to simplify various aspects of CLI parsing.

For example: There was some overly-clever code to update the `indexes`
option dest from two different options (--indexes and --pypi). This
change does the more obvious and simple thing, and as a result by
default we pass in 'https://pypi.org/simple' explicitly, instead of 
passing in None and letting the underlying code default to PyPI.

Also removes hand-written "Default: value" in the help strings,
since argparse can generate those automatically.


Eric-Arellano (approved):
This is great. Thank you

benjyw:
I tested this manually by passing in various flags, including all the ones with custom actions, and inspecting the resulting values.

pex/bin/pex.py
```diff
@@ -43,168 +49,136 @@ def log(msg, V=0):
...
-def parse_bool(option, opt_str, _, parser):
-    setattr(parser.values, option.dest, not opt_str.startswith("--no"))
-
-
-def increment_verbosity(option, opt_str, _, parser):
-    verbosity = getattr(parser.values, option.dest, 0)
-    setattr(parser.values, option.dest, verbosity + 1)
-
-
-def process_disable_cache(option, option_str, option_value, parser):
```
benjyw, line 55: Note that this was unused anywhere.

pex/bin/pex.py
```diff
@@ -43,168 +49,136 @@ def log(msg, V=0):
...
-    else:
-        indexes = getattr(parser.values, option.dest, [])
-        if _PYPI not in indexes:
-            indexes.append(_PYPI)
-        setattr(parser.values, option.dest, indexes)
 
+_DEFAULT_MANYLINUX_STANDARD = "manylinux2014"
 
-def process_find_links(option, option_str, option_value, parser):
-    find_links = getattr(parser.values, option.dest, [])
```
benjyw, line 78: All this did was remove dups, which seems unnecessary, so we just use normal `action='append'` now.

benjyw:
cc @jsirois, this is worth a careful review.

jsirois:
Oops! I totally missed this. I'll give it a good look. Thanks.

pex/bin/pex.py
```diff
@@ -235,84 +209,72 @@ def configure_clp_pex_resolution(parser):
...
-        action="callback",
-        callback=process_manylinux,
-        help=(
-            "Whether to allow resolution of manylinux wheels for linux target "
-            "platforms; Default: allow manylinux wheels compatible with {}".format(
-                _DEFAULT_MANYLINUX_STANDARD
-            )
-        ),
+        action=ManylinuxAction,
+        help=("Whether to allow resolution of manylinux wheels for linux target platforms."),
```
jsirois, line 264: The parens could go.

pex/bin/pex.py
```diff
@@ -738,6 +681,15 @@ def _safe_link(src, dst):
     os.symlink(src, dst)
 
 
+def compute_indexes(options):
+    # type: (Namespace) -> List[str]
+
+    indexes = ([_PYPI] if options.pypi else []) + (options.indexes or [])
+    seen = {}  # type: Dict[str, str]
+    deduped_indexes = [seen.setdefault(x, x) for x in indexes if x not in seen]
+    return deduped_indexes
```
jsirois, line 690: How about just `return list(OrderedSet(indexes))`?

jsirois (approved):
Thanks - this lgtm!

benjyw:
I wonder if the next Pex release should bump the minor version, since this is a non-trivial change (even if it has no intended user-facing implications). We're not officially semver-ing in this repo, but unofficially it might be good to use this to imply that a potentially more disruptive than usual thing has happened.

PR #1087, opened 2020-10-20
Fix `--pex-python` when it's the same as the current interpreter

Eric-Arellano:
Before, this unexpectedly fails:

```
$ PATH=/Library/Developer/CommandLineTools/usr/bin/python3 PEX_IGNORE_RCFILES=true /Users/eric/.pyenv/versions/3.8.5/bin/python ./pex --interpreter-constraint='CPython>=3.7' --python-path='/Users/eric/.pyenv/versions/3.8.5/bin/python' -- -c 'import sys; print(sys.executable)'
/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.7/bin/python3.7
```

But this succeeds:

```
$ PATH=/Library/Developer/CommandLineTools/usr/bin/python3 PEX_IGNORE_RCFILES=true /Library/Developer/CommandLineTools/usr/bin/python3 ./pex --interpreter-constraint='CPython>=3.7' --python-path='/Users/eric/.pyenv/versions/3.8.5/bin/python' -- -c 'import sys; print(sys.executable)'
/Users/eric/.pyenv/versions/3.8.5/bin/python3.8
```

This issue is that we remove the current interpreter from our candidates if they are valid with the interpreter constraints. When `--pex-path` has only one entry, this results in the `paths` variable being empty, so that it looks like the user never specified `--pex-path`, and we fall back to `$PATH`.

### DIFF
--- a/pex/interpreter.py
+++ b/pex/interpreter.py
@@ -312,7 +312,8 @@ class PythonInterpreter(object):
     @staticmethod
     def _paths(paths=None):
         # type: (Optional[Iterable[str]]) -> Iterable[str]
-        return paths or os.getenv("PATH", "").split(os.pathsep)
+        # NB: If `paths=[]`, we will not read $PATH.
+        return paths if paths is not None else os.getenv("PATH", "").split(os.pathsep)
 
     @classmethod
     def iter(cls, paths=None):
--- a/pex/pex_bootstrapper.py
+++ b/pex/pex_bootstrapper.py
@@ -72,33 +72,40 @@ def iter_compatible_interpreters(
         # type: () -> Iterator[InterpreterOrError]
         seen = set()
 
-        paths = None  # type: Optional[MutableSet[str]]
-        if path:
-            paths = OrderedSet(os.path.realpath(p) for p in path.split(os.pathsep))
+        normalized_paths = (
+            OrderedSet(os.path.realpath(p) for p in path.split(os.pathsep)) if path else None
+        )
 
+        # Check if the current interpreter is valid. If `path` was specified, the current
+        # interpreter must be included in the value; otherwise, we always yield the interpreter.
         current_interpreter = PythonInterpreter.get()
         if not _valid_path or _valid_path(current_interpreter.binary):
-            if paths:
-                # Prefer the current interpreter if present on the `path`.
+            if normalized_paths:
                 candidate_paths = frozenset(
                     (current_interpreter.binary, os.path.dirname(current_interpreter.binary))
                 )
-                candidate_paths_in_path = candidate_paths.intersection(paths)
+                candidate_paths_in_path = candidate_paths.intersection(normalized_paths)
                 if candidate_paths_in_path:
                     for p in candidate_paths_in_path:
-                        paths.remove(p)
+                        normalized_paths.remove(p)
                     seen.add(current_interpreter)
                     yield current_interpreter
             else:
-                # We may have been invoked with a specific interpreter, make sure our sys.executable is
-                # included as a candidate in this case.
                 seen.add(current_interpreter)
                 yield current_interpreter
 
-        for interp in PythonInterpreter.iter_candidates(paths=paths, path_filter=_valid_path):
-            if interp not in seen:
-                seen.add(interp)
-                yield interp
+        # NB: We only check for other interpreters if there are remaining candidate interpreters
+        # _or_ if the user left off `path`, meaning to default to looking at $PATH.
+        #
+        # There is a tricky edge case where `path` was specified, but its only entry is the current
+        # interpreter so it was already removed.
+        if normalized_paths or path is None:
+            for interp in PythonInterpreter.iter_candidates(
+                paths=normalized_paths, path_filter=_valid_path
+            ):
+                if interp not in seen:
+                    seen.add(interp)
+                    yield interp
 
     def _valid_interpreter(interp_or_error):
         # type: (InterpreterOrError) -> bool
--- a/pex/pex_bootstrapper.py
+++ b/pex/pex_bootstrapper.py
@@ -76,8 +76,7 @@ def iter_compatible_interpreters(
             OrderedSet(os.path.realpath(p) for p in path.split(os.pathsep)) if path else None
         )
 
-        # Check if the current interpreter is valid. If `path` was specified, the current
-        # interpreter must be included in the value; otherwise, we always yield the interpreter.
+        # Prefer the current interpreter, if valid.
         current_interpreter = PythonInterpreter.get()
         if not _valid_path or _valid_path(current_interpreter.binary):
             if normalized_paths:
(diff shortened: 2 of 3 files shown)


pex/pex_bootstrapper.py
```diff
@@ -72,33 +72,40 @@ def _iter_interpreters():
...
-        for interp in PythonInterpreter.iter_candidates(paths=paths, path_filter=_valid_path):
-            if interp not in seen:
-                seen.add(interp)
-                yield interp
+        # NB: We only check for other interpreters if there are remaining candidate interpreters
+        # _or_ if the user left off `path`, meaning to default to looking at $PATH.
+        #
+        # There is a tricky edge case where `path` was specified, but its only entry is the current
+        # interpreter so it was already removed.
+        if normalized_paths or path is None:
```
benjyw, line 102: AFAICT this is the only actual logic change? The other changes (renaming of `paths` to `normalized_paths`) here makes it hard to see.

benjyw: I don't actually understand why this fixes the problem. Or, more precisely, why there was a problem in the first place. AFAICT we only remove the current interpreter if we're about to yield it. So it should be yielded by this function, and then I don't see why it matters what this `iter_candidates` does? 

If I can't figure out that edge case by reading the code and the comment then the comment needs elaborating and, ideally, the code needs simplifying...

pex/pex_bootstrapper.py
```diff
@@ -72,33 +72,40 @@ def _iter_interpreters():
...
 
-        paths = None  # type: Optional[MutableSet[str]]
-        if path:
-            paths = OrderedSet(os.path.realpath(p) for p in path.split(os.pathsep))
+        normalized_paths = (
+            OrderedSet(os.path.realpath(p) for p in path.split(os.pathsep)) if path else None
+        )
 
+        # Check if the current interpreter is valid. If `path` was specified, the current
+        # interpreter must be included in the value; otherwise, we always yield the interpreter.
```
benjyw, line 80: The comment you deleted below has important information now missing here, namely that we yield the current interpreter *before all others*, so that it's preferred.

pex/pex_bootstrapper.py
```diff
@@ -72,33 +72,40 @@ def _iter_interpreters():
...
+            if normalized_paths:
                 candidate_paths = frozenset(
                     (current_interpreter.binary, os.path.dirname(current_interpreter.binary))
                 )
-                candidate_paths_in_path = candidate_paths.intersection(paths)
+                candidate_paths_in_path = candidate_paths.intersection(normalized_paths)
                 if candidate_paths_in_path:
                     for p in candidate_paths_in_path:
-                        paths.remove(p)
+                        normalized_paths.remove(p)
```
benjyw, line 90: I know this was the case before as well, but I think I see a problem here: we remove not just the full path to the interpreter, but also the directory containing it. What if there are other interpreters to consider in that directory?

benjyw: Also, since we do the `seen` check above, why do we even bother removing anything from `normalized_paths`?

pex/pex_bootstrapper.py
Eric-Arellano: I'll add that back. 

FYI the part I wanted to remove was ` # We may have been invoked with a specific interpreter` as I believe that's misleading; regardless of the current interp is via shebang discovery or specific interpreter, we should consider it.

pex/pex_bootstrapper.py
```diff
@@ -72,33 +72,40 @@ def _iter_interpreters():
...
                 yield current_interpreter
 
-        for interp in PythonInterpreter.iter_candidates(paths=paths, path_filter=_valid_path):
-            if interp not in seen:
-                seen.add(interp)
-                yield interp
+        # NB: We only check for other interpreters if there are remaining candidate interpreters
+        # _or_ if the user left off `path`, meaning to default to looking at $PATH.
+        #
+        # There is a tricky edge case where `path` was specified, but its only entry is the current
```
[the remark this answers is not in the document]
Eric-Arellano: `its` is being used as the possessive here, rather than the contraction `it is`.

pex/pex_bootstrapper.py
Eric-Arellano: > why do we even bother removing anything from normalized_paths?

Good question. I suspect performance? @jsirois do you remember the motivation in https://github.com/pantsbuild/pex/pull/783?

pex/pex_bootstrapper.py
Eric-Arellano: > So it should be yielded by this function, and then I don't see why it matters what this iter_candidates does?

We yield current interpreter, then remove it from the `normalized_paths`. Now, `normalized_paths=[]`. This means that `iter_candidates` will now see that it's empty, and will look like the user did not specify `--pex-path`, which triggers `iter_candidates` looking up the `$PATH`.

I'll revisit this all for readability.

pex/pex_bootstrapper.py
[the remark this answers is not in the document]
benjyw: Oh oops, I read it as the contraction for some reason.

pex/pex_bootstrapper.py
benjyw: OK, that was extremely non-obvious, but makes sense.

pex/pex_bootstrapper.py
benjyw: Yes, and we are always invoked with a specific interpreter, in some sense...

pex/pex_bootstrapper.py
benjyw: Not doing that would remove this weird case, no?

Eric-Arellano: Indeed. If we don't remove, we avoid this edge case without needing any special `if` statement and confusing comment.

pex/interpreter.py
```diff
@@ -312,7 +312,8 @@ def get(cls):
     @staticmethod
     def _paths(paths=None):
         # type: (Optional[Iterable[str]]) -> Iterable[str]
-        return paths or os.getenv("PATH", "").split(os.pathsep)
+        # NB: If `paths=[]`, we will not read $PATH.
+        return paths if paths is not None else os.getenv("PATH", "").split(os.pathsep)
```
Eric-Arellano, line 316: This is the new bug fix. The rest of the changes are light refactoring / tests.

pex/pex_bootstrapper.py
Eric-Arellano: I fixed this another way, but we may still want to remove this code. 

I think I'm not comfortable removing it until we hear back from John, though, so let's save for a followup? Iirc, he's said in the past that interpreter discovery is expensive, hence optimizations like preferring the current interpreter.

benjyw: I still think this is actually wrong though. We remove the directory, when there may be other interpreters in it.

benjyw: Still, that is the current behavior, so we can fix in a followup.

benjyw (approved):

pex/pex_bootstrapper.py
jsirois: This is definitely a bug: #1109 

That said, the if statement is still needed. If the function is called with a non `None` `path` then we must respect that and only include the current interpreter if it is a member of that `path`.

Issue #430, opened 2017-11-15
Make whether to take min or max version interpreter configurable in pex CLI


CMLivingston:
During interpreter selection portions of the pex codebase (for building pexes and for running them), the logic is currently selecting the lowest-version interpreter when multiple compatible interpreters are present. 

A useful feature would be to have a command line flag that allows the user to specify whether to take the min or max versioned interpreter amongst multiple compatible interpreters. This would allow greater flexibility in a given organization's  interpreter management strategies.

A clear example of when this filtering occurs is in the interpreter selection logic of the `build_pex` function in bin/pex.py.

PR #1088, opened 2020-10-20
Prefer more recent patch versions in interpreter discovery.

Eric-Arellano:
Currently, our naive `min()` means that Pex will choose 3.6.3 instead of 3.6.8, for example. This is not ideal because the newer patch will have important bug fixes, and should still be backward compatible.

For macOS users, there's an added benefit of this PR that the problematic system Pythons usually use older patch versions. A user is likely to have installed a more modern patch with both Homebrew and Pyenv; this PR means that Pex will correctly prefer those over the broken system Python.

This does not solve https://github.com/pantsbuild/pex/issues/430; we will still always prefer the minimum major/minor version, i.e. `3.6.1` is preferred over `3.6.0`, and both are preferred over `3.7.*`.

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -807,7 +807,11 @@ def build_pex(reqs, options, cache=None):
                 )
             )
 
-    interpreter = PythonInterpreter.safe_min(interpreters) if interpreters else None
+    interpreter = (
+        PythonInterpreter.latest_release_of_min_compatible_version(interpreters)
+        if interpreters
+        else None
+    )
 
     try:
         with open(options.preamble_file) as preamble_fd:
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -807,7 +807,7 @@ def build_pex(reqs, options, cache=None):
                 )
             )
 
-    interpreter = min(interpreters) if interpreters else None
+    interpreter = PythonInterpreter.safe_min(interpreters) if interpreters else None
 
     try:
         with open(options.preamble_file) as preamble_fd:
--- a/pex/interpreter.py
+++ b/pex/interpreter.py
@@ -29,7 +29,17 @@ from pex.util import CacheHelper
 from pex.variables import ENV
 
 if TYPE_CHECKING:
-    from typing import Callable, Dict, Iterable, Iterator, MutableMapping, Optional, Tuple, Union
+    from typing import (
+        Callable,
+        Dict,
+        Iterable,
+        Iterator,
+        MutableMapping,
+        Optional,
+        Sequence,
+        Tuple,
+        Union,
+    )
 
     PathFilter = Callable[[str], bool]
 
@@ -162,7 +172,13 @@ class PythonIdentity(object):
 
     @property
     def version(self):
-        return self._version
+        # type: () -> Tuple[int, int, int]
+        """The interpreter version as a normalized tuple.
+
+        This assumes that the interpreter uses semantic versioning in the form
+        <major>.<minor>.<patch>, which is used by CPython, PyPy, and Jython.
+        """
+        return cast("Tuple[int, int, int]", self._version)
 
     @property
     def version_str(self):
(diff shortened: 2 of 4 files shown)


pex/bin/pex.py
```diff
@@ -807,7 +807,7 @@ def to_python_interpreter(full_path_or_basename):
                 )
             )
 
-    interpreter = min(interpreters) if interpreters else None
+    interpreter = PythonInterpreter.safe_min(interpreters) if interpreters else None
```
Eric-Arellano, line 810: It felt wrong to put this logic in `__lt__`. I want to call attention to the magic.

Because I deleted `__lt__`, calling `min()` will fail.

pex/interpreter.py
```diff
@@ -162,7 +172,13 @@ def platform_tag(self):
 
     @property
     def version(self):
-        return self._version
+        # type: () -> Tuple[int, int, int]
+        """The interpreter version as a normalized tuple.
+
+        This assumes that the interpreter uses semantic versioning in the form
+        <major>.<minor>.<patch>, which is used by CPython, PyPy, and Jython.
```
Eric-Arellano, line 179: Please let me know if this is not safe. The only interpreter I can think of is `IronPython`, which also uses semvar it looks like: https://ironpython.net.

jsirois (commented):
Not a review, but:

> For macOS users, there's an added benefit of this PR that the problematic system Pythons usually use older patch versions. A user is likely to have installed a more modern patch with both Homebrew and Pyenv; this PR means that Pex will correctly prefer those over the broken system Python.

So is this change just changing the semantic to latest compatible for general sanity sake or is it also actually papering over bugs with macOS interpreter selection?

Eric-Arellano:
> So is this change just changing the semantic to latest compatible for general sanity sake

Primarily I believe this is better behavior, regardless of macOS. I can't think of a reason we would not want to prefer the bug fixes in more recent patch releases.

> actually papering over bugs with macOS interpreter selection

The bugs we know about are things like PermissionDenied when installing certain dependencies. Iiuc, we haven't considered banning the macOS system interpreters, so they can always be a problem. But this hopefully reduces the frequency. Outside of an outright ban, I don't believe there's much we can do to fix these bugs, hence the `--pex-python` approach for the user to be able to avoid them.

pex/interpreter.py
Eric-Arellano: Thanks for pointing out `sys.version_info`! Will do.

pex/interpreter.py
```diff
@@ -305,6 +321,26 @@ class IdentificationError(Error):
...
 
+    @staticmethod
+    def safe_min(interps):
+        # type: (Sequence[PythonInterpreter]) -> PythonInterpreter
+        """Find the minimum major version, but use the most recent patch version within that minor
+        version.
+
+        That is, 3.6.1 < 3.6.0 < 3.7.*.
+        """
+        assert interps, "No interpreters passed to `PythonInterpreter.safe_min()`"
```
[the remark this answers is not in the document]
Eric-Arellano: Excellent idea!

pex/interpreter.py
```diff
@@ -734,11 +771,6 @@ def __eq__(self, other):
             return NotImplemented
         return self._binary == other._binary
 
-    def __lt__(self, other):
```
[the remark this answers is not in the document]
Eric-Arellano: Why keep it? I'm concerned someone will try to call `min()` in the feature without realizing this nuance.

pex/interpreter.py
```diff
@@ -305,6 +320,19 @@ class IdentificationError(Error):
     class InterpreterNotFound(Error):
         pass
 
+    @staticmethod
+    def latest_release_of_min_compatible_version(interps):
+        # type: (Sequence[PythonInterpreter]) -> PythonInterpreter
+        """Find the minimum major version, but use the most recent patch version within that minor
```
benjyw, line 326:
```suggestion
        """Find the minimum major version, but use the most recent micro version within that minor
```

pex/interpreter.py
```diff
@@ -305,6 +320,19 @@ class IdentificationError(Error):
...
+        version.
+
+        That is, 3.6.1 < 3.6.0 < 3.7.*.
```
benjyw, line 329: I wouldn't use `<` in this example, since again this is not intended to be thought of as an ordering, per se.

benjyw: So just say something like "3.6.1 is selected over 3.6.0, but 3.6.0 is selected over 3.7.*".

pex/interpreter.py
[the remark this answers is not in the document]
benjyw: Someone might want the actual minimum? I guess better to remove this after all so that it'll fail hard if our own code is doing this somewhere.

benjyw (approved):

pex/interpreter.py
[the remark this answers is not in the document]
Eric-Arellano: Exactly, that's my thinking. We can add it back if we find we do need the true minimum.

PR #1102, opened 2020-11-07
Fix ResourceWarning for unclosed /dev/null.

### LINKED ISSUE #1011: Release 2.1.14
On the docket:

* [x] Add `--use-first-matching-interpreter` to speed up performance of builds at the expense of worse compatibility at runtime. This is an important optimization for Pants's use of Pex. https://github.com/pantsbuild/pex/pull/1008

### LINKED ISSUE #1101: Unclosed resource warning for `/dev/null` in PEX teardown.
This presents as:
```
02:44:24.35 [WARN] sys:1: ResourceWarning: unclosed file <_io.TextIOWrapper name='/dev/null' mode='w' encoding='UTF-8'>
```

This is unfortunate since we really do intend to leave that filehandle open until Python tears it down :/:
https://github.com/pantsbuild/pex/blob/47047749f52d80f16dd8b4fd940e76c0c110fdd0/pex/pex.py#L458-L464

Fixing the warning will require `warnings` management or giving up on this robust way to `/dev/null` teardown and returning to the hackier construction of a `DevNull` file-like object as was done previously.

jsirois:
Since the lack of an explicit close is intended, we suppress the
warning.

Fixes #1101

### DIFF
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -4,7 +4,9 @@
 from __future__ import absolute_import, print_function
 
 import os
+import re
 import sys
+import warnings
 from distutils import sysconfig
 from site import USER_SITE
 from types import ModuleType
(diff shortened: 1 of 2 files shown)


jsirois:
N.B.: The test fails under tox envs py3{5,6,7,8,9} with the new warning supression code commented out.

illicitonion (approved):

Eric-Arellano (approved):

PR #1103, opened 2020-11-07
Deprecate `-R`/`--resources-directory`.

### LINKED ISSUE #1100: Remove `--sources-directory` / `--resources-directory` distinction.
We do nothing different when gathering these for inclusion in a PEX file. They all just get copied into the PEX file chroot.

jsirois:
Fixes #1100

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -634,8 +634,10 @@ def configure_clp():
         default=[],
         type=str,
         action="append",
-        help="Add sources directory to be packaged into the generated .pex file."
-        "  This option can be used multiple times.",
+        help=(
+            "Add a directory containing sources and/or resources to be packaged into the generated "
+            ".pex file. This option can be used multiple times."
+        ),
     )
 
     parser.add_option(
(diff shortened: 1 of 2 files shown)


pex/bin/pex.py
```diff
@@ -822,19 +827,19 @@ def to_python_interpreter(full_path_or_basename):
 
     pex_builder = PEXBuilder(path=safe_mkdtemp(), interpreter=interpreter, preamble=preamble)
 
-    def walk_and_do(fn, src_dir):
-        src_dir = os.path.normpath(src_dir)
-        for root, dirs, files in os.walk(src_dir):
+    if options.resources_directory:
+        pex_warnings.warn(
+            "The `-R/--resources-directory` option is deprecated. Resources should be added via "
+            "`-D/--sources-directory` instead."
```
Eric-Arellano, line 833: While I know we don't yet have a deprecation policy, thoughts on choosing a removal version here so that users can plan? Perhaps the version after this next one.

Eric-Arellano (approved):

pex/bin/pex.py
jsirois: Until we have a policy / infra I'd prefer not to defacto establish one ad-hoc.

PR #1110, opened 2020-11-10
Fix `iter_compatible_interpreters` with `path`.

### LINKED ISSUE #1109: Pex discards the current interpreter's PATH entry when it is a directory entry.
This means no other interpreters in the current interpreter's directory (sibling interpreters) can be found. This is an egregious problem since interpreters are commonly installed to /usr/bin; i.e.: /usr/bin/python2.7, /usr/bin/python3.6, ... etc will all be siblings and if a PEX is run with any one of them, the others won't be discoverable.

N.B.: Pex only encounters this condition when `PEX_PYTHON_PATH` / `--python-path` are used, populating the `path` function parameter below:
https://github.com/pantsbuild/pex/blob/93b0442c9574a0bba018249d842dc2ac29dc1e27/pex/pex_bootstrapper.py#L40-L45

And the bug is here where we remove both the interpreter binary path (which is ok) and the interpreter binary directory (which is the bug):
https://github.com/pantsbuild/pex/blob/93b0442c9574a0bba018249d842dc2ac29dc1e27/pex/pex_bootstrapper.py#L83-L89

jsirois:
Previously, if the current interpreter was valid and contained in a
path directory entry, all sibling interpreters in that same directory
went undiscovered. A test is added that fails without the fix.

Fixes #1109

### DIFF
--- a/pex/pex_bootstrapper.py
+++ b/pex/pex_bootstrapper.py
@@ -41,6 +41,7 @@ def iter_compatible_interpreters(
     path=None,  # type: Optional[str]
     valid_basenames=None,  # type: Optional[Iterable[str]]
     interpreter_constraints=None,  # type: Optional[Iterable[str]]
+    preferred_interpreter=None,  # type: Optional[PythonInterpreter]
 ):
     # type: (...) -> Iterator[PythonInterpreter]
     """Find all compatible interpreters on the system within the supplied constraints.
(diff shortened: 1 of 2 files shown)


Eric-Arellano (approved):
Great. Thank you!

benjyw (commented):
LGTM !

PR #1113, opened 2020-11-17
Fix `--help-variables` docs.

### LINKED ISSUE #1059: Variables.strip_defaults leads to misleading types

jsirois:
The work to fix #1059 in #1061 introduced DefaultedProperty but failed
to account for these when generating PEX_ environment variable help.

### DIFF
--- a/pex/variables.py
+++ b/pex/variables.py
@@ -147,6 +147,7 @@ class Variables(object):
         for variable_name, value in sorted(cls.__dict__.items()):
             if not variable_name.startswith("PEX_"):
                 continue
+            value = value._func if isinstance(value, DefaultedProperty) else value
             variable_type, variable_text = cls.process_pydoc(getattr(value, "__doc__"))
             yield variable_name, variable_type, variable_text
 


Eric-Arellano (approved):

jsirois:
Fixes:
```
$ ./pex-2.1.21.pex --help-variables

PEX_ALWAYS_CACHE: Represents a property with a default value.

    To determine the value of the property without the default value
    applied, access it through the class and call the `strip_default`
    method, passing the instance in question.

PEX_COVERAGE: Represents a property with a default value.

    To determine the value of the property without the default value
    applied, access it through the class and call the `strip_default`
    method, passing the instance in question.

PEX_COVERAGE_FILENAME: Filename.

    Write the coverage data to the specified filename.  If
    PEX_COVERAGE_FILENAME is not specified but PEX_COVERAGE is,
    coverage information will be printed to stdout and not saved.
...
```

Issue #1108, opened 2020-11-09
Support PEX files as resolve repositories.


jsirois:
This would allow using `--pex-repository` instead of `--index` / `--find-links` to attempt to resolve requirements from an already built PEX. This feature would support a Pants use case in https://github.com/pantsbuild/pants/issues/11105 and the underlying implementation is also needed to support #1020 correctly. In that case, each interpreter found on a given host could be tried for compatibility with a PEX by attempting to resolve all its requirements from itself using that interpreter.

jsirois:
A few notes on the resolve logic implementation:
+ At buildtime to implement `--pex-repository`:
  1. Implementing custom logic to compare requirements to `.deps/` wheel tags to interpreters and platforms recursing on wheel `Requires-Dist` metadata requirements.
  2. Leveraging pip:
      This requires the "wheels" we embed in `.deps/` - which are not actually wheely but wheely installed in individual chroots - to be uninstalled wheels. This in turn requires the ability for the PEX `.bootstrap/` runtime to contain wheel installing code. Currently it does not since this code lives in pip and we do not vendor Pip into the runtime `.bootstrap/`.
+ At runtime to implement appropriate interpreter selection:
   Again, since the PEX `.bootstrap/` does not currently contain Pip, which is large, we're pushed towards custom logic as sketched above.

Taking Pip as being too large to vendor in the PEX runtime `.bootstrap/` as a given, we'll need to implement the custom resolve logic. With that in place we can support `--pex-repository`, but only when used as the exclusive source of dists. That covers the Pants use case.

If we did leverage Pip instead to do resolves / convert `.deps/` to contain wheels instead of installed wheel chroots, we could support a mix of `--pex-respository`, `--index` and `--find-links` at build-time since, in that case, PEX files would contain wheels which could be exposed to pip directlt via `--find-links`.

Eric-Arellano:
> Taking Pip as being too large to vendor in the PEX runtime .bootstrap/ as a given

The concern is bundle size, right? Vendoring pip at build time removed a lot of complexity from Pex, and I suspect that vendoring here would avoid adding lots of complexity/edge cases to runtime.

jsirois:
> The concern is bundle size, right?

Right. For example the Pex PEX itself would go from 2,683,442 bytes to 4,126,954 bytes.

> Vendoring pip at build time removed a lot of complexity from Pex,

That is definitely not the case:
```
[...]

PR #1114, opened 2020-11-17
Implement full featured requirment parsing.

jsirois:
This is a prerequisite for fixing ##899, implementing #1108 and the API
will also be useful to Pants which currently can only handle a subset
of requirements, forcing edits of existing requirment files to use
alternative requirement forms.

### DIFF
--- a/pex/network_configuration.py
+++ b/pex/network_configuration.py
@@ -47,6 +47,16 @@ class NetworkConfiguration(
         assert cache_ttl >= 0, "The cache_ttl parameter should be >= 0; given: {}".format(cache_ttl)
         assert retries >= 0, "The retries parameter should be >= 0; given: {}".format(retries)
         assert timeout >= 0, "The timeout parameter should be > 0; given: {}".format(timeout)
+        if headers:
+            bad_headers = [header for header in headers if ":" not in header]
+            assert not bad_headers, (
+                "The following headers were malformed:\n"
+                "{bad_headers}\n"
+                "All headers must be of the form NAME:VALUE.".format(
+                    bad_headers="\n".join(bad_headers),
+                )
+            )
+
         return cls(
             cache_ttl=cache_ttl,
             retries=retries,
--- a/pex/network_configuration.py
+++ b/pex/network_configuration.py
@@ -6,6 +6,11 @@ from __future__ import absolute_import
 
 from collections import namedtuple
 
+from pex.typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from typing import Dict, Iterable, Optional
+
 
 class NetworkConfiguration(
     namedtuple(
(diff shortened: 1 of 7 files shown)


jsirois:
Sorry about the diff size here.

pex/requirements.py
```diff
@@ -5,67 +5,586 @@
...
+            project_name=project_name,
+            url=archive_or_project_path,
+            marker=marker,
+            editable=editable,
+            is_local_project=is_local_project,
+        )
+
+    # Handle PEP-440. See: https://www.python.org/dev/peps/pep-0440.
+    #
+    # The `pkg_resources.Requirement.parse` method does all of this for us (via
```
jsirois, line 447: N.B.: Although Pex vendors the latest packaging which does handle direct references, It currently does not import correctly in tests, failing on `pex.thirdparty.packaging.six` for inscrutable reasons. I cut losses and used `pkg_resources.Requirement.parse` from our old vendored version of setuptools since the `_split_direct_references` is fairly simple.

pex/requirements.py
```diff
@@ -5,67 +5,586 @@
...
+        proxies = None  # type: Optional[Dict[str, str]]
+        if network_configuration.proxy:
+            proxies = {
+                protocol: network_configuration.proxy for protocol in ("ftp", "http", "https")
+            }
+
+        self._handlers = (ProxyHandler(proxies), HTTPSHandler(context=ssl_context))
+
+    @contextmanager
+    def get_body_iter(self, url):
```
jsirois, line 88: I tested all this out - except client certs - manually using `mitmproxy`. Since `mitmproxy` is implemented in python and does have an API it may be reasonable to circle back and add that into an intergation test.

pex/network_configuration.py
```diff
@@ -53,3 +56,7 @@ def create(
             cert=cert,
             client_cert=client_cert,
         )
+
+    def headers_as_dict(self):
+        # type: () -> Dict[str, str]
+        return dict(header.split(":", 1) for header in self.headers)
```
benjyw, line 62: AFAICT we don't validate the headers anywhere, and if one doesn't contain a `:` then this will fail in an unhelpful way. Whereas previously pip would have issued a helpful `Could not parse header` warning and continued. Might be worth validating the headers in `create()` so that users get an explicit, helpful message if their header string doesn't contain a `:`.

pex/requirements.py
benjyw: Shall we open an issue for this?

benjyw (approved):
That was some major work!

pex/requirements.py
```diff
@@ -5,67 +5,582 @@
...
+        if network_configuration.client_cert:
+            ssl_context.load_cert_chain(network_configuration.client_cert)
+
+        proxies = None  # type: Optional[Dict[str, str]]
+        if network_configuration.proxy:
+            proxies = {
+                protocol: network_configuration.proxy for protocol in ("ftp", "http", "https")
+            }
+
+        self._handlers = (ProxyHandler(proxies), HTTPSHandler(context=ssl_context))
```
Eric-Arellano, line 85: Consider being more explicit:

```python
self._proxy_handler = ProxyHandler(proxies)
self._https_handler = HTTPSHandler(context=ssl_context)
```

I know right now that the only callsite uses `*self._handlers`, but I think that would be simpler to understand if it was instead `build_opener(self._proxy_handler, self._https_handler)`.

tests/test_requirements.py
```diff
@@ -0,0 +1,343 @@
...
+                SomeProject
+                SomeProject == 1.3
+                SomeProject >=1.2,<2.0
+                SomeProject[foo, bar]
+                SomeProject~=1.4.2
+                
+                SomeProject ==5.4 ; python_version < '2.7'
+                SomeProject; sys_platform == 'win32'
+                
+                SomeProject @ file:///somewhere/over/here
```
Eric-Arellano, line 164: While you're at it, consider testing PEP 440 URLs, in addition to files. We use these in our Pants docs:

```
Django@ git+https://github.com/django/django.git
Django@ git+https://github.com/django/django.git@stable/2.1.x
Django@ git+https://github.com/django/django.git@fd209f62f1d83233cc634443cfac5ee4328d98b8
```

Eric-Arellano (approved):
As always, great testing. Thank you!

tests/test_requirements.py
jsirois: OK, well the existing test _does_ have a direct URL reference and one URL is as good as another here ... but added anyhow.

pex/requirements.py
jsirois: I don't agree with this. That change would leak constructor details to the implementation which does not care what handlers are used. So the change would cause `get_body_iter` to be more explicit (and repetitve) about things it shouldn't have known in the 1st place.

pex/network_configuration.py
jsirois: Yup, good point. Fixed.

jsirois:
> Shall we open an issue for this?

I've got the change ready with the test so I won't open an issue but I will send that change in a follow-up here ~presentlyish.

Eric-Arellano (approved):

PR #1116, opened 2020-11-28
Add URLFetcher IT.

jsirois:
Add a test of proxy handling with auth and custom ca certs.

Follows up on #1114 converting most manual testing into an IT.

### DIFF
--- a/pex/testing.py
+++ b/pex/testing.py
@@ -487,9 +487,7 @@ def ensure_python_venv(version, latest_pip=True):
     else:
         subprocess.check_call([pip, "install", "virtualenv==16.7.10"])
         subprocess.check_call([python, "-m", "virtualenv", venv])
-    python, pip = tuple(
-        os.path.realpath(os.path.join(venv, "bin", exe)) for exe in ("python", "pip")
-    )
+    python, pip = tuple(os.path.join(venv, "bin", exe) for exe in ("python", "pip"))
     if latest_pip:
         subprocess.check_call([pip, "install", "-U", "pip"])
     return python, pip
--- a/pex/testing.py
+++ b/pex/testing.py
@@ -414,20 +414,44 @@ PY27 = "2.7.15"
 PY35 = "3.5.6"
 PY36 = "3.6.6"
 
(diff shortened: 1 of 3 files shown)


pex/testing.py
```diff
@@ -414,20 +414,44 @@ def bootstrap_python_installer(dest):
...
 
 def ensure_python_distribution(version):
     # type: (str) -> Tuple[str, str, Callable[[Iterable[str]], Text]]
-    if version not in _VERSIONS:
-        raise ValueError("Please constrain version to one of {}".format(_VERSIONS))
+    if version not in _ALL_PY_VERSIONS:
+        raise ValueError("Please constrain version to one of {}".format(_ALL_PY_VERSIONS))
 
-    pyenv_root = os.path.join(os.getcwd(), ".pyenv_test")
+    pyenv_root = os.path.join(root_dir(), ".pyenv_test")
```
jsirois, line 454: Changing from `getcwd()` to `root_dir()` ensures we re-use `ROOT_DIR/.pyenv_test` even when in a test that has cwd'd to a tmpdir.

pex/testing.py
```diff
@@ -455,6 +479,22 @@ def run_pyenv(args):
...
+        subprocess.check_call([python, "-m", "venv", venv])
+    else:
+        subprocess.check_call([pip, "install", "virtualenv==16.7.10"])
+        subprocess.check_call([python, "-m", "virtualenv", venv])
+    python, pip = tuple(
+        os.path.realpath(os.path.join(venv, "bin", exe)) for exe in ("python", "pip")
+    )
+    if latest_pip:
+        subprocess.check_call([pip, "install", "-U", "pip"])
+    return python, pip
```
jsirois, line 495: Previously tests were / were encouraged to use the `pip` in a pyenv Python version. That would of course pollute that pyenv Python version with whatever was pip installed there (under `ROOT_DIR/.pyenv_test`). Now, each test method that needs to `pip` install gets a fresh venv.

benjyw (approved):

Issue #1097, opened 2020-11-06
Add option to extend PATH with additional console scripts


jneuff:
**Background**
I deploy apache airflow as a PEX file with `-c airflow`. Now, when I run `airflow.pex webserver` I get `FileNotFoundError: [Errno 2] No such file or directory: 'gunicorn': 'gunicorn'`. That's because the `airflow webserver` command calls `gunicorn` to start the webserver. As `gunicorn` is part of the PEX, I can call it with `PEX_SCRIPT=gunicorn airflow.pex`, but it is not in the PATH.

**Workaround**
My current solution is to have a script called `gunicorn`:
```
#!/bin/sh

PEX_SCRIPT=gunicorn airflow.pex $@
```
Adding this script to PATH before running `airflow webserver` works.

**Proposal**
It would come in handy to have an option for PEX to add additional console scripts contained in the PEX to the PATH. Creating the PEX file could look like this:
```
pex -r requirements.txt -c airflow --extend-path-with-scripts gunicorn -o airflow.pex
```
And of course a related runtime environment variable would make sense too.

What's your opinion on this? I'd be happy to contribute this feature!

jsirois:
This makes sense.

To be clear though, on the surface its not as trivial as it may sound. Since a PEX is a zipfile, you can't add items within it to the `PATH` directly. You need to extract those items and add extracted locations to the `PATH` (and, in some cases, to `sys.path`). It turns out PEX files already extract all contained distributions to a location under `~/.pex` at runtime (if not run and extracted previously) though, so its likely the case that most heavy lifting needed to do this is in place.

That said, there will be problems here. Pretend this is all implemented and consider a typical gunicorn script that will now be on the `PATH`:
```python
$ cat /home/jsirois/.pex/installed_wheels/5b9580f6c90af9b2d97488e3d17143cca0b6de2a/gunicorn-20.0.4-py2.py3-none-any.whl/bin/gunicorn 
#!/usr/bin/python3.8
# -*- coding: utf-8 -*-
import re
import sys
from gunicorn.app.wsgiapp import run
if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
    sys.exit(run())
```

[...]

Issue #962, opened 2020-05-04
Run PEXes as normal applications


jsirois:
The PEX runtime presents several differences from the typical python runtime and this can lead to problems PEXing various python programs:
1. The need to mark a PEX as not zip-safe for application code that uses filesystem APIs to find code and resources in the application.
2. The merging of the sys.path from individual pre-installed wheel chroots can expose bugs in underlying distributions that are normally masked by being installed in the same chroot as their enclosing dependency set.

Although 1 has a workaround with `--not-zip-safe` it's often surprising to users and not easy to discover as a problem solution. Item 2 doesn't have a solution and this leads to the inability to PEX certain applications that use namespace packages inconsistently in particular (see #331 for examples).

Recent work that added the `--unzip` option showed that 1st unzipping PEXes led to better cold and warm cache startup latency. If we were to push on this, one solution to both of the problems listed above would be for PEXes to always unzip / re-package themselves in the pex cache with standard site-packages layout (ie: in one `sys.path` chroot). This should even improve PEX performance more since the `sys.path` would have less entries allowing imports provided by the PEX to be found in a single fs search instead of needing to search 1 (the pex zip itself) + N (pex dependencies) locations.

cosmicexplorer:
Would it be feasible to consider allowing the building of PEX files unzipped, which aren't intended to be run on another host? The result of something like `PEXBuilder.build_unzipped()` could be just the unzipped chroot path copied into the pex cache, which an application depending on `pex` could use to know how to invoke the unzipped PEX file.

#### Rationale
The lazy-loading ipex functionality added to pants in pantsbuild/pants#8793 currently has to go through a round trip of (when the ipex is first executed) downloading wheels => stuffing them in a pex file, and if pex files were then unconditionally unzipped on top of that it would then involve a third crossing of that zipped/unzipped bridge. Stuffing 3rdparty dependencies into the resulting pex file appears to take more time than actually resolving dependencies, especially when the resolve is mostly cached, and the resulting pex will always be run on the same host because ipex will exec the built pex immediately after it is produced.

jsirois:
The Pex API has supported this for a long time and Pants has correspondingly used it for a long time:
https://github.com/pantsbuild/pex/blob/916e61e04634c60d09ee25859c76ba4f27282e51/pex/pex_builder.py#L462-L477

cosmicexplorer:
I had forgotten the difference between `.build()` and `.freeze()`! Thank you so much!!

[...]

PR #1117, opened 2020-12-03
Prototype: Expose scripts to path #1097 

jneuff:
This PR is just intended as a basis for discussion. It provides a prototype for the feature I suggested in #1097.

For this I added two command line arguments:
- `--expose-entry-points` to set entry points (e.g. `pip:main`) to be exposed on PATH.
- `--expose-scripts` to set scripts (e.g. `pip`) to be exposed on PATH.

Also, to be able to call the PEX entry point (path to PEX file or path to unzipped PEX) I needed to record the PEX file's shebang in PEX-INFO.

If any `exposed_entry_points` or `exposed_scripts` are defined, they will be rendered as scripts under `$PEX_ROOT/bin/<hash>/<name>`. Where `<hash>` ensures the directory is unique with respect to PEX shebang, PEX entry point and exposed entry points/scripts.

For a regular PEX exposing the script `pip` such a file might look like:
```bash
#!/bin/sh

PEX_SCRIPT=pip exec /usr/bin/env python3.8 /home/user/scratch/pex-1097/my.pex $@
```
Or for an unzipped PEX exposing the entry point `pip:main`:
```bash
#!/bin/sh

PEX_MODULE=pip:main exec /usr/bin/env python3.8 /home/user/.pex/unzipped_pexes/d7e4fb19eea99444d4ce8b85e5cffb4aca950bca $@
```

It turned out, the main problem was how to call the PEX. Because when running an unzipped PEX, the path to the original PEX file is lost. That is why I had to record the shebang in PEX-INFO, to being able to call an unzipped PEX the same way as running the PEX file.

The exposed entry points work with this prototype:
```
~/scratch/pex-1097 » ./my.pex                                                                                                                                                                                           d0301508@kaxpi00000122
Python 3.8.6 (default, Sep 25 2020, 00:00:00) 
[GCC 10.2.1 20200826 (Red Hat 10.2.1-3)] on linux
(InteractiveConsole)
>>> import subprocess
>>> subprocess.check_call("pip:main")
WARNING: pip is being invoked by an old script wrapper. This will fail in a future version of pip.
Please see https://github.com/pypa/pip/issues/5599 for advice on fixing the underlying issue.
To avoid this problem you can invoke Python with '-m pip' instead of running pip directly.

Usage:   
  my.pex <command> [options]
..
```

Whereas exposed scripts do not:
```
>>> subprocess.check_call("pip")
Python 3.8.6 (default, Sep 25 2020, 00:00:00) 
[GCC 10.2.1 20200826 (Red Hat 10.2.1-3)] on linux
(InteractiveConsole)
```
Calling `pip` dropped me into another Python interpreter. Probably, I broke resolving scripts. I'll dig into it.

Obviously, a lot is missing here, tests for instance. As I said this is just a prototype to serve as a basis for discussion. So, I'd be glad to get some feedback on this.

### DIFF
--- a/pex/expose_scripts.py
+++ b/pex/expose_scripts.py
@@ -1,0 +1,70 @@
+import hashlib
+import itertools
+import os
+
+from pathlib import Path
+
+
+SCRIPT_TEMPLATE = """#!/bin/sh
+
+{pex_var}={entry} exec {executable} {pex_path} $@
+"""
+
+
+def get_hash(shebang, pex_path, entry_points, scripts):
+    entry_points = sorted(entry_points)
+    scripts = sorted(scripts)
+
+    hasher = hashlib.md5()
+    hasher.update(shebang.encode("utf-8"))
+    hasher.update(pex_path.encode("utf-8"))
+    for item in itertools.chain(entry_points, scripts):
+        hasher.update(item.encode("utf-8"))
+
+    return hasher.hexdigest()
+
+
+def render_wrapper_scripts(shebang, pex_path, bin_path, entry_points, scripts):
+    executable = shebang[2:]
+    bin_path.mkdir(mode=0o750, parents=True, exist_ok=True)
+    # TODO: what should be the script name for entry_points?
+    for item in entry_points:
+        item_path = bin_path / item
+        contents = SCRIPT_TEMPLATE.format(
+            pex_var="PEX_MODULE", entry=item, executable=executable, pex_path=pex_path
+        )
+        with item_path.open("w") as f:
+            f.write(contents)
+        item_path.chmod(0o770)
+
+    for item in scripts:
+        item_path = bin_path / item
+        contents = SCRIPT_TEMPLATE.format(
+            pex_var="PEX_SCRIPT", entry=item, executable=executable, pex_path=pex_path
+        )
+        with item_path.open("w") as f:
+            f.write(contents)
+        item_path.chmod(0o770)
+
+
+def expose_scripts(pex_path, pex_info):
+    shebang = pex_info.shebang
+    exposed_entry_points = pex_info.exposed_entry_points
+    exposed_scripts = pex_info.exposed_scripts
+
+    if len(exposed_entry_points) == 0 and len(exposed_scripts) == 0:
+        return
+
+    pex_path = Path(pex_path).resolve()
+    pex_path_str = str(pex_path)
+    bin_hash = get_hash(shebang, pex_path_str, exposed_entry_points, exposed_scripts)
+    bin_path = Path(pex_info.pex_root) / "bin" / bin_hash
+    render_wrapper_scripts(
+        shebang,
+        pex_path_str,
+        bin_path,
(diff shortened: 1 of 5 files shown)


pex/expose_scripts.py
```diff
@@ -0,0 +1,70 @@
+import hashlib
+import itertools
+import os
+
+from pathlib import Path
+
+
+SCRIPT_TEMPLATE = """#!/bin/sh
```
jsirois, line 8: This is not ideal. For example, it would fail in any Python [distroless](https://github.com/GoogleContainerTools/distroless) image whereas the corresponding python console script it wraps would not.

pex/expose_scripts.py
```diff
@@ -0,0 +1,70 @@
+import hashlib
+import itertools
+import os
+
+from pathlib import Path
```
jsirois, line 5: Pex still supports Python 2.7 so you need to astick to those APIs - this will fail the 2.7 CI shards as-is.

pex/expose_scripts.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def get_hash(shebang, pex_path, entry_points, scripts):
+    entry_points = sorted(entry_points)
+    scripts = sorted(scripts)
+
+    hasher = hashlib.md5()
+    hasher.update(shebang.encode("utf-8"))
+    hasher.update(pex_path.encode("utf-8"))
+    for item in itertools.chain(entry_points, scripts):
```
jsirois, line 21: This hashing scheme is foiled by setting an entry_point for the module 'pex' today and then changing that to be the 'pex' console script tomorrow. You might form a dict using the key names corresponding to the 4 slots and then hash its sorted key json encoding.

jsirois (commented):
Thanks for taking a stab at this! IIUC you don't need the `--exposed-entry-points` functionality, just the `--exposed-scripts`. If so, I have a change coming this weekend to support #962 that creates a venv on from a PEX at runtime. To do this, all the .whl bin/ scripts that are python scripts (console scripts), are re-written to have a shebang that is the selected runtime interpreter. You should probably wait on tht to land and see if it doesn't solve your use case. I'll include you on the review.

jsirois:
@jneuff if you have time to look at #1128 and in particular this test working backwards to the `--bin-path` feature it tests, I'd be grateful: https://github.com/pantsbuild/pex/pull/1128/files#diff-f6529bfd4644104d30930f6bf2eeafdf75f87ae12659a9185ac45c3fcdb2b8f9R180

jneuff:
@jsirois Thanks for the feedback. Your venv feature sounds quite interesting! That should probably cover my use-cases. I'll take a look at your PR.

jneuff:
I am closing this PR as the intended use-cases are well covered by the new venv pex-tool feature.

jsirois:
Thanks for taking a stab at this @jneuff - it was definitely part of the motivation to finally implement #962 and solve #1097 along with other common issues running applications as PEX zip files.

PR #1121, opened 2020-12-04
Fix req parsing for local archives & projects.

jsirois:
Previously, we did not handle the case where a local project directory
basename was not a well formed requirement project name. Pro-actively
parse the local project directory basename to separate out any
requirements parts for further parsing and add regression tests for
this.

### DIFF
--- a/pex/requirements.py
+++ b/pex/requirements.py
@@ -377,15 +377,48 @@ def _try_parse_project_name_from_path(path):
     return None
 
 
-def _try_parse_pip_local_project_marker(path):
-    # type: (str) -> Tuple[str, Optional[Marker]]
+def _try_parse_pip_local_formats(
+    path,  # type: str
+    basepath=None,  # type: Optional[str]
+):
+    # type: (...) -> Tuple[Optional[str], Optional[Marker]]
     project_requirement = os.path.basename(path)
+    match = re.match(
+        r"""
+        ^
+        (?P<directory_name>[^!=<>\[;]+)
+        (?P<requirement_parts>[!=<>\[;].+)?
+        $
+        """,
+        project_requirement,
+        re.VERBOSE,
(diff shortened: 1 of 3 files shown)


jsirois:
This came to light in master CI where a tmpdir happened to be generated with a `_` at the end of its name: https://travis-ci.com/github/pantsbuild/pex/jobs/454555160#L1888

Looked like:
```
=================================== FAILURES ===================================
_________________________ test_resolve_extra_setup_py __________________________
...
E           ParseError: <string> line 1:
E           /tmp/tmpW8tdb_[foo]
E           Problem parsing '/tmp/tmpW8tdb_[foo]' as a requirement: Invalid requirement, parse error at "'/tmp/tmp'"
```

jsirois:
Alrighty - this is now green and good to go for review.

pex/requirements.py
```diff
@@ -377,15 +377,64 @@ def _try_parse_project_name_from_path(path):
     return None
 
 
-def _try_parse_pip_local_project_marker(path):
-    # type: (str) -> Tuple[str, Optional[Marker]]
+def _try_parse_pip_local_formats(
+    path,  # type: str
+    basepath=None,  # type: Optional[str]
+):
+    # type: (...) -> Tuple[Optional[str], Optional[Marker]]
```
stuhood, line 384: Worth a dataclass return type?

stuhood (approved):

pex/requirements.py
jsirois: Pex still needs to be Python 2.7 compatible - that's why you see type comments and not type annotations.

jsirois: ... and since this is a private datatype, ginning up a namedtuple or similar does not seem worth it.

PR #1122, opened 2020-12-04
Add a `requires_dists` function.

jsirois:
This is needed to support #1020 and #1108.

### DIFF
--- a/pex/dist_metadata.py
+++ b/pex/dist_metadata.py
@@ -7,11 +7,27 @@ from __future__ import absolute_import
 import email
 
 from pex.third_party.packaging.specifiers import SpecifierSet
-from pex.third_party.pkg_resources import DistInfoDistribution, Distribution
+from pex.third_party.pkg_resources import DistInfoDistribution, Distribution, Requirement
 from pex.typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
-    from typing import Optional
+    from typing import Dict, Iterator, Optional
+
+
+_PKG_INFO_BY_DIST = {}  # type: Dict[Distribution, Optional[email.message.Message]]
+
+
+def _parse_pkg_info(dist):
+    # type: (Distribution) -> Optional[email.message.Message]
+    global _PKG_INFO_BY_DIST
+    if dist not in _PKG_INFO_BY_DIST:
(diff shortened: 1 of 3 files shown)


pex/dist_metadata.py
```diff
@@ -23,12 +39,40 @@ def requires_python(dist):
...
+    See:
+    + https://www.python.org/dev/peps/pep-0345/#requires-dist-multiple-use
+    + https://www.python.org/dev/peps/pep-0314/#requires-multiple-use
+
+    :param dist: A distribution to check for requirement metadata.
+    :return: All requirements found.
+    """
+    pkg_info = _parse_pkg_info(dist)
+    if pkg_info is None:
+        return
```
stuhood, line 71: Unsure of the error handling strategy in here, but is a dist not having PKG-INFO something that should be silently ignored like this?

pex/dist_metadata.py
```diff
@@ -7,11 +7,27 @@
...
-    from typing import Optional
+    from typing import Dict, Iterator, Optional
+
+
+_PKG_INFO_BY_DIST = {}  # type: Dict[Distribution, Optional[email.message.Message]]
+
+
+def _parse_pkg_info(dist):
+    # type: (Distribution) -> Optional[email.message.Message]
+    global _PKG_INFO_BY_DIST
```
stuhood, line 22: Because you don't assign to `_PKG_INFO_BY_DIST` (only internally mutate it), the global keyword shouldn't be necessary here?

I guess that in this case this is an alternative to the `memoized` decorator?

stuhood (approved):

pex/dist_metadata.py
jsirois: The `memoized` is a Pants thing, not a Pex thing, but you're correct - will fix.

pex/dist_metadata.py
jsirois: I think so, yes. If the distribution has no metadata, that means it expresses no opinon on a Python interpreter requirement, which is all we care about (it can have metadata and not an express an opinion on this field as tests exercise). It is true that distributions should always have metadata and that if they do not, a minimal file should be produced by Pip when it installs the distribution, but there is a range of interesting non-compliant package material out on PyPI and banking on Pip to fill in minimal metadata for those seems like a bet we don't need to make.

PR #1125, opened 2020-12-04
Fix `PythonInterpreter` binary normalization.

jsirois:
Previously this always used `os.path.realpath` which follows symlinks
and breaks out of venvs created by Python 3. Although this is fine for
PEX use since it scrubs the `sys.path` anyhow, it will not be fine for
upcoming use when creating venvs from PEX files.

### DIFF
--- a/pex/interpreter.py
+++ b/pex/interpreter.py
@@ -308,8 +308,103 @@ class PythonInterpreter(object):
     _PYTHON_INTERPRETER_BY_NORMALIZED_PATH = {}  # type: Dict
 
     @staticmethod
-    def _normalize_path(path):
-        return os.path.realpath(path)
+    def _read_pyvenv_home(path):
+        # type: (str) -> Optional[str]
+        # See: https://www.python.org/dev/peps/pep-0405/#specification
+        pyvenv_cfg_path = os.path.join(path, "pyvenv.cfg")
+        if os.path.isfile(pyvenv_cfg_path):
+            with open(pyvenv_cfg_path) as fp:
+                for line in fp:
+                    name, _, value = line.partition("=")
+                    if name.strip() == "home":
+                        return value.strip()
+        return None
+
+    @classmethod
+    def _find_pyvenv_home(cls, maybe_venv_python_binary):
+        # type: (str) -> Optional[str]
+        # A pyvenv is identified by a pyvenv.cfg file with a home key in one of the two following
+        # directory layouts:
+        #
+        # 1. <venv dir>/
+        #      bin/
+        #        pyvenv.cfg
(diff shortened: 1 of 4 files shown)


pex/interpreter.py
```diff
@@ -308,8 +308,20 @@ class PythonInterpreter(object):
     _PYTHON_INTERPRETER_BY_NORMALIZED_PATH = {}  # type: Dict
 
     @staticmethod
-    def _normalize_path(path):
-        return os.path.realpath(path)
+    def canonicalize_path(path):
```
stuhood, line 311: The stdlib (rightly I think) refers to what `os.path.realpath` does as "canonicalizing", so it might be clearer to name this method something else... maybe just `abspath` or `absolute_path`...?

stuhood (commented):
At a functional level: should this be following all symlinks _except_ the venv symlink? Or would it be relatively harmless to have multiple copies of the same interpreter referred to from multiple places? Would `python3.7`+ `python3` symlinks to the same interpreter trigger that?

pex/interpreter.py
jsirois: Well, the point of the name as the docs clarify is that this is canonical for PythonInterpreter irrespective of what other libs may think. The intent is opaque canonicalization and mentioning abs defeats that. I could switch back to normalize language, but that also runs up against os.path.normpath expectations so would seem no more acceptable given your logic. Can you think of a word you like not used by the stdlib that keeps this opaque as to impl but clear as to intent?

jsirois:
> At a functional level: should this be following all symlinks except the venv symlink? Or would it be relatively harmless to have multiple copies of the same interpreter referred to from multiple places? Would python3.7+ python3 symlinks to the same interpreter trigger that?

Its relatively harmless. If you had N symlinks to an interpreter you'd get back N + 1PythonInterpreter objects when you iterated over a given PATH containing  the N symlinks and the original. In the cold-cache case this means N extra (parallelized) subprocesses - so there is wasted work there. In the warm cache case this has ~no perf effect.

That said, thanks for the nudge. I think it might be reasonable to implement following all symlinks except https://www.python.org/dev/peps/pep-0405/ venv symlinks to the out-of-venv base interpreter by looking for a sibling `pyvenv.cfg` or one up one level.

jsirois:
Updated to resolve all symlinks saving for canonical Python 3 venv binary symlinks to the home binary. Please forgive the force-push. The change back to `os.realpath` in the non-venv case eliminated the test thrash which otherwise showed as zig-zag between the 1st and second commits.

stuhood (approved):
Thanks!

Issue #1115, opened 2020-11-25
Removing "activation" time from PEX


stuhood:
As shown in #930, in some cases the activation of a PEX takes 50% of the total runtime, primarily on the task of "resolving" the dependencies to use within that run. This resolution is necessary for cross-platform and portable PEX files which might contain multiple copies of certain wheels.

Two potential ways to eliminate this time might be:
1. Adding explicit support for building non-portable PEX files. Resolution is not necessary if the PEX is only meant to "work on my machine" (frequently the case in Pants). When compared to a virtualenv, a PEX would still have the advantage of being relocatable on the filesystem.
2. Preserving the resolutions computed at PEX construction time (in the `PEX-INFO`, perhaps) for use verbatim at runtime. This would involve doing a fuzzy dictionary lookup of the target platforms for the PEX to determine which set of contained wheels to use.

In both cases, one challenge would be giving good error messages if a PEX had been moved to an incompatible platform. If the PEX explicitly embedded the platform(s) it was built to target, it could quickly fail if was invoked on an incompatible platform (without consulting its list of wheels).

jsirois:
> PEX takes 50% of the total runtime, primarily on the task of "resolving" the dependencies to use within that run. 

How confident are you of your interpretation of the speedscope? FWICT > 50% of "resolving" is importing which was the initial focus of #930. Importing we cannot get around multiplatform PEX or not except by eliminating expensive imports from the bootstrap.

stuhood:
Somewhat confident. Were you able to get it to load?

The `activate` phase of the profile does not actually involve any importing, afaict: the majority of the time under `activate` is spent in `resolve->...->_compute_dependencies->...->parse_requirement->...`. You can _also_ see a large amount of time in `_find_and_load`, but that is independent of the time spent in `activate`.

jsirois:
This is what I'm looking at:

![image](https://user-images.githubusercontent.com/157586/100168638-7c381180-2e76-11eb-8a4f-674d4a17a821.png)

stuhood:
@jsirois : That's not the callstack: i.e. `activate` doesn't call `_find_and_load`.

In this screenshot, the left 25% of the script is under `_find_and_load`, the middle 50% is under `activate`, and the right 25% is under `pytest` itself.
[...]

PR #1128, opened 2020-12-08
Add a venv tool.

jsirois:
Add a `venv` tool to create a virtual environment from a PEX file. The
virtual environment is seeded with just the PEX user code and
distributions applicable to the selected interpreter for the local
machine. The virtual environment does not have Pip installed by default
although that can be requested with `--pip`.

The virtual environment comes with a `__main__.py` at the root of the 
venv to emulate a loose pex that can be run with `python venv.dir` just
like a loose pex. This entry point supports all the behavior of the
original PEX file not related to interpreter selection, namely support
for PEX_SCRIPT, PEX_MODULE, PEX_INTERPRETER and PEX_EXTRA_SYS_PATH.

A sibling `pex` script is linked to `__main__.py` to provide the
maximum performance entrypoint that always avoids interpreter
re-execing and thus yields equivalent performance to a pure virtual
environment.

Work towards #962 and #1115.

### DIFF
--- a/pex/tools/commands/venv.py
+++ b/pex/tools/commands/venv.py
@@ -1,0 +1,236 @@
+# Copyright 2020 Pants project contributors (see CONTRIBUTORS.md).
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+from __future__ import absolute_import
+
+import errno
+import os
+import shutil
+import zipfile
+from argparse import ArgumentParser, Namespace
+from textwrap import dedent
+
+from pex import pex_builder, pex_warnings
+from pex.common import chmod_plus_x, safe_mkdir
+from pex.environment import PEXEnvironment
+from pex.pex import PEX
+from pex.tools.command import Command, Error, Ok, Result
+from pex.tools.commands.virtualenv import PipUnavailableError, Virtualenv
+from pex.tracer import TRACER
+from pex.typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from typing import Tuple
+
+
+def _copytree(
+    src,  # type: str
+    dst,  # type: str
+    exclude=(),  # type: Tuple[str, ...]
+    collisions_ok=False,  # type: bool
+):
+    # type: (...) -> None
+    safe_mkdir(dst)
+    link = True
+    for root, dirs, files in os.walk(src, topdown=True, followlinks=False):
+        if src == root:
+            dirs[:] = [d for d in dirs if d not in exclude]
+            files[:] = [f for f in files if f not in exclude]
+
+        for d in dirs:
+            try:
+                os.mkdir(os.path.join(dst, os.path.relpath(os.path.join(root, d), src)))
+            except OSError as e:
+                if e.errno != errno.EEXIST:
+                    raise e
+
+        for f in files:
+            src_entry = os.path.join(root, f)
+            dst_entry = os.path.join(dst, os.path.relpath(src_entry, src))
+            try:
+                if link:
+                    try:
+                        os.link(src_entry, dst_entry)
+                        continue
+                    except OSError as e:
+                        if e.errno != errno.EXDEV:
+                            raise e
+                        link = False
+                shutil.copy(src_entry, dst_entry)
+            except OSError as e:
+                if e.errno == errno.EEXIST:
+                    pex_warnings.warn(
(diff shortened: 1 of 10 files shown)


jsirois:
Commits are reviewable in sequence individually if that suits.

jsirois:
This is hefty, but also not as big as it looks. Of the +3,371 −49 delta +2,633 comes from embedding the virtualenv.py script for Python 2.7 support.

Thanks in advance for your patience on the remaining diff though.

jsirois:
Going into draft mode whil I work out CI failures that do not repro on my machine.

jsirois:
OK - this is now ready for review. Its based on top of #1131 which form the 1st two commits here. Please focus on the 7 commits starting at 06eff63 "Automation for embedding of virtualenv 16.7.10.".

Thanks again in advance.

pex/tools/commands/venv.py
```diff
@@ -0,0 +1,236 @@
...
+if TYPE_CHECKING:
+    from typing import Tuple
+
+
+def _copytree(
+    src,  # type: str
+    dst,  # type: str
+    exclude=(),  # type: Tuple[str, ...]
+    collisions_ok=False,  # type: bool
+):
```
stuhood, line 31: A comment explaining "why not `shutil.copytree`" might be good.

pex/tools/commands/venv.py
```diff
@@ -0,0 +1,236 @@
...
+        )
+
+    def run(
+        self,
+        pex,  # type: PEX
+        options,  # type: Namespace
+    ):
+        # type: (...) -> Result
+
+        venv = Virtualenv.create(options.venv[0], interpreter=pex.interpreter, force=options.force)
```
stuhood, line 118: Since this actually does create the venv on disk as a sideeffect, it might be worth mentioning as step 0 in here.

pex/tools/commands/venv.py
```diff
@@ -0,0 +1,236 @@
...
+                pex.path(), pex_info, venv.site_packages_dir, exclude=("__main__.py",)
+            )
+        else:
+            _copytree(
+                src=pex.path(),
+                dst=venv.site_packages_dir,
+                exclude=(pex_info.internal_cache, pex_builder.BOOTSTRAP_DIR, "__main__.py"),
+            )
+
+        for dist in pex.activate():
```
stuhood, line 133: I had thought of "activate" as meaning "add to the pythonpath of the current process", so it's interesting to see it used like this. Would splitting that method into "iterate over what would have been activated" vs "activate" make sense?

pex/tools/commands/venv.py
```diff
@@ -0,0 +1,236 @@
...
+                    )
+                )
+                sys.exit(1)
+
+            pex_script = pex_overrides.get("PEX_SCRIPT")
+            if pex_script:
+                script_path = os.path.join(bin_dir, pex_script)
+                os.execv(script_path, [script_path] + sys.argv[1:])
+
+            # TODO(John Sirois): Support `-c`, `-m` and `-` special modes when PEX_INTERPRETER is
```
stuhood, line 188: TODOs with github issues are a bit easier to maintain and add information to.

stuhood (approved):
Thanks!

pex/tools/commands/venv.py
jsirois: That might make sense although it is neither here nor there for tool use where the side-effect of the distributions getting added to the sys.path here has no relevant effect:
https://github.com/pantsbuild/pex/blob/12ad6a63f988c07065cebbfb6aa4102a6d3fa094/pex/environment.py#L476

pex/tools/commands/venv.py
jsirois: Agreed - done.

pex/tools/commands/venv.py
jsirois: I'm not sure about "side-effect" except in the technical sense that all IO is that way - the method is "create" indicating creation is the main point. That said - done.

pex/tools/commands/venv.py
jsirois: Yup, added.

pex/tools/commands/venv.py
jsirois: This is a good deal thornier than it may look. The code paths involved are relevant in the resolve-from-pex work though so I'll defer to that work.

Issue #1111, opened 2020-11-10
Release 2.1.22


jsirois:
On the docket:
+ [x] Fix `--help-variables` docs. #1113
+ [x] pex binary hangs on startup at atomic_directory #1119
+ [x] pex vendoring does not (always) isolate itself from a colliding setuptools install in site-packages #1031
+ [x] Remove long deprecated support for _pex module. #1135

jsirois:
Truncating this docket over to #1138 to get out this release and 2.1.23 in #1134 with the Pip upgrade to unblock macOS Big Sur users.

Issue #1086, opened 2020-10-20
Support dependency lock files to save the results of package resolution


jriddy:
This is feature request to support _lock files_ to save and re-use the results of a package resolution. 

Having to resolve packages each time pex is executed is time consuming and difficult, especially if you end up in workflows that call pex a lot (like in pants). And this is kinda unfortunate, because the results of resolutions can be fairly easily saved, which could in turn save users a lot of time.

### What are lockfiles?

Lockfiles save important data about the exact dependency packages needed to reproduce an environment deterministically. This is more than just a pinned version. This is an exact match, with checksums and source information. Many, many packaging solutions use these (cargo, npm, [pipfile](https://github.com/pypa/pipfile), poetry, gradle, gem) to gain true determinism in builds.

It's easy to see why pex has never done this because it could be argued that the pex format is itself a kind of lockfile, and you can re-use its resolutions as with `--requirements-pex`. Pants seems to use pex like this, but this is kind of inefficient, and results in fairly coarse-grained caching effects.

If instead we saved the resolution as a json/toml/yaml lockfile of some kind, caching could be based purely on questions like "does what i have in my cache for this name match the checksum in the lockfile." 

### What about constraints files?

These are not lockfiles, and were never intended to be. The only difference between them and requirements files is that they don't force the installation of what's inside them, but [only constrain the versions](https://pip.pypa.io/en/stable/user_guide/#constraints-files) of packages _if they are installed_.

In contrast, a lockfile demands that its _exact set of sources_ get installed, and doesn't allow for any deviation or decision making beyond that. That the [pants documenation](https://www.pantsbuild.org/docs/python-third-party-dependencies#using-a-lockfile-strongly-recommended) calls `constraints.txt` a lockfile is a bit misleading, especially in light of how pex will attempt to perfom resolutions with it. A lockfile is the result of a resolution, not an input to one.

### What should lockfiles contain?

At the base, there should be metadata about the platform and implementation it was generated on, and then a list of packages with their checksums and urls they were downloaded from. 

### How would we use this?

I could see a workflow like this...

```
pex --generate-lockfile pex.lock -r requirements.txt --no-output-file  # Generate the lockfile and exit
pex -L pex.lock -r requirements.txt -O  # same thing with short options

pex --from-lockfile pex.lock --output-file app.pex  # create the pex using the exact dependencies as specified in the lock file
pex -l pex.lock -o app.pex  # same thing with short options
```

[...]

Issue #1084, opened 2020-10-20
Possible bug with cache TTL expiration resulting in redownloading files


jriddy:
Using pex version 2.1.19

The current default cache-ttl of 1 hr (3600 s) is far too short for `pants` use cases where you end up changing dependencies a lot over a short period of time. This results in frequent unnecessary re-downloading and rebuilding of library wheels/sdists when you're changing dependencies frequently (something that can happen very easily and almost inadvertently with pants' dependency inference). I spent a day or two trying to figure out why it was taking 5 min to build constraints files all the time

This issue is compounded by pants not having a way to control this flag (gonna make an issue over there for that), nor is there any documentation on how to set this value to never expire that I can find.

That said, I don't fully understand why TTL caching is even at play here. When you download python packages, you have hashes in the filenames that you can check against what you have in your cache. If the hashes match...there doesn't need to be a download regardless of TTL or cache expiry issues. Perhaps there are other issues at play when it comes to determining _which_ file to download, but once that determination is made, it seems like we could safely let checksums rule the day.

I can provide logs of this and the trouble that it causes if needed.

jsirois:
It probably makes sense to decouple Pants from the PEX defaults here with Pants work to plumb these options.

Moving to your hashes / why is TTL in play: unless a requirement is exact, once a TTL expires, the ranged requirement sort of has to be looked up again. I'd be surprised if the new PIp resolver does anything different here compared to the old Pip resolver (which Pex currently uses).

jriddy:
Thanks for the quick feedback

> why is TTL in play

I can see why it is in play for to check for new requirements, but if that check results in a found link to a file we've already downloaded, why does it have to download it again? But come to think of it, I think pip shares this behavior, with perhaps a more reasonable default.

> the ranged requirement sort of has to be looked up again
I'm using pinned requirements in a constraints file. I don't really want a resolution, I want the dependencies i've specified and already resolved previously (although I know pex and internally pip still have to locate files with matching tags and find wheels/sdists etc).

I think what we're really missing here for these use cases is a _lock file_.

jsirois:
> I think what we're really missing here for these use cases is a lock file.

Pants has that: https://www.pantsbuild.org/docs/python-third-party-dependencies#using-a-lockfile-strongly-recommended

jsirois:
... as does pex which Pants passes through to using `pex --constraints=...`

jriddy:
[...]

PR #1133, opened 2020-12-12
Upgrade Pex to Pip 20.3.1.

jsirois:
Since Pip now natively supports passing multiple --platform options on
the download CLI, the only adjustments needed were to remove the
--headers and --cache-ttl support that was provided by a custom patch
that never made it into Pip due to security concerns as well as an
adjustment to the wheel install code due to the 2020-resolver being more
strict.

A new --resolver-version option is added to toggle between the 
legacy-resolver and the new 2020-resolver with the legacy-resolver being
used by default to allow Pex users to transition. A follow-up release 
will add a deprecation notice / warning for those still using the 
legacy-resolver.


jsirois:
This is reviewable but will not be merged until after the 2.1.22 release (#1111) to go out in its own dedicated 2.1.23 release (#1134).

pex/bin/pex.py
```diff
@@ -826,11 +836,14 @@ def to_python_interpreter(full_path_or_basename):
         pex_builder.add_from_requirements_pex(requirements_pex)
 
     with TRACER.timed("Resolving distributions ({})".format(reqs + options.requirement_files)):
+        if options.cache_ttl:
+            pex_warnings.warn("The --cache-ttl option is deprecated and no longer has any effect.")
```
stuhood, line 840: Worth referencing on https://github.com/pantsbuild/pex/issues/1084 I suspect.

Rather than fine-tuning TTLs moving forward, I expect that true lockfile support (#1086 - which would include the hashes/URLs to fetch, rather than actually running a resolve) would be the best optimization.

stuhood (approved):

pex/bin/pex.py
jsirois: The TTLs were actually an anti-optimization in practice (except for Twitter's hack to reduce load on find-links repos). Any positive TTL causes Pip to ignore its cache once the TTL expires whereas a 0 TTL causes it to use the cache with a conditional GET as long as the resource stays the same. So this appears to be a strict win except for the Twitter case. I've updated #1084 with some comments about this upgrade. And agreed, #1086 will be the solution Pants was looking for although fwict it will involve working outside Pip. Command line tests show that Pip still hits the network with --no-deps and --hash-checking enabled. You get a fixed resolve, but its slower than need be.

PR #1143, opened 2020-12-15
Upgrade to a patched Pip 20.3.3.

jsirois:
This brings in a fix for `pip download --constraint` as well as support
for Python 3.10.

See: https://github.com/pypa/pip/issues/9283


stuhood (approved):
Thanks!

PR #1149, opened 2020-12-17
Implement PEX_INTERPRETER special mode support.

### LINKED ISSUE #1136: Add support for PEX_INTERPRETER `-c`, `-m` and `-` special modes to PEX venvs.
#1128 lands a Venv tool that creates a venv from a PEX file with an entrypoint that fully emulates PEX behavior except for handling `-c`, `-m` and `-` special modes when `PEX_INTERPRETER` is activated. Implement these modes so that a PEX venv can be counted on to work exactly like a PEX file.

jsirois:
This brings the venv pex script up to parity with PEX files.

Fixes #1136

### DIFF
--- a/pex/tools/commands/venv.py
+++ b/pex/tools/commands/venv.py
@@ -189,14 +189,47 @@ class Venv(Command):
                 script_path = os.path.join(bin_dir, pex_script)
                 os.execv(script_path, [script_path] + sys.argv[1:])
 
-            # TODO(John Sirois): Support `-c`, `-m` and `-` special modes when PEX_INTERPRETER is
-            # activated like PEX files do: https://github.com/pantsbuild/pex/issues/1136
-            pex_interpreter = pex_overrides.get("PEX_INTERPRETER", "").lower()
+            pex_interpreter = pex_overrides.get("PEX_INTERPRETER", "").lower() in ("1", "true")
+            PEX_INTERPRETER_ENTRYPOINT = "code:interact"
             entry_point = (
-                "code:interact"
-                if pex_interpreter in ("1", "true")
-                else pex_overrides.get("PEX_MODULE", {entry_point!r} or "code:interact")
+                PEX_INTERPRETER_ENTRYPOINT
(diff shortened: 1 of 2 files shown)


pex/tools/commands/venv.py
```diff
@@ -189,14 +189,47 @@ def run(
...
+            pex_interpreter = pex_overrides.get("PEX_INTERPRETER", "").lower() in ("1", "true")
+            PEX_INTERPRETER_ENTRYPOINT = "code:interact"
             entry_point = (
-                "code:interact"
-                if pex_interpreter in ("1", "true")
-                else pex_overrides.get("PEX_MODULE", {entry_point!r} or "code:interact")
+                PEX_INTERPRETER_ENTRYPOINT
+                if pex_interpreter
+                else pex_overrides.get("PEX_MODULE", {entry_point!r} or PEX_INTERPRETER_ENTRYPOINT)
             )
```
stuhood, line 198: Are we trending toward this codepath replacing the existing codepath, or are we expecting them to live in parallel for the medium future? Not terribly complicated, but if it's going to continue to grow, some sort of de-duplication might be in order.

stuhood (approved):

pex/tools/commands/venv.py
jsirois: I expect these to live in parallel for the foreseeable future. That aside, fhe factored out code could never use anything but the stdlib which we'd need infra to check / enforce and it might have to get tortured slightly to deal with PEX needing to demote its bootstrap at the right point in time in the different branches. Not sure about the last bit - may be a non-problem. May be worth doing, but the test coverage here seems sufficient to punt doing that.

stuhood: Agreed: great, coverage.

jsirois: Ah yeah - the grow question - no growth. This should be the last for PEX_INTERPRETER handling until some new Python release adds new interpreter options that make sense for PEX-as-an-interpreter to support.

PR #1152, opened 2020-12-17
Remove redundant dep edge label info.

jsirois:


jsirois:
Slightly better.

Before:
old deps

After:
new deps

Eric-Arellano (approved):
This is awesome. Cool thing to come back to from my time off :)

pex/tools/commands/graph.py
```diff
@@ -68,7 +68,12 @@ def _create_dependency_graph(pex):
                 graph.add_edge(
                     start=dist.project_name,
                     end=req.project_name,
-                    label=str(req) if (req.specifier or req.marker) else None,
+                    label="{specifier}{marker}".format(
+                        specifier=req.specifier if req.specifier else "",
+                        marker="; {}".format(req.marker) if req.marker else "",
+                    )
+                    if (req.specifier or req.marker)
+                    else None,
```
Eric-Arellano, line 76: Formatting nit that I'm more saying as an FYI: if you wrap this whole thing in parentheses, Black will format this block with proper indentation.

jsirois: I think we've had this chat before - I consider whatever an auto-formatter does as proper - that's the very deal you make when you sign up for auto formatting - you let go of formatting.

Eric-Arellano: Yeah, you're right, we have. I view the extra parentheses as giving a small hint to Black so that it can do its thing. But that's fine - not blocking of course.

PR #1153, opened 2020-12-21
Support a --venv mode similar to --unzip mode.

### LINKED ISSUE #962: Run PEXes as normal applications

### LINKED ISSUE #1097: Add option to extend PATH with additional console scripts

### LINKED ISSUE #1115: Removing "activation" time from PEX

jsirois:
The new --venv execution mode builds a PEX file that includes pex.tools
and extracts itself into a venv under PEX_ROOT upon 1st execution or any
execution that might select a diffrent interpreter than the default.

In order to speed up the local build and execute case, --seed mode is
added to seed the PEX_ROOT caches that will be used at runtime. This is
important for --venv mode since venv seeding depends on the selected
interpreter and one is already selected during the PEX file build
process.

Fixes #962
Fixes #1097
Fixes #1115

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -11,12 +11,11 @@ from __future__ import absolute_import, print_function
 import os
 import sys
 import tempfile
-import zipfile
 from argparse import Action, ArgumentDefaultsHelpFormatter, ArgumentParser, ArgumentTypeError
 from textwrap import TextWrapper
 
 from pex import pex_warnings
-from pex.common import die, safe_delete, safe_mkdtemp
+from pex.common import atomic_directory, die, open_zip, safe_mkdtemp
 from pex.inherit_path import InheritPath
 from pex.interpreter import PythonInterpreter
 from pex.interpreter_constraints import (
@@ -27,7 +26,7 @@ from pex.jobs import DEFAULT_MAX_JOBS
 from pex.network_configuration import NetworkConfiguration
 from pex.orderedset import OrderedSet
 from pex.pex import PEX
-from pex.pex_bootstrapper import iter_compatible_interpreters
+from pex.pex_bootstrapper import ensure_venv, iter_compatible_interpreters
 from pex.pex_builder import PEXBuilder
 from pex.pip import ResolverVersion
 from pex.platforms import Platform
@@ -35,6 +34,7 @@ from pex.resolver import Unsatisfiable, parsed_platform, resolve_multi
 from pex.tracer import TRACER
 from pex.typing import TYPE_CHECKING
 from pex.variables import ENV, Variables
+from pex.venv_bin_path import BinPath
 from pex.version import __version__
 
 if TYPE_CHECKING:
@@ -86,6 +86,17 @@ class HandleTransitiveAction(Action):
 
     def __call__(self, parser, namespace, value, option_str=None):
         setattr(namespace, self.dest, option_str == "--transitive")
+
+
+class HandleVenvAction(Action):
+    def __init__(self, *args, **kwargs):
+        kwargs["nargs"] = "?"
+        kwargs["choices"] = (BinPath.PREPEND.value, BinPath.APPEND.value)
+        super(HandleVenvAction, self).__init__(*args, **kwargs)
+
+    def __call__(self, parser, namespace, value, option_str=None):
+        bin_path = BinPath.FALSE if value is None else BinPath.for_value(value)
+        setattr(namespace, self.dest, bin_path)
 
 
 class PrintVariableHelpAction(Action):
@@ -335,6 +346,19 @@ def configure_clp_pex_options(parser):
         help="Whether or not the pex file should be unzipped before executing it. If the pex file will "
         "be run multiple times under a stable runtime PEX_ROOT the unzipping will only be "
         "performed once and subsequent runs will enjoy lower startup latency.",
+    )
+
+    group.add_argument(
+        "--venv",
+        dest="venv",
+        metavar="{prepend,append}",
+        default=False,
+        action=HandleVenvAction,
+        help="Convert the pex file to a venv before executing it. If 'prepend' or 'append' is "
+        "specified then all scripts and console scripts provided by distributions in the pex file "
+        "will be added to the PATH. If the the pex file will be run multiple times under a stable "
+        "runtime PEX_ROOT the venv creation will only be done once and subsequent runs will enjoy "
+        "lower startup latency.",
     )
 
     group.add_argument(
@@ -713,6 +737,16 @@ def configure_clp():
     )
 
     parser.add_argument(
+        "--seed",
+        "--no-seed",
+        dest="seed",
+        action=HandleBoolAction,
+        default=False,
+        help="Seed local Pex caches for the generated PEX and print out the command line to run "
+        "directly from the seed with.",
+    )
+
+    parser.add_argument(
         "--help-variables",
         action=PrintVariableHelpAction,
         nargs=0,
@@ -824,7 +858,7 @@ def build_pex(reqs, options, cache=None):
         path=safe_mkdtemp(),
         interpreter=interpreter,
         preamble=preamble,
-        include_tools=options.include_tools,
+        include_tools=options.include_tools or options.venv,
     )
 
     if options.resources_directory:
@@ -844,6 +878,8 @@ def build_pex(reqs, options, cache=None):
     pex_info = pex_builder.info
     pex_info.zip_safe = options.zip_safe
     pex_info.unzip = options.unzip
+    pex_info.venv = bool(options.venv)
+    pex_info.venv_bin_path = options.venv
     pex_info.pex_path = options.pex_path
     pex_info.always_write_cache = options.always_write_cache
     pex_info.ignore_errors = options.ignore_errors
@@ -1005,14 +1041,31 @@ def main(args=None):
 
         if options.pex_name is not None:
             log("Saving PEX file to %s" % options.pex_name, V=options.verbosity)
-            tmp_name = options.pex_name + "~"
-            safe_delete(tmp_name)
             pex_builder.build(
-                tmp_name,
+                options.pex_name,
                 bytecode_compile=options.compile,
                 deterministic_timestamp=not options.use_system_time,
             )
-            os.rename(tmp_name, options.pex_name)
+            if options.seed:
+                pex_path = pex.path()
+                with TRACER.timed("Seeding local caches for {}".format(pex_path)):
+                    if options.unzip:
+                        unzip_dir = pex.pex_info().unzip_dir
+                        with atomic_directory(unzip_dir, exclusive=True) as chroot:
+                            if chroot:
+                                with TRACER.timed("Extracting {}".format(pex_path)):
+                                    with open_zip(options.pex_name) as pex_zip:
+                                        pex_zip.extractall(chroot)
+                        print("{} {}".format(pex.interpreter.binary, unzip_dir))
+                    elif options.venv:
+                        with TRACER.timed("Creating venv from {}".format(pex_path)):
+                            print(ensure_venv(pex))
+                    else:
+                        with TRACER.timed(
+                            "Extracting code and distributions for {}".format(pex_path)
+                        ):
+                            pex.activate()
+                        print(os.path.abspath(options.pex_name))
         else:
             if not _compatible_with_current_platform(interpreter, options.platforms):
                 log("WARNING: attempting to run PEX with incompatible platforms!", V=1)
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -12,6 +12,7 @@ import os
 import sys
 import tempfile
 from argparse import Action, ArgumentDefaultsHelpFormatter, ArgumentParser, ArgumentTypeError
+from shlex import shlex
 from textwrap import TextWrapper
 
 from pex import pex_warnings
@@ -38,7 +39,7 @@ from pex.venv_bin_path import BinPath
 from pex.version import __version__
 
 if TYPE_CHECKING:
-    from typing import List
+    from typing import List, Iterable
     from argparse import Namespace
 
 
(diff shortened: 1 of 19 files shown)


pex/tools/commands/venv.py
```diff
@@ -61,12 +63,206 @@ def _copytree(
...
-                        "Failed to overwrite {} with {}: {}".format(dst_entry, src_entry, e)
+                if e.errno != errno.EEXIST:
+                    raise e
+
+
+class CollisionError(Exception):
+    """Indicates multiple distributions provided the same file when merging a PEX into a venv."""
+
+
+def populate_venv_with_pex(
```
jsirois, line 74: This was mainly extracted from the Venv Command so that pex_bootstrapper could use it. The only real change was to add `record_provenance` which I used to track down a collision in building a venv from Toolchain requirements. In that case two different distributions were providing the same script - `jsondiff` (one from https://pypi.org/project/jsondiff/ and one from https://pypi.org/project/jsonpatch/). The `bin/` scripts are namespaced even more loosely than the global project package namespace (basically not at all) so this seems likely to keep happening. That drove me to default collisions being ok for the `--venv` mode and just emitting warnings.

pex/tools/commands/venv.py
```diff
@@ -61,12 +63,206 @@ def _copytree(
...
+                        sys.stderr.write("Argument expected for the -c option\\n")
+                        sys.exit(2)
+                    filename = "-c <cmd>"
+                    content = args[1]
+                    sys.argv = ["-c"] + args[2:]
+                elif arg == "-":
+                    content = sys.stdin.read()
+                else:
+                    with open(arg) as fp:
+                        content = fp.read()
```
Eric-Arellano, line 221: You may want to open this with `rb`. From the docs https://docs.python.org/3/library/functions.html#compile:

> source can either be a normal string, a byte string, or an AST object.

Why do it? I found in https://github.com/pantsbuild/pants/pull/11213/commits/d8b50388d12e204e953dde50d6fa6c2d127a0bb0 for Pants' dep inference that opening with `rb` avoided Unicode errors.

pex/venv_bin_path.py
```diff
@@ -0,0 +1,37 @@
+# Copyright 2020 Pants project contributors (see CONTRIBUTORS.md).
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+from __future__ import absolute_import
+
+
+class BinPath(object):
```
Eric-Arellano, line 7: This looks similar to `Enum` - once we drop Py2, would we want to switch this to that? If so, a TODO would be great.

pex/variables.py
```diff
@@ -93,6 +95,25 @@ def strip_default(self, instance):
...
 
+    def value_or(
+        self,
+        instance,  # type: _O
+        fallback,  # type: _P
+    ):
+        # type: (...) -> _P
+        """Return the value of this property without the default value applied or else the fallback.
+
+        If the property is not set `fallback` will be validated and returned.
```
Eric-Arellano, line 106:
```suggestion
        If the property is not set, `fallback` will be validated and returned.
```

pex/variables.py
```diff
@@ -93,6 +95,25 @@ def strip_default(self, instance):
         except NoValueError:
             return None
 
+    def value_or(
+        self,
+        instance,  # type: _O
+        fallback,  # type: _P
+    ):
+        # type: (...) -> _P
+        """Return the value of this property without the default value applied or else the fallback.
```
Eric-Arellano, line 104: To confirm, this is _without_ the default? That's not a typo?

If so, perhaps:

> Return the value of this property, without the default value applied, and falling back to `fallback`.

Eric-Arellano (approved):
Awesome

pex/tools/commands/venv.py
```diff
@@ -61,12 +63,206 @@ def _copytree(
                         link = False
                 shutil.copy(src_entry, dst_entry)
             except OSError as e:
-                if e.errno == errno.EEXIST:
-                    pex_warnings.warn(
-                        "Failed to overwrite {} with {}: {}".format(dst_entry, src_entry, e)
+                if e.errno != errno.EEXIST:
+                    raise e
```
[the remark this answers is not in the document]
jsirois: Because the only error type considered tolerable by the API is exists (collisions_ok). Anything else is an error - say out of disk space.

pex/bin/pex.py
```diff
@@ -337,6 +348,19 @@ def configure_clp_pex_options(parser):
...
+        "--venv",
+        dest="venv",
+        metavar="{prepend,append}",
+        default=False,
+        action=HandleVenvAction,
+        help="Convert the pex file to a venv before executing it. If 'prepend' or 'append' is "
+        "specified then all scripts and console scripts provided by distributions in the pex file "
+        "will be added to the PATH. If the the pex file will be run multiple times under a stable "
+        "runtime PEX_ROOT the venv creation will only be done once and subsequent runs will enjoy "
+        "lower startup latency.",
```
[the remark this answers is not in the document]
jsirois: I don't really buy the confusion - there seems to be no other possible interpretation. I added more words though.

pex/tools/commands/venv.py
```diff
@@ -61,12 +63,206 @@ def _copytree(
...
-                    if not collisions_ok:
-                        raise e
+                )
+                sys.exit(1)
+            os.environ["PATH"] = os.pathsep.join(PATH)
+
+        PEX_EXEC_OVERRIDE_KEYS = ("PEX_INTERPRETER", "PEX_SCRIPT", "PEX_MODULE")
+        pex_overrides = dict(
+            (key, os.environ.pop(key)) for key in PEX_EXEC_OVERRIDE_KEYS if key in os.environ
+        )
```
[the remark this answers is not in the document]
jsirois: I preferred avoiding the escaping gore, but done.

pex/tools/commands/venv.py
jsirois: The relevant doc is here and different: https://docs.python.org/2.7/library/functions.html#compile

pex/venv_bin_path.py
jsirois: Yes, this should go to enum as should InheritPath. This just moved out of pex.tools.commands.venv so I'll leave as-is. There are alot of things that can change once we drop Python 2.7 including namedtuple -> dataclass and more. That will be a whole project.

pex/variables.py
jsirois: Fixed.

pex/variables.py
jsirois: Confirmed. I do not see how the new words make things more clear.

pex/bin/pex.py
```diff
@@ -1005,14 +1041,31 @@ def warn_ignore_pex_root(set_via):
...
+                with TRACER.timed("Seeding local caches for {}".format(pex_path)):
+                    if options.unzip:
+                        unzip_dir = pex.pex_info().unzip_dir
+                        with atomic_directory(unzip_dir, exclusive=True) as chroot:
+                            if chroot:
+                                with TRACER.timed("Extracting {}".format(pex_path)):
+                                    with open_zip(options.pex_name) as pex_zip:
+                                        pex_zip.extractall(chroot)
+                        print("{} {}".format(pex.interpreter.binary, unzip_dir))
+                    elif options.venv:
```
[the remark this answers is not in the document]
jsirois: Ah yes, meant to do so. Fixed with a mutex arg group

pex/pex_builder.py
```diff
@@ -521,21 +546,20 @@ def build(self, filename, bytecode_compile=True, deterministic_timestamp=False):
...
         if not self._frozen:
             self.freeze(bytecode_compile=bytecode_compile)
+        tmp_zip = filename + "~"
         try:
-            os.unlink(filename + "~")
+            os.unlink(tmp_zip)
             self._logger.warning(
-                "Previous binary unexpectedly exists, cleaning: %s" % (filename + "~")
+                "Previous binary unexpectedly exists, cleaning: {}".format(tmp_zip)
             )
```
stuhood, line 554: Should this be above the `unlink` call? Wouldn't matter unless it was going to fail, but that's the most useful time to be warned.

pex/tools/commands/venv.py
```diff
@@ -61,12 +63,206 @@ def _copytree(
...
+                    index=index, dst=dst, srcs="\n\t".join(srcs)
+                )
+            )
+        message = "\n".join(message_lines)
+        if not collisions_ok:
+            raise CollisionError(message)
+        pex_warnings.warn(message)
+
+    # 2. Add a __main__ to the root of the venv for running the venv dir like a loose PEX dir
+    # and a main.py for running as a script.
```
stuhood, line 138: Is it worth checking for a collision with `__main__.py`?

pex/tools/commands/venv.py
stuhood: > That drove me to default collisions being ok for the --venv mode and just emitting warnings.

It looks like the default for the option is `False`?

pex/tools/commands/venv.py
```diff
@@ -61,12 +63,206 @@ def _copytree(
...
+            venv_python=venv_python,
+            venv_bin_dir=venv_bin_dir,
+            venv_dir=venv_dir,
+            bin_path=bin_path,
+            entry_point=pex_info.entry_point,
+            exec_ast=(
+                "exec ast in globals_map, locals_map"
+                if venv.interpreter.version[0] == 2
+                else "exec(ast, globals_map, locals_map)"
+            ),
```
stuhood, line 255: Maybe not worth it yet, but if all of the templated variables are strings, you might consider extracting this as a "real" Python file to allow for mypy/linting, and then loading the file as a resource to template it using a slightly heavier template syntax (something like mustache, but cheaper hopefully).

pex/tools/commands/virtualenv.py
```diff
@@ -91,7 +91,7 @@ def create(
             interpreter.execute(args=["-m", "venv", "--without-pip", venv_dir])
         else:
             virtualenv_py = resource_string(__name__, "virtualenv_16.7.10_py")
-            with named_temporary_file(mode="w") as fp:
+            with named_temporary_file(mode="wb") as fp:
```
stuhood, line 94: Is the `bytes`/`str`-ness of `resource_string` version dependent?

stuhood (approved):
Thanks!

pex/interpreter.py
```diff
@@ -607,7 +607,11 @@ def create_interpreter(stdout, check_binary=False):
         # would otherwise be unstable.
         #
         # See cls._REGEXEN for a related affordance.
-        path_id = binary.replace(os.sep, ".").lstrip(".")
+        #
+        # N.B.: The path for --venv mode interpreters can be quite long; so we just used a fixed
+        # length hash of the interpreter binary path to ensure uniqeness and not run afoul of file
```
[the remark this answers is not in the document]
jsirois: Fixed,

pex/bin/pex.py
```diff
@@ -1005,14 +1041,31 @@ def warn_ignore_pex_root(set_via):
...
-            tmp_name = options.pex_name + "~"
-            safe_delete(tmp_name)
             pex_builder.build(
-                tmp_name,
+                options.pex_name,
                 bytecode_compile=options.compile,
                 deterministic_timestamp=not options.use_system_time,
             )
-            os.rename(tmp_name, options.pex_name)
+            if options.seed:
```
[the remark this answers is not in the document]
jsirois: Done.

pex/pex_builder.py
jsirois: It would have to be more complex than that since this is in the ask for forgiveness later style. This was existing so I'll leave as-is.

pex/tools/commands/venv.py
jsirois: There can be none. A venv is layed out as:
```
<venv dir>/
  bin/
  lib/
  include/
  pyvenv.cfg
```
The `__main__.py` and the `pex` symlink are dropped in the `<venv dir>` root, not in `lib/.../site-packages` which is the place where collisions could occur.

pex/tools/commands/venv.py
jsirois: Nope, `collisions_ok=True` is the default.

pex/tools/commands/venv.py
jsirois: Perhaps worth it down the line as with the venv `__main__.py` and the `PythonInterpreter` identification code, etc...

pex/tools/commands/virtualenv.py
jsirois: It doesn't look like it - at least as far back as 38.2.0 which was my lazy check. I'm forgetting now why I hit an error here in this PR and not when the class was introduced.

Issue #1155, opened 2020-12-28
pex fails to run on windows: "No module named 'fcntl'"


mgrandi:
```plaintext

PS C:\Users\mgrandi> py -3 -m venv .\Temp\testvenv
PS C:\Users\mgrandi> .\Temp\testvenv\Scripts\Activate.ps1
(testvenv) PS C:\Users\mgrandi> pip3 install pex
Collecting pex
  Using cached pex-2.1.24-py2.py3-none-any.whl (2.5 MB)
Installing collected packages: pex
Successfully installed pex-2.1.24
WARNING: You are using pip version 20.2.3; however, version 20.3.3 is available.
You should consider upgrading via the 'c:\users\mgrandi\temp\testvenv\scripts\python.exe -m pip install --upgrade pip' command.
(testvenv) PS C:\Users\mgrandi> pex --help
Traceback (most recent call last):
  File "C:\Python39\lib\runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Python39\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\mgrandi\Temp\testvenv\Scripts\pex.exe\__main__.py", line 4, in <module>
  File "c:\users\mgrandi\temp\testvenv\lib\site-packages\pex\bin\pex.py", line 19, in <module>
    from pex.common import die, safe_delete, safe_mkdtemp
  File "c:\users\mgrandi\temp\testvenv\lib\site-packages\pex\common.py", line 9, in <module>
    import fcntl
ModuleNotFoundError: No module named 'fcntl'
(testvenv) PS C:\Users\mgrandi> python --version
Python 3.9.0
(testvenv) PS C:\Users\mgrandi> pip3 list
Package    Version
---------- -------
pex        2.1.24
pip        20.2.3
setuptools 49.2.1
WARNING: You are using pip version 20.2.3; however, version 20.3.3 is available.
You should consider upgrading via the 'c:\users\mgrandi\temp\testvenv\scripts\python.exe -m pip install --upgrade pip' command.
(testvenv) PS C:\Users\mgrandi> $PSVersionTable.PSVersion

Major  Minor  Build  Revision
-----  -----  -----  --------
5      1      19041  610


(testvenv) PS C:\Users\mgrandi> [environment]::OSVersion.Version

Major  Minor  Build  Revision
-----  -----  -----  --------
10     0      19042  0

```

jsirois:
[...]

Issue #111, opened 2015-05-28
pex fails in a non-elevated Windows session


pfmoore:
I just tried one of the simple examples from the pex documentation on Windows:

```
PS 19:27 (VE-1) C:\Work\Scratch
->pip install --pre pex
You are using pip version 6.1.1, however version 7.0.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
Collecting pex
  Downloading pex-1.0.0.dev3-py2.py3-none-any.whl (89kB)
    100% |################################| 90kB 1.4MB/s
Requirement already satisfied (use --upgrade to upgrade): setuptools<16,>=2.2 in c:\users\gustav\appdata\local\temp\ve-1\lib\site-packages (from pex)
Installing collected packages: pex
Successfully installed pex-1.0.0.dev3
PS 19:27 (VE-1) {00:03} C:\Work\Scratch
->pex requests psutil
Traceback (most recent call last):
  File "C:\Apps\Python34\Lib\runpy.py", line 170, in _run_module_as_main
    "__main__", mod_spec)
  File "C:\Apps\Python34\Lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Users\Gustav\AppData\Local\Temp\VE-1\Scripts\pex.exe\__main__.py", line 9, in <module>
  File "C:\Users\Gustav\AppData\Local\Temp\VE-1\lib\site-packages\pex\bin\pex.py", line 506, in main
    pex_builder = build_pex(reqs, options, resolver_options_builder)
  File "C:\Users\Gustav\AppData\Local\Temp\VE-1\lib\site-packages\pex\bin\pex.py", line 444, in build_pex
    interpreter = interpreter_from_options(options)
  File "C:\Users\Gustav\AppData\Local\Temp\VE-1\lib\site-packages\pex\bin\pex.py", line 437, in interpreter_from_options
    interpreter = resolve(interpreter, WHEEL_REQUIREMENT)
  File "C:\Users\Gustav\AppData\Local\Temp\VE-1\lib\site-packages\pex\bin\pex.py", line 410, in resolve_interpreter
    installer_provider)
  File "C:\Users\Gustav\AppData\Local\Temp\VE-1\lib\site-packages\pex\bin\pex.py", line 380, in _resolve_and_link_interpreter
    _safe_link(target_location, target_link)
[...]

PR #1157, opened 2020-12-30
Add Windows support to file locking

chrisjbremner:
While this project does not officially support Windows, it was compatible with `1.x.x` versions of this package, and there's minimal changes necessary to restore that functionality in future `2.x.x` versions as well. The only noticeable incompatibility at the moment is  around the use of `fcntl` for file locking, which is not available in Windows (`msvcrt` is the alternative), which was discussed in https://github.com/pantsbuild/pex/issues/1155. This PR adds logic to choose the correct package for file locking based on what platform is being used.

I was able to successfully build and test a `.pex` file on Windows 10 after making this change. As noted [here](https://docs.python.org/3/library/os.html#os.symlink), a user will need to build the `.pex` as an Administrator, Developer Mode, or enable the `SeCreateSymbolicLinkPrivilege` privilege in order for symlinking to occur correctly.

### DIFF
--- a/pex/common.py
+++ b/pex/common.py
@@ -6,7 +6,6 @@ from __future__ import absolute_import, print_function
 import atexit
 import contextlib
 import errno
-import fcntl
 import os
 import re
 import shutil
@@ -21,7 +20,13 @@ from contextlib import contextmanager
 from datetime import datetime
 from uuid import uuid4
 
+from pex.compatibility import WINDOWS
 from pex.typing import TYPE_CHECKING
+
+if WINDOWS:
+    import msvcrt
+else:
+    import fcntl
 
 if TYPE_CHECKING:
     from typing import Any, DefaultDict, Iterable, Iterator, NoReturn, Optional, Set, Sized
@@ -393,7 +398,10 @@ def atomic_directory(target_dir, exclusive, source=None):
         if lock_fd is None:
             return
         try:
-            fcntl.lockf(lock_fd, fcntl.LOCK_UN)
+            if WINDOWS:
+                msvcrt.locking(lock_fd, msvcrt.LK_UNLCK, 1)
+            else:
+                fcntl.lockf(lock_fd, fcntl.LOCK_UN)
         finally:
             os.close(lock_fd)
 
@@ -410,7 +418,18 @@ def atomic_directory(target_dir, exclusive, source=None):
         # N.B.: Since lockf operates on an open file descriptor and these are guaranteed to be
         # closed by the operating system when the owning process exits, this lock is immune to
         # staleness.
-        fcntl.lockf(lock_fd, fcntl.LOCK_EX)  # A blocking write lock.
+        if WINDOWS:
+            while True:
+                # Force the non-blocking lock to be blocking. LK_LOCK is msvcrt's implementation of
+                # a blocking lock, but it only tries 10 times, once per second before rasing an
+                # OSError.
+                try:
+                    msvcrt.locking(lock_fd, msvcrt.LK_NBLCK, 1)
+                    break
+                except OSError:
+                    safe_sleep(1)
+        else:
+            fcntl.lockf(lock_fd, fcntl.LOCK_EX)  # A blocking write lock.
         if atomic_dir.is_finalized:
             # We lost the double-checked locking race and our work was done for us by the race
             # winner so exit early.
--- a/pex/common.py
+++ b/pex/common.py
@@ -424,9 +424,12 @@ def atomic_directory(target_dir, exclusive, source=None):
                 # a blocking lock, but it only tries 10 times, once per second before rasing an
                 # OSError.
                 try:
-                    msvcrt.locking(lock_fd, msvcrt.LK_NBLCK, 1)
+                    msvcrt.locking(lock_fd, msvcrt.LK_LOCK, 1)
                     break
-                except OSError:
+                except OSError as ex:
+                    # Deadlock error is raised after failing to lock the file
+                    if ex.errno != errno.EDEADLOCK:
+                        raise
                     safe_sleep(1)
         else:
             fcntl.lockf(lock_fd, fcntl.LOCK_EX)  # A blocking write lock.


jsirois:
Thanks @chrisjbremner. I hope it was clear from #1155 that the blocking issue from the Pex side is having CI set up. I do not want to continue to give a false impression that Pex supports Windows. It does not. Even with this fix it does not. Only with this fix + a CI setup for Windows that's green can Pex support Windows in good faith now and going forward. What's you're plan or suggestion to get the CI side of this squared away?

pex/common.py
```diff
@@ -410,7 +418,18 @@ def unlock():
         # N.B.: Since lockf operates on an open file descriptor and these are guaranteed to be
         # closed by the operating system when the owning process exits, this lock is immune to
         # staleness.
-        fcntl.lockf(lock_fd, fcntl.LOCK_EX)  # A blocking write lock.
+        if WINDOWS:
+            while True:
+                # Force the non-blocking lock to be blocking. LK_LOCK is msvcrt's implementation of
```
jsirois, line 423: Presumably then using `LK_LOCK` is a bit better here. You still need to loop, but presumably its slightly more efficient to let Windows do the 1st 10 tries before sleeping here and looping than to do the 1st 10 tries here manually.

chrisjbremner: Makes sense, it also raises a clearer error (`OSError: [Errno 36] Resource deadlock avoided`) when the locking operation times out after 10 tries, so we can use that directly rather than an `OSError` catchall. Fixed in https://github.com/pantsbuild/pex/pull/1157/commits/edc84354c38a4d422fc551f58564a8323cadbf06

chrisjbremner:
> What's your plan or suggestion to get the CI side of this squared away?

I am willing to try and help set up some CI for this. I haven't worked much with Travis before, but can give it a shot. What's the process for getting started?

pex/common.py
jsirois: Excellent - that's much better.

jsirois:
> What's the process for getting started (do I need access to somewhere else to add a pipeline)?

That's kind of the point - no clue here - I've never set up a Windows CI for an OSS project and haven't used Windows in ~15 years.

It does look like Travis may support Windows now: https://docs.travis-ci.com/user/reference/windows/. If you want to try that you'd just make the appropriate edits to the .travis.yml in this commit.

But Travis is not a requirement. I've seen https://www.appveyor.com/ used over in the Pip project and I'm sure there are other options too. If you're willing to dive in I can provide any support needed to click buttons to enable services for the pantsbuild org / the pex repo.

chrisjbremner:
Sounds good, I'll survey other OSS projects to get an idea of what they do, then let you know.

Is CI a blocker for merging of this PR, or just to add the `Operating System :: Microsoft :: Windows` classifier to the project?

jsirois:
> Is CI a blocker for merging of this PR

It's a blocker for this PR. All other PRs must be green and this one is only fake green. We really need to know it works independently of you or I. Thanks for taking a crack at all this.

chrisjbremner:
Ok, I've created https://github.com/pantsbuild/pex/pull/1158 for the Windows CI effort

chrisjbremner:
I've spent quite a while trying to figure out why some tests are failing on Windows and it seems to be isolated to Tox (wheels get downloaded, but not installed for a currently unknown reason), since I am able to build a Pex successfully outside of Tox. While I will continue trying to get Windows CI to work, I don't necessarily think it should be a requirement to merge this PR. Given the presence of `pex.compatibility.WINDOWS`, it seems like Windows was previously intended to be supported, so wouldn't it make sense to implement this fix so that anyone who depended on it in previous versions (like myself) can continue to use it as they were?

jsirois:
Since you are our current Windows expert, you should feel free to not use tox for the windows CI. Whatever works and can also be easily run by a Windows maintainer on their own machine is great. That could simply be a PowerShell script or batch file that wraps up the testing and can be used both locally and in CI.

As to needing CI green to land any Windows fixes, I really don't want to budge on that. We need to make forward progress, and the only way to do that is with a green base. Forward progress for you is modern Pex working on Windows - I understand - but that is not forward progress for the project - its just the status quo - Windows is still unsupported and will break at any minute.

> Given the presence of pex.compatibility.WINDOWS, it seems like Windows was previously intended to be supported, ...

No. See the limp along comment below, but I've been the primary maintainer for 2+ years and - at least to me - no open source project should ever support something it doesn't test (you can be assured I was not around to accept the compatibility change you referenced).

> so wouldn't it make sense to implement this fix so that anyone who depended on it in previous versions (like myself) can continue to use it as they were?

For that - you can just use older versions. If you need newer features of Pex then we really need you, or some one, to own getting CI working; otherwise, we just re-engage limp along mode where Pex _happens_ to work on Windows but is not at all supported.

jcar87:
Landed here after trying things out on WIndows (didn't really check it wasn't supported).

This PR was helpful and I was able to hack around and factor it into `atomic_directory.py`. There's quite a few other things that needed tweaks - although I suspect I was only using a fraction of the functionality and barely scratching the surface. Mostly what I found had to do with deriving the location of the interpreter (there's  usually no `bin/python3`, but rather `Scripts/python.exe`), and some interesting issue with path comparisons with strings failing because Windows paths contain (or may contain) `\` instead of forward slashes. 

@jsirois, I can see that at some point progress was made towards Windows compatibility (as per this comment https://github.com/pantsbuild/pex/issues/111#issuecomment-1251798488). Is the CI issue mentioned in the past "easier" now that GitHub actions are used?  

Was wondering since some of the tweaks I've made could potentially be worked into proper patches that don't cause regressions for already supporting platforms, but would be good to have CI test coverage - even if pex isn't fully functional on Windows yet

jsirois:
@jcar87 the issue has never really been getting Windows boxes for CI, its been my insistence that Windows support is added with green CI to back it up. No one has been willing to put in all that work.

I made significant progress (https://github.com/jsirois/pex/tree/Windows/wip) but when I was last diverted to other priorities, there were still ~80 integration tests failing (I believe I got all unit tests passing).

My stance on only going forward with Windows support under a fully green CI PR remains and I will circle back to that branch, but that's cold comfort with no schedule. If you want to put in the work to meet the fully green CI goal, that works too of course. It's alot of work though!

jsirois:
@chrisjbremner thanks again for the attempt here. After failing my own effort at landing Windows support fully green with CI in 2022, I'm starting afresh with an incremental approach much like yours here - although there will be CI with a progress indication (100% red today, 75% tomorrow, etc until green). The 1st increment in #2663 lands Windows lock file support which `msvcrt.locking` is no longer sufficient for (Pex has gone on to use shared vs exclusive locks). I'll close your PR as a result.

Issue #899, opened 2020-02-20
Top level environment markers are lost in built PEXes.


jsirois:
When any requirements passed to the Pex CLI have environment markers, these should be recorded in the resulting PEX / PEX-INFO but currently are not. This can lead to erroneous runtime activation failure:
```
$ python -mpex --python=python2.7 --python=python3.6 "subprocess32==3.2.7; python_version<'3'" -o maybe_subprocess32.pex
DEPRECATION: Python 2.7 reached the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 is no longer maintained. A future version of pip will drop support for Python 2.7. More details about Python 2 support in pip, can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support
DEPRECATION: Python 2.7 reached the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 is no longer maintained. A future version of pip will drop support for Python 2.7. More details about Python 2 support in pip, can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support
DEPRECATION: Python 2.7 reached the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 is no longer maintained. A future version of pip will drop support for Python 2.7. More details about Python 2 support in pip, can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support
$ ./maybe_subprocess32.pex 
Python 2.7.17 (default, Oct 22 2019, 09:14:09) 
[GCC 9.2.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
(InteractiveConsole)
>>> 
$ python3.6 ./maybe_subprocess32.pex 
Failed to execute PEX file. Needed manylinux2014_x86_64-cp-36-cp36m compatible dependencies for:
 1: subprocess32==3.2.7
    But this pex only contains:
      subprocess32-3.2.7-cp27-cp27mu-linux_x86_64.whl
$ unzip -qc maybe_subprocess32.pex PEX-INFO | jq .
{
  "always_write_cache": false,
  "build_properties": {
    "class": "CPython",
    "pex_version": "2.1.3",
    "platform": "manylinux2014_x86_64",
    "version": [
      2,
      7,
      17
    ]
  },
  "code_hash": "da39a3ee5e6b4b0d3255bfef95601890afd80709",
  "distributions": {
[...]

PR #1160, opened 2021-01-04
Support project_name_and_version metadata.

jsirois:
In addition to extracting this metadata from Distributions, retrofit support
for loading metadata from uninstalled wheels and sdists which will be needed
to fully support requirements parsing needs.

Work towards #899, #1020 and #1108.


stuhood (approved):
Thanks!

PR #1166, opened 2021-01-11
Refresh and type OrderedSet.

jsirois:
Since Python 2.7 has OrderedDict, just use that for the implementation
and add typing support.

### DIFF
--- a/pex/orderedset.py
+++ b/pex/orderedset.py
@@ -10,71 +10,65 @@
 
 from __future__ import absolute_import
 
+from collections import OrderedDict
+
 from pex.compatibility import MutableSet
+from pex.typing import TYPE_CHECKING, Generic
+
+if TYPE_CHECKING:
+    from typing import Any, Iterable, Optional, TypeVar
+
+    _I = TypeVar("_I")
 
 
-class OrderedSet(MutableSet):
-    KEY, PREV, NEXT = range(3)
-
+class OrderedSet(MutableSet, Generic["_I"]):
     def __init__(self, iterable=None):
-        self.end = end = []
-        end += [None, end, end]  # sentinel node for doubly linked list
-        self.map = {}  # key --> [key, prev, next]
+        # type: (Optional[Iterable[_I]]) -> None
+        self.map = OrderedDict()  # type: OrderedDict[_I, None]
         if iterable is not None:
-            self |= iterable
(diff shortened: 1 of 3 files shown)


jsirois:
This is taking over for #1164.

pex/orderedset.py
```diff
@@ -10,71 +10,65 @@
...
-class OrderedSet(MutableSet):
-    KEY, PREV, NEXT = range(3)
 
+class OrderedSet(MutableSet, Generic["_I"]):
     def __init__(self, iterable=None):
-        self.end = end = []
-        end += [None, end, end]  # sentinel node for doubly linked list
-        self.map = {}  # key --> [key, prev, next]
+        # type: (Optional[Iterable[_I]]) -> None
+        self.map = OrderedDict()  # type: OrderedDict[_I, None]
```
stuhood, line 27: If the API is changing anyway (public `self.end` going away), it seems like maybe a good time to hide `self.map`, unless it's consumed within the codebase?

pex/orderedset.py
```diff
@@ -10,71 +10,65 @@
...
-            return "%s()" % (self.__class__.__name__,)
-        return "%s(%r)" % (self.__class__.__name__, list(self))
+            return "{}()".format(
+                self.__class__.__name__,
+            )
+        return "{}({!r})".format(self.__class__.__name__, list(self))
 
     def __eq__(self, other):
         if isinstance(other, OrderedSet):
             return len(self) == len(other) and list(self) == list(other)
```
stuhood, line 73: Directly comparing `self.map` should be cheaper than converting to a list, as `OrderedDict` It takes ordering into account. Downside might be that Python3's dict does _not_ take ordering into account in eq, so would need a comment to fix that when Python2 support is dropped.

stuhood (approved):

pex/orderedset.py
jsirois: SGTM - I changed it back from self._data to self.map at the last minute just to be nice to reviewers.

pex/orderedset.py
jsirois: Fixed. I bit the bullet and re-implemented `__eq__` completely taking away its ability to compare `OrderedSet([1, 2, 3]) == {1, 2, 3}` which was all of strange, wrong and unused.

PR #1169, opened 2021-01-12
Type BuildAndInstallRequest.

jsirois:
Work towards #899.

### DIFF
--- a/pex/resolver.py
+++ b/pex/resolver.py
@@ -347,6 +347,19 @@ class BuildRequest(object):
         # type: (str) -> BuildResult
         return BuildResult.from_request(self, dist_root=dist_root)
 
+    def __repr__(self):
+        # type: () -> str
+        return (
+            "{class_name}("
+            "target={target!r}, source_path={source_path!r}, fingerprint={fingerprint!r}"
+            ")"
+        ).format(
+            class_name=self.__class__.__name__,
+            target=self.target,
+            source_path=self.source_path,
+            fingerprint=self.fingerprint,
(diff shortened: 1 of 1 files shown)


jsirois:
Depends on #1167 (2f2fadd) and #1168 (2229d24). Please just review 3ebfda2.

pex/resolver.py
```diff
@@ -347,32 +376,46 @@ def from_request(cls, build_request, dist_root):
...
     def finalize_build(self):
-        self.atomic_dir.finalize()
+        # type: () -> Iterator[InstallRequest]
+        self._atomic_dir.finalize()
         for wheel in os.listdir(self.dist_dir):
             yield InstallRequest.create(self.request.target, os.path.join(self.dist_dir, wheel))
 
 
-class InstallRequest(namedtuple("InstallRequest", ["target", "wheel_path", "fingerprint"])):
+class InstallRequest(object):
```
Eric-Arellano, line 410: Does this need `__eq__` and `__hash__`? May be worth doing `__repr__` too.

Eric-Arellano: Ditto on the other classes in this file that don't use namedtuple anymore.

Eric-Arellano (approved):
Excellent! Thanks!

pex/resolver.py
jsirois: They do not need `__eq__` / `__hash__` but `{Build,Install}Request` could both stand a `__repr__` for the verbose logs.

PR #1177, opened 2021-01-17
Platform can report supported_tags.

jsirois:
We need this in order to support resolving foreign platforms from PEX
file repositories just like we support resolving foreign platforms from
traditional index servers and find links repositories via Pip.

We use `pip debug` for this even though Pip warns against relying on the
output of that command. Since Pip is vendored, this is only a concern
when we upgade our vendored Pip. If Pip does yank the debug output we
need or alter its format, we can either implement our own logic using
packaging.tags (~150 LOC) or adapt our parsing logic respectively.

Work towards #1108.


jsirois:
Reviewers, don't be afraid of the diff size. 909 lines are test data - a huge list of expected platform tags.

pex/pip.py
```diff
@@ -297,24 +299,55 @@ def _spawn_pip_isolated(
...
+        abi,  # type: str
+        manylinux=None,  # type: Optional[str]
+    ):
+        # type: (...) -> Iterator[str]
+        if manylinux and platform.startswith("linux"):
+            yield "--platform"
+            yield platform.replace("linux", manylinux, 1)
+
+        yield "--platform"
+        yield platform
```
Eric-Arellano, line 340: Should that second part be in an else clause? Otherwise we'd get `--platform` twice.

Eric-Arellano: If that's intentional, a brief comment acknowledging this is intentional would help.

pex/pip.py
```diff
@@ -521,6 +555,31 @@ def spawn_install_wheel(
...
+    def spawn_debug(
+        self,
+        platform,  # type: str
+        impl,  # type: str
+        version,  # type: str
+        abi,  # type: str
+        manylinux=None,  # type: Optional[str]
+    ):
+        # type: (...) -> Job
+        # N.B.: Pip gives fair warning:
```
Eric-Arellano, line 567: Maybe worth adding something like this: "However, we can use it because we vendor a pinned version of pip".

tests/test_platform.py
```diff
@@ -41,4 +44,939 @@ def test_platform_create_bad_platform_empty_fields():
...
+    platform = Platform.create("macosx-10.13-x86_64-cp-36-m")
+
+    # A golden file test. This could break if we upgrade Pip and it upgrades packaging which, from
+    # time to time, corrects omissions in tag sets.
+    assert (
+        tuple(
+            itertools.chain.from_iterable(
+                tags.parse_tag(tag)
+                for tag in (
+                    "cp36-cp36m-macosx_10_13_x86_64",
```
Eric-Arellano, line 60: Could you please save this to something like a .txt file and load as a fixture? I think it will keep this test file much easier to work with and to read.

Eric-Arellano (approved):
Thanks, John!

pex/pip.py
jsirois: Sure. It won't be brief though. This was the reason for the 1st year of running with patched Pip when Pex 1st switched to Pip to support Twitter's use case. It'll take a few paragraphs but be a bit more direct than the 5 tests that fail when you remove it.

pex/pip.py
jsirois: I'll pass on this since it's redundant if you're reading this file. It's central to all aspects of calling pip that we have a pinned version - vendored or not - so we can rely on flags for all other three commands as well.

tests/test_platform.py
jsirois: Yeah, sg.

Eric-Arellano (approved):
Thanks!

PR #1178, opened 2021-01-19
PEXEnvironment for a DistributionTarget.

jsirois:
Instead of requiring a PythonIntepreter to resolve distributions in a
PEXEnvironment, a DistributionTarget is enough. This allows for
resolving distributions from a PEXEnvironment given only a Platform
which we need to support resolving from a `--pex-repository` for
foreign platforms in #1108.


jsirois:
Alright, after a bit of splitting-up-a-branch thrash, this is ready for review.

stuhood (approved):

PR #1180, opened 2021-01-19
Fix plumbing of 2020-resolver to Pip.

### LINKED ISSUE #1179: Pex does not use the pip 2020-resolver when resolving for Python 2.7
It turns out Pip defaults to the legacy resolver under Python 2.7 and the 2020 resolver under anything else. As such, we need to actively pass --use-feature=2020-resolver when our resolve target is a Python 2.7 interpreter.

jsirois:
Previously this was not working for Python 2.7.

Fixes #1179

### DIFF
--- a/pex/pip.py
+++ b/pex/pip.py
@@ -28,20 +28,15 @@ if TYPE_CHECKING:
 
 class ResolverVersion(object):
     class Value(object):
-        def __init__(
-            self,
-            value,  # type: str
-            pip_args=None,  # type: Optional[Iterable[str]]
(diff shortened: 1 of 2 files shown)


jsirois:
@butlern this fix has an integration test drawn straight from your example. Thanks again for digging on this.

Eric-Arellano (approved):
Thank you for getting to the bottom of this! And thank you @butlern for the report!

butlern:
💯 Thanks for the quick turnaround!

PR #1182, opened 2021-01-20
Support resolving from a PEX file repository.

### LINKED ISSUE #1108: Support PEX files as resolve repositories.

jsirois:
Introduce a `--pex-repository` option to the Pex CLI to switch
requirement resolution from using index servers and find-links
repositories to using a local PEX file with pre-resolved requirements.
This can be useful when a number of projects share a consistent resolve
via a shared requirement file. You can resolve the full requirement file
into a requirements PEX and then later resolve just the portions needed
by each individual project from the fully resolved requirements PEX.

Fixes #1108

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -30,7 +30,7 @@ from pex.pex_bootstrapper import ensure_venv, iter_compatible_interpreters
 from pex.pex_builder import PEXBuilder
 from pex.pip import ResolverVersion
 from pex.platforms import Platform
-from pex.resolver import Unsatisfiable, parsed_platform, resolve_multi
+from pex.resolver import Unsatisfiable, parsed_platform, resolve_from_pex, resolve_multi
 from pex.tracer import TRACER
 from pex.typing import TYPE_CHECKING
 from pex.variables import ENV, Variables
@@ -173,6 +173,18 @@ def configure_clp_pex_resolution(parser):
         dest="indexes",
         type=str,
         help="Additional cheeseshop indices to use to satisfy requirements.",
+    )
+
+    parser.add_argument(
+        "--pex-repository",
+        dest="pex_repository",
+        metavar="FILE",
+        default=None,
+        type=str,
+        help=(
+            "Resolve requirements from the given PEX file instead of from --index servers or "
+            "--find-links repos."
+        ),
     )
 
     default_net_config = NetworkConfiguration.create()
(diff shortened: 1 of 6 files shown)


jsirois:
Depends on #1178 - please review starting at 06ddbe1.

pex/distribution_target.py
```diff
@@ -94,17 +94,21 @@ def requirement_applies(
         requirement,  # type: Requirement
         extras=None,  # type: Optional[Tuple[str, ...]]
     ):
-        # type: (...) -> bool
+        # type: (...) -> Optional[bool]
```
jsirois, line 97: This tri-state is needed to support build-time resolves for foreign platforms. We do not have information to build an environment markers dictionary without a foreign platform's target interpreter present. As such, we can only answer "Maybe" (None) to questions about environment marker applicability. We use this tristate to include distributions without checking environment markers when we can't when building a PEX (i.e.: we err on being overly inclusive). At runtime on the foreign platform, the resolve will be exact an winnow out any extra distributions we shipped over-eagerly. So we get a correct runtime resolve at the cost of potentially including more foreign platform distributions than needed. We can do no better here and this is a fairly narrow edge case.

pex/environment.py
```diff
@@ -335,29 +383,37 @@ def _evaluate_marker(
...
         self,
         requirement,  # type: Requirement
+        resolved_dists_by_key,  # type: MutableMapping[Distribution, _RequirementKey]
+        required,  # type: Optional[bool]
         required_by=None,  # type: Optional[Distribution]
     ):
-        # type: (...) -> Iterator[Union[Distribution, _DistributionNotFound]]
+        # type: (...) -> Iterator[_DistributionNotFound]
+        requirement_key = _RequirementKey.create(requirement)
+        if requirement_key in resolved_dists_by_key:
```
jsirois, line 406: This change to populate `resolved_dists_by_key` as a side-effect and short circuit drops a large, connected resolve graph from 1.2s to 0.6s in my stress testing (and that down from 15s for a normal network-based resolve using Pip with a warm cache). In other words, just resolving the graph minimally is an important optimization that seems to justify this ugliness.

tests/test_resolver.py
```diff
@@ -528,3 +532,260 @@ def test_resolve_arbitrary_equality_issues_940():
...
+    assert 2 == len(distribution_locations_by_key["pysocks"]), (
+        "PySocks has a non-platform-specific Python 2.7 distribution and a non-platform-specific "
+        "Python 3 distribution; so we expect to resolve two distributions - one covering "
+        "Python 2.7 and one covering local Python 3.6 and our cp36 foreign platform."
+    )
+
+    assert "cryptography" in distribution_locations_by_key
+    assert 3 == len(distribution_locations_by_key["cryptography"]), (
+        "The cryptography requirement of the security extra is platform specific; so we expect a "
+        "unique distribution to be resolved for each of the three distributin targets"
```
benjyw, line 675: s/distributin/distribution/

benjyw (approved):
Looks great! This is going to be a big deal and a real differentiator for Pants.

tests/test_resolver.py
jsirois: Fixed.

pex/environment.py
```diff
@@ -88,6 +114,28 @@ class ResolveError(Exception):
...
+    def satisfied_keys(self):
+        # type: () -> Iterator[_RequirementKey]
+
+        # If we resolve a requirement with extras then we've satisfied resolves for the powerset of
+        # the extras.
+        # For example, if we resolve `cake[birthday,wedding]` then we satisfy resolves for:
+        # `cake[]`
+        # `cake[birthday]`
+        # `cake[wedding]`
+        # `cake[birthday,wedding]`
```
Eric-Arellano, line 132: Great explanation. Thanks!

pex/resolver.py
```diff
@@ -1382,3 +1392,83 @@ def install(
...
+
+    direct_requirements = _parse_reqs(requirements, requirement_files, network_configuration)
+    direct_requirements_by_key = OrderedDict()
+    for direct_requirement in direct_requirements:
+        if isinstance(direct_requirement, LocalProjectRequirement):
+            raise Untranslatable(
+                "Cannot resolve local project {path} from the PEX repository {pex}.".format(
+                    path=direct_requirement.path, pex=pex
+                )
+            )
```
Eric-Arellano, line 1419: This is because the entire category of resolves is not valid?

Is the "so what?" here to either a) remove the local project from the input requirements or b) stop using `--pex-respository`? It would be helpful to add that, if so.

pex/bin/pex.py
```diff
@@ -175,6 +175,18 @@ def configure_clp_pex_resolution(parser):
         help="Additional cheeseshop indices to use to satisfy requirements.",
     )
 
+    parser.add_argument(
+        "--pex-repository",
+        dest="pex_repository",
+        metavar="FILE",
+        default=None,
+        type=str,
```
Eric-Arellano, line 183: Can you envision wanting to support multiple PEX repositories in the future?

Eric-Arellano (commented):
Exciting!

pex/resolver.py
jsirois: Correct - the entire category is invalid. That is the so what, but what possible other so what could it be? In other words, is there value in adding a so what when there is no alternative we can suggest besides the obvious ~ "If it hurts don't do it.". I really dislike the precedent of so much hand-holding. I totally agree that providing an alternative when there is one is something we should always do. I think explicating "False" means "No" - which is what this would be tantamount to - is just going to far.

pex/bin/pex.py
jsirois: Absolutely not - that would be incorrect. The only reason it's safe to resolve from a single `--pex-repository` is because it contains the result of a single Pip resolve that took into account all the constraints at once. The minute you mix in a second source - be it another PEX repository, and --index or a --find-links repo, you force PEX to produce invalid results or else learn a full implementation of a resolve algorithm. The former is clearly not acceptable and the latter would pretty much fully negate the benefit of embedding Pip and delegating to it to do real resolves.

pex/resolver.py
jsirois: Hmm, on a re-read the error is unclear, should be:
```
Cannot resolve local projects from PEX repositories.  Asked to resolve {path} from {pex}.
```
At that point I stand behind the opinion above that the list of alternatives is undue handholding.

jsirois: Ok, thanks for making me look at the words more. I think this is an improvement that says everything the user needs to know: https://github.com/pantsbuild/pex/pull/1184

pex/bin/pex.py
jsirois: One way to look at it is Pex uniformly only supports a single Pip resolve. In a normal resolve this is done by Pip which happens to take in potentially multiple sources to run its singular logic against. In a `--pex-repository` resolve its done by Pex, but its not a real resolve, its just Pex picking the subset of the single resolve Pip did at some point in the past to build the PEX repository.

PR #1183, opened 2021-01-20
Suppress `pip debug` warning.

jsirois:
Pex accepts this warning in order to implement the `--pex-repository`
feature. User's should not have to see the warning since they are just
along for the ride.


jsirois:
In support of #1108.

benjyw (approved):

PR #1184, opened 2021-01-21
Improve PEX repository error for local projects.

jsirois:
Make it clear local projects can't be resolved from PEX repositories in
general. Previously you might take the error to mean the particular
local project was not resolvable from the particular PEX repository.


Eric-Arellano (approved):
Thanks! This is indeed an improvement, the original one left me wondering if it's because Pex never can do it vs. an error in the operation.

A tweak, take it or leave it:

> PEX cannot resolve local projects when using the option `--pex-repository`. You asked to resolve {path} from {pex}.

PR #1207, opened 2021-02-02
Fixup repo root detection code in tests.

jsirois:


jsirois:
#1205 & #1206 were running into issues detecting the repo root in tests under Python 2.7 on macOS only. This fixes that.

Eric-Arellano (approved):
Thanks for breaking this out

PR #1208, opened 2021-02-02
Replace Travis with GH Actions. 

jsirois:


jsirois:
The upshot is ~10 minutes overall faster than when we were paying Travis for extra shards: down from ~28 minutes to ~18 minutes.

jsirois:
N.B.: The obvious things, but you can't do them:
1. Use yaml achors to DRY - Github does not allow this.
2. Use env vars to DRY - this does not work for `runs-on` and at least one other field I forget ... `name`?
3. Use other actions from your own: There is no support for uses in action.yml

So this 1st land is verbose. I may be able to use a matrix to simplify, not sure yet. I'm loathe to introduce a script to generate the yml - I'd prefer something readable in one location over indirection.

.github/actions/run-tox/action.yml
```diff
@@ -0,0 +1,23 @@
...
+      run: |
+        python -mpip install -U pip
+        python -mpip install 'tox<=3.14.5' 'virtualenv<20.0.5'
+      shell: bash
+    - name: Execute tox -e ${{ inputs.tox-env }}
+      run: |
+        if [[ "$(uname -s)" == "Darwin" ]]; then
+          # Needed for Pillow to build successfully:
+          #   https://github.com/python-pillow/Pillow/issues/3438#issuecomment-543812237
+          export CPATH="$(xcrun --show-sdk-path)/usr/include"
```
Eric-Arellano, line 20: You <3 macOS, am I right? 😅 ...

Eric-Arellano (approved):
Thank you, John! Blog post sounds very justified.

> I'm loathe to introduce a script to generate the yml - I'd prefer something readable in one location over indirection.

That's fine to start without. I've personally found that the script has made iterating on Pants's CI much easier, but also Pants's full CI is 900 lines rather than 300.

jsirois:
> That's fine to start without. I've personally found that the script has made iterating on Pants's CI much easier,

Yeah, and I've found the opposite. The Pants CI scares me away as a casual user. There is a tradeoff to the indirection IMO.

.github/actions/run-tox/action.yml
jsirois: Yup. I was very happy to see Github changes 10x multiplier to use mac machines for private repos. Really make you think about why submit yourself to the torture if you're not developing desktop apps or phone apps. It simply makes no sense at all.

PR #1209, opened 2021-02-03
Leverage GH actions job matrix.

jsirois:
This allows DRYing up the CI config a good bit.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,58 +1,55 @@
 name: CI
 on: [push, pull_request]
 jobs:
-  format-check:
-    name: TOXENV=format-check
+  checks:
+    name:  TOXENV=${{ matrix.tox-env }}
     runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        include:
+          - check-name: Formatting
+            python-version: 3.9
+            tox-env: format-check
+          - check-name: Types
+            python-version: 3.9
+            tox-env: typecheck
+          - check-name: Vendoring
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -23,6 +24,12 @@ classifiers = [
   "Programming Language :: Python :: 3.7",
   "Programming Language :: Python :: 3.8",
   "Programming Language :: Python :: 3.9",
+  "Programming Language :: Python :: Implementation :: CPython",
+  "Programming Language :: Python :: Implementation :: PyPy",
+  "Topic :: Software Development :: Build Tools",
+  "Topic :: System :: Archiving :: Packaging",
+  "Topic :: System :: Software Distribution",
+  "Topic :: Utilities",
```
Eric-Arellano, line 32: Eh, this doesn't really fit into this PR. But fine with me if you want to save trees.

Eric-Arellano (approved):
Nice. Thanks, John.

Maybe tweak the title to include "and add Pypy3"? I'm thinking about looking at git history for `tox.ini` - this current title is misleading for why the file was changed. Alternatively, break out a dedicated PR.

pyproject.toml
jsirois: The PyPy totally does and was the impetous. Pex now has official PyPy support since it covers pypy 2 and 3. I'll break out this and the pypy3 changes to 1 PR. The trees are a big problem I'll continue to think on. There has to be some good way to put backpressure into the system that forces us to account for the damage.

PR #1213, opened 2021-02-04
Automate most of a Pex release.

jsirois:
This leaves out both CHANGES transcription to the Github Release body
and an update of the RELEASE.rst to reflect this new infra until after
any kinks are worked out.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,67 @@
+name: Release
+on:
+  push:
+    tags:
+      - v[0-9]+.[0-9]+.[0-9]+
+jobs:
+  pypi:
+    name: Publish sdist and wheel to PyPI
+    runs-on: ubuntu-20.04
+    environment: Release
+    steps:
+      - name: Checkout Pex
+        uses: actions/checkout@v2
+      - name: Setup Python 3.9
+        uses: actions/setup-python@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,67 @@
+name: Release
+on:
+  push:
+    tags:
+      - v[0-9]+.[0-9]+.[0-9]+
+jobs:
+  pypi:
+    name: Publish sdist and wheel to PyPI
+    runs-on: ubuntu-20.04
+    environment: Release
```
jsirois, line 10: N.B.: This both exposes the PYPI_{USERNAME ,PASSWORD} secrets, which are configured here: https://github.com/pantsbuild/pex/settings/environments, and requires a committer approve the release since I configured the `Release` environment to have required reviewers: https://docs.github.com/en/actions/reference/environments#required-reviewers.

We'll see what that's like and if it should stay.

Eric-Arellano (commented):
Could you please update release.rst with the new instructions?

jsirois:
> Could you please update release.rst with the new instructions?

NACK. See the description. I purposefully do not want to do that until the kinks are worked out. I have a feeling there will be thrash.

Eric-Arellano (approved):

PR #1214, opened 2021-02-04
Upgrade tox and cleanup config.

jsirois:
Newer tox supports environment expansion which helps DRY up the config.
Further remove unused and long broken environments and supporting
scripts.


Eric-Arellano (approved):
Nice

PR #1216, opened 2021-02-05
Fix run-tox action tox version.

jsirois:
This makes sure we install the tox version specified in tox.ini.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,7 +40,6 @@ jobs:
         with:
           tox-env: package
       - name: Determine Tag
-        id: determine-tag
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -40,7 +40,6 @@ jobs:
         with:
           tox-env: package
       - name: Determine Tag
-        id: determine-tag
```
jsirois, line 43: The release workflow uses `>> ${GITHUB_ENV}` + env vars to communicate between steps; so the id was not needed. That's only needed to refer to outputs as is done above in the run-tox action using the set-output workflow command.

jsirois:
I'm going to TBR this to get the release out and start tweaking the release automation if needed.

PR #1219, opened 2021-02-05
Make `PexInfo.pex_hash` calculation more robust. 

### LINKED ISSUE #1218: PexInfo.pex_hash does not include metadata
Generally a PEX file is uniqued by a combination of its user code and dependencies. Currently, this is what the `pex_hash` is comprised of:
https://github.com/pantsbuild/pex/blob/d205e2db713ce30d70cd2df9704468f8a3b3f9e0/pex/pex_builder.py#L508-L512

This scheme is foiled by PEX files with the same code and deps but different entry points (or any other metadata for that matter). This sort of thing comes up frequently in Pants with the BUILD file pattern:
```
pex_library()
pex_binary(...)
pex_binary(...)
```
Due to the vagaries of how `./pants run` happens to work today, all `pex_binary` targets here share the same code (none) and deps but each has a different entrypoint.

jsirois:
Previously the PexInfo metadata itself was left out of the calculation.
This could lead to `unzip` and `venv` mode collisions.

Add a previously failing test and fix.

Fixes #1218

### DIFF
--- a/pex/pex_builder.py
+++ b/pex/pex_builder.py
@@ -504,13 +504,7 @@ class PEXBuilder(object):
 
     def _prepare_code(self):
         self._pex_info.code_hash = CacheHelper.pex_code_hash(self._chroot.path())
-
-        hasher = hashlib.sha1()
-        hasher.update("code:{}".format(self._pex_info.code_hash).encode("utf-8"))
-        for location, sha in sorted(self._pex_info.distributions.items()):
-            hasher.update("{}:{}".format(location, sha).encode("utf-8"))
-        self._pex_info.pex_hash = hasher.hexdigest()
-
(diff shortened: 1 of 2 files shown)


Eric-Arellano (approved):
Thank you. To double check, we're making a reasonable gamble that this should be safe from a backwards compatibility perspective, right? That is, it's unlikely we'll get hash conflicts with the prior scheme.

jsirois:
> To double check, we're making a reasonable gamble that this should be safe from a backwards compatibility perspective, right? 

Right. Since this scheme includes more keys (all `PexInfo` keys instead of just code_hash + distributions) its impossible for the input to the sha1 to be the same.

> That is, it's unlikely we'll get hash conflicts with the prior scheme.

Roughly 1 chance in 2^160.

PR #1228, opened 2021-02-07
Upate the release docs.

jsirois:


Eric-Arellano (commented):
> Bump the version in pex/version.py and update CHANGES.rst in a local commit:

The "local" isn't very clear to me. This is still the normal fork PR workflow, no? If so, how about:

> Bump the version in pex/version.py and update CHANGES.rst. Open a PR with these changes and merge to the `master` branch.

jsirois:
> The "local" isn't very clear to me. This is still the normal fork PR workflow, no?

No. The doc was from the days when you just did this as a local commit exactly as it says, then tagged, then pushed to origin.
I'll update though since we've been recently in the habit of dedicating a PR to this.

Eric-Arellano (approved):

PR #1234, opened 2021-02-16
Convert most of Pex to attrs.

jsirois:
This leaves aside pex.third_party and pex.vendor since there are
circularity issues with attempting to use pex.third_party.attr from
either as both modules work to provide the pex.third_party importer.

The namedtuple use in pex.common is also left aside. Currently the
module is caught up in pex.third_party importer bootstrapping, but that
could be fixed. More critically, pex.common is used in the
latency-sensitive pex bootstrapping code where pex.third_party imports
need to be avoided for quick re-exec handoffs.


jsirois:
The 1st commit, 0224664, is being reviewed in #1233. The rest of the commits should be reviewed here.

jsirois:
This adds 99K to the 2.9M Pex PEX.

pex/resolver.py
```diff
@@ -314,14 +276,8 @@ def from_request(
...
-    def __init__(
-        self,
-        request,  # type: BuildRequest
-        atomic_dir,  # type: AtomicDirectory
-    ):
-        # type: (...) -> None
-        self.request = request
-        self._atomic_dir = atomic_dir
+    request = attr.ib()  # type: BuildRequest
+    _atomic_dir = attr.ib()  # type: AtomicDirectory
```
jsirois, line 280: N.B.: attrs is smart enough to make an underscored attribute available by its non-underscored name as a kwarg. Publically constructible via kwarg, "private thereafter".

pex/platforms.py
```diff
@@ -108,41 +92,34 @@ def from_tag(cls, tag):
...
-    @property
-    def abi(self):
-        # type: () -> str
-        return cast(str, super(Platform, self).abi)
+    def __attrs_post_init__(self):
+        # type: () -> None
+        if self.impl == "cp" and not self.abi.startswith(self.interpreter):
+            # N.B. This permits CPython users to pass in simpler extended platform
+            # strings like `linux-x86_64-cp-27-mu` vs e.g. `linux-x86_64-cp-27-cp27mu`.
+            object.__setattr__(self, "abi", self.interpreter + self.abi)
```
jsirois, line 117: N.B.: `object.__setattr__` in `__attrs_post_init__` is a native way (in their docs) to do frozen after init in attrs.

jsirois:
The most convenient docs are the API reference here: https://www.attrs.org/en/stable/api.html

jsirois:
Now re-based against #1233 which has landed on master.

Eric-Arellano (approved):
Excellent. Good idea!

pex/requirements.py
```diff
@@ -587,25 +519,29 @@ def _parse_requirement_line(
...
     project_name_extras_and_marker = _try_parse_pip_local_formats(
         local_requirement, basepath=basepath
     )
     maybe_abs_path, extras, marker = (
-        project_name_extras_and_marker
+        (
+            project_name_extras_and_marker.project_name,
+            project_name_extras_and_marker.extras,
+            project_name_extras_and_marker.marker,
+        )
```
cosmicexplorer, line 534:
```suggestion
        project_name_extras_and_marker.astuple()
```
It looks like we do this `xxx, extras, marker = (... if ... else ...)` deconstruction at least two times in this file, but I think that the `ProjectNameExtrasAndMarker.astuple()` method exists for exactly this reason.

pex/requirements.py
```diff
@@ -560,7 +488,11 @@ def _parse_requirement_line(
             parsed_url.fragment
         )
         project_name, extras, marker = (
-            project_name_extras_and_marker
+            (
+                project_name_extras_and_marker.project_name,
+                project_name_extras_and_marker.extras,
+                project_name_extras_and_marker.marker,
+            )
```
cosmicexplorer, line 495:
```suggestion
            project_name_extras_and_marker.astuple()
```
Note that in this one we want `extras` to default to `None` instead of `()`, so we would have to modify `astuple` to allow overrides e.g. `project_name_extras_and_marker.astuple(extras=None)`. That's unless `extras` is correct as *either* `()` or `None` here. But `extras` differs from the other change in this file in that way.

cosmicexplorer (commented):
I'm very glad I saw this PR when I did! Going to use this to convince [spack](https://github.com/spack/spack) folks to pick up the `attrs` lib!

If this PR is very time-sensitive, consider these comments non-blocking, and I'll post followup issues/PRs for them. If there's time for discussion, consider these comments as "normal" (aka feel free to discard/override at your discretion).

pex/platforms.py
cosmicexplorer: This conversation can be resolved at your discretion.

It looks like you left this explicit comment https://github.com/pantsbuild/pex/pull/1234#issuecomment-780024126 pointing to the API reference for `attrs`, so it feels like your choice to avoid it in the code itself was intentional. `attrs` is canonical enough of a library that I think it's relatively straightforward to later readers to web search for `attrs` directly, which would allow them to understand the rationale for using this method. I would still have left a comment in the code saying:

```python
            # N.B.: object.__setattr__ in __attrs_post_init__ is a native way (in their docs at https://www.attrs.org/en/stable/api.html) to do frozen after init in attrs.
```

but I would not have any difficulty obtaining context later if you chose otherwise.

pex/network_configuration.py
```diff
@@ -4,46 +4,44 @@
...
+    """
+
+    retries = attr.ib(default=5)  # type: int
+    timeout = attr.ib(default=15)  # type: float
+    proxy = attr.ib(default=None)  # type: Optional[str]
+    cert = attr.ib(default=None)  # type: Optional[str]
+    client_cert = attr.ib(default=None)  # type: Optional[str]
+
+    @retries.validator
+    def _validate_retries(self, attribute, value):
```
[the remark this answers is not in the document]
cosmicexplorer: (can be resolved at will)

pex/pip.py
```diff
@@ -126,7 +126,7 @@ def _calculate_env(
 
         if network_configuration.client_cert:
             assert not isolated
-            yield "PIP_CLIENT_CERT", network_configuration.cert
+            yield "PIP_CLIENT_CERT", network_configuration.client_cert
```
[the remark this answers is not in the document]
cosmicexplorer: (this conversation can be resolved if you're comfortable with it, but i'd like to know ask @Eric-Arellano whether he would find this particular proposed change useful)

pex/requirements.py
cosmicexplorer: (this convo can be resolved at your discretion, but I'm pretty confident this change would make the result more readable. i would like to either use `.astuple()` here, or remove the awkward `if ... else ...` construct altogether)

pex/requirements.py
cosmicexplorer: (see https://github.com/pantsbuild/pex/pull/1234#discussion_r577298669 -- i'm pretty confident this is a useful change, but please apply discretion)

pex/requirements.py
jsirois: It does but it's only typed to Tuple[Any].

pex/platforms.py
jsirois: Yeah - I view that sort of thing as babying readers. If you adopt a library you should not regurgitate its docs. You should learn the library. I'm a bit more OK with babying reviewers on the 1st intro of the library - thus the PR comment.

pex/pip.py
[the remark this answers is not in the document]
jsirois: That seems over the top. If you wanted newtypes you could just `class CABundle(str): ...` right?

tests/test_pex.py
```diff
@@ -35,6 +34,7 @@
     temporary_content,
     write_simple_pex,
 )
+from pex.third_party.pkg_resources import Distribution
```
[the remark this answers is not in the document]
jsirois: I'm not sure how wise that is to do. When we drop support for python 2.7 it'll come back out as a normal import.

pex/pip.py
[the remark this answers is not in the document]
jsirois: Since this is the one and only consumer of cert / client_cert, I think even newtypes are over the top. I agree though that if this were exposed to users in an API a newtype would pull its weight and potentially prevent N flip-flops instead of the only possible one, which is right here.

pex/requirements.py
jsirois: Yeah, good point. extras is in fact correct as either () or None since both the receiver at line 516 and the receiver at line 544 take `Optional[Iterable[str]]`. I had already been down this road as is shown by the fact I forgot to delete the currently unused typed astuple method here:
https://github.com/pantsbuild/pex/blob/bd542066f827676ede273a296d12b508a79c8ff0/pex/requirements.py#L343-L345

I'll just use that method, although the reason I didn't was I thought it didn't pull its weight and this repitition was more clear.

pex/requirements.py
jsirois: Aha - I did not read this comment well. I glossed an assumed you meant the attrs API `astuple`: `attr.astuple(attrs_class_instance)`.  That is unfortunately typed to Tuple[Any] with the standard typeshed and that is where I started with all this. It took quite a bit of thrashing with mypy to learn this was the issue. That led to the explicit tuple style here then to writing my own astuple method that was properly typed, then reconsidering how smart that was but forgetting to delete my custom astuple.

jsirois: Fixed.

pex/requirements.py
jsirois: Fixed.

PR #1238, opened 2021-02-17
Fix venv pex script for use with multiprocessing.

### LINKED ISSUE #1236: The `--venv` mode `pex` script does not have a __name__ == '__main__' gaurd breaking multiprocessing.
The multiprocessing module relies on being able to import the file run as `__main__` from spawned subprocesses without the side-effect of actually running `__main__` code. See "Safe importing of main module" here: https://docs.python.org/3/library/multiprocessing.html#the-spawn-and-forkserver-start-methods

To play well with multiprocessing, the `pex` script should get a `__main__` guard.

This was spotted over in #1235.

jsirois:
Add a test that passes for multiprocessing fork mode but fails for both
spawn and forkserver modes and fix the pex script to gaurd against non
`__name__ == '__main__'` mode to fix.

Fixes #1236

### DIFF
--- a/pex/tools/commands/venv.py
+++ b/pex/tools/commands/venv.py
@@ -140,108 +140,109 @@ def populate_venv_with_pex(
         """\
         #!{venv_python} -sE
 
-        import os
-        import sys
+        if __name__ == "__main__":
+            import os
+            import sys
 
-        python = {venv_python!r}
-        if sys.executable != python:
-            sys.stderr.write("Re-execing from {{}}\\n".format(sys.executable))
-            os.execv(python, [python, "-sE"] + sys.argv)
+            python = {venv_python!r}
+            if sys.executable != python:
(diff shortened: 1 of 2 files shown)


pex/tools/commands/venv.py
```diff
@@ -140,108 +140,109 @@ def record_provenance(src_to_dst):
...
-        bin_path = os.environ.get("PEX_VENV_BIN_PATH", {bin_path!r})
-        if bin_path != "false":
-            PATH = os.environ.get("PATH", "").split(os.pathsep)
-            if bin_path == "prepend":
-                PATH.insert(0, bin_dir)
-            elif bin_path == "append":
-                PATH.append(bin_dir)
-            else:
+        if __name__ == "__main__":
+            import os
```
jsirois, line 144: From here down was just an indent - GitHub diffing is poor here.

Eric-Arellano (approved):
Thank you, John!

PR #1240, opened 2021-02-18
Ensure PEX files aren't symlinked in venv.

### LINKED ISSUE #1239: The `--seed` mode for a `--venv` PEX is unsafe.
Currently, `--seed`ing of a `--venv` PEX operates on the PEXBuilder chroot and not the final PEX zip:
https://github.com/pantsbuild/pex/blob/37c9ff0fa815ac1f03a73d3f2e33fb20352c8843/pex/bin/pex.py#L1072-L1085
https://github.com/pantsbuild/pex/blob/37c9ff0fa815ac1f03a73d3f2e33fb20352c8843/pex/bin/pex.py#L1124-L1127

Since the PEXBuilder in question is setup to use symlinks:
https://github.com/pantsbuild/pex/blob/37c9ff0fa815ac1f03a73d3f2e33fb20352c8843/pex/bin/pex.py#L869-L875

That means any loose sources and resources added to the PEXBuilder are copied straight from their symlinks:
https://github.com/pantsbuild/pex/blob/37c9ff0fa815ac1f03a73d3f2e33fb20352c8843/pex/tools/commands/venv.py#L103-L109
https://github.com/pantsbuild/pex/blob/37c9ff0fa815ac1f03a73d3f2e33fb20352c8843/pex/tools/commands/venv.py#L56-L64

On linux, when the symlink is on the same filesystem as the PEX_ROOT, os.link is used to make the copy and os.link on a symlink makes a new symlink. If the target of the symlink is later deleted; then the symlink in the seeded venv is left dangling.

jsirois:
Ensure PEX file contents are materialized in venvs as durable files.
Previously, distributions were handled correctly but PEX sources and
resources were not. A test is added that exercised the bug but is now
fixed.

Fixes #1239

### DIFF
--- a/tests/tools/commands/test_venv.py
+++ b/tests/tools/commands/test_venv.py
@@ -5,7 +5,9 @@ from __future__ import absolute_import
 
 import multiprocessing
 import os
+import shutil
 import subprocess
+import sys
 import tempfile
 from subprocess import CalledProcessError
 from textwrap import dedent
@@ -14,6 +16,7 @@ import pytest
 
 from pex.common import safe_open, temporary_dir, touch
 from pex.executor import Executor
+from pex.pex_builder import CopyMode, PEXBuilder
 from pex.testing import run_pex_command
 from pex.tools.commands.virtualenv import Virtualenv
 from pex.typing import TYPE_CHECKING, cast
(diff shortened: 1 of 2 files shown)


tests/tools/commands/test_venv.py
```diff
@@ -341,3 +344,25 @@ def bar():
...
+    pex_builder.build(pex_file, bytecode_compile=False)
+    assert 42 == subprocess.Popen(args=[pex_file]).wait()
+
+    venv = os.path.join(str(tmpdir), "a.venv")
+    subprocess.check_call(
+        args=[sys.executable, "-m", "pex.tools", pex_builder.path(), "venv", venv]
+    )
+    venv_pex = os.path.join(venv, "pex")
+    shutil.rmtree(src)
+    assert 42 == subprocess.Popen(args=[venv_pex]).wait()
```
jsirois, line 368: This was the failing line prior to the fix. The main module was not found.

Eric-Arellano (approved):
Great, thank you.

PR #1242, opened 2021-02-19
The `pex` venv script handles entrypoints like PEX.

### LINKED ISSUE #1241: The venv `pex` script handles entrypoint functions differently from PEX.
Currently `PEX` uses the return value of the entry point function:
https://github.com/pantsbuild/pex/blob/2df133378a4f9e8bb3c6966d824465e57c66ba72/pex/pex.py#L476-L479
https://github.com/pantsbuild/pex/blob/2df133378a4f9e8bb3c6966d824465e57c66ba72/pex/pex.py#L645-L651

Which was introduced here:
https://github.com/pantsbuild/pex/commit/eac023361fb5e524f2f31916dd186621cec46d6d#diff-07a90ebaf117fe2a7d1010d8dbe6ed42fab40761f07622b3dba999b2e2dbce78R349-R351

And the `pex` venv script does not:
https://github.com/pantsbuild/pex/blob/2df133378a4f9e8bb3c6966d824465e57c66ba72/pex/tools/commands/venv.py#L236-L248

After consideration it seems sticking with the change in #605 and bringing the `pex` venv script in line with the `PEX` behavior is the best way forward. Existing users expecting this behavior won't be broken this way and the surprisingly (to me) flexible behavior of [`sys.exit`](https://docs.python.org/2.7/library/sys.html#sys.exit) makes this work like you want in most imaginable scenarios (clearly by design on the part of stdlib authors). The behavior though should be better documented.

jsirois:
Previously behavior diverged for entrypoint functions returning a value.
Now both PEXes and the `pex` venv script behave the same and the
behavior is documented and tested.

Fixes #1241

### DIFF
--- a/docs/buildingpex.rst
+++ b/docs/buildingpex.rst
@@ -186,17 +186,18 @@ get pydoc help on the ``flask.app`` package in Flask:
 
 and so forth.
 
-Entry points can also take the form ``package:target``, such as ``sphinx:main`` or ``fabric.main:main`` for Sphinx
-and Fabric respectively.  This is roughly equivalent to running a script that does ``from package import target; target()``.
+Entry points can also take the form ``package:target``, such as ``sphinx:main`` or
+``fabric.main:main`` for Sphinx and Fabric respectively.  This is roughly equivalent to running a
+script that does ``import sys, from package import target; sys.exit(target())``.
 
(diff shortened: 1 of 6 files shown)


Eric-Arellano (approved):
Thank you, this sounds good.

PR #1244, opened 2021-02-20
Fix `--runtime-pex-root` leak in pex bootstrap.

jsirois:
When PEX bootstrap involved finding an interpreter, the
`--runtime-pex-root` was not propagated and interpreter identification
files would leak into the default `PEX_ROOT`.

Add a failing test that is fixed by proper PEX_ROOT propagation during
PEX bootstrapping.


jsirois:
After the weekend bump.

Eric-Arellano (approved):

PR #1245, opened 2021-02-22
Support venvs that can outlive their base python.

### LINKED ISSUE #1230: Support control of venv creation mode `--copies` vs. `--symlinks`
Currently venvs created by Pex use the default creation mode of the venv or virtualenv module depending on which is used. The venv module supports forcing `--copies` or `--symlinks` for Python3.5+ and the virtualenv module we vendor supports `--always-copy` (there is no correspnding `--always-symlink`).

jsirois:
Add support for creating venvs using copies instead of symlinks.

Fixes #1230

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -372,6 +372,17 @@ def configure_clp_pex_options(parser):
         "multiple times under a stable runtime PEX_ROOT, the venv creation will only be done once "
         "and subsequent runs will enjoy lower startup latency.",
     )
+    group.add_argument(
+        "--venv-copies",
+        "--no-venv-copies",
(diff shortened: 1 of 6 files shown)


jsirois:
In combination with #1246 and #1247 this change allows for Pants use of installed venvs. Its infeasible to actually use these today due to poor performance of materializing a full venv in a Pants process chroot, but this is generically useful Pex functionality and Pants will solve its perf problem here one day.

jsirois:
Note that in the Pants case, this change solves problems with storing directories with abosute symlinks - which standard venvs are.

Eric-Arellano (approved):

PR #1246, opened 2021-02-22
Support pre-compiling a venv.

jsirois:
This is useful when using Pex to install a system wide application to a
location where not all users will have permission to create bytecode
files.

### DIFF
--- a/pex/tools/commands/venv.py
+++ b/pex/tools/commands/venv.py
@@ -315,6 +315,12 @@ class Venv(Command):
             default=False,
             help="Create the venv using copies of system files instead of symlinks",
         )
+        parser.add_argument(
+            "--compile",
+            action="store_true",
+            default=False,
+            help="Compile all `.py` files in the venv.",
+        )
 
     def run(
         self,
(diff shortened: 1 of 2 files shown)


jsirois:
In combination with #1245 and #1247 this change allows for Pants use of installed venvs. Its infeasible to actually use these today due to poor performance of materializing a full venv in a Pants process chroot, but this is generically useful Pex functionality and Pants will solve its perf problem here one day.

jsirois:
Note that in the Pants case this is a performance optimization. Its dwarfed by the slowness of materializing the full venv in a Pants process chroot, but will be important once that perf issue is solved.

Eric-Arellano (approved):
This will run into similar issues as https://github.com/pantsbuild/pants/issues/11339 in terms of breaking byte-for-byte reproducibility, right? This sounds sensible for Pex to have either way, but wondering if this will be safe to use as-is with Pants.

jsirois:
> but wondering if this will be safe to use as-is with Pants

Safe - I think yes. As efficient as possible - No.

IIUC the issue with byte for byte reproducibility is only in avoiding un-necessary work. If you change a comment in code that goes into a venv, then build a venv with pre-compiled code - that precompiled code will be the same .... mod a timestamp. So when you don't fix the compilation timestamp you suffer downstream rules re-running when they need not have.

jsirois:
Tracking a stable timestamp mode as part of an umbrella to audit Pex artifact reproducibility in #1249.

PR #1247, opened 2021-02-22
Support venv relocation.

jsirois:
Venv scripts aren't relocatable without edits since their shebangs
record the absolute path of the venv interpreter. However, they should
work when invoked indirectly by the moved venv's interpreter. This
already worked for venv console scripts but it did not work for the
`pex` venv script, which re-execs when invoked with an unexpected
interpreter.

Add a relocation test that fails and fix the `pex` venv script to be
more flexible in its venv interpreter identification.

### DIFF
--- a/pex/tools/commands/venv.py
+++ b/pex/tools/commands/venv.py
@@ -147,22 +147,23 @@ def populate_venv_with_pex(
             import os
             import sys
 
-            python = {venv_python!r}
+            venv_dir = os.path.abspath(os.path.dirname(__file__))
+            venv_bin_dir = os.path.join(venv_dir, "bin")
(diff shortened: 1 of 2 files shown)


jsirois:
In combination with #1245 and #1246 this change allows for Pants use of installed venvs. Its infeasible to actually use these today due to poor performance of materializing a full venv in a Pants process chroot, but this is generically useful Pex functionality and Pants will solve its perf problem here one day.

jsirois:
Note that in the Pants case this solves the issue that a venv is created and stored in one process execution chroot directory, but ~never again materialized to a process execution chroot directory with the same path.

Eric-Arellano (approved):

PR #1250, opened 2021-02-22
Fix Pex isolation.

### LINKED ISSUE #1232: When Pex is run from a Pex PEX its isolation is broken.
As noted here: https://github.com/pantsbuild/pex/commit/4562d7d386a56095612b6f7f9b8eb8261e45c1af#r47156419

jsirois:
Previously isolation was only correct for Pex when it was loaded from a
distribution installed loose on a filesystem. Now isolation works when
Pex is loaded from a PEX zipfile as well.

Fixes #1232

### DIFF
--- a/pex/third_party/__init__.py
+++ b/pex/third_party/__init__.py
@@ -375,7 +375,18 @@ def isolated():
 
         pex_path = os.path.join(vendor.VendorSpec.ROOT, "pex")
         with _tracer().timed("Hashing pex"):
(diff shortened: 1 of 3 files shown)


jsirois:
Alot easier to fix than to test as will be obvious from the test commit.

jsirois:
The relevant backstory is in #881

benjyw (approved):

PR #1254, opened 2021-02-25
Guard against too long `--venv` mode shebangs.

### LINKED ISSUE #1252: Venv `pex` and bin scripts can run afoul of shebang length limits.
These limits can be as low as 128 bytes for the whole shebang line on Linux depending on how the kernel is configured: https://www.in-ulm.de/~mascheck/various/shebang/.

On my Arch Linux machine I find:
```
$ grep BINPRM_BUF_SIZE /usr/include/linux/binfmts.h 
#define BINPRM_BUF_SIZE 256
``` 

But on Debian 10.7 Buster you get:
```
$ docker run --rm -it debian:10.7
root@f5735d61a8c7:/# apt update && apt install --yes linux-source
root@f5735d61a8c7:/# grep BINPRM_BUF_SIZE /usr/include/linux/binfmts.h
#define BINPRM_BUF_SIZE 128
```

jsirois:
The default path length to a `--venv` mode PEX python used in shebangs
for the venv `pex` script and venv console scripts is 102 characters
starting from the PEX_ROOT. Add to that 2 characters for '#!' and 4 for
the trailing ' -sE' we add and that leaves just 20 characters for the
path to the PEX_ROOT (e.g.: /home/jane/.pex) before these scripts start
failing on machines where the shebang length limit is 128 bytes, which
is not uncommon amongst Linux distributions.

Fix this for forseeable cases by keeping the existing cache structure,
which has a primary key path component of the pex hash which is cheap to
compute (pre-computed), convenient for debugging and ~guaranteed unique,
and introducing a "jump table" of venv dir symlinks in
PEX_ROOT/venvs/short. This jump table is computed once at venv
materialization time with a much shorter hash that is checked for
collisions and expanded as needed.

Fixes #1252

### DIFF
--- a/pex/pex_bootstrapper.py
+++ b/pex/pex_bootstrapper.py
@@ -3,11 +3,13 @@
 
 from __future__ import absolute_import
 
+import errno
+import hashlib
 import os
 import sys
 
 from pex import pex_warnings
-from pex.common import atomic_directory, die
+from pex.common import atomic_directory, die, pluralize, safe_mkdir
 from pex.inherit_path import InheritPath
 from pex.interpreter import PythonInterpreter
 from pex.interpreter_constraints import UnsatisfiableInterpreterConstraintsError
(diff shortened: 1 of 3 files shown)


jsirois:
The savings here is 67 characters. For a new typical shebang length of ~60 characters.

Eric-Arellano (approved):
Excellent fix, thanks John!

PR #1258, opened 2021-03-03
Use pytest-xdist to speed up ITs.

jsirois:


jsirois:
The worst improvement appears to be ~50%.

Eric-Arellano (approved):

Issue #1249, opened 2021-02-22
Audit Pex reproducibility.


jsirois:
Check that Pex can reproducibly create identical artifacts:
+ [ ] PEX files
+ [ ] Venvs

This will be a combination of hardening existing tests and code for the PEX file case (see #716 and #734) as well as adding a new option and code for the venv case (and maybe the unzip case).

PR #1259, opened 2021-03-03
Fix PEX file reproducibility.

jsirois:
Two sources of non-reproducibility remained:
1. When zipping a PEX file, symlinked directory walks were not sorted.
2. When installing wheels in chroots, Pip direct_url.json retained a
   system-specific path that varied as PEX_ROOT varied.

The first is fixed with an explicit sort and the second is fixed by
simply deleting problematic direct_url.json files post-install.

Concludes the #1249 audit for PEX files.

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -7,6 +7,7 @@ import errno
 import filecmp
 import functools
 import glob
+import itertools
 import json
 import multiprocessing
 import os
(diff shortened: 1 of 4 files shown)


jsirois:
This uses the test modification made by @Eric-Arellano over in #734 mostly wholesale.

tests/test_integration.py
```diff
@@ -1763,37 +1764,54 @@ def test_issues_539_abi3_resolution():
...
+            for index in range(3):
+                path = os.path.join(td, "{}.pex".format(index))
+                paths.append(path)
+                # We sleep to ensure that there is no non-reproducibility from timestamps or
+                # anything that may depend on the system time. Note that we must sleep for at least
+                # 2 seconds, because the zip format uses 2 second precision per section 4.4.6 of
+                # https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT.
+                if index > 0:
+                    safe_sleep(2)
+                create_pex(path, seed=(index * 497) + 4)
```
jsirois, line 1787: This seemed and I assume was mostly arbitrary; however, switching to just using the index masked a bug that only surfaced in Python 3.5 over in Chroot.zip now fixed in this PR.

Eric-Arellano (approved):
Excellent! Thoughts on a test for this direct_url.json rewrite?

jsirois:
> Excellent! Thoughts on a test for this direct_url.json rewrite?

There is no rewrite - its just a delete if the contained "url" if "file://...". The test you wrote does not pass without this change, so its tested.

PR #1264, opened 2021-03-08
Turn on xdist for unit tests.

jsirois:
This required one workaround for pkg_resources.


jsirois:
Hrm, apparently flaky still. Retracting.

PR #1265, opened 2021-03-08
Support console scripts found in the PEX_PATH.

### LINKED ISSUE #1257: Support console scripts found in the PEX_PATH.
Right now, when building up a PEX with PEXBuilder, only the immediate distributions added to the PEXBuilder are searched for the console script entry point. Pants could use this to support a few binaries it assembles via PEX_PATH to avoid various invalidations.

jsirois:
Include the distributions found in any PEXes present on the PEX_PATH in
the search for console scripts at PEX build time.

Fixes #1257

### DIFF
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -78,7 +78,7 @@ class PEX(object):  # noqa: T000
         self._pex_info = PexInfo.from_pex(self._pex)
         self._pex_info_overrides = PexInfo.from_env(env=env)
         self._vars = env
-        self._envs = []  # type: List[PEXEnvironment]
+        self._envs = None  # type: Optional[Iterable[PEXEnvironment]]
         self._activated_dists = None  # type: Optional[Iterable[Distribution]]
         if verify_entry_point:
             self._do_entry_point_verification()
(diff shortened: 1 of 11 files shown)


pex/pex.py
```diff
@@ -95,27 +95,38 @@ def interpreter(self):
         # type: () -> PythonInterpreter
         return self._interpreter
 
+    @property
```
Eric-Arellano, line 98: Thoughts on adding `@memoized_property` to Pex? This code will run twice iiuc.

tests/test_pex_builder.py
```diff
@@ -266,3 +267,24 @@ def verify(pb):
             pb3.build(target)
             pb4 = build_from_req_pex(td4, target)
         verify(pb4)
+
+
+def test_prex_builder_script_from_pex_path(tmpdir):
```
Eric-Arellano, line 272:
```suggestion
def test_pex_builder_script_from_pex_path(tmpdir):
```

Eric-Arellano (approved):

pex/pex.py
jsirois: It caches manually in `self._envs` variable using the `if self._envs is None` guard.

tests/test_pex_builder.py
jsirois: Thanks - you know my answer there - I'll save some trees.

PR #1268, opened 2021-03-12
Ensure Pip logs to stderr.

### LINKED ISSUE #1267: Pex verbose log output partially goes to stdout.
When Pex logging is turned up to verbosity level 3 or higher, Pip verbose logging is engaged and some of Pip's logging goes to stdout. Since Pip views the mixed stdout / stderr logging discipline as a feature and not a bug (C.F.: https://github.com/pypa/pip/issues/9420), Pex should deal with this problem on its side.

jsirois:
Fixes #1267


stuhood (approved):
Thanks!

PR #1273, opened 2021-03-16
allow cli arguments to be specified in a file

### LINKED ISSUE #1271: Allow command-line arguments to be read from a file
Various tools, including [gcc](https://gcc.gnu.org/onlinedocs/gcc/Overall-Options.html#Overall-Options) and [clang](https://llvm.org/docs/CommandLine.html#response-files), allow command-line arguments to be read from a file.  This is typically done by designating a special character (@) to indicate that the argument is to be interpreted as a file of arguments.  This behavior is helpful when the command-line length may become greater than the maximum size allowed by the system.

Would there be interest in adding the ability to pass command-line argument from a file to pex?  I would be happy to submit a pull request with the needed changes if there is interest. I suspect the needed changes would be minimal as ArgumentParser supports this behavior with its [`fromfile_prefix_chars`](https://docs.python.org/3/library/argparse.html#fromfile-prefix-chars) parameter.  

I know I occasionally have the need to pass a large number of resource directories to the pex command and worry about reaching the limits of my systems command-line length.

jjhelmus:
Allow command line arguments to be specified in a file which is passed
on the command line with an @ symbol before the filename.

Add a corresponding tests for this behavior which verifies that two
requirements files can be passed in this manner.

closes #1271

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -637,7 +637,11 @@ def configure_clp():
         "sources, requirements, their dependencies and other options."
     )
 
-    parser = ArgumentParser(usage=usage, formatter_class=ArgumentDefaultsHelpFormatter)
+    parser = ArgumentParser(
+        usage=usage,
+        formatter_class=ArgumentDefaultsHelpFormatter,
+        fromfile_prefix_chars="@",
+    )
 
     parser.add_argument("--version", action="version", version=__version__)
     parser.add_argument("requirements", nargs="*", help="Requirements to add to the pex")
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -635,6 +635,9 @@ def configure_clp():
         "%(prog)s [-o OUTPUT.PEX] [options] [-- arg1 arg2 ...]\n\n"
         "%(prog)s builds a PEX (Python Executable) file based on the given specifications: "
         "sources, requirements, their dependencies and other options."
+        "\n"
+        "Command-line options can be provided in one or more files by prefixing the filenames "
+        "with an @ symbol. These files must contain one argument per line."
     )
 
     parser = ArgumentParser(
--- a/tests/test_pex_binary.py
+++ b/tests/test_pex_binary.py
@@ -124,6 +124,17 @@ def test_clp_constraints_txt():
     assert options.constraint_files == ["requirements1.txt"]
 
 
+def test_clp_arg_file():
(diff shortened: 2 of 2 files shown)


jjhelmus:
Adding something to the help text mentioning that arguments can be passed by a file prefixed with `@` seems like a good idea but I was not sure the best location to put this text.

pex/bin/pex.py
```diff
@@ -637,7 +637,11 @@ def configure_clp():
         "sources, requirements, their dependencies and other options."
```
jsirois, line 637: Right here is probably the best place to mention the @ handling. If you want to add in some newlines so the mention gets its own paragraph - that works for me.

jsirois (approved):
Thanks @ jjhelmus! I'll merge after the `@` mention is added.

jsirois:
The `(macos-10.15) TOXENV=py39-integration` shard is apparently hung - ~2:40:00 runtime so far and no logs from the test step, I'll cancel / restart the workflow.

jsirois:
@jjhelmus - I've circled back around here because I've found a use for `@` in a particular option's values (e.g.: `--scie-busybox @,!@ansible`). That trips up the existing from file integration you added. Pex never breaks backward compatibility, so I won't get rid of from file support, but ... I dug in here and it seems I can preserve support with different corner case semantics. This is based on the following assumptions:
+ You never pass `--real-option-that-takes-a-value @fromfile`
+ You don't care about the ordering guaranty Argparse from file support gives in terms of where the arguments from files are inserted in the arg list

Do those assumptions fix how you use this feature?

I've tested out a 2-pass parse where I 1st parse without from file support, in which case any @fromfile values are interpreted as requirements (the positional args `pex` accepts). I filter the requirements collecting the @fromfile values to form an extra args list. I then argparse a second time with those extra args and finally merge the resulting options with the 1st parsed options, extending any list-valued options and over-writing the rest. This is definitely a breaking change depending on use, but I'm not seeing how this would break a reasonable use.

Thanks for your time.

benjyw:
FWIW, in Pants you can escape a literal `@` with a second `@`: `@@valuebeginningwithampersand`. This assumes that an at-file won't have `@` as its leading character, which seems extremely reasonable to assume. So maybe add that?

jsirois:
That would defeat the point of option values that use an @ syntax. You could never actually write @, you'd always have to write @@.

PR #1274, opened 2021-03-16
Fix module entrypoints.

### LINKED ISSUE #1018: Issue when running a module inside pex file
I met some strange behaviors when running a module inside pex file. I found a simple reproducible example as below.

Python: 3.8.3
Pex: 2.1.15

I have a `test.py` under a folder `src`,
```
[cwd]
└── src
    └── test.py
```

`test.py`:
```python
import pickle

def add(a: int, b: int) -> int:
    return a+b

def main() -> None:

    pickle_add = pickle.dumps(add)
    add_clone = pickle.loads(pickle_add)
    print(add_clone(1, 2))

if __name__ == "__main__":
    main()
```

To build a pex file with entry point:
    
    pex -D ./src -e test:main -o test_with_ep.pex

To build a pex file without entry point:

    pex -D ./src -o test_no_ep.pex

No issue when running:

    python test_with_ep.pex      
    3

But when running `test_no_ep.pex`:

    python test_no_ep.pex -m test
      File "/private/tmp/test_no_ep.pex/.bootstrap/pex/pex.py", line 423, in execute
      File "/private/tmp/test_no_ep.pex/.bootstrap/pex/pex.py", line 356, in _wrap_coverage
      File "/private/tmp/test_no_ep.pex/.bootstrap/pex/pex.py", line 387, in _wrap_profiling
      File "/private/tmp/test_no_ep.pex/.bootstrap/pex/pex.py", line 477, in _execute
      File "/private/tmp/test_no_ep.pex/.bootstrap/pex/pex.py", line 520, in execute_interpreter
      File "/private/tmp/test_no_ep.pex/.bootstrap/pex/pex.py", line 584, in execute_module
      File "/Users/yuhan/.pyenv/versions/3.8.3/lib/python3.8/runpy.py", line 210, in run_module
        return _run_code(code, {}, init_globals, run_name, mod_spec)
      File "/Users/yuhan/.pyenv/versions/3.8.3/lib/python3.8/runpy.py", line 87, in _run_code
        exec(code, run_globals)
      File "/private/tmp/test_no_ep.pex/test.py", line 13, in <module>
      File "/private/tmp/test_no_ep.pex/test.py", line 8, in main
    _pickle.PicklingError: Can't pickle <function add at 0x1042aef70>: attribute lookup add on __main__ failed

Problem could be solved if I move that function `main()` into another file and import it in `test.py`. Then I could run that module in pex file.

Do I miss something here ?

jsirois:
Previously we did not invoke module entrypoints identically to Python
which could lead to unexpected errors using the pickle module.

Now we match Python for all cases except a PEX in zipapp mode with a
module entrypoint. This is a case that is also not handled by Python
itself since its zipapp module requires a function entrypoint; ie
mod:func.

Fixes #1018

### DIFF
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -660,6 +660,14 @@ class PEX(object):  # noqa: T000
         if ":" in entry_point:
             return self.execute_pkg_resources(entry_point)
 
+        # When running as a zipapp we can't usefully `alter_sys` to point to the module we're
+        # executing as argv[0] since that module will be contained in a zipfile. In other words, if
+        # we did do this, sys.argv[0] would be `/path/to/pex.zip/path/to/module.py` and that value
+        # would not be usable in the standard way. Its not a file you can read or re-execute
+        # against like a loose python module source file would be.
+        #
+        # Python itself disallows this case altogether in the standard zipapp module (probably for
+        # similar reasons). See: https://docs.python.org/3/library/zipapp.html#cmdoption-zipapp-m
         alter_sys = not zipfile.is_zipfile(self._pex)
         return self.execute_module(entry_point, alter_sys)
 
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -7,6 +7,7 @@ import os
 import re
 import sys
 import warnings
+import zipfile
 from distutils import sysconfig
 from site import USER_SITE
 from types import ModuleType
@@ -576,7 +577,7 @@ class PEX(object):  # noqa: T000
         log("  * - paths that do not exist or will be imported via zipimport")
 
     def execute_interpreter(self):
-        # type: () -> Optional[str]
+        # type: () -> Any
         args = sys.argv[1:]
         if args:
             # NB: We take care here to setup sys.argv to match how CPython does it for each case.
@@ -588,7 +589,7 @@ class PEX(object):  # noqa: T000
             elif arg == "-m":
                 module = args[1]
                 sys.argv = args[1:]
-                return self.execute_module(module)
+                return self.execute_module(module, alter_sys=True)
             else:
                 try:
                     if arg == "-":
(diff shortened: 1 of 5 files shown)


tests/test_pex.py
```diff
@@ -626,7 +626,9 @@ def test_execute_interpreter_dashm_module():
         stdout, stderr = process.communicate()
 
         assert 0 == process.returncode
-        assert b"foo.bar one two\n" == stdout
+        assert "{} one two\n".format(os.path.join(pex_chroot, "foo/bar.py")) == stdout.decode(
```
jsirois, line 629: N.B.: The prior expected output does not match what you get for `python -m` - it now does match, with argv0 being the path of the -m module source file.

pex/pex.py
```diff
@@ -654,20 +655,25 @@ def execute_content(
...
+        alter_sys = not zipfile.is_zipfile(self._pex)
+        return self.execute_module(entry_point, alter_sys)
+
+    def execute_module(
+        self,
+        module_name,  # type: str
+        alter_sys,  # type: bool
+    ):
+        # type: (...) -> None
+        self.demote_bootstrap()
```
stuhood, line 672: A few inline comments on why the various execution modes need to `demote` vs `alter_sys` vs etc wouldn't hurt.

tests/test_integration.py
```diff
@@ -3210,3 +3214,93 @@ def test_console_script_from_pex_path(tmpdir):
...
+                    main()
+                """
+            )
+        )
+    expected_output = b"3\n"
+
+    # There are 9 ways we can invoke the module above using Pex corresponding to the 2D matrix with
+    # axes: {zipapp, unzip, venv} x {entrypoint-function, entrypoint-module, ad-hoc (-m) module}.
+    # Of these, zipapp + entrypoint-module is the only combination where we can't both satisfy being
+    # able to re-exec a PEX based on argv[0] and support pickling.
```
stuhood, line 3247: Could this test be phrased as a nested loop over these dimensions, and a conditional for the exceptional case? Or the case itself be extracted into a helper method instead?

stuhood (approved):
Thanks!

tests/test_integration.py
jsirois: It probably could, but there really isn't a whole lot repeated or shared so it would get harder to read. The --venv --seed mode is an outlier from all other cases for one.

pex/pex.py
jsirois: All execution modes demote so there is no vs.

I'll add verbage explaining why the zipapp case does not alter_sys though.

jsirois: Fixed.

PR #1279, opened 2021-03-18
Expose Pex tools via a pex-tools console script.

jsirois:


jsirois:
I was trying to debug some PEXes produced by Pants in process execution chroots and wanted to be able to say `pex-tools <pex> ...` without having to care if I could `PEX_TOOLS=1 <pex> ...` (i.e.: not have to answer "Does the pex --include-tools?") or have to be in a pex clone where I could `python -mpex.tools /path/to/chroot/<pex> ...`.

Eric-Arellano (approved):

PR #1287, opened 2021-03-25
Improve resolve error information.

jsirois:
When a resolve failed at the stage of determining roots it left alot
of context on the floor. Add that context in the error message when
available.


jsirois:
This is motivated by making a hard to debug macOS / CI specific issue here: https://github.com/pantsbuild/pants/pull/11773/checks?check_run_id=2162600254#step:9:87 easier to debug.

Eric-Arellano (approved):
Thank you!

PR #1292, opened 2021-03-25
Fix Packaging check to be complete.

jsirois:
Previously the check only included distribution building; now it
includes building of the Pex PEX for distribution as well.


jsirois:
OK - there was a little bit of thrash but this is ready for review.

Eric-Arellano (approved):

PR #1296, opened 2021-03-29
Fix wheel building for certain macOS system interpreters.

### LINKED ISSUE #1294: Pex builds wheels with bad platform tags on some mac machines.
This was seen over in Pants in https://github.com/pantsbuild/pants/issues/11061. In that case, the mac system interpreter was built with a `MACOSX_DEPLOYMENT_TARGET` of `10.14.6` leading to wheels built with that interpreter having an invalid platform tag of `macosx_10_14_6_x86_64`. Pex should either detect these badly configured / behaving interpreters and exclude them from discovery / fail fast when directly selected or else find a way to un-break these interpreters.

jsirois:
Some macOS system interpreters mis-report their platform for the
purposes of wheel generation and consumption. Ensure the all macOS
interpreters report a proper platform with the major and minor version
components of the mac release via a sysconfig backdoor present in
CPython 2.7 through 3.9.

Add a previously failing test that is fixed by this change.

Fixes #1294

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -3305,3 +3305,21 @@ def test_execute_module_issues_1018(tmpdir):
         args=["-D", src_dir, "-m", "issues_1018", "-o", with_module_venv_pex, "--venv"]
     ).assert_success()
     assert expected_output == subprocess.check_output(args=[with_module_venv_pex])
+
+
+def test_invalid_macosx_platform_tag(tmpdir):
+    # type: (Any) -> None
+    if not any((3, 8) == pi.version[:2] for pi in PythonInterpreter.iter()):
+        pytest.skip("Test requires a system Python 3.8 interpreter.")
+
+    repository_pex = os.path.join(str(tmpdir), "repository.pex")
+    ic_args = ["--interpreter-constraint", "==3.8.*"]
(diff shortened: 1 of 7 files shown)


jsirois:
Alright - ready for review. The 1st commit adds the test and has a red CI run because of that new test on macOS. Second commit fixes.

tests/test_integration.py
```diff
@@ -3305,3 +3305,21 @@ def main():
         args=["-D", src_dir, "-m", "issues_1018", "-o", with_module_venv_pex, "--venv"]
     ).assert_success()
     assert expected_output == subprocess.check_output(args=[with_module_venv_pex])
+
+
+def test_invalid_macosx_platform_tag(tmpdir):
+    # type: (Any) -> None
+    if not any((3, 8) == pi.version[:2] for pi in PythonInterpreter.iter()):
+        pytest.skip("Test requires a system Python 3.8 interpreter.")
```
Eric-Arellano, line 3313: Should this only run on macOS to save CI resources?

Eric-Arellano (approved):
Thank you for fixing this! Great documentation.

tests/test_integration.py
jsirois: Yeah, good idea.

PR #1299, opened 2021-03-30
Add support for `--seed verbose`.

jsirois:
This enables seeding and finding out all pertinent information about
the seed atomically. Previously, seed users like Pants would need to
concoct fragile or unsafe ways to determine the PEX_ROOT used by a
seeded PEX.

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -8,6 +8,7 @@ sources, requirements and their dependencies.
 
 from __future__ import absolute_import, print_function
 
+import json
 import os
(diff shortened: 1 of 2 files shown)


jsirois:
The motivation for this is over here in Pants: https://github.com/pantsbuild/pants/blob/2b412dbfe1b803f8502cda29b6dce46690137592/src/python/pants/backend/python/util_rules/pex.py#L852-L871

Eric-Arellano (approved):

PR #1301, opened 2021-03-30
Spruce up `pex` and `pex-tools` CLIs.

jsirois:
Add uniform `-V` / `--version` support and ensure pex-tools displays
argument defaults.

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -691,7 +691,7 @@ def configure_clp():
         fromfile_prefix_chars="@",
     )
 
-    parser.add_argument("--version", action="version", version=__version__)
+    parser.add_argument("-V", "--version", action="version", version=__version__)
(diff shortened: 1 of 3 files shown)


jsirois:
> Thanks. Consider updating the PR title to more explicitly mention -V/--version. Right now, this sounds purely like a refactor.

Since its in the body, I think that's fine. To be clear I'm pretty sure we both have different rules guiding our commit messages. IIUC you're eyeing the changelog. I'm following the Linux guidelines - for no great reason - and restricting to a <= 50 chars subject and a body wrapped at 72 chars.

If I've stated our cross purposes correctly, I'm happy to expand the actual changelog with this info.

Eric-Arellano:
> IIUC you're eyeing the changelog

Yep, my PR titles always are worded for the changelog, which is why they often exceed your char limit. Ack on using something like TownCrier being a better alternative to that.

PR #1307, opened 2021-04-01
Migrate to softprops/action-gh-release.

jsirois:
Eric noticed both the actions/create-release and
actions/upload-release-asset GitHub repositories had been archived. This
converts our two-step create and upload into one step using the
reccomendd replacement action.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -73,13 +73,12 @@ jobs:
         with:
           tox-env: package
       - name: Create ${{ needs.determine-tag.outputs.release-tag }} Release
-        id: create-release
-        uses: actions/create-release@v1
+        uses: softprops/action-gh-release@v1
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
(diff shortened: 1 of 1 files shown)


jsirois:
Thanks for pointing this our @Eric-Arellano - assuming this works, things got simpler anyhow.

Eric-Arellano (approved):
Neat! Thanks.

How do you test this when iterating? I.e. a dry release. That was something holding me back from implementing stable Pex release in Pants.

jsirois:
> How do you test this when iterating?

I don't. I did have the same nerves though and implemented a manual release using:
https://github.com/pantsbuild/pex/blob/8d953ba407cc75f56a3697a72b8b1cb34195d777/.github/workflows/release.yml#L6-L10

I had to use that in my 1st release or two when sorting things out.

jsirois:
@Eric-Arellano that worked just fine: https://github.com/pantsbuild/pex/runs/2242481723?check_suite_focus=true
You should have no troubles with Pants PEX.

Issue #793, opened 2019-11-15
`pex -r <(pip freeze)` syntax stopped working in 2.0.0


zachmoshe:
So far I was using pex with a temporary requirements file like this: 
`pex -r <(pip freeze)`  and it worked nicely.

After updating to 2.0, I started getting this: 
`ERROR: Could not open requirements file: [Errno 9] Bad file descriptor: '/dev/fd/63'`

I'm on MacOS, with pex upgraded to the latest (2.0.0), python 3.6.5 and pip 19.3.1.

If I try on Unix (Ubuntu 16.04), python 3.6.5, pex 2.0.0, pip 19.3.1, I get a slightly different message:
`ERROR: Could not open requirements file: [Errno 2] No such file or directory: '/dev/fd/63'`

It works normally if I save the `pip freeze` output to a file and pass the filename as the argument.

Was anything changed in how pex parses arguments or opens the requirements file?
Is the (convenient) old syntax still supported?

jsirois:
This one is tricky to fix. PEX needs to read requirement files before handing them off to pip, and for process substitution psuedo-files, this closes the fd leaving pip without a valid handle. Can you instead just `pex $(pip freeze)`? Less typing and works with all versions of PEX.

zachmoshe:
makes sense. thanks!

adeandrade:
The non-working approach is still in the docs.

jsirois:
@adeandrade do you mean here?: https://github.com/pantsbuild/pex/blob/master/README.rst#simple-examples
If so would you be willing to sping up a PR?

PR #1323, opened 2021-04-06
Update docs to use `pex $(pip freeze)`

lynshi:
As mentioned in #793, `pex $(pip freeze)` should be used instead of `pex -r <(pip freeze)`. The README was updated, but the docs site was not.


Eric-Arellano (approved):
Thank you for the contribution!

docs/buildingpex.rst
```diff
@@ -100,7 +100,7 @@ way to freeze a virtualenv into a PEX file:
 
 .. code-block:: console
 
-    $ pex -r <(pip freeze) -o my_application.pex
+    $ pex $(pip freeze) -o my_application.pex
```
jsirois, line 103: You'll need to update the lead in paragraph above too: https://github.com/pantsbuild/pex/pull/1323/files#diff-ade16e5bd61e6943865733995fbf8a75c6cf7b5e270b583bc3f51606aec4011bR98-R99

As it stands now its confusing, since the paragraph is about `-r` and this freeze reccomendation has nothing to do with that any more.

lynshi: Good catch, thanks!

As you can see, I only read the example 😄

docs/buildingpex.rst
```diff
@@ -95,12 +95,12 @@ You can then import and manipulate modules like you would otherwise:
     >>> import psutil
     >>> ...
 
-Requirements can also be specified using the requirements.txt format, using ``pex -r``.  This can be a handy
+Requirements can also be specified using the requirements.txt format as generated by ``pip freeze``.  This can be a handy
```
jsirois, line 98: Hrm, this effectively wiped out a recommendation about using -r. The `pip freeze` output is a tiny subset of what can go in requirements.txt.

So, the structure before of this `Specifying requirements` section was:
1. You can specify command line requirement strings like so.
2. You can also use requirements files like so.

Now the structure is:
1. You can specify command line requirements like so.
2. A small misrepresentation of requirements.txt format / pip freeze.

How about:
1. You can specify command line requirements like so.
1a. A handy example of 1 is pex $(pip freeze) since `pip freeze` emits a list of pinned requirement strings for you like `setuptools==2.1.2` mentioned in section 1 (not these words exactly - this is just me trying to explain the structure).
2. You can also use requirements files with -r just like with Pip.

lynshi: Ah, thanks for the clarification - I see what the section was going for now. Will make a corresponding change!

Eric-Arellano (approved):
That looks better. Thanks John for catching it and @lynshi for iterating on this!

jsirois (approved):
Excellent. Thanks @lynshi.

jsirois:
Alright @lynshi this is live on latest (needed to fix a missed master -> main transition): https://pex.readthedocs.io/en/latest/buildingpex.html

lynshi:
Awesome, happy to help!

PR #1325, opened 2021-04-11
Fix pyenv shim identification.

### LINKED ISSUE #1324: Once a pyenv shim is activated and seen once by Pex it is always visible to Pex.
This happens because the target of a pyenv shims is mutable but the Pex interpreter cache is keyed by the shim (and its contents) which is immutable. Pex needs to factor in the mutable information (PYENV_VERSION, .python_version files in the CWD ancestor lineage and PYENV_ROOT/versions) into its cache key for pyenv shims.

jsirois:
Previously shims were improperly cached without regard to their
activation status. Now a dedicated Pyenv class is used to test if an
interpreter binary is a pyenv shim, and, if so, resolve it to its
activated version if any. This fixes cache keying and also improves
bulk interpreter identification by allowing de-duping active shims and
skipping inactive shims.

An existing test is updated and expanded to test for this as well as
verifying the pyenv version configuration hierarchy of methods.

Fixes #1324

### DIFF
--- a/pex/pyenv.py
+++ b/pex/pyenv.py
@@ -13,7 +13,7 @@ from pex.typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
     import attr  # vendor:skip
-    from typing import Iterator, List, Optional, Tuple
+    from typing import Iterable, Iterator, Optional, Tuple, Iterable
 else:
     from pex.third_party import attr
 
@@ -157,32 +157,38 @@ class Pyenv(object):
         # type: (Optional[str]) -> Tuple[str, ...]
         """Reports the active pyenv versions for the given starting search directory or $PWD."""
 
+        source_and_versions = None  # type: Optional[Tuple[str, Iterable[str]]]
+
         # See: https://github.com/pyenv/pyenv#choosing-the-python-version
         with TRACER.timed("Finding {} active versions...".format(self), V=6):
             shell_version = os.environ.get("PYENV_VERSION")
             if shell_version:
-                TRACER.log(
-                    "Found active pyenv version of PYENV_VERSION={}".format(shell_version), V=6
+                source_and_versions = (
+                    "PYENV_VERSION={}".format(shell_version),
+                    shell_version.split(":"),
                 )
-                return (shell_version,)
-
-            cwd = search_dir if search_dir is not None else os.getcwd()
-            TRACER.log("Looking for pyenv version files starting from {}.".format(cwd), V=6)
-
-            versions = []  # type: List[str]
-            local_version = self._find_local_version_file(search_dir=cwd)
-            if local_version:
-                versions.extend(self._read_pyenv_versions(local_version))
-                TRACER.log("Found active versions in {}: {}".format(local_version, versions), V=6)
             else:
-                global_version = os.path.join(self.root, "version")
-                if os.path.exists(global_version):
-                    versions.extend(self._read_pyenv_versions(global_version))
-                    TRACER.log(
-                        "Found active versions in {}: {}".format(global_version, versions), V=6
-                    )
+                cwd = search_dir if search_dir is not None else os.getcwd()
+                TRACER.log("Looking for pyenv version files starting from {}.".format(cwd), V=6)
 
+                local_version = self._find_local_version_file(search_dir=cwd)
+                if local_version:
+                    source_and_versions = (local_version, self._read_pyenv_versions(local_version))
+                else:
+                    global_version = os.path.join(self.root, "version")
+                    if os.path.exists(global_version):
+                        source_and_versions = (
+                            global_version,
+                            self._read_pyenv_versions(global_version),
+                        )
+
+        if source_and_versions:
+            source, versions = source_and_versions
+            TRACER.log("Found active versions in {}: {}".format(source, versions), V=6)
             return tuple(versions)
+
+        TRACER.log("Found no active pyenv versions.", V=6)
+        return ()
 
     def python(
         self,
--- a/pex/pyenv.py
+++ b/pex/pyenv.py
@@ -8,27 +8,26 @@ import re
 import subprocess
 
 from pex.common import is_exe
-from pex.compatibility import to_unicode
 from pex.tracer import TRACER
 from pex.typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
     import attr  # vendor:skip
-    from typing import Iterator, List, Optional, Text, Tuple
+    from typing import Iterator, List, Optional, Tuple
 else:
     from pex.third_party import attr
 
 
 @attr.s(frozen=True)
 class Pyenv(object):
-    root = attr.ib()  # type: Text
+    root = attr.ib()  # type: str
 
     @classmethod
     def find(cls):
         # type: () -> Optional[Pyenv]
         """Finds the active pyenv installation if any."""
         with TRACER.timed("Searching for pyenv root...", V=3):
-            pyenv_root = to_unicode(os.environ.get("PYENV_ROOT", ""))
+            pyenv_root = os.environ.get("PYENV_ROOT", "")
             if not pyenv_root:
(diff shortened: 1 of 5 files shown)


jsirois:
I think this finally deals with pyenv properly. There was simply no way to do so without acknowledging its existence with dedicated handling. I hate to do this, but, similar to macOS, pyenv shims are widely used and so Pex has to handle them to be useful to a large share of developers.

jsirois:
Even with the new checks the speed difference is in the noise. Sometimes OLD wins by a hair, sometimes NEW:
```
$ hyperfine --prepare "cd ~/dev/pantsbuild/jsirois-pex" -n "NEW" --prepare "cd ~/dev/pantsbuild/pex" -n "OLD" 'python -mpex --interpreter-constraint===3.5.9 -- -c "import sys; print(sys.executable, sys.version_info)"' 'python -mpex --interpreter-constraint===3.5.9 -- -c "import sys; print(sys.executable, sys.version_info)"'
Benchmark #1: NEW
  Time (mean ± σ):     691.6 ms ±   8.1 ms    [User: 611.5 ms, System: 82.1 ms]
  Range (min … max):   682.1 ms … 706.5 ms    10 runs
 
Benchmark #2: OLD
  Time (mean ± σ):     692.1 ms ±   5.0 ms    [User: 618.4 ms, System: 75.3 ms]
  Range (min … max):   684.1 ms … 702.7 ms    10 runs
 
Summary
  'NEW' ran
    1.00 ± 0.01 times faster than 'OLD'
```

jsirois:
Hrm. These tests work repeatedly locally on all supported Pythons. I'll need to dig more...

jsirois:
Alright - this is solid now and ready for review.

pex/pyenv.py
```diff
@@ -0,0 +1,205 @@
...
+        name = attr.ib()  # type: str
+        major = attr.ib()  # type: Optional[str]
+        minor = attr.ib()  # type: Optional[str]
+
+        _SHIM_REGEX = re.compile(
+            r"""
+            ^
+            (?P<name>
+                python |
+                pypy
```
benjyw, line 62: What about things like jython? do we care?

pex/pyenv.py
```diff
@@ -0,0 +1,205 @@
...
+        """Reports the active pyenv versions for the given starting search directory or $PWD."""
+
+        # See: https://github.com/pyenv/pyenv#choosing-the-python-version
+        with TRACER.timed("Finding {} active versions...".format(self), V=6):
+            shell_version = os.environ.get("PYENV_VERSION")
+            if shell_version:
+                TRACER.log(
+                    "Found active pyenv version of PYENV_VERSION={}".format(shell_version), V=6
+                )
+                return (shell_version,)
```
benjyw, line 167: The shell version can be multiple versions (separated by a colon), don't we need to split the string here?

```
$ pyenv shell 3.7.3 2.7.15
$ echo $PYENV_VERSION
3.7.3:2.7.15
```

pex/pyenv.py
```diff
@@ -0,0 +1,205 @@
...
+
+            cwd = search_dir if search_dir is not None else os.getcwd()
+            TRACER.log("Looking for pyenv version files starting from {}.".format(cwd), V=6)
+
+            versions = []  # type: List[str]
+            local_version = self._find_local_version_file(search_dir=cwd)
+            if local_version:
+                versions.extend(self._read_pyenv_versions(local_version))
```
benjyw, line 175: It's a little odd to return early for shell versions but not for local versions. Is there a reason to not return the result of `_read_pyenv_versions()` directly here?

pex/pyenv.py
benjyw: And would be good to test this case.

benjyw (commented):
Great stuff!! 

On the Pants side, am I right that we still have to do something about propagating PYENV_VERSION through pantsd?

pex/pyenv.py
jsirois: Pex only supports CPython and PyPY: 
https://github.com/pantsbuild/pex/blob/30bc1ab39ccec76934f8842bd333508d6e69ae43/pyproject.toml#L26-L29
https://github.com/pantsbuild/pex/blob/30bc1ab39ccec76934f8842bd333508d6e69ae43/pex/interpreter.py#L72-L75
https://github.com/pantsbuild/pex/blob/30bc1ab39ccec76934f8842bd333508d6e69ae43/pex/interpreter.py#L332-L342

pex/pyenv.py
jsirois: Aha. I did not know about that. Thanks! Added handling and a test case.

pex/pyenv.py
jsirois: I ended up doing the opposite after adding shell `:` splitting handling. All three cases now share a single log and return.

jsirois:
> On the Pants side, am I right that we still have to do something about propagating PYENV_VERSION through pantsd?

Yes. But that only covers `pyenv shell`. For full handling Pants would need to deal with the various version files. Probably the best it could do easily would be to also propagate PYENV_ROOT and PATH so Pex could find pyenv and `pyenv root` gave the correct answer. That would cover reading in the `pyenv global` config robustly for local Pants runs. As things currently stand we could only handle in-repo `.python-version` files. We'd need to find those and add them to the Process inputs. A `.python-version` at the repo root - the current canonical $PWD for all pants runs - is easy enough to find and include. I'm not sure how to handle other `.python-version` files deeper in the tree. Those cases have alot of corners and room for interpretation.

jsirois:
Urg - let me add merge protection. I read too quickly and you had not actually approved. Please let me know if there were further problems and I'll follow up fix.

pex/pyenv.py
```diff
@@ -0,0 +1,211 @@
...
+        _PYENV_CPYTHON_VERSION_LEADING_CHARS = frozenset(str(digit) for digit in range(2, 10))
+
+        def select_version(self, search_dir=None):
+            # type: (Optional[str]) -> Optional[str]
+            """Reports the active shim version for the given directory or $PWD.
+
+            If the shim is not activated, returns `None`.
+            """
+            with TRACER.timed("Calculating active version for {}...".format(self), V=6):
+                active_versions = self.pyenv.active_versions(search_dir=search_dir)
```
benjyw, line 107: Note: at first I thought this logic was incorrect, because the way  https://github.com/pyenv/pyenv#choosing-the-python-version is phrased makes it seem like the fallthrough logic is per-shim. So for example if shell is 3.7.3 and local is 3.6.6 then `python3.6` should work. But I've verified that in practice it does not, and shell overrides local overrides global for all shims at once, so we're good.

pex/pyenv.py
```diff
@@ -0,0 +1,211 @@
...
+    def as_shim(self, binary):
+        # type: (str) -> Optional[Shim]
+        """View the given binary path as a pyenv shim script if it is one."""
+        return self.Shim.parse(self, binary)
+
+    @staticmethod
+    def _read_pyenv_versions(version_file):
+        # type: (str) -> Iterator[str]
+        with open(version_file) as fp:
+            for line in fp:
```
benjyw, line 140: Why not just read the entire file into a string and split it on any whitespace?

benjyw (approved):

pex/pyenv.py
benjyw: No need to re-open to fix.

pex/pyenv.py
jsirois: Yes. I went through the same tests and read the pyenv source code to verify this.

jsirois: The unit tests cover this as well.

pex/pyenv.py
jsirois: Yeah, that would work. I started with just returning stripped lines since thats what version files generated by `pyenv global` and `pyenv local` had and only later read the docs closer and noticed the bit about `... or any whitespace`.

PR #1332, opened 2021-04-23
Update interpreter.py

### LINKED ISSUE #1331: Pex fail to build wheels when target is just a major version on macos.
From https://github.com/pantsbuild/pants/issues/11967

When `MACOSX_DEPLOYMENT_TARGET=11` we get an invalid platform tag:
```
macosx-['11'].0-x86_64
```

Caused by the fix for #1294

kaos:
closes #1331


jsirois (approved):
Thanks you very much @kaos.

PR #1367, opened 2021-06-21
Fix `--platform` resolve handling of env markers.

### LINKED ISSUE #1366: Environment markers are incorrectly evaluated for --platform resolves.
As discovered in https://github.com/pantsbuild/pants/issues/12182, Pex - via Pip - incorrectly evaluates environment markers in the context of the ambient interpreter instead of using the information from the platform specification. The Pip issue is filed over here: https://github.com/pypa/pip/issues/10050 but this will likely required working around Pip since they do not see this as a bug.

jsirois:
Previously we relied on Pip, but it turns out Pip does not correctly
handle environment markers for `--platform` + `--implementation` +
`--python-version` + `--abi` (aka "foreign") resolves. Instead of
waiting on resolution of https://github.com/pypa/pip/issues/10050, patch
Pip's vendored copy of `packaging` to see a default environment that
corresponds to what we know about the foreign platform. It turns out
what we know covers the most commonly used environment markers making
this workaround both safe and useful in practice.

Fixes #1366

### DIFF
--- a/pex/pip.py
+++ b/pex/pip.py
@@ -11,8 +11,10 @@ import os
 import re
 import subprocess
 import sys
+from abc import abstractmethod
 from collections import deque
 from contextlib import closing
+from textwrap import dedent
 
 from pex import dist_metadata, third_party
 from pex.common import atomic_directory, safe_mkdtemp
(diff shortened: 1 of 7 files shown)


pex/pip.py
```diff
@@ -180,7 +195,112 @@ def __init__(
...
 
 
+class _LogAnalyzer(object):
+    @abstractmethod
+    def analyze(self, line):
+        # type: (str) -> Union[bool, str]
+        """Analyze the given log line.
+
+        Returns useful text or else True to continue to analyze the next log line or else False to
+        stop analysis.
```
stuhood, line 205: This could maybe be more easily modeled as a coroutine, but it's not clear that the type signature would elucidate its usage enough to avoid the need for the comment.

stuhood (approved):
Thanks!

pex/pip.py
jsirois: I'm not following the coroutine bit but the types could have been better, fixed those up in 4a615cf.

stuhood: ...mm, yea. That was a half baked thought. Was thinking something like [`Coroutine[bool, str, str]`](https://docs.python.org/3/library/typing.html#typing.Coroutine)... but that would definitely be more awkward to use.

PR #1369, opened 2021-07-03
Limit GH action execution to pantsbuild.

jsirois:


jsirois:
The skip mechanism working can be seen here: https://github.com/jsirois/pex/actions/runs/996681362

jsirois:
Thanks for figuring this our Eric. This cleans up what you found a bit. I'll apply to the pants-jupyter-plugin and example-python repos as well.

Eric-Arellano (approved):

Issue #1380, opened 2021-07-14
test_plain_pex_exec_no_ppp_no_pp_no_constraints fails when using python from symlinked paths


kaos:
System: MacOS
pex: main ad450399a22a9a36fa84aff231fbb9adee1c0d28
python: installed using brew
Test command: `tox -e py38-integration -- -k pp_no_constraints` (invoked from a virtualenv, using python 3.8.5)

The expected output is correct, if symlinks are resolved...
```
# expected path: /usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/bin/python3.8
# actual path: /usr/local/opt/python@3.8/bin/python3.8

$ ll /usr/local/opt/python@3.8
lrwxr-xr-x  1 x  staff  26 Oct  6  2020 /usr/local/opt/python@3.8 -> ../Cellar/python@3.8/3.8.5

$ ll /usr/local/opt/python@3.8/bin/python3.8
lrwxr-xr-x  1 x  staff  57 Oct  6  2020 /usr/local/opt/python@3.8/bin/python3.8 -> ../Frameworks/Python.framework/Versions/3.8/bin/python3.8

```

Is there another way to invoke the tests that avoids this?

Test output:
```
[gw0] darwin -- Python 3.8.5 /Users/x/src/github/kaos/pex/.tox/py38-integration/bin/python

    def test_plain_pex_exec_no_ppp_no_pp_no_constraints():
        # type: () -> None
        with temporary_dir() as td:
            pex_out_path = os.path.join(td, "pex.pex")
            env = make_env(PEX_IGNORE_RCFILES="1")
            res = run_pex_command(["--disable-cache", "-o", pex_out_path], env=env)
            res.assert_success()
    
            stdin_payload = b"import os, sys; print(sys.executable); sys.exit(0)"
            stdout, rc = run_simple_pex(pex_out_path, stdin=stdin_payload, env=env)
            assert rc == 0
>           assert (
                PythonInterpreter.get().resolve_base_interpreter().binary.encode() in stdout
            ), "Expected the current interpreter to be used when no constraints were supplied."
E           AssertionError: Expected the current interpreter to be used when no constraints were supplied.
[...]

PR #1381, opened 2021-07-14
support executing any scripts specified in `setup(scripts=...)` from …

### LINKED ISSUE #1379: support setuptools scripts
In order to have any executable script as entry point, pex should support the `scripts` section in `setup()`.
From pantsbuild/pants#3399
Doc: [setuptools `scripts`](https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html#the-scripts-keyword-argument)

kaos:
…setup.py.

fixes #1379 

Most of the needed functionality already existed ;)

### DIFF
--- a/pex/interpreter.py
+++ b/pex/interpreter.py
@@ -777,23 +777,6 @@ class PythonInterpreter(object):
             return interpreter
 
     @classmethod
-    def from_shebang(cls, shebang):
-        # type: (str) -> Optional[PythonInterpreter]
-        if not (shebang and shebang.startswith("#!")):
-            return None
-
-        binary, _, args = shebang.partition(" ")
-        if binary == "/usr/bin/env":
-            if not args:
-                return None
-            return cast(PythonInterpreter, cls.from_env(args))
-
-        if cls._matches_binary_name(binary):
-            return cls.from_binary(binary)
-
-        return None
-
-    @classmethod
     def _spawn_from_binary(cls, binary):
         # type: (str) -> SpawnedJob[PythonInterpreter]
         canonicalized_binary = cls.canonicalize_path(binary)
--- a/pex/interpreter.py
+++ b/pex/interpreter.py
@@ -777,6 +777,23 @@ class PythonInterpreter(object):
             return interpreter
 
     @classmethod
+    def from_shebang(cls, shebang):
+        # type: (str) -> Optional[PythonInterpreter]
+        if not (shebang and shebang.startswith("#!")):
+            return None
+
+        binary, _, args = shebang.partition(" ")
+        if binary == "/usr/bin/env":
+            if not args:
+                return None
+            return cast(PythonInterpreter, cls.from_env(args))
+
+        if cls._matches_binary_name(binary):
+            return cls.from_binary(binary)
+
+        return None
+
+    @classmethod
     def _spawn_from_binary(cls, binary):
         # type: (str) -> SpawnedJob[PythonInterpreter]
         canonicalized_binary = cls.canonicalize_path(binary)
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -617,10 +617,47 @@ class PEX(object):  # noqa: T000
         dist_script = get_script_from_distributions(script_name, dists)
         if not dist_script:
             return "Could not find script {!r} in pex!".format(script_name)
-        TRACER.log("Found script {!r} in {!r}.".format(script_name, dist))
-        return self.execute_content(
-            dist_script.path, dist_script.read_contents(), argv0=script_name
-        )
+        TRACER.log("Found script {!r} in {!r}.".format(script_name, dist_script.dist))
+
+        contents = dist_script.read_contents()
+        shebang = contents.split("\n", 1)[0]
+        try:
+            interpreter = PythonInterpreter.from_shebang(shebang)
+            if interpreter and interpreter.version_string != self.interpreter.version_string:
+                TRACER.log(
+                    "Python script shebang resolved to {shebang_python}, but we will execute it with {pex_python}".format(
+                        shebang_python=interpreter.version_string,
+                        pex_python=self.interpreter.version_string,
+                    ),
+                    V=2,
+                )
+        except PythonInterpreter.InterpreterNotFound as e:
+            TRACER.log(
+                "Python script shebang indicated unknown interpreter: {shebang!r}: {e}, will execute it with {version}.".format(
+                    shebang=shebang, e=e, version=self.interpreter.version_string
+                ),
+                V=2,
+            )
+            interpreter = self.interpreter
+
+        if interpreter:
+            TRACER.log(
+                "Executing python script with {}".format(self.interpreter.version_string), V=2
+            )
+            return self.execute_content(dist_script.path, contents, argv0=script_name)
+        else:
+            TRACER.log("Executing non-python script using shebang: {}".format(shebang), V=2)
+            return self.execute_external(dist_script.path)
+
+    def execute_external(self, cmd):
+        # type: (str) -> Any
+        try:
+            # execute as a shell command, inheriting stdout and stderr
+            proc = Executor.open_process([cmd] + sys.argv[:1], shell=True)
+            proc.communicate()
+            return proc.returncode
+        except Executor.ExecutionError as e:
+            return "Could not invoke script: {}".format(e)
 
     @classmethod
     def execute_content(
@@ -633,10 +670,8 @@ class PEX(object):  # noqa: T000
         argv0 = argv0 or name
         try:
             ast = compile(content, name, "exec", flags=0, dont_inherit=1)
-        except SyntaxError:
-            return "Unable to parse {}. PEX script support only supports Python scripts.".format(
-                name
-            )
+        except SyntaxError as e:
+            return "Unable to parse {}: {}".format(name, e)
 
         cls.demote_bootstrap()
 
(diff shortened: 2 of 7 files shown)


pex/testing.py
```diff
@@ -137,8 +137,8 @@ def make_project(
             )
             """
         ),
-        "scripts/hello_world": '#!/usr/bin/env python\nprint("hello world!")\n',
-        "scripts/shell_script": "#!/usr/bin/env bash\necho hello world\n",
+        "scripts/hello_world": '#!/usr/bin/env python\nprint("hello world from py script!")\n',
+        "scripts/shell_script": "#!/usr/bin/env bash\necho hello world from shell script\n",
```
kaos, line 141: I changed the output wording slightly, so we can be sure the output comes from the correct source.
Would've liked to change the print in `do_something()` as well, but as I had issues running the tests, I didn't want to mess too much up.

pex/pex.py
```diff
@@ -617,10 +617,47 @@ def execute_script(self, script_name):
...
+                ),
+                V=2,
+            )
+            interpreter = self.interpreter
+
+        if interpreter:
+            TRACER.log(
+                "Executing python script with {}".format(self.interpreter.version_string), V=2
+            )
+            return self.execute_content(dist_script.path, contents, argv0=script_name)
```
kaos, line 647: I first sent all scripts to `execute_external`, but then that failed for python scripts importing modules from the pex.. and I didn't find a way to run that as an subprocess, so circled back to the current solution, to simply parse the contents with the current interpreter. 

Added some logging in case the script uses something like `#!/usr/bin/env python3.6` while we're running a different version...

kaos:
I'm not entirely sure I have grok:ed the ins and outs of all that is going on here.. so may be way off mark how to implement this properly. Hopefully, though, it works, or is close enough, any way :)

kaos:
Bump ;)

pex/pex.py
```diff
@@ -617,10 +617,47 @@ def execute_script(self, script_name):
...
+        else:
+            TRACER.log("Executing non-python script using shebang: {}".format(shebang), V=2)
+            return self.execute_external(dist_script.path)
+
+    def execute_external(self, cmd):
+        # type: (str) -> Any
+        try:
+            # execute as a shell command, inheriting stdout and stderr
```
jsirois, line 655: This comment just restates the code. The only interesting thing to comment on is left out. Why the [risky](https://docs.python.org/3/library/subprocess.html#security-considerations) `shell=True`? Do you know of cases where the non-python script is not extracted / installed with the executable bit set?

pex/pex.py
```diff
@@ -617,10 +617,47 @@ def execute_script(self, script_name):
...
+            proc = Executor.open_process([cmd] + sys.argv[:1], shell=True)
+            proc.communicate()
```
jsirois, line 657: You can just `return proc.wait()` instead of `proc.communicate(); return proc.returncode` - you do nothing with the return of `proc.communicate()` which is a hint.

pex/pex.py
```diff
@@ -617,10 +617,47 @@ def execute_script(self, script_name):
...
-            dist_script.path, dist_script.read_contents(), argv0=script_name
-        )
+        TRACER.log("Found script {!r} in {!r}.".format(script_name, dist_script.dist))
+
+        contents = dist_script.read_contents()
+        shebang = contents.split("\n", 1)[0]
+        try:
+            interpreter = PythonInterpreter.from_shebang(shebang)
+            if interpreter and interpreter.version_string != self.interpreter.version_string:
+                TRACER.log(
```
jsirois, line 627: Neither this check and log nor the except clause log are useful afaict. When scripts are installed in venvs its expected that their shebangs are re-written to point to the current venv interpreter: https://setuptools.readthedocs.io/en/latest/deprecated/distutils/setupscript.html?highlight=scripts#installing-scripts. In the case of a PEX file, this is the executing with the PEX file itself which is what `execute_content` does. So I think a few changes are in order:
1. Python script detection should work exactly like it does for distutils / setuptools - this will address the shebang parsing non-robustness observation aobve.
2. Python script detection only need present a bool result.

Its probably the best to add this parsing / bool directly to DistributionScript which already encapsulates finding these things:
https://github.com/pantsbuild/pex/blob/37c9ff0fa815ac1f03a73d3f2e33fb20352c8843/pex/finders.py#L18-L43

It looks like [poetry](https://python-poetry.org/docs/pyproject/#scripts) and [flit](https://flit.readthedocs.io/en/latest/pyproject_toml.html#scripts-section) only support console scripts so the setuptools / distutils behavior fully defines scripts afaict.

pex/pex.py
```diff
@@ -617,10 +617,47 @@ def execute_script(self, script_name):
...
+            return self.execute_content(dist_script.path, contents, argv0=script_name)
+        else:
+            TRACER.log("Executing non-python script using shebang: {}".format(shebang), V=2)
+            return self.execute_external(dist_script.path)
+
+    def execute_external(self, cmd):
+        # type: (str) -> Any
+        try:
+            # execute as a shell command, inheriting stdout and stderr
+            proc = Executor.open_process([cmd] + sys.argv[:1], shell=True)
```
jsirois, line 656: This happens to work, but perhaps it bothered you why it worked? PEX files are zips after all.

pex/testing.py
jsirois: What issues did you hit? Issues running tests can point to bugs in Pex itself and those are useful to know about.

jsirois (commented):
Thanks for taking a crack at this @kaos.

pex/testing.py
kaos: #1380 for one. But there were more. Mostly due to my environment, I'd suspect. We can take a deep dive in this when I get back from vacation, if you'd like :)

pex/interpreter.py
```diff
@@ -776,6 +776,23 @@ def hashbang_matches(fn):
         for interpreter in cls._identify_interpreters(filter=hashbang_matches):
             return interpreter
 
+    @classmethod
+    def from_shebang(cls, shebang):
+        # type: (str) -> Optional[PythonInterpreter]
+        if not (shebang and shebang.startswith("#!")):
+            return None
+
+        binary, _, args = shebang.partition(" ")
```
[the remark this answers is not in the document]
kaos: Ouch. Wasn't aware space after leading `#!` was accepted.

pex/pex.py
kaos: No, I do not. So, more likely my understanding of how popen and the shell parameter works that is a bit fuzzy.
I read the security implications you linked to a few times over.. a few assumptions etc led me down the wrong path, then ;)

kaos: Been thinking a bit more about this, however, and wonder what the risk is, in this particular case.
We see it as risky, to execute the first line of a script using the shell, but what about the rest of the file, then?

See if I can illustrate my point:
```bash
#!/usr/bin/env bash

sudo rm -rf /
```

If this was the console_script we're about to execute, regardless of what ill-mannered code we might put at the shebang line, does it really matter? Given what we can do once the script executes?
Feels like I'm missing something here... a chain is no stronger than its weakest link, and all that..

Not saying that we should keep `shell=True`, just wondering if the security risk involved really is applicable in this case.

jsirois: The risk is not really the point here - the point is a standard risky bit of the Popen API is being used without comment _and_ its un-needed to boot (or it should be un-needed). If it is needed, it would be great to find out why and if that's not correctable, add a comment saying why this standard risky bit is being set (ie: the execute bit is not set on all or any scripts so we can't just execute them but need to use the standard system shell as an intermediary). This seems to me similar to using `unsafe` in production rust code - you always want a comment explaining why the `unsafe` is actually safe.

kaos: Thanks. That makes sense. :)

pex/pex.py
kaos: No, not really. Which shows that I haven't fully grasped what's going on :P

pex/pex.py
```diff
@@ -617,10 +617,22 @@ def execute_script(self, script_name):
...
+
+        TRACER.log("Found script {!r} in {!r}.".format(script_name, dist_script.dist))
+        if dist_script.is_python_script():
+            return self.execute_content(
+                dist_script.path, dist_script.read_contents(), argv0=script_name
+            )
+        else:
+            return self.execute_external(dist_script.path)
+
+    def execute_external(self, cmd):
```
jsirois, line 629: This can be a staticmethod ~like `execute_content` below since it does not use self or cls.

tox.ini
```diff
@@ -53,7 +54,8 @@ commands =
         --apply \
         --dont-skip __init__.py \
         --skip-glob pex/vendor/_vendored/** \
-        --skip-glob .pyenv_test/**
+        --skip-glob .pyenv_test/** \
+        --skip-glob .env/**
```
jsirois, line 58: Seeing this added and no corresponding `.gitignore` entry added is suspicious.

pex/pex.py
jsirois: Ok. The short of it is that all the distributions needed to run a PEX are extracted from the PEX file to disk (under ~/.pex/installed_wheels) and the locations of each are added to the `sys.path`. See the last 3 lines here:
https://github.com/pantsbuild/pex/blob/78c7b01d11eda2a70d1c0fb677c52b4f36a423f6/pex/environment.py#L329-L345

This is mainly to handle native extensions which can't be loaded from inside a zip like `.py` files can.

jsirois (approved):
This LGTM @kaos. I'm happy to merge this and the rest could be optional follow up if you agree.

tox.ini
kaos: yep, my bad, unintentional checkin of that change.

kaos:
Thanks. Made two quick changes you pointed out, so nothing left for follow up, unless I'm forgetting something.. ? ;)

PR #1358, opened 2021-06-12
Do not suppress pex output in bidst_pex

zmanji:
Always print out the pex command output regardless of failure or success.

Fixes #1356

PR #1383, opened 2021-07-16
Allow the creation of several pex when the option --bdist-all is used with a setup.py defining several entry points

s-binetruy-criteo:
The option -bdist-all did not work correctly, only the first entry point produced pex file.
I propose to change die by print message when the exit code equals to 0

### DIFF
--- a/pex/commands/bdist_pex.py
+++ b/pex/commands/bdist_pex.py
@@ -135,9 +135,8 @@ class bdist_pex(Command):  # noqa
                     result,
                 )
             else:
-                die(
-                    "Successfully created pex via {}:\n{}".format(
-                        " ".join(cmd), stderr.decode("utf-8")
-                    ),
-                    result,
+                print(
+                    "Successfully created pex via {}\n".format(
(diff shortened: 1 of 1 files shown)


kamaradclimber (approved):

Djailla (approved):

jsirois (commented):
Thanks @s-binetruy-criteo. You'll need to run `tox -eformat-run` to fix this up. While doing that - what do you think about adding test coverage over in https://github.com/pantsbuild/pex/blob/main/tests/test_bdist_pex.py so this doesn't break again?

jsirois:
@s-binetruy-criteo, @kamaradclimber and @Djailla - since I have 3 bdist_pex users in the room:

What do you use this for? Did you know about [conscript](https://pypi.org/project/conscript/)? Does conscript solve your usecase?

jsirois:
Noting that this effort has moved to #1358. @s-binetruy-criteo can I close this PR?

jsirois:
Solved in #1396

Issue #1401, opened 2021-08-17
Implement platform specific requirements.txt lock file support.


jsirois:
Pex should be able to generate a lock file for all the concrete Pythons and Platforms passed to a resolve in requirements.txt `--hash` format that can be used to create a new PEX with the same 3rdparty content via `pex -r requirements.locked.txt requirements.pex` on the same machine.

jsirois:
Ok, non-strict (platform-applicable dists not actually resolved are included) mode platform specific resolve looks good to go:
```console
$ python -mpex -f find-links --constraints constraints.txt pantsbuild.pants -opants.pex
ansicolors==1.1.8 # ansicolors==1.1.8 via -> pantsbuild.pants \
    --hash:sha256=47926ab39ca3a88cc191bc07c50a63bab24f515f4ededdc9ceb46eec2602637a # file:///home/jsirois/dev/pantsbuild/jsirois-pex/find-links/ansicolors-1.1.8-py2.py3-none-any.whl \
    --hash:sha256=00d2dde5a675579325902536738dd27e4fac1fd68f773fe36c21044eb559e187 # https://files.pythonhosted.org/packages/53/18/a56e2fe47b259bb52201093a3a9d4a32014f9d85071ad07e9d60600890ca/ansicolors-1.1.8-py2.py3-none-any.whl \
    --hash:sha256=99f94f5e3348a0bcd43c82e5fc4414013ccc19d70bd939ad71e0133ce9c372e0 # https://files.pythonhosted.org/packages/76/31/7faed52088732704523c259e24c26ce6f2f33fbeff2ff59274560c27628e/ansicolors-1.1.8.zip
certifi==2021.5.30 # certifi>=2017.4.17 via -> requests via -> humbug==0.2.6 via -> pantsbuild.pants \
    --hash:sha256=24b049496cc826bc483682c45a59ea82b8832326fc35166048e3c6113ccc7551 # file:///home/jsirois/dev/pantsbuild/jsirois-pex/find-links/certifi-2021.5.30-py2.py3-none-any.whl \
    --hash:sha256=50b1e4f8446b06f41be7dd6338db18e0990601dce795c2b1686458aa7e8fa7d8 # https://files.pythonhosted.org/packages/05/1b/0a0dece0e8aa492a6ec9e4ad2fe366b511558cdc73fd3abc82ba7348e875/certifi-2021.5.30-py2.py3-none-any.whl \
[...]

PR #1406, opened 2021-08-19
Extract URLFetcher for reuse with lock files.

jsirois:
Lock file generation will need to download and hash dists when their
hashes are not presented in PEP 503 link fragments and we'll also need
to download dists directly if we optimize the resolving from a lock file
case when the lock file contains urls ala PEP-665.

Work towards #1401.

### DIFF
--- a/pex/compatibility.py
+++ b/pex/compatibility.py
@@ -108,6 +108,7 @@ if PY3:
 
     from urllib.error import HTTPError as HTTPError
     from urllib.request import build_opener as build_opener
+    from urllib.request import FileHandler as FileHandler
     from urllib.request import HTTPSHandler as HTTPSHandler
     from urllib.request import ProxyHandler as ProxyHandler
     from urllib.request import Request as Request
@@ -115,6 +116,7 @@ else:
     import urlparse as urlparse
 
     from urllib2 import build_opener as build_opener
+    from urllib2 import FileHandler as FileHandler
     from urllib2 import HTTPError as HTTPError
     from urllib2 import HTTPSHandler as HTTPSHandler
     from urllib2 import ProxyHandler as ProxyHandler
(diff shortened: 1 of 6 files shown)


pex/fetcher.py
```diff
@@ -0,0 +1,90 @@
...
+        ssl_context = ssl.create_default_context(cafile=network_configuration.cert)
+        if network_configuration.client_cert:
+            ssl_context.load_cert_chain(network_configuration.client_cert)
+
+        proxies = None  # type: Optional[Dict[str, str]]
+        if network_configuration.proxy:
+            proxies = {protocol: network_configuration.proxy for protocol in ("http", "https")}
+
+        handlers = [ProxyHandler(proxies), HTTPSHandler(context=ssl_context)]
+        if handle_file_urls:
```
jsirois, line 41: The only changes to this class besides being moved from requirements.py to here are:
1. Drop "ftp" from the proxy protocols above since we don't install the FtpHandler / never intended to support plaintext FTP.
2. Add optional FileHandler support which will be needed / useful for handling --find-links "downloads" from local find-links repos in a way that keeps manual dist hashing code uniform / simpler.

pex/compatibility.py
```diff
@@ -108,13 +108,15 @@ def exec_function(ast, globals_map):
 
     from urllib.error import HTTPError as HTTPError
     from urllib.request import build_opener as build_opener
+    from urllib.request import FileHandler as FileHandler
```
jsirois, line 111: N.B.: On general Pex requirements for the uninitiated:
1. It still supports Python 2.7
2. It ships as both a distribution on PyPI and a Pex PEX via github releases.

2 makes it extremely unattractive to vendor requests since its platform-specific and the shipped Pex PEX - which is currently pure python - would need to grow ultiplatform support that included {linux wheels, and two flavors of macOS wheels} X {2.7, 3.5...3.9}.

Eric-Arellano (approved):
Thanks for splitting this out into prework :)

PR #1409, opened 2021-08-19
Improve examples and add example subsection titles

blag:
As a new user to PEX, this example is _really_ hard to parse and figure out what it does, because everything is named `pex`:

```bash
$ pex pex -c pex -o pex.pex
```

This PR just renames each part of that command to be explicit about what it's specifying:

```bash
$ pex pex -c pex_entry -o pex-executable.pex
```

I also added subsections with titles to the examples section for quick reference, and added an example of a `console_scripts` entry point in `setup.py` and a link to the Python packaging documentation to make it even easier for other users to grok the example and get started using pex immediately, instead of having to delve deeper into the Python packaging rabbithole.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -71,7 +71,7 @@ Or instead freeze your current virtualenv via requirements.txt and execute it an
 
 .. code-block:: bash
 
-    $ pex $(pip freeze) -o my_virtualenv.pex
+    $ pex -r requirements.txt -o my_virtualenv.pex
     $ deactivate
     $ ./my_virtualenv.pex
 
@@ -90,28 +90,22 @@ Launch Sphinx in an ephemeral pex environment using the Sphinx entry point ``sph
 
     $ pex sphinx -e sphinx:main -- --help
 
-Using ``setup.py`` Entry Points
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Using Entry Points
+~~~~~~~~~~~~~~~~~~
 
-Once you have a ``pex_entry`` entry point in the `console_scripts <https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html#the-console-scripts-entry-point>`_ section of your ``setup.py``:
+Projects specifying a ``console_scripts`` entry point (with
+`Flit <https://flit.readthedocs.io/en/latest/pyproject_toml.html#scripts-section>`_,
+`Poetry <https://python-poetry.org/docs/pyproject/#scripts>`_,
+or `setuptools <https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html#the-console-scripts-entry-point>`_)
+in their configuration can build standalone executables for those entry points.
 
-.. code-block:: python
-
-    setup(
-        ...
-        entry_points = {
-            'console_scripts': [
-                'pex_entry=my_package.my_module:main',
-            ],
-        }
-        ...
-    )
-
-you can build a standalone pex binary, including the ``pex`` package itself, into ``pex-executable.pex`` using the ``pex_entry`` entry point:
+To build a standalone ``pex-tools-executable.pex`` binary, including
+the ``pex`` package itself (version ``2.1.35`` or later), using the
+``pex-tools`` `entry point <https://github.com/pantsbuild/pex/blob/c92de90f5265c074ab66a225ad08d96ad045eccd/pyproject.toml#L42>`_:
 
 .. code-block:: bash
 
-    $ pex pex -c pex_entry -o pex-executable.pex
+    $ pex "pex>=2.1.35" --console-script pex-tools --output-file pex-tools-executable.pex
 
 Specifying A Specific Interpreter
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -84,17 +90,37 @@ Launch Sphinx in an ephemeral pex environment using the Sphinx entry point ``sph
...
+        }
+        ...
+    )
+
+you can build a standalone pex binary, including the ``pex`` package itself, into ``pex-executable.pex`` using the ``pex_entry`` entry point:
 
 .. code-block:: bash
 
-    $ pex pex -c pex -o pex.pex
+    $ pex pex -c pex_entry -o pex-executable.pex
```
jsirois, line 114: The `pex_entry` console script is not a valid entrypoint for the pex distribution in any release. How about:
```
$ pex "pex>=2.1.35" -c pex-tools ...
```
The pex-tools console script is not confusingly overloading the `pex` word and it is available starting only in Pex 2.1.35 which forces a `>=...` which further disambiguates tokens.

README.rst
```diff
@@ -84,17 +90,37 @@ Launch Sphinx in an ephemeral pex environment using the Sphinx entry point ``sph
 
     $ pex sphinx -e sphinx:main -- --help
 
-Build a standalone pex binary into ``pex.pex`` using the ``pex`` console_scripts entry point:
+Using ``setup.py`` Entry Points
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Once you have a ``pex_entry`` entry point in the `console_scripts <https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html#the-console-scripts-entry-point>`_ section of your ``setup.py``:
```
jsirois, line 96: I'm not sure about the extra detail of the entry_points snippet. It's true it links to the pypa - so how could that be wrong! But, the world is already shifting https://www.python.org/dev/peps/pep-0621/#entry-points

I'm not hard against the extra detail, but hopefully its clear this just might make things _more_ confusing for, say, someone new to Python that's only used Poetry from day 1 for example.

jsirois (commented):
Thanks @blag. Before I let CI rip, a few bits to wrestle with.

README.rst
jsirois: And that's defined here btw:
https://github.com/pantsbuild/pex/blob/c92de90f5265c074ab66a225ad08d96ad045eccd/pyproject.toml#L40-L42

jsirois: Between setup.py, Poetry, PEP-621 and this, you should now be doubled down on your python packaging ecosystem frustration!

jsirois: Another thing to consider is using `--console-script` instead of `-c` and `--output-file` instead of `-o` in this 1 example, say, and then flipping to the the short forms. Those explicit flags undoubtedly also help understand what is what.

README.rst
blag: Hmm, that's a good point. Would it help if I added a note that Flit, Poetry, and other tools handle specifying entry points differently?

As much as I hope the world moves to using PEP-621/`pyproject.toml` as fast as possible, I still see a lot of `setup.py` files out there. And I would like to include at least _one_ example and/or link to documentation.

README.rst
blag: Good call on all of those points. I'm a seasoned developer and I'm still confused and frustrated with Python packaging! 😆

I'll make those changes.

README.rst
blag: Eh, I removed it, but added links for easy access to the specific parts of the documentation.

README.rst
```diff
@@ -64,14 +64,20 @@ Launch an interpreter with ``requests``, ``flask`` and ``psutil`` in the environ
...
 
+Save Dependencies From Pip
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
 Or instead freeze your current virtualenv via requirements.txt and execute it anywhere:
 
 .. code-block:: bash
 
-    $ pex $(pip freeze) -o my_virtualenv.pex
+    $ pex -r requirements.txt -o my_virtualenv.pex
```
jsirois, line 74: Here you're sensibilites are trampling others. The `$(pip freeze)` was both contributed by another non-Pex developer trying to clarify docs _and_ -r requirements.txt may not exist in your virtualenv making the lead-in sentence in combination with this example not make sense.

README.rst
```diff
@@ -84,20 +90,36 @@ Launch Sphinx in an ephemeral pex environment using the Sphinx entry point ``sph
 
     $ pex sphinx -e sphinx:main -- --help
 
-Build a standalone pex binary into ``pex.pex`` using the ``pex`` console_scripts entry point:
+Using Entry Points
+~~~~~~~~~~~~~~~~~~
+
+Projects specifying a ``console_scripts`` entry point (with
```
jsirois, line 96: The `with` implies Pex only works with those three. It also works with PDM or any other PEP-517/PEP-51 build system.

I think you really have to drop the parenthese and take a few sentences to say something like "Examples of how a console script is configured in a projects build can be found here..". But I really think, at that point, with all the misleading bits finally mopped up, you've lost the thread and drifted pretty far off base. I'm very much not convinced it makes sense to re-document like this on behalf of what's effectively required required reading / prerequisite knowledge. Perhaps just stick to your first link and use lead-in language like "One way these are specified is described here: ....".

README.rst
```diff
@@ -84,20 +90,36 @@ Launch Sphinx in an ephemeral pex environment using the Sphinx entry point ``sph
...
+`Flit <https://flit.readthedocs.io/en/latest/pyproject_toml.html#scripts-section>`_,
+`Poetry <https://python-poetry.org/docs/pyproject/#scripts>`_,
+or `setuptools <https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html#the-console-scripts-entry-point>`_)
+in their configuration can build standalone executables for those entry points.
+
+To build a standalone ``pex-tools-executable.pex`` binary, including
```
jsirois, line 102: The `including` makes it sound like the `pex` distribution is being thrown in for some random purpose, but it's the vehicle of the `pex-tools` script. How about "To build a standalone ``pex-tools-executable.pex`` binary that runs the ``pex-tools`` console script found in all ``pex`` version ``2.1.35`` and newer distributions:"

README.rst
blag: Nah, I'll just remove it.

jsirois (approved):
Thanks @blag.

PR #1411, opened 2021-08-20
Support deterministic repository extract tool.

### LINKED ISSUE #1408: Reproducibility of `pex-tools extract`
`pex-tools $file extract -f $dest` currently creates output wheels with internal timestamps tied to their creation time, rather than zeroed or cloned from the producing PEX.

jsirois:
Wheels are now extracted with a deterministic timestamp by default but
the same option that the main Pex CLI has is added to control this.

Fixes #1408

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -7,7 +7,6 @@ import errno
 import filecmp
 import functools
 import glob
-import itertools
 import json
 import multiprocessing
 import os
@@ -23,15 +22,7 @@ from zipfile import ZipFile
 
 import pytest
 
-from pex.common import (
-    safe_copy,
-    safe_mkdir,
-    safe_open,
-    safe_rmtree,
-    safe_sleep,
-    temporary_dir,
-    touch,
-)
+from pex.common import safe_copy, safe_mkdir, safe_open, safe_rmtree, temporary_dir, touch
 from pex.compatibility import WINDOWS, to_bytes
 from pex.executor import Executor
 from pex.fetcher import URLFetcher
@@ -53,10 +44,13 @@ from pex.testing import (
     IntegResults,
     WheelBuilder,
     built_wheel,
+    create_pex_command,
     ensure_python_interpreter,
     ensure_python_venv,
     get_dep_dist_names_from_pex,
+    make_env,
     make_source_dir,
+    run_command_with_jitter,
     run_pex_command,
     run_simple_pex,
     run_simple_pex_test,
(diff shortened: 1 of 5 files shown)


tests/tools/commands/test_repository.py
```diff
@@ -128,7 +128,43 @@ def test_info_verbose(pex, pex_tools_env):
...
+        args=[pex, "repository", "extract", "-f", non_deterministic_dists_dir, "--use-system-time"],
+        env=pex_tools_env,
+    )
+    with open_zip(
+        os.path.join(non_deterministic_dists_dir, "requests-2.25.1-py2.py3-none-any.whl")
+    ) as zipfile:
+        infolist = zipfile.infolist()
+        assert len(infolist) > 0
+        for info in infolist:
+            assert deterministic_date_time != info.date_time
```
Eric-Arellano, line 164: Thoughts on checking more than this, like using `filecmp` on the `.whl` itself? Perhaps create two PEXes, extract wheels, and verify that every wheel is identical between the two PEXes.

Eric-Arellano (approved):

tests/tools/commands/test_repository.py
jsirois: I mean, if they aren't, what do you propose is the next step?

Eric-Arellano: I don't have any proposal for next steps: the PR title and this test both claim that it's reproducible, so I was trying to point out it would be helpful to back that up by checking the reproducibility is more than the timestamp.

Perhaps instead rename this test to be `test_extract_determinism_timestamp`?

jsirois: It turned out wheel pack is deterministic mod timestamp so the timestamp fix does produce deterministic wheels. Since the thorny question of what to do if that had not been true is dodged, I went ahead and added full-blown determinism tests.

tests/test_integration.py
```diff
@@ -1806,10 +1765,12 @@ def explode_pex(path):
...
 
-        pexes = create_multiple_pexes()
+        pexes = run_command_with_jitter(
+            path_argument="--output-file", args=create_pex_command(quiet=True), count=3
+        )
         pex_members = {pex: explode_pex(path=pex) for pex in pexes}
-
-        for pex1, pex2 in itertools.combinations(pexes, r=2):
+        pex1 = pexes.pop()
+        for pex2 in pexes:
```
jsirois, line 1773: N.B.: The LHS was always doing too much work. For `[0, 1, 2, ... N]`, to prove each pairing is equal you just need to compare `(0, 1), (0, 2) ... (0, N)` - transitivity handles the rest. Its only if you want to show that each element is unique that you need all combinations to compare different (used elsewhere in this PR).

stuhood (commented):
Thanks a lot!

Issue #1415, opened 2021-08-23
pex pip resolver fails if a dependency is specified differently for different platforms


vitalihs:
Hello there,

We recently bumped pex from 2.1.24 to 2.1.45 and now we observe problem with the following requirements file:

```
$ cat requirements.txt
https://<internal hostname>/pip/ray-0.8.4-cp37-cp37m-linux_x86_64.whl; sys_platform == 'linux'
ray==0.8.4; sys_platform == 'darwin'
```

when we call pex like this:

```
$ pex -r requirements.txt
Traceback (most recent call last):
  File "/Users/vitali/.pyenv/versions/pex-bug/bin/pex", line 8, in <module>
    sys.exit(main())
  File "/Users/vitali/.pyenv/versions/3.7.9/envs/pex-bug/lib/python3.7/site-packages/pex/bin/pex.py", line 1149, in main
    pex_builder = build_pex(options.requirements, options, cache=ENV.PEX_ROOT)
  File "/Users/vitali/.pyenv/versions/3.7.9/envs/pex-bug/lib/python3.7/site-packages/pex/bin/pex.py", line 1045, in build_pex
    ignore_errors=options.ignore_errors,
  File "/Users/vitali/.pyenv/versions/3.7.9/envs/pex-bug/lib/python3.7/site-packages/pex/resolver.py", line 1055, in resolve_multi
    ignore_errors=ignore_errors, workspace=workspace, max_parallel_jobs=max_parallel_jobs
  File "/Users/vitali/.pyenv/versions/3.7.9/envs/pex-bug/lib/python3.7/site-packages/pex/resolver.py", line 727, in install_distributions
    for index, direct_req in enumerate(direct_reqs)
AssertionError: More than one direct requirement is satisfied by ray 0.8.4:
0. ray
1. ray==0.8.4; sys_platform == "darwin"
This should never happen since Pip fails when more than one requirement for a given project name key is supplied and applies for a given target interpreter environment.
```

Let me provide details around our specifiers for `ray` package. For the linux environments (production installations run on linux) we use an internally built wheel, because we fixed a bug in ray which is still not fixed in their official version.
For local development on MacOS we're fine to work with the stock ray package, because this bug has only performance implications.

The workaround which I applied is to change our requirements file to download the ray package for both linux and darwin from a https url, something like this:

```
$ cat requirements.txt
[...]

PR #1416, opened 2021-08-23
Don't attempt to parse binary scripts.

jsirois:
It turns out some distributuions ship native executables using
distribution "scripts" (like py-spy), and these caused script re-writing
and execution to fail. Adds a failing integration test using py-spy that
this change fixes.

Follow-up on #1381.

### DIFF
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -3,6 +3,7 @@
 
 from __future__ import absolute_import, print_function
 
+import ast
 import os
 import re
 import sys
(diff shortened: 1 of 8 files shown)


jsirois:
Un-earthed by #1415 where ray has a transitive dep on py-spy.

Eric-Arellano (approved):

pex/pex.py
```diff
@@ -619,21 +620,22 @@ def execute_script(self, script_name):
...
+    def execute_external(binary):
         # type: (str) -> Any
+        args = [binary] + sys.argv[1:]
         try:
-            proc = Executor.open_process([cmd] + sys.argv[:1])
-            return proc.wait()
+            return Executor.open_process(args).wait()
         except Executor.ExecutionError as e:
-            return "Could not invoke script: {}".format(e)
+            return "Could not invoke script {}: {}".format(binary, e)
```
kaos, line 638: Did you see a error message here, where the binary name was missing?
I did not add a test for it, but I think that the error message I noticed included the name of the executable in the error message, now resulting in duplicated information in the error message.
However, I'm not sure if this will always be the case..

jsirois: Yeah, I'm not sure of that either. So, at the risk of duplicated information, I wanted to stuff what we know in their in case the layers down to the OS involved just say something like "file not found".

jsirois: To answer your question - I didn't see anything that directly prompted this change. The real prompt was the bug fixed above on line 634 where args were not being passed through correctly (`s/[:1]/[1:]/`). That ade me revisit this function wholesale.

kaos: Oh, doh. Good catch.

PR #1417, opened 2021-08-23
Fix Pip proprietary URL env marker handling.

### LINKED ISSUE #1415: pex pip resolver fails if a dependency is specified differently for different platforms

jsirois:
Add failing tests for an url with environment marker appended that work
under Pip and previously failed under Pex.

Fixes #1415

### DIFF
--- a/tests/test_resolver.py
+++ b/tests/test_resolver.py
@@ -4,6 +4,7 @@
 import functools
 import os
 import subprocess
+import sys
 from collections import defaultdict
 from textwrap import dedent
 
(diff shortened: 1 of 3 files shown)


tests/test_resolver.py
```diff
@@ -827,3 +828,26 @@ def test_resolve_from_pex_ignore_errors(
     assert len(resolved_distributions_by_key) > 1, "We should resolve at least requests and urllib3"
     assert "requests" in resolved_distributions_by_key
     assert Requirement.parse("urllib3==1.26.1") == resolved_distributions_by_key["urllib3"]
+
+
+def test_pip_proprietary_url_with_markers_issues_1415():
+    # https://<internal hostname>/pip/ray-0.8.4-cp37-cp37m-linux_x86_64.whl; sys_platform == 'linux'
+    # ray==0.8.4; sys_platform == 'darwin'
+    resolved_dists = resolve_multi(
```
jsirois, line 836: Prior to the fix, this failed much like the OP in #1415:
```console
E               AssertionError: More than one direct requirement is satisfied by ansicolors 1.1.8:
E               0. ansicolors==1.1.8
E               1. ansicolors==1.1.8; sys_platform == "linux"
E               This should never happen since Pip fails when more than one requirement for a given project name key is supplied and applies for a given target interpreter environment.
```

Eric-Arellano (approved):

PR #1420, opened 2021-08-23
Clean up resolver internal APIs.

### LINKED ISSUE #969: Merge resolve and resolve_multi

jsirois:
Kill ~unused resolve and rename resolve_multi to resolve.
Eliminate ResolvedDistribution alias in favor of InstalledDistribution
which better reflects the shap of the distribution (not a wheel, but an
installed wheel). Fix test-use fallout.

Work towards #1401.
Closes #969

### DIFF
--- a/tests/test_resolver.py
+++ b/tests/test_resolver.py
@@ -24,8 +24,8 @@ from pex.resolver import (
     Unsatisfiable,
     download,
     install,
+    resolve,
     resolve_from_pex,
-    resolve_multi,
 )
 from pex.testing import (
     IS_LINUX,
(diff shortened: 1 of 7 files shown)


tests/test_resolver.py
```diff
@@ -814,26 +818,26 @@ def test_resolve_from_pex_ignore_errors(
...
+    assert (
+        len(installed_distributions_by_key) > 1
+    ), "We should resolve at least requests and urllib3"
+    assert "requests" in installed_distributions_by_key
+    assert Requirement.parse("urllib3==1.26.1") == installed_distributions_by_key["urllib3"]
 
 
 def test_pip_proprietary_url_with_markers_issues_1415():
-    # https://<internal hostname>/pip/ray-0.8.4-cp37-cp37m-linux_x86_64.whl; sys_platform == 'linux'
-    # ray==0.8.4; sys_platform == 'darwin'
```
Eric-Arellano, line 835: Why remove this?

Eric-Arellano (approved):
Thanks for the new type hints, looks great

tests/test_resolver.py
jsirois: It was stray from earlier today - I just used it as cut/paste from the original issue in order to recreate it in the unit test. I noticed it too late in the prior PR and so I saved a tree by waiting unitl this PR to kill it.

PR #1426, opened 2021-08-28
Cleanup test warnings.

jsirois:
Two fixes:

1. Avoid warnings about universal newlines.

   These flood the end of the unit test log for all unit test environments
   except py27 and pypy2 as it stands.

2. Fixup an errant import of collections.Sequence.

### DIFF
--- a/pex/compatibility.py
+++ b/pex/compatibility.py
@@ -141,5 +141,8 @@ else:
     from Queue import Queue as Queue
     from multiprocessing import cpu_count as cpu_count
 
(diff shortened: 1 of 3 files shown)


jsirois:
The universal newline mode warnings can be seen here in a CI burn on main with 283 entries: https://github.com/pantsbuild/pex/runs/3405797418?check_suite_focus=true#step:7:1292

Eric-Arellano (approved):

Issue #1422, opened 2021-08-25
The --venv mode hash scheme can be insensitive to interpreter.


jsirois:
Summarized here:
```console
$ pex setuptools -osetuptools.pex --venv
$ python3.7 ./setuptools.pex -c 'import sys; print(".".join(map(str, sys.version_info[:2])))'
3.7
$ python3.8 ./setuptools.pex -c 'import sys; print(".".join(map(str, sys.version_info[:2])))'
3.7
$ rm -rf ~/.pex/venvs
$ python3.8 ./setuptools.pex -c 'import sys; print(".".join(map(str, sys.version_info[:2])))'
3.8
$ python3.7 ./setuptools.pex -c 'import sys; print(".".join(map(str, sys.version_info[:2])))'
3.8
$
```

PR #1427, opened 2021-08-29
Clarify PEX_PYTHON & PEX_PYTHON_PATH interaction.

### LINKED ISSUE #431: Kill off PEX_PYTHON
The pex runtime currently supports a configuration for a PEX_PYTHON variable in a pexrc file. This variable allows a user to define a Python Interpreter to override the default interpreter when executing the pex. 

This has been replaced by a PEX_PYTHON_PATH variable, that allows the user to specify a colon-separated list of paths to Python interpreters with the aim of being able support interpreter selection against a list of interpreter override candidates.

The goal of this issue is to completely kill off PEX_PYTHON support in the pex codebase in favor of the new PEX_PYTHON_PATH pexrc variable.

jsirois:
Previously, requesting a PEX_PYTHON with an absolute path to an
interpreter might not fail if the interpreter was invalid or if it was
valid but not on a requested PEX_PYTHON_PATH. Fix up both behaviors and
close #431 as no longer relevant or well guided.

Clarifies behaviors #1422 should support in interpreter selection
effects on `--venv` mode venv directory hashes.


pex/pex_bootstrapper.py
```diff
@@ -160,39 +160,68 @@ def _select_path_interpreter(
...
+        if ENV.PEX_PYTHON_PATH:
+            constraints.append("PEX_PYTHON_PATH={}".format(ENV.PEX_PYTHON_PATH))
+        if interpreter_constraints:
+            constraints.append("Version matches {}".format(" or ".join(interpreter_constraints)))
+        return constraints
+
+    preferred_interpreter = None  # type: Optional[PythonInterpreter]
+    if ENV.PEX_PYTHON and os.path.isabs(ENV.PEX_PYTHON):
+        try:
+            preferred_interpreter = PythonInterpreter.from_binary(ENV.PEX_PYTHON)
```
stuhood, line 209: As mentioned above, should this immediately error if there are `interpreter_constraints` which don't match?

stuhood (approved):
Thanks!

pex/pex_bootstrapper.py
stuhood: Whoops: stale. It does error slightly below.

PR #1428, opened 2021-08-29
Fix `--venv` mode venv dir hash.

### LINKED ISSUE #1422: The --venv mode hash scheme can be insensitive to interpreter.

jsirois:
Previously the venv dir hash locked unconstrained PEXes into a venv using
the interpreter they were first run with. Now, when a `--venv` mode PEX
is either unconstrained or only constrained by a PEX_PYTHON_PATH that
admits the interpreter being used to execute the `--venv` mode PEX, that
interpreter is recorded in the venv dir hash to differentiate a venv
dedicated to that interpreter.

Fixes #1422


jsirois:
Depends on #1427. Please review just 815c086.

pex/variables.py
```diff
@@ -668,45 +672,118 @@ def unzip_dir(
...
+    # former we hash Pex's selection criteria and for the latter we write down the actual
+    # interpreter the user selected. N.B.: The pex_hash already includes an encoding of the
+    # interpreter constraints, if any; so we don't hash those here again for the Pex chooses case.
+
+    # This is relevant in all cases of interpreter selection and restricts the valid search path
+    # for interpreter constraints, PEX_PYTHON and otherwise unconstrained PEXes.
+    venv_contents["PEX_PYTHON_PATH"] = ENV.PEX_PYTHON_PATH
+
+    interpreter_path = None  # type: Optional[str]
+    imprecise_pex_python = ENV.PEX_PYTHON and not os.path.isabs(ENV.PEX_PYTHON)
```
stuhood, line 716: Nit: should this check `and not os.path.exists(ENV.PEX_PYTHON)` to detect use of an interpreter in the current directory?

pex/variables.py
```diff
@@ -668,45 +672,118 @@ def unzip_dir(
...
-    venv_contents = {
-        "interpreter_constraints": sorted(interpreter_constraints),
-        "PEX_PYTHON": ENV.PEX_PYTHON,
-        "PEX_PYTHON_PATH": ENV.PEX_PYTHON_PATH,
-        "strip_pex_env": strip_pex_env,
-        "pex_path": {},
-    }  # type: Dict[str, Any]
+
+    # The venv contents are affected by which PEX files are in play as well as which interpreter
+    # is selected. The former is influenced via PEX_PATH and the is influenced by interpreter
```
stuhood, line 685: and the latter

pex/variables.py
```diff
@@ -668,45 +672,118 @@ def unzip_dir(
 
 
 def venv_dir(
```
stuhood, line 674: So, this consumes the constraints, the pex_path, and the interpreter rather than actually executing interpreter selection first, right? Is that for performance reasons? Because it seems like if you ran interpreter selection first, then the venv would always be keyed by the final interpreter, and the extra parameters here wouldn't be necessary.

pex/variables.py
```diff
@@ -668,45 +672,118 @@ def unzip_dir(
...
+    # for interpreter constraints, PEX_PYTHON and otherwise unconstrained PEXes.
+    venv_contents["PEX_PYTHON_PATH"] = ENV.PEX_PYTHON_PATH
+
+    interpreter_path = None  # type: Optional[str]
+    imprecise_pex_python = ENV.PEX_PYTHON and not os.path.isabs(ENV.PEX_PYTHON)
+    if imprecise_pex_python:
+        # For PEX_PYTHON=python3.7, for example, we just write down the constraint and let the PEX
+        # runtime determine the path of an appropriate python3.7, if any (Pex chooses).
+        venv_contents["PEX_PYTHON"] = ENV.PEX_PYTHON
+    elif not has_interpreter_constraints:
```
stuhood, line 721: If there is a `PEX_PYTHON`, can't you assume (based on the errors in #1427) that it will be used here, regardless of whether there are constraints? It might error if the constraints mismatch, but...

pex/variables.py
```diff
@@ -668,45 +672,118 @@ def unzip_dir(
...
+                """\
+                Using a venv selected by PEX_PYTHON={pex_python} for {pex_file} at {venv_path}.
+
+                If `{pex_python}` is upgraded or downgraded at some later date, this venv will still
+                be used. To force re-creation of the venv using the upgraded or downgraded
+                `{pex_python}` you will need to delete it at that point in time.
+
+                To avoid this warning, either specify a Python binary with major and minor version
+                in its name, like PEX_PYTHON=python{current_python_version} or else re-build the PEX
+                with `--no-emit-warnings` or re-run the PEX with PEX_EMIT_WARNINGS=False.
```
stuhood, line 759: Would it be worth fingerprinting the interpreter here instead, and failing? Or perhaps providing an option that the caller could pass to indicate "I fingerprinted it" (rather than disabling _all_ warnings)?

Also, how does the application of constraints from #1427 play in here? Is the idea that the constraints are only applied the first time the venv is created, but then never again? It seems like applying the constraints to the single fixed interpreter to check that it still matches would be pretty cheap.

Having said that, if at some point in the future the interpreter did shift, you'd error, rather than recovering. To avoid that, the version check would need to happen before selecting the venv dir...

stuhood (commented):
Thanks. Some questions about warning vs erroring/automatically recreating the venv.

pex/variables.py
jsirois: It is for performance reasons. Even importing `PythonInterpreter` is a ~20ms perf hit and we're talking re-exec  times of 50ms.

pex/variables.py
jsirois: You can't fingerprint the intepreter because you don't have one to fingerprint. This is something like `PEX_PYTHON=python`. So you have no clue what the selects without running path logic that is expensive - Pex rips through the path and actually uses PythonInterpreter to fingerprint interpreters, discarding invalid ones, etc.

To be clear, these two warnings were just added to be nice. The locked-in bugs they point out on future upgrades or path content changes were always present.

Also all warnings equals just these 2. The venv mode does not have pex warnings since there is no pex code in a venv! The main point of the venvs being to remove all Pex overhead.

stuhood: > You can't fingerprint the intepreter because you don't have one to fingerprint. This is something like PEX_PYTHON=python.

You can't do it here, but you could do it in `venv` startup code (`__main__.py` I think?).

stuhood (approved):

pex/variables.py
jsirois: > You can't do it here, but you could do it in venv startup code (__main__.py I think?).

Can you spell that out in detail? Keep in mind, the venv startup code is in a venv that has no Pex bootstrap code in it - just `__main__.py`. To boot, its _in_ the venv hash dir; so if it calculated a different answer for the hash, it would have to re-build the venv there and re-exec itself. Worse, it has to do this ever single time, not just the 1st.

stuhood: > Can you spell that out in detail? Keep in mind, the venv startup code is in a venv that has no Pex bootstrap code in it - just __main__.py. 

I started to type out something something like:
```
version = .'.join(map(str, sys.version_info[:3]))
if not any(ic.matches(version) for ic in interpreter_constraints):
    raise ...
```
... but I suppose that that matching code would require `setuptools`.

So yea, got it.

jsirois: Even if you could come up with the code - you still haven't detailed how you'd not create the new venv hash dir and re-exec there just 1 time and not suffer that penalty on every run of the venv pex.

stuhood: > Even if you could come up with the code - you still haven't detailed how you'd create the new venv hash dir and re-exec there just 1 time and not suffer that penalty on every run of the venv pex.

There were a few different aspects to my first comment: the first was "it would probably be better to error than to warn", and the other was "it would be better still to be able to automatically use a new `venv`". I thought that I had a potential suggestion of how to implement the former thing, but you pointed out that we don't have any bootstrapped code in the `venv`, so that's out. I never had a performant way to do the latter thing.

pex/variables.py
jsirois: Fixed.

pex/variables.py
jsirois: OK. What you can do is severly limited by how expensive both imports and exec'ing a PythonInterpreter are. In particular, not being able to tell an interpeters version given its binary name is a killer.

pex/variables.py
jsirois: That's right, fixed. Although no new test added.

pex/variables.py
jsirois: That sounds right. I need to think on this though. The logic is getting pretty intricate.

jsirois: Alright. Definitely correct. Thanks for the close look. Fixed.

PR #1430, opened 2021-08-29
Refactor integration test layout.

jsirois:
The `tests/test_integration.py` file was just shy of 3600 lines and was
starting to give smart editors problems. This change moves that file to
`tests/tests/test_integration.py` and breaks out specific issue tests to
their own files as well as breaking out three larger clusters of
behavior tests: interpreter selection, reexec and reproducible builds.
The result is `tests/integration/test_integration.py` is half its former
size. More breaking up of that file along behaviors tested can be done,
but the editor crisis is averted and a sustainable pattern set.


jsirois:
This was all pure moves with the exception of editing issue test test names to remove redundant `issues_XXX` bits.

Eric-Arellano (approved):

PR #1435, opened 2021-09-01
Add `SpawnedJob.and_then` API.

jsirois:
This allows collecting results from a Job's side-effects and will be used to
convert resolve information from Pip verbose logs into lock file data.

Work towards #1401.


Eric-Arellano (approved):
Looks great, thanks for the clear tests and type hints.

PR #1441, opened 2021-09-09
LockConfiguration strict: bool -> style: enum.

jsirois:
This makes way for cross-platform as a style to be added later but
a CLI option to be plumbed now.

### DIFF
--- a/pex/pip.py
+++ b/pex/pip.py
@@ -35,6 +35,7 @@ from pex.locked_resolve import (
     LockConfiguration,
     LockedRequirement,
     LockedResolve,
+    LockStyle,
     Pin,
 )
 from pex.network_configuration import NetworkConfiguration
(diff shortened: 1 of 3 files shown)


pex/pip.py
```diff
@@ -483,7 +484,7 @@ def analyze(self, line):
                     via=via,
                 )
             )
-        elif not self._lock_configuration.strict:
+        elif LockStyle.SOURCES == self._lock_configuration.style:
```
chrisjrn, line 487: any particular reason for leading with the enum value rather than the config variable? This reads like defensive Java coding to me rather than solid Python code

Eric-Arellano (approved):

chrisjrn (approved):
this seems fine

pex/pip.py
jsirois: Its a habit from tests - lead with expected value. Here - if you're used to that style, you can skip reading the RHS of the line - it puts the important bit 1st. Clearly eyes live in beholders though.

PR #1455, opened 2021-09-16
Factor out resolve configuration.

jsirois:
This sets the stage for PEX_TOOLs that can invoke resolves using the
standard CLI flags users have come to rely on.

Work towards #1401.


jsirois:
Lots of delta lines, but this is just a move of options setup out of `pex/bin/pex.py` to `pex/resolve/resolve_options.py` along with associated test moves.

Eric-Arellano (approved):

PR #1458, opened 2021-09-17
Refactor `PEX_TOOLS` `Command` infrastructure.

jsirois:
This will allow almost all of its machinery to be re-used for commands
that do not necessarily operate on a built PEX and using a different
`__main__` than `PEX_TOOLS`.

Work towards #1401.

### DIFF
--- a/pex/tools/command.py
+++ b/pex/tools/command.py
@@ -3,19 +3,26 @@
 
 from __future__ import absolute_import, print_function
 
+import functools
 import json
+import logging
 import os
 import subprocess
 import sys
-from argparse import ArgumentParser, Namespace
+from abc import abstractmethod
+from argparse import ArgumentDefaultsHelpFormatter, ArgumentParser, Namespace
 from contextlib import contextmanager
 
 from pex.common import safe_open
 from pex.pex import PEX
-from pex.typing import TYPE_CHECKING
+from pex.typing import TYPE_CHECKING, Generic, cast
+from pex.version import __version__
 
 if TYPE_CHECKING:
-    from typing import Any, Dict, Iterable, Iterator, IO, NoReturn, Optional
+    import attr  # vendor:skip
+    from typing import Any, Dict, Iterable, Iterator, IO, NoReturn, Optional, Type, TypeVar
+else:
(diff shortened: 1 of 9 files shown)


pex/tools/command.py
```diff
@@ -120,17 +128,29 @@ def show_help(
...
-        self,
-        pex,  # type: PEX
-        options,  # type: Namespace
-    ):
-        # type: (...) -> Result
-        pass
+    options = attr.ib()  # type: Namespace
+
+
+class PEXCommand(Command):
```
jsirois, line 149: Everything in this file, except this here PEXCommand will be able to be moved wholesale to the `pex.commands` package for re-use by the `pex3` cli. I'll do that pure move in a follow-up with a minimal pex3 CLI that just supports `-h`/`-V` to start.

jsirois:
N.B.: The existing PEX_TOOLS infra cannot be used to carry lockfile commands since those do resolving which pulls in the vendored Pip dependency. This would burden all `--include-tools` / `--venv` PEXes with many MB more code they will never need.  As such, carving out the future Pex CLI in a `pex3` console script that will later `pex3` -> `pex` with a deprecation period.

pex/tools/command.py
Eric-Arellano: pex3? I know we discussed that in meeting, but I thought you'll be using PEX_TOOLS for now. Is that in the roadmap for near future?

Eric-Arellano (approved):

jsirois:
> I know we discussed that in meeting, but I thought you'll be using PEX_TOOLS for now. Is that in the roadmap for near future?

See https://github.com/pantsbuild/pex/pull/1458#issuecomment-921356715, but PEX_TOOLS can't be used for the lock commands. I was sloppy in my thinking there, forgetting about the build-time nature of the new commands needed. All PEX_TOOLS command are runtime in that they operate on an already built PEX file - no Pip in sight.

Pex 3 is not on the roadmap in any sense except that it will exist at some point. I need a new console script so I have to pick a name, `pex3` seems to make sense. It will be thrown away of course, but for now.

Issue #1402, opened 2021-08-17
Implement platform agnostic requirements.txt lock file support.


jsirois:
Pex should be able to generate a lock file for all the Pythons and Platforms implicated by `--python`, `--interpreter-constraint` and `--platform` in a resolve in requirements.txt `--hash` format. The locked requirements.txt should only be constrained by the valid interpreter versions implied by `--python`, `--interpreter-constraint` and `--platform` and the input requirements and constraints.

jsirois:
Depends on #1401 for the lock datastructure and requirements.txt `--lock` emitting code.

jsirois:
The relevant Pip debug log line for the `download` command for foreign wheels is matched by:
```python
match = re.search(
    r"Skipping link: none of the wheel's tags \([^)]+\) are compatible \(run pip "
    r"debug --verbose to show compatible tags\): (?P<url>[^\s]+) ",
    line,
)
```

jsirois:
Notes:

There are legacy sdists whose setup.py asserts it's being run by an appropriate Python interpreter and this causes problems during a download where Pip is forced to run `python setup.py egg_info` to get dependency metadata (For a bit on why its forced to do this and can't use the PKG-INFO that ships in sdists, see [1][1]). An example of an sdist like this is functools32: https://github.com/michilu/python-functools32/blob/ad90fa86e2f4f494a3aedb0571274f3bbc6d7ab5/setup.py#L8-L10 This is a problem for universal locks. There are only two ways to solve afaict:
1. Patch pip to just use the PKG-INFO in the sdist when present and thus skip executing `python setup.py egg_info`. This has all the problems mentioned in [1][1].
2. Patch marker evaluation to consider interpreter constraints. Currently marker evaluation is patched to always return `True` so that all possible dependency branches for all possible target environments are gathered (thus forming a universal lock). If the resolve is constrained by an effective `Requires-Python` though, this implies a universe of `python_version` and `python_full_version` environment values that should evaluate `True` as well as the converse. So some way to both form these universes and then evaluate just those two markers and no others could trim out these problem dependencies in the 1st place.

\[1]: https://discuss.python.org/t/why-isnt-source-distribution-metadata-trustworthy-can-we-make-it-so/2620

[1]: https://discuss.python.org/t/why-isnt-source-distribution-metadata-trustworthy-can-we-make-it-so/2620 "PyPA Discussion"

PR #1464, opened 2021-09-21
Introduce the `pex3 lock create` command.

jsirois:
This provides a basic platform specific lock generation facility.

For now, the lock is only useful when a single platform or interpreter
is selected by the target options. A follow-up will introduce a lock
format that can store multiple locks allowing use and export of these
as individual units.

Work towards #1401.


pex/cli/commands/lock.py
```diff
@@ -0,0 +1,111 @@
...
+            "--style",
+            default=LockStyle.STRICT,
+            choices=LockStyle.values(),
+            type=LockStyle.for_value,
+            help=(
+                "The style of lock to generate. The {strict!r} style is the default and generates "
+                "a lock file that contains exactly the distributions that would be used in a local "
+                "resolve. If an sdist would be used, the sdist is included, but if a wheel would "
+                "be used, an accompanying sdist will not be included. The {sources} style includes "
+                "locks containing wheels and the associated sdists when available.".format(
```
Eric-Arellano, line 57: Will `strict` include wheels for other platforms? (https://github.com/pantsbuild/pants/issues/12458)

I traced the code back to this and can't tell:

https://github.com/pantsbuild/pex/blob/0ab02beb91b406563b88049fd470beca82a12d7d/pex/pip.py#L472-L478

Eric-Arellano (approved):
🎉

pex/cli/commands/lock.py
jsirois: It will not. This PRs issue, #1401, is all about platform specific locks. In order to get the Pip log data for other platforms, there are 4 Pip runtime patches that need to be applied and a new regex added here after that is done.

jsirois: So, I mean, yes - it will include wheels for other platforms when `--platform` is specified. I assumed you meant though for all possible other platforms, i.e.: totally platform agnostic. That's #1402.

Eric-Arellano: Yeah I did mean #1402. Cool, thanks for clarifying!

Issue #1414, opened 2021-08-22
Support selecting the appropriate lock for a resolve.


jsirois:
The most generic Pex resolves can have both multiple local interpreters in-play as well as multiple foreign platforms. It's likely the only reasonable was to store the locks for these is in multiple lockfiles or one lockfile with multiple sections. When the multi-lockfile complex later needs to be used to do a resolve, Pex will need to pick which lock file to use. This can likely be automatically determined with the following code from the locking experiment that was run:
```python
def select_lock(self, target):
    # type: (DistributionTarget) -> str
    with TRACER.timed("Picking locked resolve for {}".format(target)):
        supported_tags = target.get_supported_tags()
        ranked_resolves = {}  # type: Dict[Tuple[int, Platform], Iterable[BuiltArtifact]]
        for platform, resolve in self.resolves.items():
            resolve_rank = None  # type: Optional[int]
            for artifact in resolve:
                artifact_rank = None  # type: Optional[int]
                for tag in artifact.compatibility_tags:
                    try:
                        rank = supported_tags.index(tag)
                        artifact_rank = (
                            rank if artifact_rank is None else min(rank, artifact_rank)
                        )
                    except ValueError:
                        pass
                if artifact_rank is None:
                    resolve_rank = None
                    break
                resolve_rank = (
                    artifact_rank if resolve_rank is None else min(resolve_rank, artifact_rank)
                )
            if resolve_rank is not None:
                ranked_resolves[(resolve_rank, platform)] = resolve

        if not ranked_resolves:
            raise ValueError(
                "There were {} resolves but none of these support {}".format(
                    len(self.resolves), target
                )
            )

        (final_rank, platform), locked_requirements = sorted(
            # TODO(John Sirois): XXX: Explicit tie break for rank.
            ranked_resolves.items(),
            key=lambda item: item[0][0],
        )[0]
        TRACER.log(
            "Selected resolve generated by {} with a rank of {} from {}".format(
                platform, final_rank, ", ".join(sorted(map(str, self.resolves.keys())))
            )
        )
    return locked_requirements
```
Here the driving concept is to:
1. Confirm a given resolve is compatible by looking at each distribution in the resolve and confirming it has at least 1 artifact tha is compatible with the distribution target.
2. Giving each compatible resolve found in 1 a rank based on the specificity of its compatible artifacts tags.

The example above just used the most-specific tag, but a production implementation should probably use a global figure. Maybe the sum of all compatible artifacts most specific tags.

[...]

PR #1470, opened 2021-09-25
Make Enums sortable.

jsirois:
Upcoming lock file work will benefit from this. Instead of just sorting
by the string value of an `Enum.Value`, introduce an ordinal to sort by
and take the opportunity to dry up the values method.

Work towards #1401 and #1414.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -71,7 +71,8 @@ commands =
 [testenv:typecheck]
 deps =
     attrs==21.2.0  # This version should track the version in pex/vendor/__init__.py.
-    mypy==0.800
+    mypy[python2]==0.910
+    types-mock
 commands =
     bash scripts/typecheck.sh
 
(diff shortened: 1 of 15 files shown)


jsirois:
The 1st commit (57645f9) is from #1469 which this PR depends on. Please just review the second commit (65a8a6e).

tox.ini
```diff
@@ -71,7 +71,8 @@ commands =
 [testenv:typecheck]
 deps =
     attrs==21.2.0  # This version should track the version in pex/vendor/__init__.py.
-    mypy==0.800
+    mypy[python2]==0.910
```
jsirois, line 74: I mistakenly thought this upgrade was needed to support @total_ordering used in `Enum.Value` in this PR.

stuhood (approved):

Issue #1403, opened 2021-08-17
Consider implementing PEP-665 lock file format support.


jsirois:
[PEP-665](https://www.python.org/dev/peps/pep-0665/) is a nascent lock file format standard. It's specification currently supports both platform specific (#1401) and platform agnostic (#1402) locks. Support PEP-665 as an optional output format for Pex lock files and be able to consume these lock files as well instead of using `-r`.

There is separable work here.
1. Emit the lock file and be able to ingest it (not necessarily any PEP-665 lock, just the ones Pex generates). Do not leverage the url data, just let a normal locked resolve with hashes proceed.
2. a. Use the url, dependency and compatibility data in the lock to skip resolution (pip) altogether and just parallel-download the needed distributions. | b. Be able to consume any PEP-665 lock.

jsirois:
Depends on #1401 for lock datastructure.

jsirois:
Closing since PEP-665 was rejected: https://discuss.python.org/t/pep-665-take-2-a-file-format-to-list-python-dependencies-for-reproducibility-of-an-application/11736/140

PR #1471, opened 2021-09-26
Introduce the `pex3 lock export` command.

### LINKED ISSUE #1401: Implement platform specific requirements.txt lock file support.

### LINKED ISSUE #1414: Support selecting the appropriate lock for a resolve.

jsirois:
This switches `lock create` to store lock files in a Pex-specific format
that retains all information needed to later use portions of the lock.
The `export` subcommand then uses this facility to select an appropriate
lock and then translate it to Pip requirements format for export.

Closes #1401
Closes #1414


jsirois:
Thanks in advance - this is large. Even the individual commits are a bit, although conceptually each is tight. The trio of Lockfile, JSON code and the lock export command get a bit tangled in the tests; so it seemed best to stop all the PR splitting and land this final set of commits as a unit.

tests/cli/commands/lockfile/test_lockfile.py
```diff
@@ -0,0 +1,283 @@
...
+        with open(locked_requirements_file, "w") as fp:
+            selected[target].emit_requirements(fp)
+        return run_pex_command(
+            args=["-r", locked_requirements_file, "--", "-c", "import p537"],
+            python=target.get_interpreter().binary,
+        )
+
+    use_lock(py37).assert_success()
+
+    # N.B.: We created a lock above that falsely advertises there is a solution for Python 2.7.
```
jsirois, line 277: N.B.: Poetry has the same problem. It's inherent to doing anything but creating a strict lock.

pex/cli/commands/lockfile/json_codec.py
```diff
@@ -0,0 +1,283 @@
...
+                )
+            return value
+        except KeyError:
+            raise ParseError(
+                "The object at '{path}' in {source} did not have the expected key "
+                "{key!r}.".format(path=path, source=source, key=key)
+            )
+
+    try:
+        resolver_version = ResolverVersion.for_value(get("resolver_version"))
```
stuhood, line 103: 👍

pex/cli/commands/lock.py
```diff
@@ -3,28 +3,45 @@
...
+    from typing import List, DefaultDict
 
-class Lock(OutputMixin, BuildTimeCommand):
+
+class LockFormat(Enum["LockFormat.Value"]):
+    class Value(Enum.Value):
+        pass
+
+    PIP = Value("pip")
+    PEP_665 = Value("pep-665")
```
stuhood, line 36: So, is the implication of `PEP-665` being an export format that it didn't work as an internal representation of the file? That would likely be useful feedback for them.

stuhood (approved):
Thanks!

pex/cli/commands/lockfile/lockfile.py
```diff
@@ -0,0 +1,168 @@
...
+        Mapping,
+        Optional,
+        Tuple,
+    )
+else:
+    from pex.third_party import attr
+
+
+@attr.s(frozen=True)
+class _RankedLock(object):
```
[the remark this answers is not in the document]
kaos: Does the ranking described in #1414 answer that, perhaps?

pex/cli/commands/lock.py
jsirois: It is not, I was just making space for someone else to pick up #1403 in a clear way. The spec isn't real though yet, so at bottom, it makes most sense for Pex to store its lock with all the data it needs right now in a real format. The `pex_version` field in that format allows us to fully switch to PEP-665 as a native format if it makes sense in the future. That aside, the code doesn't currently produce enough data to produce a valid PEP 665 lock yet. That will need to combine the metadata from the downloaded local distributions in a Downloaded result with the LockedResolves found in a Downloaded to produce the needs / needed_by for each project.

pex/cli/commands/lockfile/lockfile.py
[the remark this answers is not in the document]
jsirois: Agreed. Warm up for adding the doc below in the form of the answers.

> From reading, it seems like you are generating multiple conceptual resolves which are each independently usable, and then selecting from them for a particular consumer.

Locks aside - this is fundamentally what a PEX is. It contains multiple resolves and selects the correct one at runtime. A lock file for a Pex then has to do the same thing.

> Does that mean there is never an exact match for the consumer?

No, although it's likely you don't have the correct notion of what exact match means. For a lock generated for --platform you can have an exact match if you export for the same --platform. And that's about it. Otherwise, the lock was generated for an interpreter and the only way to be sure the match is exact in that case is if the interpreter is the exact same interpreter, which means the exact same ordered set of supported tags and the exact same environment marker environment dict.

That aside and accepting the vague notion of exact match, if you did only support exact matches, then you could never support LockStyle.SOURCES or the forthcoming LockStyle.CROSS_PLATFORM. Even those aside, with a LockStyle.STRICT for a set of requirements that encloses a - say - universal wheel set - you'd not want to fail an export just because the target being used on this machine doesn't match the one on the machine used to generate the lock in the 1st place.

pex/cli/commands/lock.py
jsirois: To put a finer point on this: way back when PEP 665 was brought up I reviewed it and the thread at that time. My conclusion then was that it covered everything Pex needed. That conclusion was based on my lock experiments - so real data - and has not changed.

stuhood: > It is not, I was just making space for someone else to pick up #1403 in a clear way. The spec isn't real though yet, so at bottom, it makes most sense for Pex to store its lock with all the data it needs right now in a real format. The `pex_version` field in that format allows us to fully switch to PEP-665 as a native format if it makes sense in the future.

Chickens and eggs. =)

> That aside, the code doesn't currently produce enough data to produce a valid PEP 665 lock yet. That will need to combine the metadata from the downloaded local distributions in a Downloaded result with the LockedResolves found in a Downloaded to produce the needs / needed_by for each project.

Got it. Thanks!

jsirois: > Chickens and eggs. =)

I'm not ruling out being God here and upending chickens and eggs, its just that it seems most appropriate to do this via an _export_ to PEP-665 being the 1st instance of it in the wild and not the fundamental underlying lock format for Pex. Hopefully that makes sense and you agree its the only sane way to proceed with vaporware.

stuhood: Yea, not a blocking comment. Talked with @chrisjrn about this at some point, but even if PEP-665 had had a reference implementation of the format, it's unlikely that that reference implementation would have been drop-in for any consumers (PEX included), so the vaporware-ness seems like an optimistic guess that someone will try implementing the spec and report back about whether it was possible/useful.

Eric-Arellano:
(I plan to look tomorrow. Feel free to merge without me and I'll still leave feedback for follow up)

pex/cli/commands/lock.py
jsirois: For a bit of clarity I'll rename this to `ExportFormat`.  The export does not even need to be to a proper lock, it could be to a pinned constraints file for example.

Eric-Arellano (approved):
This looks really excellent. Great tests and very helpful docstring

tests/cli/commands/lockfile/test_json_codec.py
```diff
@@ -0,0 +1,528 @@
...
+
+
+def test_roundtrip(tmpdir):
+    # type: (Any) -> None
+
+    lockfile = Lockfile.create(
+        pex_version="1.2.3",
+        resolver_version=ResolverVersion.PIP_2020,
+        requirements=(
+            Requirement.parse("ansicolors"),
```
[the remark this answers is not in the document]
jsirois: I think you're missing the point of the pointless test values here. My error is in hewing too close to the norm. I should be using foo, bar, baz or Seinfeld characters to really underscore the values absolutely don't matter for the purposes of this test. Only that the values round-trip matters.

tests/cli/commands/lockfile/test_json_codec.py
```diff
@@ -0,0 +1,528 @@
...
+                                url="file://find-links/requests-2.0.0.tar.gz",
+                                fingerprint=Fingerprint(algorithm="sha512", hash="123"),
+                            ),
+                        ),
+                        via=("direct", "from", "a", "test"),
+                    ),
+                ],
+            ),
+            LockedResolve.from_platform_tag(
+                platform_tag=tags.Tag("cp37", "cp37m", "manylinux1_x86_64"),
```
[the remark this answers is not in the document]
jsirois: Ditto.

pex/cli/commands/lock.py
```diff
@@ -76,11 +126,43 @@ def add_extra_arguments(
...
+                requirements.append(parsed_requirement.requirement)
+        if local_projects:
+            return Error(
+                "Cannot create a lock for local project requirements. Given {count}:\n"
+                "{projects}".format(
+                    count=len(local_projects),
+                    projects="\n".join(
+                        "{index}.) {project}".format(index=index, project=project.path)
+                        for index, project in enumerate(local_projects, start=1)
+                    ),
```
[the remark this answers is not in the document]
jsirois: I do too - sometimes apparently. Sometimes I do this and sometimes a numbered list. What I don't like is arbitrariness and fashion. What I should do is introduce a helper that creates a list using my whim style of the day that day and then just always use the utility instead of dithering with style back and forth every other week.

tests/cli/commands/lockfile/test_json_codec.py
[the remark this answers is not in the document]
Eric-Arellano: Agreed! Not blocking, but might be worth doing that. Feel free to include in a related follow up to save a tree.

Issue #1412, opened 2021-08-22
Support lock file updates


jsirois:
Given a lockfile and the original set of input requirements (which should likely be encoded in the lockfile), Pex should allow updating the lock file in part or in full. In full is easy, that's just re-running lock generation using the original requirements. It turns out in part is straight-forward as well. Just take a list of project names that should be updated (these must appear in the lock) and create a pinned constraints file from the lock with all projects _except_ those to be updated and include that constraints file in the resolve. The only corner to deal with is when the original lock also contains constraints. If there is any overlap, Pex can fail fast.

jsirois:
Depends on #1401 for the basic CLI lock request option and infra.

jsirois:
Instead of failing fast for constraint conflict on update I'm going with a warning. If you're specifically asking for `pex3 lock update -p urllib3 lock.json` you probably really mean it even if the original `lock.json` has a constraint on `urllib3`.

jsirois:
OK, this worked out better than hoped. You can update the whole lock, update as per cargo with `-p <project1> -p <project2> ...` and upgrade or downgrade with more control with `-p <project><specifier>`, e.g.: `-p 'urllib3<1.26'`

PR #1472, opened 2021-09-27
Introduce try_/catch for Error handling.

jsirois:
This allows use of Error for short circuiting computations without all
the boilerplate of try/except.

Work towards #1412.

### DIFF
--- a/pex/commands/command.py
+++ b/pex/commands/command.py
@@ -28,30 +28,32 @@ from pex.version import __version__
 
 if TYPE_CHECKING:
     import attr  # vendor:skip
-    from typing import Any, Dict, Iterable, Iterator, IO, NoReturn, Optional, Type, TypeVar
+    from typing import (
+        Any,
+        Callable,
+        Dict,
+        Iterable,
+        Iterator,
+        IO,
+        NoReturn,
+        Optional,
+        Type,
+        TypeVar,
+        Union,
+    )
 else:
     from pex.third_party import attr
 
 
+@attr.s(frozen=True)
 class Result(object):
-    def __init__(
-        self,
-        exit_code,  # type: int
-        message="",  # type: str
-    ):
-        # type: (...) -> None
-        self._exit_code = exit_code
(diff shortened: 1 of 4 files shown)


pex/commands/command.py
```diff
@@ -88,6 +84,53 @@ def __init__(
...
+
+@attr.s(frozen=True)
+class ResultError(Exception):
+    """Wraps an Error in an exception for use in short-circuiting via `try_` / `catch`."""
+
+    error = attr.ib()  # type: Error
+
+
+def try_(result):
+    # type: (Union[_T, Error]) -> _T
```
Eric-Arellano, line 99: How about calling it `unwrap()` from Rust? Seems more descriptive and avoids confusion with the std lib `try`.

Also you could avoid the coupling with `catch` in the docstring - it's often helpful to combine `catch()` and `unwrap()`, but it's not necessary.

pex/commands/command.py
```diff
@@ -88,6 +84,53 @@ def __init__(
...
+    :param result: a result containing a successful value or else an error.
+    :return: the result if not an error.
+    :raise: :class:`ResultError` if the result is an error.
+    """
+    if isinstance(result, Error):
+        raise ResultError(error=result)
+    return result
+
+
+def catch(
```
Eric-Arellano, line 114: This name makes sense to me.

pex/commands/command.py
jsirois: The `try_` is actually from rust! Before `?` there was only the `try!` macro.

jsirois: And this unwrap does nothing like the panic a non-Ok Result unwrap does. I'll think on alternative names.

kaos: As one unfamiliar with rust, I don't see the connection between them wrap and catch.. I think try_ makes sense, as it models a similar behavior as the standard try/catch.

Eric-Arellano (approved):

PR #1478, opened 2021-10-01
Introduce the `pex3 lock update` command.

### LINKED ISSUE #1412: Support lock file updates

jsirois:
Invoking `pex3 lock update <lockfile>` re-creates a lockfile using its
original resolve settings allowing freshening a locked resolve with all
the latest compatible distributions en-masse.

Invoking `pex3 lock update -p <project> [-p <project> ...] <lockfile>`
allows doing so, but only for targeted projects used by the locked
resolve. This can be useful to perform a conservative compatible upgrade
or downgrade to pick up a security fix or back away from a new exploit
respectively.

Closes #1412.


pex/cli/commands/lock.py
```diff
@@ -115,6 +127,47 @@ def _add_export_arguments(cls, export_parser):
...
+        update_parser.add_argument(
+            "--strict",
+            "--no-strict",
+            "--non-strict",
+            action=HandleBoolAction,
+            default=True,
+            type=bool,
+            help=(
+                "Require all target platforms in the lock be updated at once. If any target "
+                "platform in the lock file does not have a representative local interpreter to"
```
stuhood, line 151:
```suggestion
                "platform in the lock file does not have a representative local interpreter to "
```

pex/cli/commands/lock.py
```diff
@@ -220,3 +277,149 @@ def _export(self):
...
+                )
+
+        targets_to_update = OrderedDict()  # type: OrderedDict[DistributionTarget, LockedResolve]
+        for locked_resolve in lock_file.locked_resolves:
+            target = targets_by_platform_tag.get(locked_resolve.platform_tag)
+            if target:
+                targets_to_update[target] = locked_resolve
+
+        if not targets_to_update:
+            return Ok()
```
stuhood, line 321: Should this fail instead with "no targets matched"?

pex/cli/commands/lock.py
```diff
@@ -220,3 +277,149 @@ def _export(self):
...
+                    )
+        if not dry_run:
+            lockfile.store(
+                attr.evolve(
+                    lock_file,
+                    pex_version=__version__,
+                    locked_resolves=SortedTuple(locked_resolve_by_platform.values()),
+                ),
+                lockfile_path,
+            )
```
stuhood, line 424: The render+commit step here at the end might be good in a separate function, perhaps one that calls the non-sideffecting bits from the first 2/3 of the function?

pex/cli/commands/lockfile/updater.py
```diff
@@ -0,0 +1,205 @@
...
+    When a set of updates is specified, the updater ensures all other project distributions remain
+    unchanged.
+    """
+
+    @classmethod
+    def create(
+        cls,
+        requirements,  # type: Iterable[Requirement]
+        constraints,  # type: Iterable[Requirement]
+        updates,  # type: Iterable[Requirement]
```
stuhood, line 69: Is it assumed that `updates` are always a (fuzzy) subset of `requirements`? Or will this do the right thing if given new requirements that did not exist in the set before?

pex/cli/commands/lockfile/updater.py
```diff
@@ -0,0 +1,205 @@
...
+        for update in updates:
+            constraint_req = Requirement.parse(update)
+            project_name = ProjectName(constraint_req.project_name)
+            original_constraint = original_constraints.get(project_name)
+            if original_constraint:
+                logger.warning(
+                    "Over-riding original constraint {original} with {override}.".format(
+                        original=original_constraint, override=constraint_req
+                    )
+                )
```
stuhood, line 89: Should this explode? If you were using a lockfile, it would imply that your constraints file was not being used for that usecase, and instead would likely be a hand-written thing used to pin down your transitive deps: overwriting it here would be strange.

Alternatively, should it try to merge the requirements, and fail if it can't?

pex/cli/commands/lockfile/updater.py
```diff
@@ -0,0 +1,205 @@
...
+            )
+
+        assert 1 == len(updated_lock_file.locked_resolves)
+        updated_resolve = updated_lock_file.locked_resolves[0]
+
+        updated_requirements_by_project_name = OrderedDict(
+            (updated_requirement.pin.project_name, updated_requirement)
+            for updated_requirement in updated_resolve.locked_requirements
+        )  # type: OrderedDict[ProjectName, LockedRequirement]
```
stuhood, line 166: Will this handle deleting a `LockedRequirement` if it is no longer needed?

stuhood (approved):
Thanks!

pex/cli/commands/lockfile/updater.py
jsirois: Yes. That's covered by this test: https://github.com/pantsbuild/pex/pull/1478/files#diff-8c268962cf5d39934d82d9fe9fe14fc7032ef0df2f946c5077d4c40ca0578bdfR310

pex/cli/commands/lockfile/updater.py
jsirois: > Should this explode? If you were using a lockfile, it would imply that your constraints file was not being used for that usecase, and instead would likely be a hand-written thing used to pin down your transitive deps: overwriting it here would be strange.

It doesn't seem strange to me:

1. You use a constraints file and create a lock on some day.
2. Later, you learn one requirement you over-zealously had pinned in the constraints file has a security issue and by upgrading or downgrading you can avoid the security issue and still maintain compatibility.

So you use `-p` to over-ride the constraint. Top actually pass in a new, edited constraints file means re-running the whole lock and you don't want to do that.

What is strange is that the constraints recorded in the lock file are not edited by the over-ride. I'll fix that.

pex/cli/commands/lockfile/updater.py
jsirois: I'm not sure what the right thing is. Right now, disjoint updates will just go unused. The update mechanism is just to re-run the resolve using the original requirements + a full set of constraints covering the previous lock pinned transitive closure with an override for each `-p`. So if a `-p` is not in the original transitive closure, it still get written into the constraints file; it will just never be used in the course of the resolve.

pex/cli/commands/lock.py
jsirois: Perhaps. Succeed or fail though, a message calling the no-op out is definitely needed.

pex/cli/commands/lock.py
jsirois: Fixed.

pex/cli/commands/lockfile/updater.py
Eric-Arellano: That seems fine, although probably worth documenting in `-p` that values for requirements not in the lockfile will be ignored.

Eric-Arellano (approved):
Really neat stuff

pex/cli/commands/lock.py
jsirois: Kept this Ok, but added a message indicating no lock update was performed and how to possibly change this by tweaking --python-path.

pex/cli/commands/lock.py
jsirois: Re-factored.

pex/cli/commands/lockfile/updater.py
jsirois: Ok, I used the precedent set by cargo here and now fail fast.

Eric-Arellano (approved):

PR #1495, opened 2021-10-29
Introduce the `PEX` env var.

### LINKED ISSUE #1485: How to know whether we are running from within pex
Hello there!
I recently bumped the pex version of our project and figured out that the hack (import _pex) that we were using to detect whether or not we are running from within a pex does not work anymore. Here is the initial discussion about it started in 2019 by one of my colleagues: https://github.com/pantsbuild/pex/issues/721
Is there a new way to detect that we are in a pex ?
Thanks in advance.

jsirois:
This is now always exposed to Pex applications so they can uniformly
detect that they are running from a PEX vs. from a virtual environment
or similar.

Closes #1485

### DIFF
--- a/docs/recipes.rst
+++ b/docs/recipes.rst
@@ -3,6 +3,14 @@
 PEX Recipes and Notes
 =====================
 
+PEX-aware application
+---------------------
+
+If your code benefits from knowing whether it is running from within a PEX or not, you can inspect
+the ``PEX`` environment variable. If it is set, it will be the absolute path of the PEX your code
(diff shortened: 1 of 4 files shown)


jsirois:
Hopefully this does the trick @nateagr.

jsirois:
The CI break is un-related. Fixing in #1496 and will re-base.

Eric-Arellano (approved):
Thanks!

nateagr:
> Hopefully this does the trick @nateagr.

Perfect! Thanks a lot @jsirois

PR #1106, opened 2020-11-07
Clear any prior vendorings when (re)vendoring.

jsirois:
This allows for re-vendored distributions losing files as well as changing
the set of vendored distributions altogether.

Fixes #1096

PR #1496, opened 2021-10-29
Fix vendoring constraints.

jsirois:
Ever since #1106 these were not being used.


jsirois:
Errors like this popped up since our vendored packaging 20.8 depends on `pyparsing>=2.0.2` and pyparsing ~just released a breaking 3.x series: https://github.com/pantsbuild/pex/runs/4050132167?check_suite_focus=true#step:4:284

Eric-Arellano (approved):

PR #1497, opened 2021-10-29
Add a fan-in aggregate status step.

jsirois:
This will be used as the single status check in GH branch protection
rules.


jsirois:
An example of failure is here - I used a cancel to test: https://github.com/pantsbuild/pex/actions/runs/1400792877

jsirois:
This also needs #1496 - macos Python 3.9 IT shard keeps timing our over there :/...

Eric-Arellano (approved):

Issue #1284, opened 2021-03-22
Consider removing support for `--requirements-pex`.


jsirois:
This was likely only ever used by Pants which has moved away from it with https://github.com/pantsbuild/pants/pull/11773.

The equivalent functionality can be had with --pex-repository for the single requirements pex case with a bit of juggling:
```
pex --pex-repository requirements.pex $(
  pex-tools requests.pex repository info -v | jq -r '.project_name + "==" + .version'
) ...
```

PR #1517, opened 2021-11-11
pex builder dup dist resilience

kaos:
Avoid a `FileExists` error when using symlink copy mode, in case a dist is present in a pex loaded using the `--requirements-pex` option is loaded again from other sources.

closes pantsbuild/pants#13587.


jsirois:
N.B.: #1284. I'm going to look over in Pants to see why this is still being used / if there is an alternative.

kaos:
I'll close this, as there'll be a fix for this directly in Pants instead.

PR #1532, opened 2021-12-13
Improve space efficiency of `--venv` mode.

jsirois:
Previously `--venv` mode auto-venvs were constructed just like PEX_TOOLS
venvs - using hard links (or copies if needed). Although this is
required for robust PEX_TOOLS venvs that don't collapse when you delete
the PEX_ROOT, it is not required for `--venv` mode auto-venvs which are
themselves stored in the PEX_ROOT. Now we symlink all `--venv` mode
auto-venv 3rdparty distributions to save a minor amount of space over
hard linking individual `.py` files, but a significant amount of space
in the shared `.pyc` files generated when the venv is later executed in
the case where many venvs share 3rdparty distribution dependencies and
interpreter major/minor versions.

### DIFF
--- a/pex/compatibility.py
+++ b/pex/compatibility.py
@@ -7,6 +7,7 @@
 from __future__ import absolute_import
 
 import os
+import re
 import sys
 from abc import ABCMeta
 from io import BytesIO
@@ -153,3 +154,16 @@ MODE_READ_UNIVERSAL_NEWLINES = "rU" if PY2 else "r"
 def get_stdout_bytes_buffer():
     # type: () -> BytesIO
     return cast(BytesIO, getattr(sys.stdout, "buffer", sys.stdout))
+
+
+if PY3:
+    is_valid_python_identifier = str.isidentifier
+else:
+
+    def is_valid_python_identifier(text):
+        # type: (str) -> bool
+
+        # N.B.: Python 2.7 only supports ASCII characters so the check is easy and this is probably
+        # why it's nt in the stdlib.
+        # See: https://docs.python.org/2.7/reference/lexical_analysis.html#identifiers
+        return re.match(r"^[_a-zA-Z][_a-zA-Z0-9]*$", text) is not None
--- a/pex/tools/commands/venv.py
+++ b/pex/tools/commands/venv.py
@@ -1,21 +1,24 @@
 # Copyright 2020 Pants project contributors (see CONTRIBUTORS.md).
 # Licensed under the Apache License, Version 2.0 (see LICENSE).
 
-from __future__ import absolute_import
+from __future__ import absolute_import, print_function
 
 import errno
+import itertools
 import logging
 import os
 import shutil
 from argparse import ArgumentParser
-from collections import defaultdict
+from collections import Counter, defaultdict
 from textwrap import dedent
 
 from pex import pex_warnings
 from pex.commands.command import Error, Ok, Result
 from pex.common import chmod_plus_x, pluralize, safe_delete, safe_mkdir, safe_rmtree
+from pex.compatibility import is_valid_python_identifier
 from pex.enum import Enum
 from pex.environment import PEXEnvironment
+from pex.orderedset import OrderedSet
 from pex.pex import PEX
 from pex.tools.command import PEXCommand
 from pex.tools.commands.virtualenv import PipUnavailableError, Virtualenv
@@ -24,9 +27,20 @@ from pex.typing import TYPE_CHECKING
 from pex.venv_bin_path import BinPath
 
 if TYPE_CHECKING:
+    import typing
     from typing import Iterable, Iterator, Optional, Tuple
 
 logger = logging.getLogger(__name__)
+
+
+def _relative_symlink(
+    src,  # type: str
+    dst,  # type: str
+):
+    # type: (...) -> None
+    dst_parent = os.path.dirname(dst)
+    rel_src = os.path.relpath(src, dst_parent)
+    os.symlink(rel_src, dst)
 
 
 # N.B.: We can't use shutil.copytree since we copy from multiple source locations to the same site
@@ -36,6 +50,7 @@ def _copytree(
     src,  # type: str
     dst,  # type: str
     exclude=(),  # type: Tuple[str, ...]
+    symlink=False,  # type: bool
 ):
     # type: (...) -> Iterator[Tuple[str, str]]
     safe_mkdir(dst)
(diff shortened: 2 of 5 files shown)


jsirois:
The example I used was from toolchainlabs Pants integration running `./pants test ::` after removing the lmdb and Pex caches. This is a case that generates many `--venv` mode test PEXes that share a significant number of distribution dependencies.

Before:
```
$ du -sh ~/.cache/pants/lmdb_store ~/.cache/pants/named_caches/pex_root
179M	/home/jsirois/.cache/pants/lmdb_store
3.4G	/home/jsirois/.cache/pants/named_caches/pex_root
```
After:
```
$ du -sh ~/.cache/pants/lmdb_store.tc.venv_symlinks ~/.cache/pants/named_caches/pex_root.tc.venv_symlinks
179M	/home/jsirois/.cache/pants/lmdb_store.tc.venv_symlinks
1.1G	/home/jsirois/.cache/pants/named_caches/pex_root.tc.venv_symlinks
```

jsirois:
Hrm, failing under CPython 2.7 only. Works For PyPy2, PyPy3 and CPython 3.5+. Digging...

pex/compatibility.py
```diff
@@ -153,3 +154,16 @@ def cpu_count():
...
+
+if PY3:
+    is_valid_python_identifier = str.isidentifier
+else:
+
+    def is_valid_python_identifier(text):
+        # type: (str) -> bool
+
+        # N.B.: Python 2.7 only supports ASCII characters so the check is easy and this is probably
+        # why it's nt in the stdlib.
```
kaos, line 167: In case there is another commit.. here's a typo `nt => not` ;)

pex/tools/commands/venv.py
```diff
@@ -45,34 +60,38 @@ def _copytree(
...
+                                raise e
+                            link = False
+                    shutil.copy(src_entry, dst_entry)
             except OSError as e:
                 if e.errno != errno.EEXIST:
                     raise e
 
+        if symlink:
+            # Once we've symlinked the top-level directories and files, we've "copied" everything.
+            return
```
kaos, line 93: What is the purpose of this `if` ?

jsirois: This function is called in one of two modes. If `symlink` is False (the old way), the src is copied to dst, but to save space, hard links are used when possible. When you do this you need to do a full fs walk all the way down src. If `symlink` is `True` then symlinks are used, and to symlink the contents of src you need only symlink each top level src member; then you're done and you can return early without walking further down the src/ tree.

jsirois: That really just re-states the comment in 5x as many words. Maybe drawing things will help - that helps me.

kaos: Yes, I was aware of that. I had to open up the whole file to spot my "mistake". I failed to notice that this if-block was in the scope of the `for` loop. Sorry for the noise :)

kaos: Heh, I actually see now, that there's a two nested for loops, but only one of those are shown in the diff...

jsirois: Ah, yeah. I pretty much <3 GitHub Diffs as much as I <3 macOS.

stuhood: This bit (only copying/linking the root directories, rather than every inner node) is one reason why I think that this improves space usage: you're right that hardlinking did not create new inodes per file, but it did create new inodes per directory, so the directory listings ended up fully duplicated.

Either way: nice!

pex/tools/commands/venv.py
```diff
@@ -24,18 +27,30 @@
...
+
 # N.B.: We can't use shutil.copytree since we copy from multiple source locations to the same site
 # packages directory destination. Since we're forced to stray from the stdlib here, support for
 # hardlinks is added to provide a measurable speed up and disk space savings when possible.
 def _copytree(
     src,  # type: str
     dst,  # type: str
     exclude=(),  # type: Tuple[str, ...]
+    symlink=False,  # type: bool
 ):
```
stuhood, line 54: Why copy the internals of the `src` and `dst` rather than directly creating `dst` as a symlink? In cases where we are creating symlinks, it's not like we can delete `src` afterward, so `src` must be long lived.

...if this is explained below in the "In the symlink case, in order to share all generated" comment, then I'm not quite understanding it.

pex/tools/commands/venv.py
```diff
@@ -99,25 +119,83 @@ def record_provenance(src_to_dst):
...
+            symlink=False,
         )
     )
 
     with open(os.path.join(venv.venv_dir, pex_info.PATH), "w") as fp:
         fp.write(pex_info.dump())
 
+    # Since the pex distributions are all materialized to ~/.pex/installed_wheels, which we control,
+    # we can optionally symlink to take advantage of sharing generated *.pyc files for auto-venvs
+    # created in ~/.pex/venvs.
```
stuhood, line 139: These symlinks are created entirely within `~/.pex`, right? That seems safe.

Will this impact `pex-tools venv` creating a venv in a user specified location? In that case it might be worth considering whether hardlinks still make sense.

stuhood (approved):
Thanks!

pex/tools/commands/venv.py
jsirois: If you track callers, the `pex-tools` venv does still use hardlinks/copies.

pex/tools/commands/venv.py
jsirois: Its not explained, but backing up to the locally assumed domain specific knowledge of a venv and site-packages, `dst` is `site-packages` and `src` is some installed wheel chroot. Those typically look like:
```
ls -l ~/.pex/installed_wheels/23de6e1192493377da687afdbb2a2cf11e6d613a/Django-3.2.10-py3-none-any.whl/
total 12
drwxr-xr-x  2 jsirois jsirois 4096 Dec 13 10:26 bin
drwxr-xr-x 18 jsirois jsirois 4096 Dec 13 10:25 django
drwxr-xr-x  2 jsirois jsirois 4096 Dec 13 10:26 Django-3.2.10.dist-info
```
So when you symlink this into `site-packages` you need:
```
site-packages/
  django -> ...
  Django-3.2.10.dist-info -> ...
  foo_dist -> ...
  foo-dist-1.0.0.dist-info -> ...
```
  
So, basically, each `src` has more than one item that must be linked into one big flat `dst`. There are many `src`s.

jsirois:
> This bit (only copying/linking the root directories, rather than every inner node) is one reason why I think that this improves space usage: you're right that hardlinking did not create new inodes per file, but it did create new inodes per directory, so the directory listings ended up fully duplicated.

As I mention in the description it does save some space, but its only ~10% of the total saved, which is mainly in the shared pyc. Using the same repo as the total `./pants test ::` savings shown above, here's an example of the breakdown:

Initial normal venv with no pyc:
```
$ pex --python python3.9 -r 3rdparty/python/requirements.txt --constraint 3rdparty/python/constraints.txt -o tc.pex --venv
$ PEX_TOOLS=1 ./tc.pex venv --collisions-ok --no-emit-warnings --force starts_off_with_no_pyc.venv
$ du -sh starts_off_with_no_pyc.venv
320M	starts_off_with_no_pyc.venv
```
With all .pyc:
```
$ starts_off_with_no_pyc.venv/bin/python -m compileall -q starts_off_with_no_pyc.venv
$ du -sh starts_off_with_no_pyc.venv
416M	starts_off_with_no_pyc.venv
```
So, in this example, 94M of pyc.

Now, isolating inode size stacking up:
```
$ PEX_TOOLS=1 ./tc.pex venv --collisions-ok --no-emit-warnings --force starts_off_with_no_pyc.venv
$ du -sh starts_off_with_no_pyc.venv
320M	starts_off_with_no_pyc.venv
$ cp -al starts_off_with_no_pyc.venv starts_off_with_no_pyc.hardlinks.venv
$ du -sh starts_off_with_no_pyc.venv starts_off_with_no_pyc.hardlinks.venv
320M	starts_off_with_no_pyc.venv
19M	starts_off_with_no_pyc.hardlinks.venv
```

So, in this example, 19M of inodes.

pex/tools/commands/venv.py
stuhood: Got it. Thanks!

jsirois:
> you're right that hardlinking did not create new inodes per file

FWIW I never said that to my knowledge. It was always clear to me there were [number of files] inodes created in the hardlink scheme vs the number of entries in the top level dir only for the symlink scheme. The difference was not significant though; so when initially implementing PEX_TOOLS venv support, I stuck to hardlink / copy since that initial impl needed to not collapse when the PEX_ROOT got nuked. The --venv mode in the main pex tool came later. It was at that point in time that I was both lazy, but more importantly, ignorant of the pyc bit. That came as a surprise investigating https://github.com/pantsbuild/requirements-perf just a few days ago.

PR #1538, opened 2021-12-17
Always pass absolute cert path to Pip.

### LINKED ISSUE #1537: The --cert option does not pass absolute path of cert which causes PEP-517 wheel builds with setup_requires to fail
A user with an internal index that uses a custom cert found this fails:
```
REQUESTS_CA_BUNDLE=cert.pem pip wheel --isolated 'avro-python3==1.10.0' --no-cache  --index=https://pypi.org/simple/ --extra-index-url=https://...
```

Wheras this works:
```
REQUESTS_CA_BUNDLE=$PWD/cert.pem pip wheel --isolated 'avro-python3==1.10.0' --no-cache  --index=https://pypi.org/simple/ --extra-index-url=https://
```

Right now, Pex passes --cert as-is without 1st absolutizing the path, so we get the former and fail:
https://github.com/pantsbuild/pex/blob/7ff27a66cd0a8e6ae848dd1a329732163f77da8e/pex/pip.py#L141-L151

jsirois:
Previously, if the `--cert` path supplied was relative, `--isolated`
builds would fail when the `--cert` was passed to Pip via env var and
the isolated build needed to resolve `setup_requires` using the cert
since the isolated build would have a fresh new CWD. Add a failing test
for this and fix.

Fixes #1537

### DIFF
--- a/pex/pip.py
+++ b/pex/pip.py
@@ -134,25 +134,30 @@ class PackageIndexConfiguration(object):
         yield "--timeout"
         yield str(network_configuration.timeout)
 
-        if network_configuration.proxy:
-            yield "--proxy"
-            yield network_configuration.proxy
-
     @staticmethod
     def _calculate_env(
         network_configuration,  # type: NetworkConfiguration
         isolated,  # type: bool
     ):
         # type: (...) -> Iterator[Tuple[str, str]]
+        if network_configuration.proxy:
+            # We use the backdoor of the universality of http(s)_proxy env var support to continue
+            # to allow Pip to operate in `--isolated` mode.
+            yield "http_proxy", network_configuration.proxy
+            yield "https_proxy", network_configuration.proxy
+
         if network_configuration.cert:
             # We use the backdoor of requests (which is vendored by Pip to handle all network
             # operations) support for REQUESTS_CA_BUNDLE when possible to continue to allow Pip to
             # operate in `--isolated` mode.
-            yield ("REQUESTS_CA_BUNDLE" if isolated else "PIP_CERT"), network_configuration.cert
+            yield (
+                ("REQUESTS_CA_BUNDLE" if isolated else "PIP_CERT"),
+                os.path.abspath(network_configuration.cert),
+            )
 
         if network_configuration.client_cert:
             assert not isolated
-            yield "PIP_CLIENT_CERT", network_configuration.client_cert
+            yield "PIP_CLIENT_CERT", os.path.abspath(network_configuration.client_cert)
 
     @classmethod
     def create(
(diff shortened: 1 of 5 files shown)


pex/pip.py
```diff
@@ -134,25 +134,30 @@ def maybe_trust_insecure_host(url):
...
+            yield "https_proxy", network_configuration.proxy
+
         if network_configuration.cert:
             # We use the backdoor of requests (which is vendored by Pip to handle all network
             # operations) support for REQUESTS_CA_BUNDLE when possible to continue to allow Pip to
             # operate in `--isolated` mode.
-            yield ("REQUESTS_CA_BUNDLE" if isolated else "PIP_CERT"), network_configuration.cert
+            yield (
+                ("REQUESTS_CA_BUNDLE" if isolated else "PIP_CERT"),
+                os.path.abspath(network_configuration.cert),
```
jsirois, line 155: With this abspath removed, the new test fails with:
```
...
E           OSError: Could not find a suitable TLS CA certificate bundle, invalid path: cert
...
```
This reproduces the reported problem back over in Pants where --cert is always passed as a relative path since the cert is materialized in a sandbox.

tests/integration/test_integration.py
```diff
@@ -1437,7 +1369,9 @@ def req(
             EXAMPLE_PYTHON_REQUIREMENTS_URL,
             fetcher=URLFetcher(
                 NetworkConfiguration(
-                    proxy="{proxy_auth}@localhost:{port}".format(proxy_auth=proxy_auth, port=port),
+                    proxy="http://{proxy_auth}@localhost:{port}".format(
```
jsirois, line 1372: This was just to prove to myself that Fetcher / urllib could work with the protocol-prfixed form. Pip, it turns out, requires that unlike Fetcher / urllib; so FAPP users must pass using `--proxy http(s)://...` syntax.

jsirois: Re-confirmed direct use of the pip CLI, whose --help lies here:
```
$ PYTHONPATH=pex/vendor/_vendored/pip python -mpip -h
...
General Options:
  -h, --help                  Show help.
...
  --proxy <proxy>             Specify a proxy in the form [user:passwd@]proxy.server:port.
```
also requires `http(s)://` or else fails with:
```
$ PYTHONPATH=pex/vendor/_vendored/pip python -mpip --proxy localhost:8888 download --no-deps ansicolors --dest wheels
...
  File "/home/jsirois/dev/pantsbuild/jsirois-pex/pex/vendor/_vendored/pip/pip/_vendor/requests/adapters.py", line 414, in send
    raise InvalidURL(e, request=request)
pip._vendor.requests.exceptions.InvalidURL: Not supported proxy scheme localhost
```

jsirois: Fixed Pex --proxy help to not parrot Pip --help.

stuhood (approved):
Thanks!

PR #1543, opened 2021-12-18
Add knob for `--venv` site-packages symlinking.

### LINKED ISSUE #1542: Some python tools and libraries don't work with symlinked venv site-packages contents.
Notably, pylint, which is used by Pants in a `--venv` PEX: https://github.com/PyCQA/pylint/issues/1470

Pex should gain a knob to work around this as needed.

jsirois:
Previously, `--venv` mode PEXes always used symlinks for site-packages
dependencies; now there is a `--venv-site-packages-copies` option to
allow the choice of hard links / copies when symlinks are not understood
by projects installed in site-packages.

Fixes #1542

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -184,7 +184,21 @@ def configure_clp_pex_options(parser):
         action=HandleBoolAction,
         help=(
             "If --venv is specified, create the venv using copies of base interpreter files "
-            "instead of symlinks."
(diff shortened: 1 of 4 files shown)


stuhood (approved):

jsirois:
I could imagine that too, but we only have 1 data point so far so we're really just guessing at the utility of the more complex (for implementers and users alike) option.

PR #1548, opened 2021-12-20
Work around setuptools 60 release in test.

jsirois:
The release added a warning that trips up a test. Pin low to avoid this
since setuptools version is incidental to the test.

See: https://github.com/pypa/setuptools/blob/main/CHANGES.rst#v6000


jsirois:
An example of the test failure can be seen here: https://github.com/pantsbuild/pex/runs/4588485105?check_suite_focus=true#step:7:611

stuhood (approved):

PR #1545, opened 2021-12-19
Fix Pex to identify Python 3.10 interpreters.

jsirois:
Although Pex could work with a Python 3.10 interpreter it could not
identify one when searching a path / using `--interpreter-constraint`.

Fix this and update all tests using the 3.8 pyenv interpreter to use
a 3.10 pyenv interpreter instead. As a side-effect, 3.10 identification
is tested.

Fixes #1544

PR #1549, opened 2021-12-20
Update CI handling of pyenv IT interpreters.

jsirois:
In #1545 I missed updating CI handling. Although this led to no errors,
the cache key bump at least, was needed to save cache space and purge
the old 3.8 pyenv interpreter stored there.


jsirois:
Requires #1548 to get CI green 1st.

tdyas (approved):

PR #1554, opened 2021-12-22
Allow using `pex3` CLI with the PEX from GitHub Releases

Eric-Arellano:
For Pants lockfile generation, we need to be able to run the `pex3` console script. To do that, users simply set `PEX3_CLI`.

## Alternatives considered

Releasing a `pex3` binary. It would be inefficient to download both `pex` and `pex3`.

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -653,6 +653,12 @@ def _compatible_with_current_platform(interpreter, platforms):
 
 
 def main(args=None):
+    if "PEX3_CLI" in os.environ:
+        import pex.cli.pex
+
+        pex.cli.pex.main()
+        return
+
     args = args[:] if args else sys.argv[1:]
     args = [transform_legacy_arg(arg) for arg in args]
     parser = configure_clp()


pex/bin/pex.py
```diff
@@ -653,6 +653,12 @@ def _compatible_with_current_platform(interpreter, platforms):
 
 
 def main(args=None):
```
Eric-Arellano, line 655: Putting the override here means that the override works for both `python -m pex` and the binary created by `scripts/package.py`.

pex/bin/pex.py
```diff
@@ -653,6 +653,12 @@ def _compatible_with_current_platform(interpreter, platforms):
...
+    if "PEX3_CLI" in os.environ:
```
Eric-Arellano, line 656: Lmk if this should check whether the value == 1 or "true" etc.

jsirois, line 656: Have you tried `PEX_SCRIPT=pex3 ./pex.pex`? `pex --help-variables` is invaluable.

Eric-Arellano: Doh. Thanks. 

`PEX_SCRIPT=pex3 python3 -m pex -h` doesn't work but that doesn't matter to me. `PEX_SCRIPT=pex3 dist/pex` does 🙌

pex/bin/pex.py
jsirois: `python -m pex.cli` is already set up for that style access.

pex/bin/pex.py
jsirois: Re the 1st command:

PEX_ are Pex runtime variables, these take effect at Pex runtime, i.e.: when a Pex boots up.

`python -m module` uses Python to boot (via the runpy machinery) not Pex.

PR #1561, opened 2022-01-09
Expose more Pip options.

### LINKED ISSUE #802: Better support broken `setup.py` sdists.
The `setup.py` of projects can be broken in various ways that are either intended / the lesser of evils or else inobvious to the maintainer. In particular, since `setup.py` is executed to produce distributions, it can have implicit dependencies (expected available imports) that can only be satisfied by the `setuptools` `setup_requires` extension to `distutils`. That mechanism is imperfect even when used since it downloads and installs requirements OOB from the frontend installer - say pip or pex - not respecting the frontend configuration (indexes, find_links, etc). Some examples of `setup.py` files with external deps beyond `setuptools` and `wheel`:

1. statsmodels:
https://github.com/statsmodels/statsmodels/blob/75edeb6cc6ce095d6a5c4b1006c1c7756019bf0c/setup.py
2. hgvs:
https://github.com/counsyl/hgvs/blob/aebe5bd9683f4b5937fd653ce4b13fcd4f3ebb10/setup.py

In the 1st example, an attempt at `setup_requires` is used for `numpy` but there is an unguarded import later in the file that uses numpy leading to an unuseable sdist except in environments where numpy is pre-installed.

In the second example, pip is imported. In general pip will be present, but this is not necessarily true, in which case, again, the sdist is unuseable in enviornments without pip pre-installed.

There are two broad approaches here:
1. Fix the project's build:
  Preferrably convert it to use a [PEP-517][2] / [PEP-518][3] build.
2. Arrange for the implicit dependencies to be pre-installed.

The former is always preferred. [PEP-517][2] / [PEP-518][3] are designed to solve this problem cleanly. However, in the case where an older release must be used, it's very unlikely a backport fix can be arranged even if the project maintainers will accept a fix at HEAD.

PEX coud support the latter though. Imagine the following pex command line:
```
pex --build-dep=statsmodels|0.8.0|numpy==1.6.2 some_project
```

The main idea is there would be new CLI options / resolver api args that would allow specification of a list of build dependencies for a distribution known to be present in a resolve. In practice this would mean an initial pex execution like the following would lead to knowledge of the implicit build dependency:
```
pex  some_project
```
The failure to create a pex from `some_project` would include the the fact that a build of the statsmodel 0.8.0 sdist failed due to the missing implicit numpy dependency. This would allow crafting the second attempt command line. Pex, in the second attempt, would 1st download the statsmodel 0.8.0 sdist and then build a wheel from it using pip / [PEP-517][2] / [PEP-518][3] by unpacking the sdist and inserting a `pyproject.toml` with the appropriate entries before handing that unpacked directory to `pip wheel`. Unfortunately, Pex would need to hand-roll the download using [PEP-503][1] since `pip download` doesn't just download, but also runs `python setup.py egg_info` on downloaded sdists with no `pyproject.toml`.

After downloading the sdist and building a wheel this way, Pex could then proceed with a normal resolve, adding in a `find_links` repo pointing to the chroot containing each wheel pre-built in this way. 

[1]: https://www.python.org/dev/peps/pep-0503/
[2]: https://www.python.org/dev/peps/pep-0517/
[3]: https://www.python.org/dev/peps/pep-0518/

### LINKED ISSUE #1202: Add --prefer-binary to pip
I recently encountered an issue with pex trying to download an unfortunate package from the pypi ecosystem: boilerpipe. boilerpipe sets an unbounded dep on JPype1. The most recent JPype1 is py3 only. But there appear to only be sdists on pypi. This results, when running pex with py2 interpreter, in pip downloading the latest JPype sdist and the wheel building fails because JPype1 uses pathlib in its setup.py :(

We have run our own cheeseshop in the past to workaround these unfortunate problems with the pypi ecosystem, but we'd like to not have to maintain our cheeseshop as much and move to pypi more and more. With our cheeseshop in place for special case pre-built wheel packages, we could specify both it and pypi if pex supported --prefer-binary.

The proposed option can be off by default, just plumbed through so that we can access it from the cli and then i can issue a seperate pants ticket to add that option in pants.

### LINKED ISSUE #1560: Pex bypass/avoid using pep517
We are seeing a set of issues crop up over the week because setuptools over v60.0 introduced breaking changes with regards to the version of distutils used. Setting the env variable `SETUPTOOLS_USE_DISTUTILS=stdlib` seems to be a workaround, but I would like to know if there is a documented way to avoid using pep517 via pex? 

When pex is invoked with a constraints file, or pinned requirement of an older setuptools, it's respected all the way until a dependent package that does not fully conform to pep517 comes along, i.e., it has a `pyproject.toml` file, but no `build-system` section and `build-backend` entry. 

When it does come across such a package, pip is invoked with the following arguments - `'setuptools>=40.8.0' wheel
`. This is causing the latest setuptools to get pulled and installed.

Please excuse any mistaken assumptions I may have made along the way, and I would really appreciate any direction on this matter.

jsirois:
Expose Pip's `--prefer-binary`, `--[no]-use-pep517` and
`--no-build-isolation` options to provide users more control over
distribution resolution and wheel building. This allows more corner
cases in the ecosystem to be PEXed.

Closes #802
Closes #1202
Closes #1560


jsirois:
The 2nd commit (9e5fd38) was broken out to #1564 since it's independent bitrot that also affects #1563. I'll rebase this once that's in, but the 1st commit is good to go for review.

stuhood (approved):

Issue #1556, opened 2021-12-22
`pex3 lock create` does not work with VCS requirements as input


Eric-Arellano:
This works:

```
pex 'ansicolors @ git+https://github.com/jonathaneunice/colors.git@c965f5b9103c5bd32a1572adb8024ebe83278fb0'
```

But this fails:

```
❯ pex3 lock create 'ansicolors @ git+https://github.com/jonathaneunice/colors.git@c965f5b9103c5bd32a1572adb8024ebe83278fb0'
Traceback (most recent call last):
  File "/Users/ericarellano/.local/bin/pex3", line 8, in <module>
    sys.exit(main())
  File "/Users/ericarellano/.local/pipx/venvs/pex/lib/python3.9/site-packages/pex/cli/pex.py", line 29, in main
    result = catch(command.run)
  File "/Users/ericarellano/.local/pipx/venvs/pex/lib/python3.9/site-packages/pex/commands/command.py", line 130, in catch
    return func(*args, **kwargs)
  File "/Users/ericarellano/.local/pipx/venvs/pex/lib/python3.9/site-packages/pex/cli/command.py", line 71, in run
    return subcommand_func(self)
  File "/Users/ericarellano/.local/pipx/venvs/pex/lib/python3.9/site-packages/pex/cli/commands/lock.py", line 199, in _create
    create(
  File "/Users/ericarellano/.local/pipx/venvs/pex/lib/python3.9/site-packages/pex/cli/commands/lockfile/__init__.py", line 113, in create
    downloaded = resolver.download(
  File "/Users/ericarellano/.local/pipx/venvs/pex/lib/python3.9/site-packages/pex/resolver.py", line 1097, in download
    build_requests, download_results = _download_internal(
  File "/Users/ericarellano/.local/pipx/venvs/pex/lib/python3.9/site-packages/pex/resolver.py", line 1002, in _download_internal
    download_results = download_request.download_distributions(
  File "/Users/ericarellano/.local/pipx/venvs/pex/lib/python3.9/site-packages/pex/resolver.py", line 134, in download_distributions
    return list(
  File "/Users/ericarellano/.local/pipx/venvs/pex/lib/python3.9/site-packages/pex/jobs.py", line 510, in execute_parallel
    yield spawn_result.spawned_job.await_result()
  File "/Users/ericarellano/.local/pipx/venvs/pex/lib/python3.9/site-packages/pex/jobs.py", line 214, in await_result
    job.wait()
[...]

PR #1563, opened 2022-01-10
Reject VCS requirements in locks.

### LINKED ISSUE #1562: Support VCS requirements 1st class
Right now Pex requirement parsing lumps these under `URLRequirement` precluding any special handling required. One way or another #1556 will require recognizing VCS requirements.

jsirois:
The current pex lock infrastructure cannot handle VCS requirements in
the same way it can't handle local project requirements. Unlike local
project requirements though, the failure to handle these requirements
was uncontrolled. Add explicit parsing of VCS requirements to allow them
to be singled out and rejected along side local project requirements.

This addresses the UX for the failure noted in #1556 but may not be the
final word there. If there is a desire to actually handle VCS
requirements in locks, the `VCSRequirement` parsing added here can be
expanded to extract the needed extra data (commit id certainly) to build
that support.

Closes #1562


jsirois:
This is good to go for review despite red. The failure is independent bitrot being addressed in #1564. I'll re-base this (and #1561) once this lands.

stuhood (commented):
Content looks good, and the parsing is a great step toward https://github.com/pantsbuild/pex/issues/1562. But I do wonder if this shouldn't be a strongly worded warning rather than an error, or perhaps whether failing should be behind an option (on by default is fine).

Although these requirements will be uncontrolled, as we migrate users in Pants to lockfiles by default, erroring for this case will inhibit some users from using lockfiles who might otherwise have been able to lock the rest of their requirements. So it might be good for Pants to be able to detect the requirement, warn the user with a ticket number, but then pass an option to PEX to skip erroring so that the rest of their requirements can be locked at least.

Thoughts?

jsirois:
> But I do wonder if this shouldn't be a strongly worded warning rather than an error

We can do nothing except error right now. There will be extensive work needed to support not erroring. You may not be aware, but Pants currently does not error ... but does the wrong thing and silently locks the latest publically available version of the project from PyPI instead of grabbing the requested VCS version and locking that.

jsirois:
> ... and the parsing is a great step toward #1562

The parsing is exactly #1562. It resolves that. The notes at the bottom of that issue are contingent on the answer to #1556 which I expect is do all the hard work. But we could also just proceed as is with failing and circle back later and advance the state of the art. The current failure is slightly better than the Pants state with the silent ignore what you said behavior.

jsirois:
@stuhood if it helps focus: This just takes the current backtrace hard error and turns it into a human-friendly hard-error. It does not preclude any further better support and the work done here to identify VCS requirements in the 1st place is needed for any efforts of that sort. So ... this should be completely uncontroversial.

stuhood (approved):
Ok, thanks.

I think that I interpreted:
> Unlike local project requirements though, the failure to handle these requirements was uncontrolled.

... as meaning that we would still successfully build/install the requirement in question, but then not actually successfully lock as described in https://github.com/pantsbuild/pex/issues/1562#issuecomment-1009199447... i.e., that re-running immediately would get you what you wanted, but that re-running at some point in the future might not.

Eric-Arellano:
Thank you, John. This looks great.

PR #1564, opened 2022-01-10
Fix bdist_pex tests.

jsirois:


jsirois:
Broken out of #1561. This is also needed to get #1563 green.

stuhood (approved):

PR #1566, opened 2022-01-10
Fix locks to exclude build requirements.

### LINKED ISSUE #1565: pex3 lock create includes build requirements
```
$ pex3 lock create pantsbuild.pants
```

Includes `setuptools` twice in the locked resolves.

> Looks like PEP-517 build requirements are getting recorded: https://github.com/pantsbuild/pex/blob/6720d4c64fb033e8418b2e14535512baa01b9a2d/pex/vendor/_vendored/pip/pip/_internal/pyproject.py#L118-L130
-- @jsirois 

```
{
  "artifacts": [
    {
      "algorithm": "sha256",
      "hash": "60d78588f15b048f86e35cdab73003d8b21dd45108ee61a6693881a427f22073",
      "url": "https://files.pythonhosted.org/packages/4b/b9/71687c5d2034c863db1db2e0704f5e27581ff3cb44d7f293968c5e08ceb3/setuptools-57.5.0-py3-none-any.whl"
    }
  ],
  "project_name": "setuptools",
  "requirement": "setuptools<58.0,>=56.0.0",
  "version": "57.5",
  "via": [
    "pantsbuild.pants"
  ]
}
{
  "artifacts": [
    {
      "algorithm": "sha256",
      "hash": "68eb94073fc486091447fcb0501efd6560a0e5a1839ba249e5ff3c4c93f05f90",
      "url": "https://files.pythonhosted.org/packages/eb/53/0dd4c7960579da8be13fa9b2c2591643d37f323e3d79f8bc8b1b6c8e6217/setuptools-60.5.0-py3-none-any.whl"
    }
  ],
  "project_name": "setuptools",
  "requirement": "setuptools>=40.8.0",
  "version": "60.5",
  "via": []
}
```

jsirois:
Previously build requirements resolved recursively via Pip PEP-517
handling would show up in locks.

Fixes #1565


jsirois:
Depends on #1561 to add tests (they'll use --force-pep517).

jsirois:
The test failure is unrelated bitrot being fixed in #1564.

stuhood (approved):

pex/pip.py
```diff
@@ -429,20 +431,48 @@ def _extract_resolve_data(url):
...
+        # 2.) "... Added <requirement pin> from <url> ... to build tracker ..."
+        # 3.) Lines related to extracting metadata from <requirement pin>'s artifact
+        # 4.) "... Removed <requirement pin> from <url> ... from build tracker ..."
+        #
+        # The lines in section 3 can contain this same pattern of lines if the metadata extraction
+        # proceeds via PEP-517 which recursively uses Pip to resolve build dependencies. We want to
+        # ignore this recursion since a lock should only contain install requirements and not build
+        # requirements (If a build proceeds differently tomorrow than today then we don't care as
+        # long as the final built artifact hashes the same. In other words, we completely rely on a
+        # cryptographic fingerprint for reproducibility and security guarantees from a lock).
```
kaos, line 449: Oh, I'm not that familiar with Python package builds. My experience from building binaries (such as for C applications, for instance) is that even when the source input is the same, the binary output usually differ (due to embedded timestamps if nothing else). Is this not the case for Python applications/libraries? (as that would break the fingerprint)

jsirois: Its should not be the case for Pex. There are some standards out in the world now that help even beyond python and like this one which we also use:
https://github.com/pantsbuild/pex/blob/7cfb36d9ad094b5dcfe0ae443f3470e244ea67ed/pex/tools/commands/repository.py#L248-L254
https://github.com/pantsbuild/pex/blob/56455c4f307c024b89738d5ad59468600a8c4c2f/pex/bin/pex.py#L724-L728

We have semi-extensive tests for this, but I'm pretty sure we don't set `SOURCE_DATE_EPOCH` for our `pip download ...` and `pip wheel ...` commands, both of which might build an sdist into a wheel.

jsirois: Ah, right. We do not in fact set `SOURCE_DATE_EPOCH` when running `pip download ...` or `pip wheel ...`; so the resulting wheels do have non-reproducible timestamps for their zip entries, but since PEX files don;t use wheels, but _installed_ wheels, the non-reproducible timestamps in the wheels we installed from evaporate. If and when Pex starts doing all the gymnastics to support VCS requirements and local project requirements in locks, then that code will need to handle this. This code, though, just uses fingerprints of downloaded sdists or wheels. The building going on here is just to extract metadata for the resolve; i.e.: Find out the install_requires dependencies of sdists to continue the recursive resolve process. The only thing that needs to be reproducible for that are the Requires-Dist metadata lines.

Issue #1361, opened 2021-06-14
Support creating a PEX from a venv.


jsirois:
This would close the circle and we have all the infra to do this. Currently resolution from a fixed set of Distributions is buried in PEXEnvirronment but could be lifted out. The Virtualenv class would just need to grow `iter_distributions` to make all this work. That could be implemented via `pkg_resources.WorkingSet()` in a subprocess using the venv interpreter since a `WorkingSet` is an iterator of `Distibution`s.

jsirois:
The wrinkle here is only dist-info distributions work with PEXes isolation scheme. That isolation scheme requires a wheel name and that can only be reconstructed from a dist-info distribution with a WHEEL metadata file with a Tag entry. If the venv is populated via Pip - a common case to be sure - any sdists are installed via a different scheme that uses egg-info and provides know way to directly derive a wheel name.

jsirois:
I guess the safe option for sdists is to adopt the most specific platfrom tag of the venv's interpreter. That would treat every distribution in the venv installed from an sdist as if it were platform specific even when this was un-true. The upshot would be the resulting PEX file is not as portable as it should be. For example, if the venv constinaed one distribution installed from an sdist that was in fact universal (py2.py3), the PEX created from that venv would only be able to run under the venv interpreter (say CPython 3.9 on Linux). To mitigate surprise, a warning could be emitted for each distribution gathered into the PEX that was derived from an sdist and not a wheel.

Eric-Arellano:
> To mitigate surprise, a warning could be emitted for each distribution gathered into the PEX that was derived from an sdist and not a wheel.

That sounds reasonable. 

Maybe point out that building a wheel beforehand can provide a workaround. It's likely not obvious to everyone that you can prebuild an sdist and have Pex use the resulting wheel.

jsirois:
> It's likely not obvious to everyone that you can prebuild an sdist and have Pex use the resulting wheel.

It might not be obvious, but Pex is not involved here. They created the venv with some tool. Pip, Poetry - who knows. What exactly would you say to a Poetry user for example? (I don't think there is actually an answer to that one). How about a Pip user? (There is an answer to that one).

jsirois:
[...]

PR #1571, opened 2022-01-19
Support iterating a venv's distributions.

jsirois:
Work towards #1361.

### DIFF
--- a/pex/tools/commands/virtualenv.py
+++ b/pex/tools/commands/virtualenv.py
@@ -4,6 +4,7 @@
 from __future__ import absolute_import
 
 import fileinput
+import json
 import logging
 import os
 import re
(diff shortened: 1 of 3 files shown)


pex/tools/commands/virtualenv.py
```diff
@@ -219,6 +264,47 @@ def iter_executables(self):
...
+                                )
+                            )
+                    """.format(
+                        setuptools_path=setuptools_path
+                    )
+                ),
+            ]
+        )
+        for line in stdout.splitlines():
+            project_name, version, sys_path_entry = line.split()
```
stuhood, line 303:
```suggestion
            project_name, version, sys_path_entry = line.split(maxsplit=2)
```
In case `sys_path_entry` contains spaces?

stuhood (approved):

pex/tools/commands/virtualenv.py
jsirois: Yup - good catch. Bug fixed with a previously failing test.

PR #1575, opened 2022-01-22
Introduce iter_compatible_versions.

jsirois:
This will be used to filter distributions in platform agnostic resolves
when interpreter constraints are specified.

Work towards #1402.

### DIFF
--- a/pex/interpreter_constraints.py
+++ b/pex/interpreter_constraints.py
@@ -157,6 +157,15 @@ class Lifecycle(Enum):
     EOL = Value("eol")
 
 
+# This value is based off of:
+# 1. Past releases: https://www.python.org/downloads/ where the max patch level was achieved by
+#    2.7.18.
+# 2. The 3.9+ annual release cycle formalization: https://www.python.org/dev/peps/pep-0602/ where
+#    the last bugfix release will be at a patch level of ~10 and then 3.5 years of security fixes
+#    as needed before going to EOL at the 5-year mark.
+DEFAULT_MAX_PATCH = 30
+
+
 @attr.s(frozen=True)
 class PythonVersion(object):
     lifecycle = attr.ib()  # type: Lifecycle.Value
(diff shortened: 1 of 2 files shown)


pex/interpreter_constraints.py
```diff
@@ -138,3 +146,71 @@ def create_message(self, preamble=None):
...
+
+
+# TODO(John Sirois): Integrate a `pyenv install -l` based lint / generate script for CI / local
+# use that emits the current max patch for these versions so we automatically stay up to date
+# mod dormancy in the project.
+
+COMPATIBLE_PYTHON_VERSIONS = tuple(
+    # Each 5 units of padding costs `iter_compatible_versions` ~2 extra ms and a padding of 10 gets
+    # us through most of a brand new stable release's likely active lifecycle.
+    version.pad(10)
```
Eric-Arellano, line 194: We used to use a lower number for Pants, but @benjyw updated it to 30 with the argument it's worse to have Pex/Pants not work with some future patch release like 3.10.11 than for it to be a little slower: https://github.com/pantsbuild/pants/blob/523bdb2c2844dd56e266499796dcb712ae26c694/src/python/pants/backend/python/util_rules/interpreter_constraints.py#L38

We know from Pants that people often do use older versions for longer than at least I would have anticipated.

The EOL stuff looks great, but maybe increase this number? Or let you choose the padding per major.minor: padding for `3.7.12` of 10 is probably too high, whereas for `3.11.0` it's not enough.

Eric-Arellano (approved):

pex/interpreter_constraints.py
benjyw: I had a similar comment on this code in #1577

benjyw: 30 might have been excessive, but it depends on the perf hit. Doing this per major.minor would mitigate that.

jsirois: I've gone with 30 but left the door open to plumbing an option as a backstop which is used in tests right away.

Eric-Arellano (approved):
Thanks for updating patch versions. 

The other thing to consider is Python 3.12+. I don't expect many pure-Pex users to be stuck on an older Pex version, but I could see this happening to Pants users. In Pants-land, we solved this with an option: https://www.pantsbuild.org/docs/reference-python#section-interpreter-versions-universe. This is probably fine to land as is, but might be worth exposing an option.

jsirois:
> The other thing to consider is Python 3.12+.

But Pex has a runtime and that is only tested to work through some fixed major / minor (3.10 currently). Even adding 3.11 like I did is dubious.

Eric-Arellano:
Good point. Thanks!

PR #1577, opened 2022-01-22
Make resolution of DistributionTargets lazy.

jsirois:
Break up TargetConfiguration into just the configuration data and a
resolve method that can be used to request DistributionTargets only
when needed. This divide will be needed for platform agnostic locking
where interpreter constraints need to be known and used but
interpreters do not need to be resolved using those constraints.

Work towards #1402.


jsirois:
The 1st commit (7818a36) is out for review in #1575. Please focus on the second commit (2bc7854).

pex/distribution_target.py
```diff
@@ -183,3 +187,43 @@ def __eq__(self, other):
...
         return hash(self._tup())
+
+
+@attr.s(frozen=True)
+class DistributionTargets(object):
+    interpreters = attr.ib(default=())  # type: Tuple[PythonInterpreter, ...]
+    platforms = attr.ib(default=())  # type: Tuple[Optional[Platform], ...]
+    assume_manylinux = attr.ib(default=None)  # type: Optional[str]
+
+    @property
```
jsirois, line 198: From here down was moved from https://github.com/pantsbuild/pex/pull/1577/files#diff-ad403368ac5713232cee63c69f5236bf3839256ad17ffe17c9eac45117fe05afL45

pex/resolve/target_configuration.py
```diff
@@ -3,74 +3,130 @@
...
+        # type: () -> DistributionTargets
+        """Resolves the distribution targets satisfying the target configuration.
+
+        :raise: :class:`InterpreterNotFound` specific --python interpreters were requested but could
+            not be found.
+        :raise: :class:`InterpreterConstraintsNotSatisfied` if --interpreter-constraint were
+            specified but no conforming interpreters could be found.
+        """
+
+        # TODO(#1075): stop looking at PEX_PYTHON_PATH and solely consult the `--python-path` flag.
```
jsirois, line 58: This was moved from https://github.com/pantsbuild/pex/pull/1577/files#diff-6530c790096dac6af237266218e9b5934e25fee63241b7937ae66acc63581562L171

jsirois:
Almost all just re-shuffling. The two key moved bits have comments pointing out the move sources.

pex/interpreter_constraints.py
```diff
@@ -138,3 +133,71 @@ def create_message(self, preamble=None):
...
+# TODO(John Sirois): Integrate a `pyenv install -l` based lint / generate script for CI / local
+# use that emits the current max patch for these versions so we automatically stay up to date
+# mod dormancy in the project.
+
+COMPATIBLE_PYTHON_VERSIONS = tuple(
+    # Each 5 units of padding costs `iter_compatible_versions` ~2 extra ms and a padding of 10 gets
+    # us through most of a brand new stable release's likely active lifecycle.
+    version.pad(10)
+    for version in (
+        PythonVersion(Lifecycle.EOL, 2, 7, 18),
```
benjyw, line 183: Is that 2 ms per (major,minor) version, or total? If the former, should the padding be per-(major,minor) version? There will never be a 2.7.28. If the latter, there's probably no benefit.

benjyw (approved):
Hard to follow in detail, but looks right in shape.

pex/interpreter_constraints.py
jsirois: Its total for bumping the pad by 5. The 2.7.28 case is already handled since EOL gets no padding. I'll continue the rest of the discussion of this commit under its review in #1575.

PR #1579, opened 2022-01-22
Introduce the universal lock mode.

### LINKED ISSUE #1402: Implement platform agnostic requirements.txt lock file support.

jsirois:
The universal lock mode allows locking distributions for all platforms
while still providing the ability to constrain interpreter versions if
desired.

Closes #1402


pex/pip.py
```diff
@@ -588,18 +611,179 @@ def create(
...
+                                        set(
+                                            (version[0], version[1])
+                                            for version in  python_full_versions
+                                        )
+                                    )
+                                    python_majors = sorted(
+                                        set(version[0] for version in  python_full_versions)
+                                    )
+
+                                # 1.) Universal dependency environment marker applicability.
```
jsirois, line 660: The 1, 2, 3 starting here is the only magic to get `pip download` working for universal resolves. In short, there are 3 items to deal with:
1. Requires-Dist: Need to only partially evaluate env markers for python_version / python_full_version (if --interpreter-constraint) but not anything else.
2. Wheel tags: Need to ignore these saving for the major / minor version part of the python tag if --interpreter-constraint.
3. Requires-Python: Need to ignore this / substitute info from --interpreter-constraint to select.

pex/cli/commands/lock.py
```diff
@@ -191,9 +196,32 @@ def add_extra_arguments(
...
+                        universal=LockStyle.UNIVERSAL.value,
+                        were="were" if len(target_configuration.pythons) > 1 else "was",
+                        num_pythons=len(target_configuration.pythons),
+                        num_platforms=len(target_configuration.platforms),
+                    )
+                )
+            lock_configuration = LockConfiguration(
+                style=LockStyle.UNIVERSAL,
+                requires_python=tuple(
+                    str(interpreter_constraint.specifier)
```
jsirois, line 215: N.B.: Right now this is lossy on the project_name portion of the interpreter requirement; i.e.: for an explicit 'PyPy>=3' or 'CPython>=3' We'll resolve distributions not implied by the implementation constraint. I'd like to circle back to handle that: #1580

pex/pip.py
```diff
@@ -588,18 +611,179 @@ def create(
...
                     isolated_pip_builder.add_dist_location(dist=dist_location)
+                env_var_names = dict(
+                    patched_markers_env_var_name=cls._PATCHED_MARKERS_FILE_ENV_VAR_NAME,
+                    skip_markers_env_var_name=cls._SKIP_MARKERS_ENV_VAR_NAME,
+                    python_versions_env_var_name=cls._PYTHON_VERSIONS_FILE_ENV_VAR_NAME,
+                )
                 with named_temporary_file(prefix="", suffix=".py", mode="w") as fp:
                     fp.write(
                         dedent(
                             """\
```
jsirois, line 622: This dynamic patching only has 4 items, but its pretty large. I'd like to follow up and break the patches out to a (re)source file: #1581

pex/pip.py
```diff
@@ -588,18 +611,179 @@ def create(
...
+                                # Python 3.10 but a marker says a dependency is only for
+                                # 'python_version < "3.6"' we still want to lock that dependency
+                                # subgraph too.
+                                def patch_marker_evaluate():
+                                    from pip._vendor.packaging import markers
+
+                                    original_get_env = markers._get_env
+                                    original_eval_op = markers._eval_op
+                                    
+                                    skip = object()
```
stuhood, line 673: Would having empty `python_versions` or `python_full_versions` every result in "accidentally" skipping this behavior? Or should the empty set have the same behavior as skipping anyway?

stuhood (approved):
Thanks!

pex/pip.py
jsirois: Its not clear to me at all his work deals with this universal stuff. I don't think it does. That's why I pinged. I think there are really just 2 ways to lessen maintenance burden:
1. Pip comes out with lockfile support that suits Pex needs and we switch to that.
2. Pex moves away from Pip and builds around resolvelib directly.

Given my experience with Pip maintainers (and even that aside and putting myself in their shoes) and I think both are about as hard / unlikely to happen so I'd lean towards #2 without having given it alot of thought.

pex/pip.py
jsirois: Yeah, same as skip - this is not accidental. An unconstrained (no `--interpreter-constraint`) universal resolve should skip (ignore) `python_version` and `python_full_version` environment markers, and resolve every single wheel available. There is a test for this in this PR with 5 wheels downloaded for the `--interpreter-constraint` supplied case (`python_full_versions` not empty) and 31 wheels for the `--interpreter-constraint` not present case (`python_full_versions` is empty).

pex/cli/commands/lock.py
```diff
@@ -99,8 +99,13 @@ def _add_create_arguments(cls, create_parser):
                 "a lock file that contains exactly the distributions that would be used in a local "
                 "resolve. If an sdist would be used, the sdist is included, but if a wheel would "
                 "be used, an accompanying sdist will not be included. The {sources} style includes "
-                "locks containing wheels and the associated sdists when available.".format(
-                    strict=LockStyle.STRICT, sources=LockStyle.SOURCES
+                "locks containing both wheels and the associated sdists when available. The "
+                "{universal} style generates a universal lock for all possible target interpreters "
+                "and platforms, although the scope can be constrained via one or more "
+                "--interpreter-constraint.".format(
```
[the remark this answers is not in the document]
jsirois: Well, someone else needs to give this guidance I think - you won't like my opinion. My opinion is that strict is preferred - its the only true lock. When you admit sources in addition you admit untested paths. So you've locked, but you have no clue if:
1. The locked sdist, when built a year later, even though it matches the fingerprint at lock time, doesn't unleash havoc with malicious code that does not match the wheel contents.
2. The locked sdist has different Requires-Dist / Requires-Python. This is much better than 1 - you won't suffer an attack, but you will suffer a resolve failure since there will be dependencies missing from the lock.

When you go further and admit universal, its the same problem en-masse.

If you want me to write down all this and are OK with the opinion, I can do that - but none of this is specific to Pex or even Python, this is true conceptually for any locking system where there is more than 1 distribution.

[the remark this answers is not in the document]
Eric-Arellano: I do think your opinion would be worth sharing, John. I suspect not every Pex user will realize this tradeoff, many still are learning about supply chain attacks in the first place.

For example, something like:

> It is often necessary to use {strict} or {universal} for the lock to work in other environments, such as a lock generated on macOS to work on Linux. However, this extra compatibility means that ...

Eric-Arellano (approved):
Exciting!

pex/cli/commands/lock.py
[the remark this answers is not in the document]
jsirois: Alright then. Will do.

I truly think anyone enabling locks and not thinking about what they're doing is being very American and not very Japanese. There should be more thoughtfulness, deliberation, learning and craft instead of check a box slap-dash get-er-done, but it is true that people don't seem to self-educate that well.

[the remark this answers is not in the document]
Eric-Arellano: > but it is true that people don't seem to self-educate that well.

I view this feature and this soon-to-be help message as part of how we get people on the path to learning. We get them to think about something they may have never considered in the first place because they don't know what questions to ask. By using Pex/Pants, you become a better engineer and are exposed to important practices.

As discussed before, I don't think we need to / want to explain everything in-depth. Instead, it can be particularly compelling to introduce a topic and then link to more in-depth things for people to keep learning.

Eric-Arellano (approved):
Thank you!

pex/pip.py
```diff
@@ -610,18 +633,179 @@ def create(
...
+                                    Wheel.support_index_min = lambda *args, **kwargs: 0
+
+                                    if python_versions:
+                                        import re
+
+                                        def supported(self, *_args, **_kwargs):
+                                            if not hasattr(self, "_versions"):
+                                                versions = set()
+                                                for pyversion in self.pyversions:
+                                                    if pyversion[:2] in ("cp", "pp", "py"):
```
jsirois, line 748: https://github.com/pantsbuild/pex/runs/4933490418?check_suite_focus=true#step:7:935 exposed a bug here. This only considers `<python>` from the `<python>-<abi>-<platfrom>` tag triple, but when `<abi>` is abi3, `<python>` evaluation should only consider the major version of `<python>` - namely `3`. So, right now, since this lock is for `==3.9.*` the compatible `cryptography-35.0.0-cp36-abi3-macosx_10_10_x86_64.whl` wheel if filtered out and an attempt to build cryptography with the sdist proceeds and fails.

Eric-Arellano: > evaluation should only consider the major version of <python> - namely 3.

It also needs to consider the <python> prefix if `--interpreter-constraint`, right? `cp36-abi3` is compatible with 3.6+ but not 3.5.

jsirois: Can you cite your source for that? Pip (packaging 20.8 really), disagrees (Here my `pip` tox env is Python 3.10.):
```
$ tox -epip -- debug -v | grep abi3 | grep linux_x86_64
  cp310-abi3-linux_x86_64
  cp39-abi3-linux_x86_64
  cp38-abi3-linux_x86_64
  cp37-abi3-linux_x86_64
  cp36-abi3-linux_x86_64
  cp35-abi3-linux_x86_64
  cp34-abi3-linux_x86_64
  cp33-abi3-linux_x86_64
  cp32-abi3-linux_x86_64
```
 So packaging takes it back from present to the birth of abi3 in Python 3.2.

jsirois: Keep in mind, over inclusion is fine (you've already signed away your right to worry about un-vetted dists by choosing universal); so If there is a cp3{0,1} wheel out there that claims abi3; so be it. At resolve time, packaging will be used to evaluate tags and reject such a tag and so your resolve will either fail (again - you bough into this by choosing universal), or you'll proceed to get the sdist and build that (another thing you bought into).

Eric-Arellano: >> cp36-abi3 is compatible with 3.6+ but not 3.5.

> Can you cite your source for that?

PEP 652: https://www.python.org/dev/peps/pep-0652/#stable-abi

> The Stable ABI is not generally forward-compatible: an extension built and tested with CPython 3.10 will not generally be compatible with CPython 3.9.

> At resolve time, packaging will be used to evaluate tags and reject such a tag and so your resolve will either fail (again - you bough into this by choosing universal)

I wonder if it's possible to reject the lock, i.e. more eager than failing at resolution time. If you set `--interpreter-constraint="CPython==3.6.*"` and there is only a `cp38-abi3` wheel available, it would be neat if we could fail to lock.

jsirois: That's the wrong PEP / standard. The _Limited_ ABI is what abi3 is all about:
+ https://www.python.org/dev/peps/pep-0652/#stable-abi-vs-limited-api
+ https://www.python.org/dev/peps/pep-0384

The fact Pip / packaging emit my sample tag set above for a Python 3.10 interpreter implies they view things this way at least.

jsirois: > ... it would be neat if we could fail to lock.

cosmicexplorer:
Apologies for missing out on reviewing this, but will now scan it closely for relation to my current pip work. I appreciate pinging me.

PR #1586, opened 2022-01-24
Extract --pex-repository resolving from resolver.

jsirois:
The resolver module is huge and the code for resolution from a PEX had
nothing in common except the result type which is now extracted as well.


jsirois:
This is all pure moves.

benjyw (approved):

Issue #1587, opened 2022-01-25
Pex does not support mac universal2 wheels


jsirois:
A Pants user [reported](https://github.com/pantsbuild/pants/issues/14244) cryptography being built by Pex instead of its published wheel being used: https://pypi.org/project/cryptography/36.0.1/#files. That wheel is `cryptography-36.0.1-cp36-abi3-macosx_10_10_universal2.whl`.

This was fixed in Pip 21.0.1 here via a packaging upgrade: https://github.com/pypa/pip/commit/00fb5a0b224cde08e3e5ca034247baadfb646468 That is the last version of packaging to support Python 2.7 so we're in luck and can just merge the patch in our fork.

thor:
If necessary, I can test this locally with an M1 device for `pex`. We've recently stumbled into this while trying to make `universal2` compatible executables.

jsirois:
@thor that would be great. The fix is in Pex 2.1.64 to test: https://pypi.org/project/pex/2.1.64/

jsirois:
Missed upgrading Pex's vendored copy: https://github.com/pantsbuild/pants/issues/14244#issuecomment-1021386040

Issue #1583, opened 2022-01-22
Support resolving using a Pex lock file.


jsirois:
Instead of exporting the lock file to `requirements.txt` format and letting Pip do the resolve, record the `Requires-Dist` metadata in the lock so that Pex can calculate the resolve (subset) using the logic it already has for `--pex-repository` and then parallel-download all required dists. The backend to this can then be the current Pip build and install phase (`BuildAndInstallRequest`).

jsirois:
N.B.: This is also required to support the final phase of #1556 which requires Pex to do the hashing and hash-checking of loose source project requirements (local and VCS projects).

jsirois:
Perf is decent.

Cold:
```
$ hyperfine -w1 -r3 -p 'rm -rf ~/.pex' 'python3.7 -mpex jupyter-server -o jupyter-server.pex -c jupyter --venv --layout packed' 'python3.7 -mpex --lock lock.json jupyter-server -o jupyter-server.pex -c jupyter --venv --layout packed'
Benchmark 1: python3.7 -mpex jupyter-server -o jupyter-server.pex -c jupyter --venv --layout packed
  Time (mean ± σ):      9.887 s ±  0.139 s    [User: 21.607 s, System: 2.452 s]
  Range (min … max):    9.786 s … 10.046 s    3 runs
 
Benchmark 2: python3.7 -mpex --lock lock.json jupyter-server -o jupyter-server.pex -c jupyter --venv --layout packed
  Time (mean ± σ):      6.420 s ±  0.603 s    [User: 39.429 s, System: 5.474 s]
  Range (min … max):    5.763 s …  6.949 s    3 runs
 
Summary
  'python3.7 -mpex --lock lock.json jupyter-server -o jupyter-server.pex -c jupyter --venv --layout packed' ran
    1.54 ± 0.15 times faster than 'python3.7 -mpex jupyter-server -o jupyter-server.pex -c jupyter --venv --layout packed'
```

Where cold standard breaks down as:
```
pex: Building pex: 9465.6ms
pex:   Resolving distributions (jupyter-server): 9463.6ms
pex:     Resolving requirements.: 9462.6ms
pex:       Resolving for:
  /home/jsirois/.pyenv/versions/3.7.12/bin/python3.7: 7902.4ms
pex:       Calculating project names for direct requirements:
[...]

PR #1590, opened 2022-01-25
Prepare the 2.1.64 release.

### LINKED ISSUE #1573: Release 2.1.64
On the docket:
+ [ ] Pex does not support mac universal2 wheels #1587

jsirois:
Closes #1573

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,14 @@
 Release Notes
 =============
 
+2.1.64
+------
+
+This release brings support for mac universal2 wheels.
+
+* Update vendored Pip to 386a54f0. (#1589)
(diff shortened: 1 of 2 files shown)


Eric-Arellano (approved):
Thanks. 

Would it make sense to include the universal lock change in this release? Iirc, you're not announcing those tho because it's experimental.

jsirois:
>

I don't think so. Primarily because #1587 is a pressing bug. That aside though, the universal lock is not usable right now since it can't be read. I've got that 1/2 done (extra metadata is now written down in the lock #1585) but the resolve side of using the lock remains (See: #1583). As such no one would benefit from including the ability to write down a fancy lock but not actually use it afaict.

PR #1591, opened 2022-01-25
Upgrade vendored packaging to 20.9.

### LINKED ISSUE #1587: Pex does not support mac universal2 wheels

jsirois:
Also upgrade wheel to latest which upgrades its vendoring of packaging
to 20.9 as well.

Fixes #1587


jsirois:
And this time we have an M1 user having vetted this all: https://github.com/pantsbuild/pants/issues/14244#issuecomment-1021476700

stuhood (approved):

PR #1593, opened 2022-01-25
Prepare the 2.1.65 release.

### LINKED ISSUE #1588: Release 2.1.65
On the docket:
+ [x] Upgrade vendored packaging to 20.9. #1591

jsirois:
Closes #1588

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,16 @@
 Release Notes
 =============
 
+2.1.65
+------
+
+This release really brings support for mac universal2 wheels. The fix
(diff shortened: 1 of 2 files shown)


Eric-Arellano (approved):

jsirois:
It's always mac and it's always the 3.10 IT shard, but more than a fair share of these runs never terminate (they're killed after 6hrs IIRC). This has been true for a few months at least. :/ I <3 macOS.

Eric-Arellano:
We have that in Pants CI too...that macOS shards sometimes just hang. It's been happening ever since I remember adopting GH Actions

PR #1599, opened 2022-02-11
Respect `PEX_ROOT` in `PEXEnvironment.mount`.

### LINKED ISSUE #1598: Spurious warning when building pex with a non-default empty PEX_ROOT
Tested with 2.1.65

If `PEX_ROOT` is set and the default `~/.pex` is not writable, `pex` will warn that `PEX_ROOT` is not writable but still write to it. Once the cache has been populated, `pex` will no longer warn.

It seems deleting `$PEX_ROOT/venvs` is what triggers the warning.

```
$ chmod -w ~/.pex
$ PEX_ROOT=$(mktemp -d)
$ pex -o /tmp/out pex
/tmp/pex/lib/python3.6/site-packages/pex/pex_info.py:384: PEXWarning: PEX_ROOT is configured as /home/user/.pex but that path is un-writeable, falling back to a temporary PEX_ROOT of /tmp/tmp3pjjrz96 which will hurt performance.
  "performance.".format(pex_root=pex_root, tmp_root=tmp_root)
$ ls $PEX_ROOT
http  installed_wheels  interpreters  isolated  pip.pex  venvs
$ pex -o /tmp/out pex    # finishes with no warning
```

jsirois:
There was a narrow case of loose layout PEXes that would not mount in
the `PEX_ROOT` (when set in the environment) that this fixes. That
narrow case was exercised by the `pip.pex` PEX that Pex itself creates
to perform resolves with.

Fixes #1598

### DIFF
--- a/pex/environment.py
+++ b/pex/environment.py
@@ -216,16 +216,18 @@ class PEXEnvironment(object):
         target=None,  # type: Optional[DistributionTarget]
     ):
         # type: (...) -> PEXEnvironment
-        pi = pex_info or PexInfo.from_pex(pex)
-        pex_hash = pi.pex_hash
+        if not pex_info:
+            pex_info = PexInfo.from_pex(pex)
+            pex_info.update(PexInfo.from_env())
+        pex_hash = pex_info.pex_hash
         if pex_hash is None:
             raise AssertionError(
                 "There was no pex_hash stored in {} for {}.".format(PexInfo.PATH, pex)
             )
-        pex_root = pi.pex_root
+        pex_root = pex_info.pex_root
         pex = maybe_install(pex=pex, pex_root=pex_root, pex_hash=pex_hash) or pex
         target = target or DistributionTarget.current()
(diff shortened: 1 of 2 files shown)


pex/environment.py
```diff
@@ -216,16 +216,18 @@ def mount(
         target=None,  # type: Optional[DistributionTarget]
     ):
         # type: (...) -> PEXEnvironment
-        pi = pex_info or PexInfo.from_pex(pex)
-        pex_hash = pi.pex_hash
+        if not pex_info:
+            pex_info = PexInfo.from_pex(pex)
+            pex_info.update(PexInfo.from_env())
```
jsirois, line 221: This is an unsatisfying fix, but there is now (previously failing) test coverage. A more wholistic re-factor of PexInfo / Variables loading is in order, but that's a delicate change that will need to be isolated and deliberated.

Eric-Arellano (approved):

Eric-Arellano (approved):

Issue #1597, opened 2022-02-11
Support specifying foreign platforms in full detail.


jsirois:
We've had two reports of `--platform` use failing due to `python_full_version` environment markers being present in a resolve. This appears to be due to the vagaries of `poetry export` and an artificial issue, but poetry is popular so it seems Pex should provide some work around.

There are two components to decide any resolve supplied by the interpreter being resolved for:
1. The environment https://www.python.org/dev/peps/pep-0508/#environment-markers
2. The compatible tags: https://www.python.org/dev/peps/pep-0425/#use

Pex currently derives as much of the foreign platform environment as it can from the `--platform`:
https://github.com/pantsbuild/pex/blob/cc3859bdbf068d299053f3557a79db3d51864592/pex/platforms.py#L264-L346

And it uses `pip debug -v` to get a list of the wheel tags supported by the foreign interpreter:
https://github.com/pantsbuild/pex/blob/cc3859bdbf068d299053f3557a79db3d51864592/pex/platforms.py#L129-L262

What is probably needed is an escape hatch that allows both 1 and 2 to be specified or extended manually via CLI args. To support this sanely, it probably makes sense to provide a Pex tool that can be run on the foreign platform to generate this data. Simple cases like `python_full_version` could of course be handled by hand, but more complex cases would fall apart in typo hell quite quickly.

jsirois:
#1596 was the 1st report of this.

Eric-Arellano:
> This appears to be due to the vagaries of poetry export and an artificial issue, but poetry is popular so it seems Pex should provide some work around.

I agree, thanks. 

Also note that Pants 2.10 will be shipping soon with multiple resolves as a flagship feature, using Poetry for the lockfile generation. Unlike tool lockfiles, these lockfiles can be used with `--platforms` via `pex_binary` and `python_awslambda`. 

The feature is disabled by default, but the current plan is to strongly encourage it in docs. So I expect the # of cases to increase.

> The feature is disabled by default, but the current plan

A very imperfect plan. Beyond this issue + VCS reqs + `[python-repos]`, Poetry has also been generating for Toolchain some lockfiles with nonsensical env markers like `py==1.11.0; python_version >= "3.6" and python_full_version < "3.0.0" and implementation_name == "pypy" or python_full_version >= "3.5.0" and python_version >= "3.6" and implementation_name == "pypy" `. These cause the build to fail.

[...]

PR #1601, opened 2022-02-11
Extract PEP-508 `MarkerEnvironment`.

jsirois:
Work towards #1597.


pex/pep_508.py
```diff
@@ -0,0 +1,157 @@
+# Copyright 2022 Pants project contributors (see CONTRIBUTORS.md).
```
Eric-Arellano, line 1: Given that the class is called `MarkerEnvironment`, how come you didn't go with `marker_environment.py`?

Eric-Arellano (approved):

pex/pep_508.py
jsirois: In my head - markers == PEP-508. It's true its just a subset of that  PEP though.

PR #1603, opened 2022-02-12
Upgrade isort and black to latest possible.

jsirois:
For isort this is latest stable but for black we have to hang back
pre-stable since stable drops support for Python 2.7.


jsirois:
The only real change is support for sorting imports in all locations.

Eric-Arellano (approved):

PR #1605, opened 2022-02-14
Use setproctitle to sanitize `ps` info.

### LINKED ISSUE #1604: Be able to see what .pex file is run from the list of system processes
**Is your feature request related to a problem? Please describe.**

When running a PEX file:

```
$ python3 dist/project/bar.pex --input="baz"
```

you don’t see much helpful information on `ps ww` command:

```
46131 s010  S+     0:00.78 /usr/bin/python3 /Users/user-name/.pex/unzipped_pexes/e9eb91dcb789973ca601f8bfb4401555f188d841 --input=baz
```

That is, you don't know what `.pex` file is being run by looking at this process.

**Describe the solution you'd like**

The use case is that at a given point of time there may be multiple PEX files running, and it would be handy to see which ones are running from the `ps` output (or somehow else). Perhaps it would be possible to give the unzipped pex a less obscure name. So instead of `unzipped_pexes/2312ebcec214434cf2349791389503f03e1033a7` the PEX could possibly unpack it to `unzipped_pexes/2312ebcec214434cf2349791389503f03e1033a7/bar.pex`.

**Describe alternatives you've considered**

One can inspect the `/Users/user-name/.pex/unzipped_pexes/e9eb91dcb789973ca601f8bfb4401555f188d841/PEX-INFO` file's `entry_point` property to see what module is being run, but this is not visible from the `ps` output directly. This is tedious.

**Additional context**

It's possible that user may have a few long running tasks (started from `.pex` files) and it would be handy to see if any of them are running by doing `ps` and `grep`.

It is possible to set the process title using a 3rd party package (https://pypi.org/project/setproctitle) as well, however, this implies setting it for each `main()` in all the PEXed programs.

https://github.com/pantsbuild/pants/blob/4b335cdccfebcbd1f92caaa5fef7d5ca161c7226/src/python/pants/pantsd/pants_daemon.py#L173-L174

jsirois:
When the setproctitle distribution is part of a PEX's activated
distributions, use it to set the process title to include the PEX file
used to launch the application instead of the final unzipped or venv
directory in the PEX_ROOT.

Fixes #1604

### DIFF
--- a/docs/recipes.rst
+++ b/docs/recipes.rst
@@ -3,6 +3,36 @@
 PEX Recipes and Notes
 =====================
 
+Long running PEX applications and daemons
+-----------------------------------------
+
+If your PEXed application will run a long time, at some point you'll likely need to debug or
+otherwise inspect it using operating system tools. Unless you built your application as a
+non-``--venv`` ``--layout loose`` PEX, its final process information will be inscrutable in ``ps``
+output since all other PEX forms re-execute themselves against an installed version of themselves in
+the configured ``PEX_ROOT``.
+
+You'll see something like this as a result:
+
+.. code-block:: bash
+
+    $ ./my.pex --foo bar &
+    $ ps -o command | grep pex
+    /home/jsirois/.pyenv/versions/3.10.2/bin/python3.10 /home/jsirois/.pex/unzipped_pexes/94790b07dc3768a9926dab999b41a87e399e0aa9 --foo bar
+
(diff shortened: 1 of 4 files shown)


Eric-Arellano (approved):
Very cool! Great docs.

I'm surprised macOS didn't give you more trouble this time 🙃

jsirois:
I have learned to reach for the bsd man pages (https://ss64.com/osx/ps.html) by now, so I was pretty sure the `ps` command would work, but, yeah, I was also ready to be burned.

PR #1607, opened 2022-02-15
Introduce `pex3 interpreter inspect`.

jsirois:
This command can be used to get a full marker environment to
support #1597 complete remote platform specification. Its similar
to the `interpreter` PEX tool but works without having to point
at a PEX which will be what a user of #1597 will want.

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -287,8 +287,11 @@ def configure_clp_pex_options(parser):
 def configure_clp_pex_environment(parser):
     # type: (ArgumentParser) -> None
     group = parser.add_argument_group(
-        "PEX environment options",
-        "Tailor the interpreter and platform targets for the PEX environment.",
+        "PEX target environment options",
+        "Specify which target environments the PEX should run on. If more than one interpreter or "
+        "platform is specified a multi-platform PEX will be created that can run on all specified "
+        "targets. N.B.: You may need to adjust the `--python-shebang` so that it works in all "
+        "the specified target environments.",
     )
 
     target_options.register(group)
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -289,7 +289,7 @@ def configure_clp_pex_environment(parser):
     group = parser.add_argument_group(
         "PEX target environment options",
         "Specify which target environments the PEX should run on. If more than one interpreter or "
-        "platform is specified a multi-platform PEX will be created that can run on all specified "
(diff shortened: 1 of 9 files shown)


pex/bin/pex.py
```diff
@@ -287,8 +287,11 @@ def configure_clp_pex_options(parser):
 def configure_clp_pex_environment(parser):
     # type: (ArgumentParser) -> None
     group = parser.add_argument_group(
-        "PEX environment options",
-        "Tailor the interpreter and platform targets for the PEX environment.",
+        "PEX target environment options",
+        "Specify which target environments the PEX should run on. If more than one interpreter or "
+        "platform is specified a multi-platform PEX will be created that can run on all specified "
```
Eric-Arellano, line 292:
```suggestion
        "platform is specified, a multi-platform PEX will be created that can run on all specified "
```

pex/cli/commands/interpreter.py
```diff
@@ -0,0 +1,150 @@
...
+                        if self.options.markers:
+                            interpreter_info[
+                                "marker_environment"
+                            ] = interpreter.identity.env_markers.as_dict()
+                        self.dump_json(self.options, interpreter_info, out)
+                    else:
+                        out.write(interpreter.binary)
+                    out.write("\n")
+            except UnsatisfiableInterpreterConstraintsError as e:
+                return Error(str(e))
```
Eric-Arellano, line 148: Where does the error come from? `interpreter.resolve_base_interpreter()`? It looks everything else is only creating a dictionary and accessing attr fields, which I wouldn't expect to be fallible.

If possible, it would be helpful to make this `try except` block tighter.

pex/resolve/target_options.py
```diff
@@ -28,11 +32,9 @@ def register(parser):
         type=str,
         action="append",
         help=(
-            "The Python interpreter to use to build the PEX (default: current interpreter). This "
-            "cannot be used with `--interpreter-constraint`, which will instead cause PEX to "
```
Eric-Arellano, line 32: I'm not following why you removed `--interpreter-constraint` comment?

Eric-Arellano (approved):
This makes me even more excited for pex3 🎉

pex/resolve/target_options.py
jsirois: Because it was untrue.

pex/bin/pex.py
jsirois: Fixed.

Eric-Arellano (approved):

PR #1608, opened 2022-02-16
Refactor DistributionTarget.

jsirois:
Refactor DistributionTarget into Target and two concreate subclasses.
This unifies the interface of a Target to make way for a third Target
implementation powered by `pex3 interpreter inspect` data.

Work towards #1597.

PR #1609, opened 2022-02-16
Add support for `--complete-platform`.

### LINKED ISSUE #1597: Support specifying foreign platforms in full detail.

jsirois:
This allows specifying a foreign platform completely such that any
resolve requirement strings can be parsed and the only limitation in a
foreign platform resolve is the availablility of pre-built wheels for
that platform.

Fixes #1597


jsirois:
N.B.: The 1st commit (21edb24) is up for review in #1608.

benjyw (approved):

benjyw (approved):

PR #1614, opened 2022-02-19
Expand `--platform` syntax: support full versions.

jsirois:
Although `--complete-platform` allows every foreign platform resolve with
available wheels to succeed, often `--platform` would be enough if the
`python_full_version` marker environment entry could be derived from it.
Expand the syntax accepted for the PYVER field to support dotted
versions and, when 3-component dotted version is supplied, create a
`python_full_version` marker environment entry.

This should prove to be a more workable corrective than creating and
consuming a `--complete-platform` as introduced in #1609 for most users 
affected by #1597.

### DIFF
--- a/pex/interpreter.py
+++ b/pex/interpreter.py
@@ -294,6 +294,13 @@ class PythonIdentity(object):
         # type: () -> Iterator[Platform]
         """All platforms supported by the associated interpreter ordered from most specific to
         least."""
+        yield Platform(
+            platform=self._platform_tag,
+            impl=self.python_tag[:2],
+            version=self.version_str,
+            version_info=self.version,
+            abi=self.abi_tag,
+        )
         for tag in self._supported_tags:
             yield Platform.from_tag(tag)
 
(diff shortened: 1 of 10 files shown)


jsirois:
I hadn't considered tweaking platform when working #1597 / #1609 since it comes from days of yore. It turned out to be doable though with out breaking old platform strings or any of the guarantees needed in the code base.

pex/interpreter.py
```diff
@@ -294,6 +294,13 @@ def iter_supported_platforms(self):
         # type: () -> Iterator[Platform]
         """All platforms supported by the associated interpreter ordered from most specific to
         least."""
+        yield Platform(
```
benjyw, line 297: I'm unable to grok whether this could duplicate one of the other platforms yielded below, and whether this would matter if so?

benjyw (approved):

pex/interpreter.py
jsirois: It can't, although it wouldn't matter. This platform gets a 3 component full Python version unlike all the rest which are derived from wheel tags which only ever have a 2 component version.

PR #1616, opened 2022-02-20
Lift `Requires-Python` evaluation to Target.

jsirois:
This will be needed to resolve from lock files.


benjyw (approved):

Eric-Arellano (commented):
Thanks for the excellent comment.

PR #1625, opened 2022-02-23
Add support for selecting urls from a lock.

jsirois:
This is the 1st half of resolving from a lock. The second half is
fetching the urls, building any fetched sdists and then installing all
the resulting wheels.

Work towards #1583


jsirois:
Thanks in advance - this is pretty big. The feature - selecting artifacts from a lock - is big and the test is bigger.

jsirois:
The flit fix is out for independent review / land in #1626.

pex/resolve/locked_resolve.py
```diff
@@ -178,6 +194,138 @@ def select_artifact(self, target):
...
+        average_requirement_rank = sum(requirement_ranks) / float(len(requirement_ranks))
+
+        # N.B.: Lowest rank means highest rank value. I.E.: The 1st tag is the most specific and
+        # the 765th tag is the least specific.
+        largest_value = target.supported_tags.lowest_rank.value
+
+        return cls(
+            target_specificity=(largest_value - average_requirement_rank) / largest_value,
+            downloadable_artifacts=tuple(downloadable_artifacts),
+        )
```
Eric-Arellano, line 323: This method also seems like it could benefit from a unit test, that we get the math right.

pex/resolve/locked_resolve.py
```diff
@@ -282,3 +430,203 @@ def emit_artifact(
...
+        prefer_older_binary=False,  # type: bool
+    ):
+        # type: (...) -> Union[Resolved, Error]
+
+        is_local_interpreter = isinstance(target, LocalInterpreter)
+        if not use_wheel:
+            if not build:
+                return Error(
+                    "Cannot both ignore wheels (use_wheel=False) and refrain from building "
+                    "distributions (build=False)."
```
Eric-Arellano, line 450: How challenging would it be to mention the lockfile's path in this error message? I'm thinking something like:

> Cannot resolve lockfile at /path/to/lock.txt: Cannot both ignore wheels ...

This isn't as relevant when using PEX directly via CLI, but would help when using via Pants.

pex/resolve/locked_resolve.py
```diff
@@ -282,3 +430,203 @@ def emit_artifact(
...
+                            vias="\n    ".join(rr.render_via() for rr in resolve_requests),
+                        )
+                    )
+                continue
+
+            resolved_artifacts.append(best_match)
+
+        if errors:
+            return Error(
+                "Failed to resolve all requirements for {target}:\n"
```
Eric-Arellano, line 614: Ditto that it would be helpful to mention the lockfile path.

Eric-Arellano (approved):
It's long but very easy to read! I love the abstractions you've built up like `Error`.

pex/resolve/locked_resolve.py
jsirois: This is easy enough to do - I'll follow up this PR with a smaller PR to plumb this.

pex/targets.py
```diff
@@ -136,6 +136,9 @@ def __str__(self):
         # type: () -> str
         return str(self.platform.tag)
 
+    def render_description(self):
+        raise NotImplementedError()
```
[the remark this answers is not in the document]
jsirois: Only because enough situations rebuff usage that I just no longer use at all. Here its the attr.s class decorator that foils MyPy when combined with ABC / vice-versa. If you don't use ABC and you just `@abstractmethod`, MyPy is fooled.

To boot, this is a sealed class. Type checking is great, but if you need it for code local to 1 file, you're moving too fast.

pex/resolve/resolved_requirement.py
```diff
@@ -39,6 +39,12 @@ def as_requirement(self):
             "{project_name}=={version}".format(project_name=self.project_name, version=self.version)
         )
 
+    def __str__(self):
+        # type: () -> str
+        return "{project_name} {version}".format(
```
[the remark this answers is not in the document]
jsirois: I 1st considered using as_requirement directly since the `__str__` of requirement is what you suggested. I wrote more code only because using a requirement for a realized value was unintuitive to me. The requirement in reality was almost never ansicolors==1.1.8 but maybe `ansicolors>=1,<2` so saying `ansicolors==1.8` says, at 1st blush, I required 1.1.8. I did not. The resolve reified my `ansicolors>=1,<2` requirement to ansicolors 1.1.8.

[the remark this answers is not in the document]
Eric-Arellano: Ah. Makes sense. Thanks for explaining that!

pex/resolve/lock_resolver.py
```diff
@@ -0,0 +1,129 @@
+# Copyright 2022 Pants project contributors (see CONTRIBUTORS.md).
```
Eric-Arellano, line 1: Did you mean to add this?

pex/resolve/locked_resolve.py
```diff
@@ -178,6 +194,138 @@ def select_artifact(self, target):
...
+    def version(self):
+        # type: () -> Union[packaging_version.LegacyVersion, packaging_version.Version]
+        return self.locked_requirement.pin.version.parsed_version
+
+    def select_higher_rank(
+        self,
+        other,  # type: _ResolvedArtifact
+        prefer_older_binary=False,  # type: bool
+    ):
+        # type: (...) -> _ResolvedArtifact
```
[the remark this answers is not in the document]
jsirois: Will do in a follow up.

PR #1627, opened 2022-02-23
Plumb a lock source to resolve error messages.

jsirois:
The Lockfile class already has a source field that callers of
`LockedResolve.resolve` will be able to pass in.


Eric-Arellano (approved):
Thanks! It'd be great to update 435-448 too, those two errors.

jsirois:
> Thanks! It'd be great to update 435-448 too, those two errors.

It would not. That data does not come from any lockfile; so it's be noise. That data comes from Pex CLI flags.

PR #1629, opened 2022-02-24
Move the Result family out of command.

jsirois:
This code is now more widely used than in just leaf command
implementations.


pex/result.py
```diff
@@ -0,0 +1,99 @@
+# Copyright 2022 Pants project contributors (see CONTRIBUTORS.md).
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+from __future__ import absolute_import, print_function
```
jsirois, line 4: This file is all a pure move from command.

Eric-Arellano (approved):

PR #1634, opened 2022-02-28
Support splitting venv creation into deps & srcs.

### LINKED ISSUE #1631: Support a container layer-friendly venv installation mode.
Often application changes involve only source code changes or only dependency changes. In a re-deployment scenario for application changes like this, it would be beneficial to only re-deploy the changed part of the application; especially when the other portion of the application is large. In a container image context this issue is acute and, in particular, it would be useful if a PEX could be split into two portions representing each of these changes separately. Splitting the PEX file itself is problematic. Although Pex has the `--pex-path` / `PEX_PATH` feature to re-join a split PEX, the feature is both unreliable (there is no way of confirming the right pieces have been re-assembled and in the right order to re-form the app you expect) and it performs poorly since the re-assembly requires a PEX runtime execution which is, at minimum, O(50ms) of overhead on startup.

In a container image there is no need for a traditional PEX file at all since the ~"single file" to ship around becomes the container image; so installing a tested PEX file in a venv in the container is the recommended procedure: https://pex.readthedocs.io/en/v2.1.67/recipes.html#pex-app-in-a-container. If the installation from a PEX into a venv could be broken into two steps then separate image layers could be generated for dependencies and application code all the while ensuring these two layers belong together and form a whole app as intended.

So, the idea is, support something like:
```Dockerfile
FROM python:slim as deps
COPY /app.pex /
RUN PEX_TOOLS=1 /app.pex venv --deps-only --compile /app

FROM python:slim as source
COPY /app.pex /
RUN PEX_TOOLS=1 /app.pex venv --source-only --compile /app

FROM python:slim
COPY --from=deps /app /app
COPY --from=sources /app /app
ENTRYPOINT ["/app/pex"]
```

jsirois:
Add a `--scope` option to the `venv` Pex tool that defaults to
installing `all` of a PEX into a venv (the long standing behavior), but
allows splitting up the install into `deps` and `srcs`. This splitting
should be useful for creating optimized container images.

Closes #1631


jsirois:
cc @kaos

pex/venv/pex.py
```diff
@@ -0,0 +1,466 @@
...
+import errno
+import itertools
+import os
+import shutil
+from collections import defaultdict, Counter
+from textwrap import dedent
+
+from pex import pex_warnings
+from pex.common import safe_mkdir, pluralize, chmod_plus_x
+from pex.compatibility import is_valid_python_identifier
```
jsirois, line 15: This file was a move and splitup into functions from the `pex.tools.commands.venv` module.

jsirois:
Thanks in advance for looking at another largish change. Most of this was a move that is poorly tracked by the GH diff viewer from `pex.tools.commands.venv` to `pex.venv.pex`.

tests/integration/tools/commands/test_venv.py
```diff
@@ -133,3 +136,86 @@ def test_collisions_mergeable_issue_1570(tmpdir):
...
+    execute_venv_tool(venv_dir)
+    canonical_venv_listing = recursive_listing(venv_dir)
+    venv = assert_app(venv_dir)
+    # N.B.: Some of the venv activation scripts in bin/ differ since they hard code the venv python
+    # interpreter path in them. Since we have a full canonical_venv_listing to check, we settle on
+    # testing contents are identical just for site-packages which is what is important anyhow: it's
+    # where the deps and user code are installed to.
+    canonical_venv_hash = CacheHelper.dir_hash(venv.site_packages_dir)
+
+    venv_dir_deps_then_srcs = os.path.join(str(tmpdir), "venv.deps-srcs")
```
Eric-Arellano, line 209: Probably worth also testing when you only run `--scope=deps` and only `--scope=srcs`, that it doesn't contain the full PEX.

Eric-Arellano (approved):
Very cool, thanks! I like the test haha

tests/integration/tools/commands/test_venv.py
jsirois: Yeah, will do. My tests of this were the Docker tests where I confirmed image layer re-use by hand but I never ended up encoding those. That's too much so I'll do as you say as a proxy.

docs/recipes.rst
```diff
@@ -43,17 +43,33 @@ container like so:
...
    just like the original PEX.
 2) Pre-compile all PEX Python code installed in the venv.
-3) Remove the original ``/my-app.pex`` as well as any temporary caches created by Pex in the
-   ``PEX_ROOT``, leaving just the newly installed and pre-compiled ``/my-app`` venv.
+
+Notably, the PEX venv install is done in using a
+`multi-stage build <https://docs.docker.com/develop/develop-images/multistage-build/>`_ to ensure
+only the final venv remains on disk and it uses two layers to ensure changes to application code
+do not lead to re-builds of lower layers. This accommodates the common case of modifying and
+re-deploying first party code more often than third party dependencies.
```
thejcannon, line 72: Well that's pretty sweet!

thejcannon (approved):
My novice eyes approve :smile:

Eric-Arellano (approved):
Thanks, test rewrite looks great

PR #1643, opened 2022-03-03
Force stderr to be unbuffered in tests.

jsirois:
Python 3 (Until a fix here in 3.9: https://bugs.python.org/issue13601)
switched from stderr being unbuffered to stderr being buffered by
default (See more here: https://bugs.python.org/issue13597). This can
lead to tests checking stderr failing to see what they expect if the
stderr buffer block has not been flushed. Force stderr line buffering
(which is what setting PYTHONUNBUFFERED nets you) so that tests can
rely on stderr lines being observable.


benjyw (approved):
Nice

PR #1650, opened 2022-03-07
Retain and index downloads when locking.

jsirois:
For the local case of lock create and then later resolve, it will be
good to retain the (primary) artifacts downloaded by Pip for the local
interpreter when creating the lock in order to save downloading those
artifacts again when later resolving from the lock.

Work towards #1583.

### DIFF
--- a/pex/resolve/lockfile/__init__.py
+++ b/pex/resolve/lockfile/__init__.py
@@ -3,23 +3,30 @@
 
 from __future__ import absolute_import
 
+import os
+import shutil
+
 from pex import resolver
-from pex.common import pluralize, safe_open
+from pex.common import pluralize, safe_mkdtemp, safe_open
 from pex.requirements import LocalProjectRequirement, VCSRequirement
 from pex.resolve import resolvers
-from pex.resolve.locked_resolve import LockConfiguration
+from pex.resolve.locked_resolve import Artifact, LockConfiguration
+from pex.resolve.lockfile.download_manager import DownloadedArtifact, DownloadManager
 from pex.resolve.lockfile.lockfile import Lockfile as Lockfile  # For re-export.
 from pex.resolve.requirement_configuration import RequirementConfiguration
 from pex.resolve.resolver_configuration import PipConfiguration
+from pex.resolver import Downloaded
 from pex.result import Error
 from pex.targets import Targets
 from pex.third_party.pkg_resources import Requirement
+from pex.tracer import TRACER
 from pex.typing import TYPE_CHECKING
+from pex.util import CacheHelper
 from pex.variables import ENV
 from pex.version import __version__
 
 if TYPE_CHECKING:
-    from typing import List, Text, Union
+    from typing import List, Mapping, Optional, Text, Union
 
 
 class ParseError(Exception):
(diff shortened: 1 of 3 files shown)


pex/resolve/lockfile/download_manager.py
```diff
@@ -0,0 +1,56 @@
...
+class DownloadedArtifact(object):
+    path = attr.ib()  # type: str
+
+    def fingerprint(self):
+        # type: () -> str
+        with open(os.path.join(os.path.dirname(self.path), "sha1"), "rb") as fp:
+            return str(fp.read().decode("ascii"))
+
+
+class DownloadManager(object):
```
jsirois, line 31: N.B.: The forthcoming lock resolver will have its own implementation of this; thus factoring up the logic of where to store downloads for re-use / cache-hits centrally here.

pex/resolve/lockfile/download_manager.py
```diff
@@ -0,0 +1,56 @@
...
+    def __init__(self, pex_root=None):
+        # type: (Optional[str]) -> None
+        self._download_dir = os.path.join(pex_root or ENV.PEX_ROOT, "downloads")
+
+    def store(self, artifact):
+        # type: (Artifact) -> DownloadedArtifact
+
+        download_dir = os.path.join(self._download_dir, artifact.fingerprint.hash)
+        with atomic_directory(download_dir, exclusive=True) as atomic_dir:
```
jsirois, line 40: There is actually no need for this to be exclusive in this PR; however the lock resolver will also be implementing this and fetching URLs directly from PyPI most of the time. The exclusive lock will prevent two independent processes from trying to download the same artifact from PyPI concurrently and relieve a little pressure from PyPI.

pex/resolve/lockfile/__init__.py
```diff
@@ -77,6 +84,51 @@ def store(
...
+        # type: (...) -> CreateLockDownloadManager
+
+        artifacts_by_filename = {
+            artifact.filename: artifact
+            for locked_resolve in downloaded.locked_resolves
+            for locked_requirement in locked_resolve.locked_requirements
+            for artifact in locked_requirement.iter_artifacts()
+        }
+        path_by_artifact = {
+            artifacts_by_filename[f]: os.path.join(root, f)
```
benjyw, line 104: Are we certain that every file we encounter in the filesystem walk is an artifact? No other crap can end up in there? I.e., is it worth checking that the `artifacts_by_filename[f]` doesn't blow up?

Eric-Arellano (approved):

pex/resolve/lockfile/download_manager.py
```diff
@@ -0,0 +1,56 @@
...
+    from pex.third_party import attr
+
+
+@attr.s(frozen=True)
+class DownloadedArtifact(object):
+    path = attr.ib()  # type: str
+
+    def fingerprint(self):
+        # type: () -> str
+        with open(os.path.join(os.path.dirname(self.path), "sha1"), "rb") as fp:
```
benjyw, line 27: Generally, should we use sha256 in pex for more robust security? sha1 is susceptible to practical collision attacks, and it looks like we use it a lot.

benjyw (approved):

pex/resolve/lockfile/download_manager.py
jsirois: Yeah, so we're exactly as vulnerable as git. I really don't see the practical concern on the collision end as a result; i.e.: people have much bigger problems if this is real since they ~all use git!

On the practical side of this is filesystem path name length limits. This is already a problem we run into and pex has warnings it emits when things are looking dangerous. Going to sha256 would net 24 more characters in some paths and 48 in others (venvs).

pex/resolve/lockfile/__init__.py
jsirois: We're certain. This class should probably be named `_CreateLockDownloadManager` to underscore its local-only private use, but down here we set up the empty temporary download dir: https://github.com/pantsbuild/pex/pull/1650/files#diff-eda2c2a78c6edeac766676b01baf61449994f90376ca6d9feb099e2817edaeceR175 and the Pip we vendor only plops distributions from our call to `pip download --dest ...` there.

pex/resolve/lockfile/download_manager.py
jsirois: This may be worthwhile to investigate; although using a lock file mitigates afaict since the primary cache key becomes the hash algorithm used by the artifact - generally sha256. All the rest of the keys are then just internal. To attack one of those you'd need to 1st break the artifact sha256 hash. So, for those _not_ using lock files, Pex is more vulnerable than it need be as you say. The path lengths are the sticky wicket.

jsirois: https://github.com/pantsbuild/pex/issues/1655

PR #1651, opened 2022-03-07
Extract `parse_lockable_requirements`.

jsirois:
Resolving from a lockfile will need this too.

PR #1654, opened 2022-03-07
Introduce `--lock` resolves.

### LINKED ISSUE #1583: Support resolving using a Pex lock file.

jsirois:
The Pex CLI can now resolve from a lock file created via
`pex3 lock create` when creating PEX files.

Closes #1583

### DIFF
--- a/pex/cli/commands/lock.py
+++ b/pex/cli/commands/lock.py
@@ -77,6 +77,7 @@ class Lock(OutputMixin, JsonMixin, BuildTimeCommand):
         resolver_options.register(
             cls._create_resolver_options_group(parser),
             include_pex_repository=False,
+            include_lock=False,
         )
 
     @classmethod
--- a/pex/resolve/lock_resolver.py
+++ b/pex/resolve/lock_resolver.py
@@ -1,0 +1,289 @@
+# Copyright 2022 Pants project contributors (see CONTRIBUTORS.md).
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+from __future__ import absolute_import
+
+import functools
+import hashlib
+from collections import OrderedDict
+from multiprocessing.pool import ThreadPool
+
+from pex.common import pluralize
+from pex.compatibility import cpu_count
+from pex.fetcher import URLFetcher
+from pex.network_configuration import NetworkConfiguration
+from pex.orderedset import OrderedSet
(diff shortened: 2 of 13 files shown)


jsirois:
N.B.: The 1st two commits are under review elsewhere:
e46c06d: #1650 
168d64e: #1651

Please just focus on ffd28cd.

pex/cli/commands/lock.py
```diff
@@ -77,6 +77,7 @@ def _add_resolve_options(cls, parser):
         resolver_options.register(
             cls._create_resolver_options_group(parser),
             include_pex_repository=False,
+            include_lock=False,
```
Eric-Arellano, line 80: @benjyw this is why it's an option to disable registration.

pex/resolve/lock_resolver.py
```diff
@@ -0,0 +1,289 @@
...
+from pex.targets import Target, Targets
+from pex.tracer import TRACER
+from pex.typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from typing import Dict, Iterable, Optional, Sequence, Union
+
+
+class URLFetcherDownloadManager(DownloadManager):
+    _BUFFER_SIZE = 65536
```
Eric-Arellano, line 35: I hadn't seen this number used before but knew you chose it for a reason. Fine if you don't want to burn a tree, but otherwise this would help newbs like me

```suggestion
    _BUFFER_SIZE = 65536  # i.e. max number of 16 bits.
```

Eric-Arellano (approved):
Epic! 🎉

pex/resolve/lock_resolver.py
jsirois: Yeah - not the reason. I can circle back, but, as always, buffer size selection is all about perf and nothing more. The number of bits may tangentially relate to the size of underlying os buffers, but its not the point, the point, hopefully, is you measured on some reasonably diverse set of inputs and found a best-fit. That's what this was.

stuhood:
Very exciting!

Issue #1657, opened 2022-03-09
`--lock` with no input requirements should resolve the whole lock


Eric-Arellano:
With this lockfile created via `pex3 lock create ansicolors -o colors.lock --indent=2`:

```json
{
  "allow_builds": true,
  "allow_prereleases": false,
  "allow_wheels": true,
  "build_isolation": true,
  "constraints": [],
  "locked_resolves": [
    {
      "locked_requirements": [
        {
          "artifacts": [
            {
              "algorithm": "sha256",
              "hash": "00d2dde5a675579325902536738dd27e4fac1fd68f773fe36c21044eb559e187",
              "url": "https://files.pythonhosted.org/packages/53/18/a56e2fe47b259bb52201093a3a9d4a32014f9d85071ad07e9d60600890ca/ansicolors-1.1.8-py2.py3-none-any.whl"
            }
          ],
          "project_name": "ansicolors",
          "requires_dists": [],
          "requires_python": null,
          "version": "1.1.8"
        }
      ],
      "platform_tag": [
        "cp39",
        "cp39",
        "macosx_11_0_arm64"
      ]
    }
  ],
  "pex_version": "2.1.70",
  "prefer_older_binary": false,
  "requirements": [
    "ansicolors"
  ],
  "requires_python": [],
  "resolver_version": "pip-legacy-resolver",
  "style": "strict",
  "transitive": true,
  "use_pep517": null
}
```

I get this error:

```
❯ ../pex/dist/pex --lock colors.lock
Traceback (most recent call last):
  File "/Users/ericarellano/.pex/unzipped_pexes/4bbee6e719f7c34181394d0293cbdafc19203558/.bootstrap/pex/pex.py", line 503, in execute
    exit_value = self._wrap_coverage(self._wrap_profiling, self._execute)
  File "/Users/ericarellano/.pex/unzipped_pexes/4bbee6e719f7c34181394d0293cbdafc19203558/.bootstrap/pex/pex.py", line 408, in _wrap_coverage
    return runner(*args)
  File "/Users/ericarellano/.pex/unzipped_pexes/4bbee6e719f7c34181394d0293cbdafc19203558/.bootstrap/pex/pex.py", line 439, in _wrap_profiling
    return runner(*args)
  File "/Users/ericarellano/.pex/unzipped_pexes/4bbee6e719f7c34181394d0293cbdafc19203558/.bootstrap/pex/pex.py", line 559, in _execute
[...]

PR #1658, opened 2022-03-10
`--lock` without requirements installs the entire lock.

Eric-Arellano:
Closes https://github.com/pantsbuild/pex/issues/1657.

A followup will make sure that we can support `pex --lock` when the lock is empty.

### DIFF
--- a/pex/resolve/lock_resolver.py
+++ b/pex/resolve/lock_resolver.py
@@ -119,7 +119,7 @@ def resolve_from_lock(
             parse_lockable_requirements(
                 requirement_configuration,
                 network_configuration=network_configuration,
-                fallback_requirements=lock.requirements,
+                fallback_requirements=(str(req) for req in lock.requirements),
             )
         )
 
--- a/pex/resolve/lock_resolver.py
+++ b/pex/resolve/lock_resolver.py
@@ -119,9 +119,18 @@ def resolve_from_lock(
             parse_lockable_requirements(
                 requirement_configuration,
                 network_configuration=network_configuration,
-                fallback_requirements=(str(req) for req in lock.requirements),
+                fallback_requirements=lock.requirements,
             )
         )
+        if not parsed_requirements:
+            raise ResultError(
+                Error(
+                    "No requirements requested. This should not be possible: either users will "
+                    "have specified requirements on the command line, or we will fall back to "
+                    "resolving the entire lockfile. Please file a bug at "
+                    "https://github.com/pantsbuild/pex/issues/new."
+                )
+            )
 
     errors_by_target = {}  # type: Dict[Target, Iterable[Error]]
     downloadable_artifacts = OrderedSet()  # type: OrderedSet[DownloadableArtifact]
--- a/pex/resolve/lock_resolver.py
+++ b/pex/resolve/lock_resolver.py
@@ -122,15 +122,6 @@ def resolve_from_lock(
                 fallback_requirements=lock.requirements,
             )
         )
-        if not parsed_requirements:
-            raise ResultError(
-                Error(
-                    "No requirements requested. This should not be possible: either users will "
-                    "have specified requirements on the command line, or we will fall back to "
-                    "resolving the entire lockfile. Please file a bug at "
-                    "https://github.com/pantsbuild/pex/issues/new."
-                )
-            )
 
     errors_by_target = {}  # type: Dict[Target, Iterable[Error]]
     downloadable_artifacts = OrderedSet()  # type: OrderedSet[DownloadableArtifact]
--- a/pex/resolve/lock_resolver.py
+++ b/pex/resolve/lock_resolver.py
@@ -117,7 +117,9 @@ def resolve_from_lock(
         )
         parsed_requirements = try_(
             parse_lockable_requirements(
-                requirement_configuration, network_configuration=network_configuration
+                requirement_configuration,
+                network_configuration=network_configuration,
+                fallback_requirements=(str(req) for req in lock.requirements),
             )
         )
 
--- a/pex/resolve/lockfile/__init__.py
+++ b/pex/resolve/lockfile/__init__.py
@@ -124,12 +124,15 @@ class Requirements(object):
 def parse_lockable_requirements(
     requirement_configuration,  # type: RequirementConfiguration
     network_configuration=None,  # type: Optional[NetworkConfiguration]
+    fallback_requirements=None,  # type: Optional[Iterable[str]]
 ):
     # type: (...) -> Union[Requirements, Error]
 
     parsed_requirements = []  # type: List[Union[PyPIRequirement, URLRequirement]]
     projects = []  # type: List[str]
-    for parsed_requirement in requirement_configuration.parse_requirements(network_configuration):
+    for parsed_requirement in requirement_configuration.parse_requirements(
+        network_configuration, fallback_requirements=fallback_requirements
+    ):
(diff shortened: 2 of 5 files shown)


pex/resolve/lock_resolver.py
```diff
@@ -117,7 +117,9 @@ def resolve_from_lock(
         )
         parsed_requirements = try_(
             parse_lockable_requirements(
-                requirement_configuration, network_configuration=network_configuration
+                requirement_configuration,
+                network_configuration=network_configuration,
+                fallback_requirements=(str(req) for req in lock.requirements),
```
Eric-Arellano, line 122: `Lockfile.requirements` is a sequence of `pkg_resources.Requirement`, whereas we need `pex.requirements.ParsedRequirement`. Our parser for that, `parse_requirement_strings`, expects strings.

pex/resolve/lockfile/__init__.py
```diff
@@ -155,6 +158,13 @@ def parse_lockable_requirements(
             )
         )
 
+    if not parsed_requirements:
+        return Error(
+            "No requirements requested. This should not be possible: either users will have "
+            "specified requirements on the command line, or we will fall back to resolving the "
+            "entire lockfile. Please file a bug at https://github.com/pantsbuild/pex/issues/new."
```
Eric-Arellano, line 165: Some defensive programming - note that Pex already complains if you use a lock with no requirements. But I figure this is better than the `ZeroDivisionError` from https://github.com/pantsbuild/pex/issues/1657.

pex/resolve/requirement_configuration.py
```diff
@@ -24,8 +24,12 @@ class RequirementConfiguration(object):
     requirement_files = attr.ib(default=None)  # type: Optional[Iterable[str]]
     constraint_files = attr.ib(default=None)  # type: Optional[Iterable[str]]
 
-    def parse_requirements(self, network_configuration=None):
-        # type: (Optional[NetworkConfiguration]) -> Iterable[ParsedRequirement]
+    def parse_requirements(
+        self,
+        network_configuration=None,  # type: Optional[NetworkConfiguration]
+        fallback_requirements=None,  # type: Optional[Iterable[str]]
```
Eric-Arellano, line 30: `RequirementConfiguration` is used in several places throughout the codebase, so moving `fallback_requirements` into a property on the object seemed incorrect. Lmk if there's a better factoring I should use.

pex/resolve/resolver_options.py
```diff
@@ -119,8 +120,9 @@ def register(
             default=None,
             type=str,
             help=(
-                "Resolve requirements from the given lock file instead of from --index servers, "
-                "--find-links repos or a --pex-repository."
+                "Resolve requirements from the given PEX lock file instead of from --index "
```
jsirois, line 123: PEX file is right above, a.pex is a PEX file, but Pex lock file is right here, this is a lock file created by Pex (the tool).

tests/integration/test_lock_resolver.py
```diff
@@ -111,21 +111,28 @@ def test_subset(
...
+    ]
+    run_pex_command(args).assert_success()
     pex_distributions = index_pex_distributions(urllib3_pex)
-    assert ProjectName("requests") not in pex_distributions
     assert ProjectName("urllib3") in pex_distributions
+    for project in ("requests", "idna", "chardet", "certifi"):
+        assert ProjectName(project) not in pex_distributions
+
+    # However, if no requirements are specified, resolve the entire lock.
+    args.pop(2)
```
jsirois, line 131: That's pretty obscure! I'd personally extract a nested function that took optional *requirements. Clearly I made sense of this code as it was though, so there you go.

pex/resolve/lock_resolver.py
```diff
@@ -117,7 +117,9 @@ def resolve_from_lock(
         )
         parsed_requirements = try_(
```
jsirois, line 118: This is all fine and works, but you've also pushed a concern of `lock_resolver.py` to lots of other places. It's not clear to me anyway that users of all those places have the same concerns. The Error you return from deeper down definitely makes sense when its returned from right here, I'm not sure about other sites. Did you consider just ~:
```python
if not parsed_requirements.requirements:
    parsed_requirements = Requirements.create(
        parsed_requirements=parse_requirement_strings(
            str(req) for req in lock.requirements
        ),
        parsed_constraints=parsed_requirements.parsed_constraints
    )

```

jsirois (approved):
> This mirrors the behavior of `--pex-repository`.

It actually doesn't. I checked this out on the CLI since my memory is so bad. Although zero reqs are tolerated, the result is an ~empty pex. I looked and I did go back and introduce a `PEXEnvironment.resolve()` but the --pex-repository resolver still used `PEXEnvironment.resolve_dists(...)` in all cases.

pex/resolve/lock_resolver.py
Eric-Arellano: You're right: `pex3 lock create` is using `parse_lockable_requirements` and this PR resulted in that now erroring when no requirements are specified. I do think we should error (https://github.com/pantsbuild/pex/issues/1659), but this is a bad error message. Callers should raise the error.

Re doing the parsing in `lock_resolver.py` directly, I started with that, but realized we want the logic to filter out VCS and local requirements. I took your suggestion though and reverted changes to `requirement_configuration.py`.

Thanks!

Eric-Arellano:
> I checked this out on the CLI since my memory is so bad. Although zero reqs are tolerated, the result is an ~empty pex. 

Ah my testing was too lazy. I should have tried doing imports in the REPL or inspected PEX-INFO. Thanks / my bad.

pex/resolve/lock_resolver.py
jsirois: Since filtering already occurs before a lock is created, the requirements written down there to be used as fallback are kosher. MyPy is telling you a lie this time!

pex/resolve/lock_resolver.py
```diff
@@ -117,9 +117,20 @@ def resolve_from_lock(
...
             parse_lockable_requirements(
-                requirement_configuration, network_configuration=network_configuration
+                requirement_configuration,
+                network_configuration=network_configuration,
+                fallback_requirements=lock.requirements,
             )
         )
+        if not parsed_requirements:
```
jsirois, line 125: MyPy lying to you again. This will never be None!

Eric-Arellano: This wasn't MyPy complaining - this was me being defensive to avoid the `ZeroDivisionError` in case our assertions ever are violated. 

I'll instead take your suggestion on https://github.com/pantsbuild/pex/issues/1659 to tolerate empty lockfiles. So will remove this for now.

pex/resolve/lock_resolver.py
jsirois: I guess my point is the coversion from ParsedRequirement is lossy on the typing side but it never actually matters since we vet the ParsedRequirement types before ever writing them as Requirements in the lockfile; so MyPy will always be noise here.

pex/resolve/lock_resolver.py
jsirois: Moot since it's gone, but you missed the point I think parsed_requirements is a compound object that is never falsey (None) on this line. What you meant to check was `if not parsed_requirements.requirements`. MyPy was truly lying to you since Python supports `__bool__`.

jsirois (approved):

Issue #1659, opened 2022-03-10
`pex3 lock create` should error if no requirements are specified


Eric-Arellano:
This works: `pex3 lock create -o empty.lock`

But then loading it fails:

```
❯ python3 -m pex --lock empty.lock                                                  
Traceback (most recent call last):
  File "/opt/homebrew/Cellar/python@3.9/3.9.10/Frameworks/Python.framework/Versions/3.9/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/opt/homebrew/Cellar/python@3.9/3.9.10/Frameworks/Python.framework/Versions/3.9/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/Users/ericarellano/code/pex/pex/__main__.py", line 8, in <module>
    __name__ == "__main__" and pex.main()
  File "/Users/ericarellano/code/pex/pex/bin/pex.py", line 724, in main
    do_main(
  File "/Users/ericarellano/code/pex/pex/bin/pex.py", line 745, in do_main
    pex_builder = build_pex(
  File "/Users/ericarellano/code/pex/pex/bin/pex.py", line 594, in build_pex
    resolve_from_lock(
  File "/Users/ericarellano/code/pex/pex/resolve/lock_resolver.py", line 110, in resolve_from_lock
    lock = lockfile.load(lockfile_path)
  File "/Users/ericarellano/code/pex/pex/resolve/lockfile/__init__.py", line 57, in load
    return json_codec.load(lockfile_path=lockfile_path)
  File "/Users/ericarellano/code/pex/pex/resolve/lockfile/json_codec.py", line 266, in load
    return loads(fp.read(), source=lockfile_path)
  File "/Users/ericarellano/code/pex/pex/resolve/lockfile/json_codec.py", line 138, in loads
    raise ParseError(
pex.resolve.lockfile.ParseError: Expected '.requirements' in empty.lock to have at least one requirement.
```

This makes sense to me to error - an empty lock is meaningless.

jsirois:
My only reservation with your conclusion is that an empty PEX totally makes sense and has been a foundational showcase of Pex. For example run `pex`. I was confused by what happens when you do that bitd but Wickman had wisdom. It drops you in an isolated interpreter which totally makes sense and is a large part of Pex's raison d`etre: isolation from the environment. So, consider fixing the back half instead of the front half.

[...]

PR #1660, opened 2022-03-10
Support lockfiles with no requirements

Eric-Arellano:
Closes https://github.com/pantsbuild/pex/issues/1659.


Eric-Arellano (commented):
Note that an empty lockfile looks like this:

```
{
  "allow_builds": true,
  "allow_prereleases": false,
  "allow_wheels": true,
  "build_isolation": true,
  "constraints": [],
  "locked_resolves": [],
  "pex_version": "2.1.65",
  "prefer_older_binary": false,
  "requirements": [],
  "requires_python": [],
  "style": "strict",
  "transitive": true,
  "use_pep517": null
}
```

That doesn't seem useful to me, but perhaps its use will grow over time.

jsirois (approved):

PR #1671, opened 2022-03-15
Extract Pip runtime patches.

### LINKED ISSUE #1581: Break out Pip runtime patches to a dedicated (re)source file.
There are 4 patches currently and they're all contained in a big ~200 line string. That's fairly unmaintainable.

jsirois:
This moves the Pip tool into its own package and extracts the Pip runtime
patching code into a sibling resource that can be type-checked and linted.

Fixes #1581

### DIFF
--- a/pex/pip/runtime_patches.py
+++ b/pex/pip/runtime_patches.py
@@ -1,0 +1,193 @@
+# Copyright 2022 Pants project contributors (see CONTRIBUTORS.md).
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+# Various runtime patches applied to Pip to work around issues or else bend Pip to Pex's needs.
+
+import os
+import runpy
+
+# N.B.: The following environment variables are used by the Pex runtime to control Pip and must be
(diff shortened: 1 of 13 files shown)


pex/pip/runtime_patches.py
```diff
@@ -0,0 +1,193 @@
...
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+# Various runtime patches applied to Pip to work around issues or else bend Pip to Pex's needs.
+
+import os
+import runpy
+
+# N.B.: The following environment variables are used by the Pex runtime to control Pip and must be
+# kept in-sync with `tool.py`.
+skip_markers = os.environ.pop("_PEX_SKIP_MARKERS", None)
```
jsirois, line 11: I extracted these 4 runtime control env var accesses into this block and added some type ignores for the pip imports inline below; otherwise, this file was a straight move from the source code string in Pip with `tox -eformat-run` applied.

Eric-Arellano (approved):

PR #1672, opened 2022-03-15
Ensure universal locks cover IC range.

### LINKED ISSUE #1667: Lockfiles: requirement might not be compatible with requested interpreter constraints
`pex3 lock create --style=universal --resolver-version=pip-2020-resolver --indent=2 --interpreter-constraint='CPython>=3.7' ipython` results in a lockfile that chooses `8.1.1` for `ipython`, which has: `"requires_python": ">=3.8"`. That means that the lockfile fails when run with Python 3.7.

Instead, we ideally would choose an ipython version from 7.*, the latest version compatible with 3.7. Note that this does happen if you use `--interpreter-constraint='CPython==3.7.*'`.

jsirois:
Previously any dependency that satisfied just one interpreter version
in an IC range would be picked which was problematic if that dependency
version did not satisfy all interpreters in the range. Force a universal
resolve with an IC range to satisfy every Python version in that range.

Fixes #1667

### DIFF
--- a/pex/interpreter_constraints.py
+++ b/pex/interpreter_constraints.py
@@ -190,7 +190,7 @@ COMPATIBLE_PYTHON_VERSIONS = (
     PythonVersion(Lifecycle.EOL, 3, 5, 10),
     PythonVersion(Lifecycle.EOL, 3, 6, 15),
     PythonVersion(Lifecycle.STABLE, 3, 7, 12),
-    PythonVersion(Lifecycle.STABLE, 3, 8, 11),
+    PythonVersion(Lifecycle.STABLE, 3, 8, 12),
     PythonVersion(Lifecycle.STABLE, 3, 9, 10),
     PythonVersion(Lifecycle.STABLE, 3, 10, 2),
     PythonVersion(Lifecycle.DEV, 3, 11, 0),
--- a/pex/pip/runtime_patches.py
+++ b/pex/pip/runtime_patches.py
@@ -141,7 +141,11 @@ if skip_markers:
             orig_check_requires_python = packaging.check_requires_python
 
             def check_requires_python(requires_python, *_args, **_kw):
-                return any(
+                # Ensure any dependency we lock is compatible with the full interpreter range
+                # specified since we have no way to force Pip to backtrack and follow paths for any
+                # divergences. Most (all?) true divergences should be covered by forked environment
+                # markers.
+                return all(
                     orig_check_requires_python(requires_python, python_full_version)
                     for python_full_version in python_full_versions
                 )
--- a/tests/integration/cli/commands/test_issue_1667.py
+++ b/tests/integration/cli/commands/test_issue_1667.py
@@ -66,10 +66,17 @@ def test_interpreter_constraints_range_coverage(
         ],
     ).assert_success()
 
-    run_ipython_args = [ipython_pex, "-c", "import IPython; print(IPython.__file__)"]
-    output = subprocess.check_output(args=[py37.binary] + run_ipython_args)
-    assert output.decode("utf-8").strip().startswith(pex_root)
+    def assert_pex_works(python):
+        # type: (str) -> None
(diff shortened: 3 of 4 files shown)


pex/interpreter_constraints.py
```diff
@@ -190,7 +190,7 @@ def iter_compatible_versions(
     PythonVersion(Lifecycle.EOL, 3, 5, 10),
     PythonVersion(Lifecycle.EOL, 3, 6, 15),
     PythonVersion(Lifecycle.STABLE, 3, 7, 12),
-    PythonVersion(Lifecycle.STABLE, 3, 8, 11),
+    PythonVersion(Lifecycle.STABLE, 3, 8, 12),
```
jsirois, line 193: I updated this as I contemplated making no ICs map to all interpreter versions supported by Pex. I did not make that switch, but left the update.

tests/integration/cli/commands/test_issue_1667.py
```diff
@@ -0,0 +1,63 @@
...
+        "-o",
+        lock,
+    ).assert_success()
+
+    pex_root = os.path.join(str(tmpdir), "pex_root")
+    ipython_pex = os.path.join(str(tmpdir), "ipython.pex")
+    run_pex_command(
+        args=[
+            "--interpreter-constraint",
+            ">=3.7,<3.11",
```
Eric-Arellano, line 48: If you were to run this test on a machine without Python 3.7 discoverable (such that Pex tries to install for 3.7), I suspect this test would still pass. Note that the offending lockfile from https://github.com/pantsbuild/pex/issues/1667 works when run with Python 3.8+, only not 3.7.

Perhaps one of these strategies instead?
- Force the lockfile consumption to use Python 3.7, or
- inspect the lock's selected version

pex/pip/runtime_patches.py
```diff
@@ -141,7 +141,11 @@ def patch_requires_python():
             orig_check_requires_python = packaging.check_requires_python
 
             def check_requires_python(requires_python, *_args, **_kw):
-                return any(
+                # Ensure any dependency we lock is compatible with the full interpreter range
+                # specified since we have no way to force Pip to backtrack and follow paths for any
+                # divergences. Most (all?) true divergences should be covered by forked environment
+                # markers.
+                return all(
```
Eric-Arellano, line 148: Checking my understanding: if there is no version that's compatible with the IC range, we now will fail to lock. Whereas before we'd lock w/ a version that might only work with a subset of the ICs. Right?

If so, that's maybe worth a test?

Eric-Arellano (commented):
Thank you.

tests/integration/cli/commands/test_issue_1667.py
jsirois: The pip-2020-resolver takes ~infinite time to resolve this under Python 2.7; thus this approach for now. Still looking into why that is the case.

jsirois: IOW, It works great under `tox -epy{py3,{5,6,7,8,9,10}}...`. It stalls, makes incredibly slow progress, but effectively halts, for `tox -epy{py,27}...`.

Eric-Arellano: I'm not totally following - setting `==3.7.*` causes Py2 to stall? 

Alternative wording: I'm trying to make sure we can be confident Pex is building with Python 3.7, and not using 3.8+ thanks to the loose constraint of `>=3.7,<3.11`.

jsirois: I'll give up and stick py37 back in on the consume side. It doesn't actually provide any more coverage but answers the sort of question you're having more directly I guess.

jsirois: Fixed.

Eric-Arellano (approved):
Thank you!

PR #1694, opened 2022-03-25
Fix #1693

### LINKED ISSUE #1693: "Resource deadlock avoided" when creating multiple `--venv` PEXs in parallel
Testing out some change in Pants, I create some (>10 < 60) number of processes responsible for creating a PEX venv (`--venv`) using a PEX lockfile (`--lock`) specifying some requirement strings which among the processes are likely highly overlapping yet unique subset of my lockfile.

The stacktrace is at the end of this issue. The issue is regularly reproducible in our work repo, but not reproducible in Pants' own repo with the relevant Pants code running (Pants is only relevant because that's how I'm opening the N processes).

Let me know what else you need to help debug and solve.

Stacktrace:
```
Traceback (most recent call last):                                                                                                                                            File "/home/joshuacannon/.cache/pants/named_caches/pex_root/unzipped_pexes/15004f7f47d92f2a9c0b754a1c6d2997f332ea44/.bootstrap/pex/pex.py", line 504, in execute          
    exit_value = self._wrap_coverage(self._wrap_profiling, self._execute)                                                                                                     File "/home/joshuacannon/.cache/pants/named_caches/pex_root/unzipped_pexes/15004f7f47d92f2a9c0b754a1c6d2997f332ea44/.bootstrap/pex/pex.py", line 409, in _wrap_coverage   
    return runner(*args)                                                                                                                                                      File "/home/joshuacannon/.cache/pants/named_caches/pex_root/unzipped_pexes/15004f7f47d92f2a9c0b754a1c6d2997f332ea44/.bootstrap/pex/pex.py", line 440, in _wrap_profiling  
    return runner(*args)                                                                                                                                                      File "/home/joshuacannon/.cache/pants/named_caches/pex_root/unzipped_pexes/15004f7f47d92f2a9c0b754a1c6d2997f332ea44/.bootstrap/pex/pex.py", line 560, in _execute         
    return self.execute_entry(self._pex_info.entry_point)                                                                                                                   
  File "/home/joshuacannon/.cache/pants/named_caches/pex_root/unzipped_pexes/15004f7f47d92f2a9c0b754a1c6d2997f332ea44/.bootstrap/pex/pex.py", line 696, in execute_entry
    return self.execute_pkg_resources(entry_point)
  File "/home/joshuacannon/.cache/pants/named_caches/pex_root/unzipped_pexes/15004f7f47d92f2a9c0b754a1c6d2997f332ea44/.bootstrap/pex/pex.py", line 728, in execute_pkg_resources
    return runner()
  File "/home/joshuacannon/.cache/pants/named_caches/pex_root/installed_wheels/0457d0c3fb526f3f246a04ebe5fa67dfa4f0877fe617e106f23f3b083f7b1ab1/pex-2.1.73-py2.py3-none-any.whl/pex/bin/pex.py", line 724, in main
    do_main(
  File "/home/joshuacannon/.cache/pants/named_caches/pex_root/installed_wheels/0457d0c3fb526f3f246a04ebe5fa67dfa4f0877fe617e106f23f3b083f7b1ab1/pex-2.1.73-py2.py3-none-any.whl/pex/bin/pex.py", line 745, in do_main
    pex_builder = build_pex(
  File "/home/joshuacannon/.cache/pants/named_caches/pex_root/installed_wheels/0457d0c3fb526f3f246a04ebe5fa67dfa4f0877fe617e106f23f3b083f7b1ab1/pex-2.1.73-py2.py3-none-any.whl/pex/bin/pex.py", line 594, in build_pex
    resolve_from_lock(
  File "/home/joshuacannon/.cache/pants/named_caches/pex_root/installed_wheels/0457d0c3fb526f3f246a04ebe5fa67dfa4f0877fe617e106f23f3b083f7b1ab1/pex-2.1.73-py2.py3-none-any.whl/pex/resolve/lock_resolver.py", line 198, in resolve_from_lock
    pool.map(
  File "/usr/lib/python3.8/multiprocessing/pool.py", line 364, in map
    return self._map_async(func, iterable, mapstar, chunksize).get()
  File "/usr/lib/python3.8/multiprocessing/pool.py", line 771, in get
    raise self._value
  File "/usr/lib/python3.8/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
  File "/usr/lib/python3.8/multiprocessing/pool.py", line 48, in mapstar
    return list(map(*args))
  File "/home/joshuacannon/.cache/pants/named_caches/pex_root/installed_wheels/0457d0c3fb526f3f246a04ebe5fa67dfa4f0877fe617e106f23f3b083f7b1ab1/pex-2.1.73-py2.py3-none-any.whl/pex/result.py", line 101, in catch
    return func(*args, **kwargs)
  File "/home/joshuacannon/.cache/pants/named_caches/pex_root/installed_wheels/0457d0c3fb526f3f246a04ebe5fa67dfa4f0877fe617e106f23f3b083f7b1ab1/pex-2.1.73-py2.py3-none-any.whl/pex/resolve/lock_resolver.py", line 46, in store_downloadable_artifact
    return self.store(artifact=downloadable_artifact.artifact)
  File "/home/joshuacannon/.cache/pants/named_caches/pex_root/installed_wheels/0457d0c3fb526f3f246a04ebe5fa67dfa4f0877fe617e106f23f3b083f7b1ab1/pex-2.1.73-py2.py3-none-any.whl/pex/resolve/lockfile/download_manager.py", line 40, in store
    with atomic_directory(download_dir, exclusive=True) as atomic_dir:
  File "/usr/lib/python3.8/contextlib.py", line 113, in __enter__
    return next(self.gen)
  File "/home/joshuacannon/.cache/pants/named_caches/pex_root/installed_wheels/0457d0c3fb526f3f246a04ebe5fa67dfa4f0877fe617e106f23f3b083f7b1ab1/pex-2.1.73-py2.py3-none-any.whl/pex/common.py", line 470, in atomic_directory
    fcntl.lockf(lock_fd, fcntl.LOCK_EX)  # A blocking write lock.
OSError: [Errno 35] Resource deadlock avoided
```

thejcannon:
The comments added should explain the case and solution.

This has been tested using the same setup/commands as previously used to reliably demonstrate the bug.
Additionally, I modified the code as follows
```
        assert_is_finalized = False
        while True:
            try:
                fcntl.lockf(lock_fd, fcntl.LOCK_EX)  # A blocking write lock.
            except OSError as e:
                if e.errno == errno.EDEADLK:  # Resource deadlock avoided
                    assert_is_finalized = True
                    time.sleep(1)
                    continue
                raise

            is_finalized = atomic_dir.is_finalized()
            if assert_is_finalized:
                assert is_finalized
```
and never saw the `assert`. That tells me that our assumption that in this case we don't have exclusivity and something else does is true.

### DIFF
--- a/pex/common.py
+++ b/pex/common.py
@@ -464,10 +464,19 @@ def atomic_directory(target_dir, exclusive, source=None):
             os.path.join(head, ".{}.atomic_directory.lck".format(tail or "here")),
             os.O_CREAT | os.O_WRONLY,
         )
-        # N.B.: Since lockf operates on an open file descriptor and these are guaranteed to be
-        # closed by the operating system when the owning process exits, this lock is immune to
-        # staleness.
-        fcntl.lockf(lock_fd, fcntl.LOCK_EX)  # A blocking write lock.
+        try:
+            # N.B.: Since lockf operates on an open file descriptor and these are guaranteed to be
+            # closed by the operating system when the owning process exits, this lock is immune to
+            # staleness.
+            fcntl.lockf(lock_fd, fcntl.LOCK_EX)  # A blocking write lock.
+        except OSError as e:
+            if e.errno == 35:  # Resource deadlock avoided
+                # Another thread/process is doing the work. They get the exclusivity.
+                # N.B.: This can happen if the process running this code forks, this is the same
+                # behavior as described in https://bugs.python.org/issue6721, but instead of logging
+                # doing the locking, we are.
+                return
+            raise e
         if atomic_dir.is_finalized():
             # We lost the double-checked locking race and our work was done for us by the race
             # winner so exit early.
--- a/pex/common.py
+++ b/pex/common.py
@@ -470,7 +470,7 @@ def atomic_directory(target_dir, exclusive, source=None):
             # staleness.
             fcntl.lockf(lock_fd, fcntl.LOCK_EX)  # A blocking write lock.
         except OSError as e:
-            if e.errno == 35:  # Resource deadlock avoided
+            if e.errno == errno.EDEADLK:  # Resource deadlock avoided
                 # Another thread/process is doing the work. They get the exclusivity.
                 # N.B.: This can happen if the process running this code forks, this is the same
                 # behavior as described in https://bugs.python.org/issue6721, but instead of logging


thejcannon:
```
pex/common.py:473:27: error: Module has no attribute "EDEADLK"; maybe
"EDEADLCK" or "EDEADLOCK"?  [attr-defined]
                if e.errno == errno.EDEADLK:  # Resource deadlock avoided
```

Really? https://docs.python.org/3.10/library/errno.html#errno.EDEADLK

Eric-Arellano:
I imagine `EDEADLCK` is a typo in the type stubs? Even 2.7 has `EDEADLK`. You can `# type: ignore` but would be good to open an issue with typeshed if one doesn't yet exist.

pex/common.py
```diff
@@ -464,10 +464,19 @@ def unlock():
...
-        # closed by the operating system when the owning process exits, this lock is immune to
-        # staleness.
-        fcntl.lockf(lock_fd, fcntl.LOCK_EX)  # A blocking write lock.
+        try:
+            # N.B.: Since lockf operates on an open file descriptor and these are guaranteed to be
+            # closed by the operating system when the owning process exits, this lock is immune to
+            # staleness.
+            fcntl.lockf(lock_fd, fcntl.LOCK_EX)  # A blocking write lock.
+        except OSError as e:
+            if e.errno == errno.EDEADLK:  # Resource deadlock avoided
```
Eric-Arellano, line 473: Do you understand the difference between this and `EDEADLOCK`? I don't.

thejcannon:
> I imagine `EDEADLCK` is a typo in the type stubs? Even 2.7 has `EDEADLK`. You can `# type: ignore` but would be good to open an issue with typeshed if one doesn't yet exist.

I filed https://github.com/python/typeshed/issues/7551 and then noticed https://github.com/python/typeshed/commit/af269057beea59b9b6e53626aa30b07f72db4e71 so it's just a matter of upgrade when it all is bubbled

pex/common.py
JelleZijlstra: I found that on Ubuntu they are the same (both valued at 35). On MacOS only EDEADLK exists and the value is 11. So probably EDEADLK is more portable.

(I got here through the typeshed issue.)

thejcannon: https://www.man7.org/linux/man-pages/man3/errno.3.html

>        EDEADLOCK
>              On most architectures, a synonym for EDEADLK.  On some
>              architectures (e.g., Linux MIPS, PowerPC, SPARC), it is a
>              separate error code "File locking deadlock error".

pex/common.py
```diff
@@ -464,10 +464,22 @@ def unlock():
...
+        except OSError as e:
+            deadlock_avoided_errorno = (
+                errno.EDEADLK  # type: ignore[attr-defined] # See https://github.com/python/typeshed/issues/7551
+            )
+            if e.errno == deadlock_avoided_errorno:
+                # Another thread/process is doing the work. They get the exclusivity.
+                # N.B.: This can happen if the process running this code forks, this is the same
+                # behavior as described in https://bugs.python.org/issue6721, but instead of logging
+                # doing the locking, we are.
+                return
```
thejcannon, line 481: Just saw `RuntimeError: generator didn't yield`

thejcannon: So we'll need to sleep in a loop.

thejcannon: or `yield None` as the docstring might suggest. I'll wait for @jsirois .

thejcannon:
Is there a good way to test this?

pex/common.py
thejcannon: Actually I think the return is fine? It's late on a Friday. I'll look later

thejcannon:
> Is there a good way to test this?

I suppose I could mock the lock and return the errno

jsirois (commented):
How confident are you the other thread will do the work? You seem to have ~proved this is so by experiment, but can you explain by some spec this is definitely so and the work won't be dropped? If I could not prove / reason that out, I'd want some sort of retry loop myself.

And, all that said. There are others ways to fix this depending on cause. For example, if it's known VCS requirements are involved, then an interesting tidbit there is that they are resolved in the same thread pool as URLs, but unlike URLs they fork a subprocess (`pip download`). That need not be the case. An earlier iteration of the VCS lock PR had URL fetching handled via an async map on the thread pool and `pip download` using `execute_parallel`.

thejcannon:
Now that you mention it, I think a retry loop makes the most sense. Not to mention necessary because we need to yield some value.

I'm not using VCS reqs, FWIW.

jsirois:
Ok, that should give pause then. The only relevant new thing in 2.1.74 was VCS handling. It would be great to understand things here.

thejcannon:
I'm not sure this issue is new as of 74, so much as new as of PEX lockfiles. I'll try and verify Monday.

thejcannon:
FYI I'm reproducing the issue on 73: https://github.com/thejcannon/pants/blob/lockspeedy/src/python/pants/backend/python/util_rules/pex_cli.py#L42

pex/common.py
```diff
@@ -464,10 +464,25 @@ def unlock():
             os.path.join(head, ".{}.atomic_directory.lck".format(tail or "here")),
             os.O_CREAT | os.O_WRONLY,
         )
-        # N.B.: Since lockf operates on an open file descriptor and these are guaranteed to be
-        # closed by the operating system when the owning process exits, this lock is immune to
-        # staleness.
-        fcntl.lockf(lock_fd, fcntl.LOCK_EX)  # A blocking write lock.
+        while True:
```
thejcannon, line 467: Hmm should we cap the retry/timeout here?

pex/common.py
```diff
@@ -464,10 +464,25 @@ def unlock():
...
+                # closed by the operating system when the owning process exits, this lock is immune to
+                # staleness.
+                fcntl.lockf(lock_fd, fcntl.LOCK_EX)  # A blocking write lock.
+            except OSError as e:
+                deadlock_avoided_errorno = (
+                    errno.EDEADLK  # type: ignore[attr-defined] # See https://github.com/python/typeshed/issues/7551
+                )
+                if e.errno == deadlock_avoided_errorno:
+                    # Another thread/process is doing the work. We must wait.
+                    # N.B.: This can happen if the process running this code forks, this is the same
```
jsirois, line 479: Did you mean https://bugs.python.org/issue6721? That appears to have ~nothing to do with this case. It's about mutex style locks and about hanging. We have here, on the other hand, an advisory file lock, and not hanging. To boot, sans a VCS requirement, 0 identified forks in the life-span of the thread pool.

So, something is obviously broken, but I contend we're still clueless what and how. If you agree, I'd prefer a non-misleading comment to the effect, "No clue, but we're deeming moving forward more important than grokking and this seems mostly harmless because ...?". I think I agree mostly harmless involves capping the loop at some maximum time or tries or both. I have no clue how to justify values for those limits though, since I at least, have no clue what's going on here still.

jsirois (approved):
Shipping since you're attempting to make forward progress, which I appreciate. I'm definitely not happy with the level of understanding though. I definitely feel like I have no clue what's actually going on still.

thejcannon:
@jsirois would you rather we just switch to processes and put this behind us?

thejcannon:
I'm able to reproduce this both in-Pants-without-PEX and without-Pants-and-without-PEX and the behavior we're seeing is consistent with my original theory.

Here's the reproducible steps:
    - Process runs, and opens N>1 threads ("workers")
    - Each worker forks to run some work in a subprocess. When done, continues forking to work on the next thing
    - Each (sub)work process opens N>1 threads
    - Each of those threads tries to grab one of the file locks, with contention between threads

Here it is without Pants or PEX involved (to kill it you'll need to get heavy with Ctrl+C, I didn't bother figuring out the cleanup code)
[deadlock.zip](https://github.com/pantsbuild/pex/files/8357953/deadlock.zip). At some point a thread in a worker subprocess has inherited the held lock, so the OS doesn't know how to handle that.

So, we either sleep our threads or switch to a processes, which won't share memory like the threads do.

pex/common.py
thejcannon: I'm actually thinking we shouldn't. We can't know how long this would take (maybe we're downloading `pytorch`, which is several GB) and the blocking write itself doesn't have a timeout, meaning we haven't really changed behavior all that much.

tests/test_common.py
```diff
@@ -123,6 +124,31 @@ def test_atomic_directory_empty_workdir_finalized():
             ), "When the target_dir exists no work_dir should be created."
 
 
+def test_atomic_directory_deadlock():
+    # type: () -> None
+
+    def mock_lockf(fd, cmd):  # type: (int, int) -> None
```
thejcannon, line 130: Instead of mocking we could try recreating the exact series of events, somewhat serially, which leads to this behavior but I'm not confident the code would be very helpful (likely involving some gnarly signaling between threads/processes) and I'm also not confident would reproduce this case on every OS.

pex/common.py
thejcannon: The GitHub issue should have enough context, and we can keep it open until you feel it contains enough info regarding this behavior to close it. I linked it instead of the existing NB comment.

thejcannon:
@jsirois I know you approved, but I want to double-check it's OK to merge as-is for forward progress.

thejcannon:
Current failure is due to https://github.com/psf/black/issues/2963

jsirois:
Thanks for pushing this forward @thejcannon. I've posted #1702 which doubles-back for a clearer-eyed take after fully running down the root cause here which was not the initial guess. The potential for an infinite loop in this PR seems troubling. To a much lesser extent, the contention / arbitrary delay of many, but not infinite, retries.

PR #1710, opened 2022-04-06
Document PGP release pre-reqs.

### LINKED ISSUE #1701: Release 2.1.75 tag not signed
Hi, I package pex for the AUR on Arch Linux. Arch packaging guidelines state that if upstream provides signed sources, I need to verify using the GPG key. Additionally, I cannot remove GPG verification if upstream stops signing sources.

Release tag 2.1.75 isn't signed; please sign it, preferably without the GitHub web-signing key. I'd greatly appreciate it!

jsirois:
Closes #1701

### DIFF
--- a/RELEASE.rst
+++ b/RELEASE.rst
@@ -4,6 +4,29 @@ Pex Release Process
 
 .. contents:: Table of Contents
 
+Pre-requisites
+==============
+
+PGP
+---
+
+All release tags are signed (using ``git tag --sign``) so that users of Pex can verify maintainers
+have performed & trust a release. This requires releasers having a PGP key configured with git and
+published to keyservers. An additional nicety is to configure you PGP key with GitHub for those who
+like to check provenance via a web UI.
+
+Some documentation to help you get things set up if you don't have all of these pre-requisites:
+
++ Creating a key and configuring your key with Git and GitHub is all described `here <https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification>`_
++ Publishing your key to a keyserver is described `here <https://www.gnupg.org/gph/en/manual/x457.html>`_
+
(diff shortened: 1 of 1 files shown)


Eric-Arellano (approved):
Thank you! I only have you done the MIT one, I will do the others now

RELEASE.rst
```diff
@@ -4,6 +4,29 @@ Pex Release Process
...
+published to keyservers. An additional nicety is to configure you PGP key with GitHub for those who
+like to check provenance via a web UI.
+
+Some documentation to help you get things set up if you don't have all of these pre-requisites:
+
++ Creating a key and configuring your key with Git and GitHub is all described `here <https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification>`_
++ Publishing your key to a keyserver is described `here <https://www.gnupg.org/gph/en/manual/x457.html>`_
+
+Some keyservers you probably want to publish your key to explicitly above and beyond your PGP setup's
+default configured keyserver include:
```
kaos, line 24: From the `gpg` man page I read:

> Most keyservers synchronize with each other, so there is generally no need to send keys to more than one server.

And true enough, I've sent my key to https://keys.openpgp.org/search?q=andreas.stenius%40svenskaspel.se but is also listed on https://pgp.mit.edu/pks/lookup?search=andreas.stenius%40svenskaspel.se&op=index (didn't check further)

jsirois: But some do not. Eric's went to mit but had not (yet?) synced to ubuntu when I checked earlier today; so I think I'll leave this here if you agree. I takes all of 3 seconds to run 3 send-keys commands in case the effort was wasted.

kaos: Indeed, no harm.

Issue #1696, opened 2022-03-28
Support local requirements with lockfiles


Eric-Arellano:
From John re how to implement:

> I think the same trick used for VCS basically works for local projects, but there is some juggling to do (The original standard Pex resolver does the juggling but both the --lock resolver and --pex-repository resolver refuse to do this.

Two Pants users have weighed in that this would be very useful for their repos, so Pants would love to see this feature.

jsirois:
Sounds good. IIUC @Eric-Arellano to be able to use this feature Pants will need to get out of the way re requirements parsing, or support local absolute paths in requirements in some ergonomic way. For the former, Pex supports `pex .` and for the latter `pex "pex @ file:///home/jsirois/dev/pantsbuild/pex"`.

jsirois:
Ok:
```
$ python -mpex.cli lock create . --indent 2
{
  "allow_builds": true,
  "allow_prereleases": false,
  "allow_wheels": true,
  "build_isolation": true,
  "constraints": [],
  "locked_resolves": [
    {
      "locked_requirements": [
        {
          "artifacts": [
            {
              "algorithm": "sha256",
              "hash": "0308b8310a8ac7ecbf4be61e3ac3ad6deac818a522ec76aff551b3be9c634910",
              "url": "file:///home/jsirois/dev/pantsbuild/jsirois-pex"
            }
          ],
          "project_name": "pex",
          "requires_dists": [
            "subprocess32>=3.2.7; extra == \"subprocess\" and python_version < \"3\""
          ],
          "requires_python": "!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*,<3.11,>=2.7",
          "version": "2.1.77"
        }
      ],
      "platform_tag": [
        "cp310",
        "cp310",
        "manylinux_2_35_x86_64"
      ]
    }
  ],
  "pex_version": "2.1.77",
  "prefer_older_binary": false,
  "requirements": [
    "."
  ],
  "requires_python": [],
  "resolver_version": "pip-legacy-resolver",
  "style": "strict",
  "transitive": true,
  "use_pep517": null
}
```

This runs into the same issues as #1413 though. Time to think about that.

Eric-Arellano:
Exciting! If it is amenable to land that, I think it could be useful as a precursor. While no one will be able to use it in production lock files without 1413, it can be helpful when people are debugging locally

jsirois:
[...]

PR #1716, opened 2022-04-09
Support canonicalizing absolute paths in locks.

### LINKED ISSUE #1413: Canonicalize file:// URLs in locks
In order for a lock file to be shareable across different machines when using a `--find-links` repo that points to either a network mount or else some other machine specific mount path of a known fixed set of artifacts, absolute file urls in the lock will need to be relativized to some base path that can be defined later and thus permit a canonicalized form for storage.

For example, if the resolve producing a lock was run with `-f dir1 -f dir1/di2 -f /mnt/repo`, then the following artifact links in a lock:
1. file:///home/jane/dev/project/dir1/a-1.0.tgz
2. file:///home/jane/dev/project/dir1/dir2/b-1.0.tgz
3. file:///mnt/repo/c-1.0.tgz

Could be canonicalized to:
1. file://$FL1/dir1/a-1.0.tgz
2. file://$FL2/dir1/dir2/b-1.0.tgz
3. file://$FL3/c-1.0.tgz

Then, to resolve using the generated lock, the corresponding environment variables will need to be defined. Since Pex and Pip support env var interpolation in requirements files, we can make this work. These env var values can be defined as makes sense on the machine / $PWD of the Pex resolve process is later run from. Some scheme to encode `--find-links` repos in lock files will be needed such that environment variables can be matched up to repos later. This could be a mapping of env var names to descriptions like:
```json
{
  "find_links": {
    "FL1": "The project's dir1 findl-links symlink farm prepared by our `repo-sync` infra.",
    "FL2": "The project's dir1/dir2 findl-links symlink farm prepared by our `repo-sync` infra.",
    "FL3": "The company NFS shared find-links repo mounted by our `dev-machine-init` infra."
  }
}
```
Or it could just be a facility to associate each `--find-links` used when generating a lock to be associated with a custom, descriptive env var name instead of something like the sequential `$FLN` above.

jsirois:
Although you can still create locks that contain absolute paths from the
local system, Pex now supports canonicalizing such paths for portable
locks with the `--path-mapping` option available when creating and
consuming locks.

The classic example where this is useful is a shared find-links repo
accessed as a network mount somewhere in the user's home directory. A 
portable lock for PEXes that get at least some of their distributions
resolved from such a find-links repo can be created like so:
```
pex3 lock create \
  --find-links $HOME/corp/shares/find-links \
  --path-mapping "\
CORP_FL|$HOME/corp/shares/find-links|\
The local path of the corporate find-links repo network mount.\
"
...
```

The lock file will have all paths beginning with
`$HOME/corp/shares/find-links` substituted with the `${CORP_FL}`
placeholder and the lock will centrally encode all such mappings along
with the description given.

When some other user later uses the `--lock` to create a PEX, they will
be prompted to fill in the `--path-mapping`s if they do not supply them:
```
$ python -mpex --lock lock.json 
,,,
pex.resolve.lockfile.ParseError: The lockfile given in lock.json requires definition of 1 '.path_mappings' entry: CORP_FL
Given no path mappings.
Which left the following path mappings unspecified:
CORP_FL: The local path of the corporate find-links repo network mount.
```

The problem is then corrected by giving the needed path mapping:
```
python -mpex --lock lock.json --path-mapping "CORP_FL|/tmp/find-links"
```
Here, the user mounted the corporate find-links repo in a completely
different place than the lock creator had it mounted at.

Fixes #1413

### DIFF
--- a/pex/resolve/lockfile/json_codec.py
+++ b/pex/resolve/lockfile/json_codec.py
@@ -12,6 +12,7 @@ from pex.pep_503 import ProjectName
 from pex.resolve.locked_resolve import Artifact, LockedRequirement, LockedResolve, LockStyle
 from pex.resolve.lockfile import ParseError
 from pex.resolve.lockfile.lockfile import Lockfile
+from pex.resolve.path_mappings import PathMappings
 from pex.resolve.resolved_requirement import Fingerprint, Pin
 from pex.resolve.resolver_configuration import ResolverVersion
 from pex.sorted_tuple import SortedTuple
@@ -43,6 +44,7 @@ def _load_json(
 def loads(
     lockfile_contents,  # type: Text
     source="<string>",  # type: str
+    path_mappings=PathMappings(),  # type: PathMappings
 ):
     # type: (...) -> Lockfile
 
@@ -111,7 +113,7 @@ def loads(
     ):
         # type: (...) -> Requirement
         try:
-            return Requirement.parse(raw_requirement)
+            return Requirement.parse(path_mappings.maybe_reify(raw_requirement))
         except RequirementParseError as e:
             raise ParseError(
                 "The requirement string at '{path}' is invalid: {err}".format(path=path, err=e)
@@ -128,6 +130,48 @@ def loads(
             raise ParseError(
                 "The version specifier at '{path}' is invalid: {err}".format(path=path, err=e)
             )
+
+    required_path_mappings = get("path_mappings", dict, optional=True) or {}
+    given_mappings = set(mapping.name for mapping in path_mappings.mappings)
+    unspecified_paths = set(required_path_mappings) - given_mappings
+    if unspecified_paths:
+        raise ParseError(
+            "The lockfile given in {source} requires definition of {required_count} "
+            "'.path_mappings' {entries}: {required_paths}\n"
+            "Given {given_mappings_verbiage}\n"
+            "{maybe_path_mappings}"
+            "Which left the following path mappings unspecified:\n"
+            "{unspecified_paths}".format(
+                source=source,
+                required_count=len(required_path_mappings),
+                entries="entries" if len(required_path_mappings) > 1 else "entry",
+                required_paths=", ".join(sorted(required_path_mappings)),
+                given_mappings_verbiage="the following path mappings:"
+                if path_mappings.mappings
+                else "no path mappings.",
+                maybe_path_mappings="{path_mappings}\n".format(
+                    path_mappings="\n".join(
+                        sorted(
+                            "{root}: {path}".format(root=mapping.name, path=mapping.path)
+                            for mapping in path_mappings.mappings
+                        )
+                    )
+                )
+                if path_mappings.mappings
+                else "",
+                unspecified_paths="\n".join(
+                    sorted(
+                        (
+                            "{path}: {description}".format(path=path, description=description)
+                            if description
+                            else path
+                        )
+                        for path, description in required_path_mappings.items()
+                        if path in unspecified_paths
+                    )
+                ),
+            )
+        )
 
     requirements = [
         parse_requirement(req, path=".requirements[{index}]".format(index=index))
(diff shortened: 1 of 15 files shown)


jsirois:
The existing `--find-links` option could be enhanced to support the `--path-mapping` `NAME|PATH|DESCRIPTION` format if supplied to reduce typing / redundancy, but there will be non `--find-links` uses of `--path-mapping` forthcoming to support local project requirements in #1696; so I led with the needed functionality only. This can be made more convenient later. Certainly Pants will be writing these command lines in code; so there will be no undue burden on end users over there with the current non-DRY option setup in the `--find-links` case.

pex/resolve/lockfile/json_codec.py
```diff
@@ -129,6 +131,48 @@ def parse_version_specifier(
                 "The version specifier at '{path}' is invalid: {err}".format(path=path, err=e)
             )
 
+    required_path_mappings = get("path_mappings", dict, optional=True) or {}
+    given_mappings = set(mapping.name for mapping in path_mappings.mappings)
+    unspecified_paths = set(required_path_mappings) - given_mappings
+    if unspecified_paths:
+        raise ParseError(
+            "The lockfile given in {source} requires definition of {required_count} "
```
Eric-Arellano, line 139: It may be helpful to mention how you set this value. My initial read of the error message made me think that you must set an environment variable. Whereas it sounds like you instead set the command line flag

pex/resolve/lockfile/json_codec.py
```diff
@@ -297,3 +352,8 @@ def as_json_data(lockfile):
             for locked_resolve in lockfile.locked_resolves
         ],
     }
+    if path_mappings.mappings:
+        data["path_mappings"] = {
+            path_mapping.name: path_mapping.description for path_mapping in path_mappings.mappings
+        }
```
Eric-Arellano, line 358: Generally you are always defining keys, and using empty values when relevant. Why conditionally register this key?

pex/resolve/resolver_options.py
```diff
@@ -209,6 +211,33 @@ def register(
...
+        action="append",
+        default=[],
+        type=str,
+        help=(
+            "A mapping of the form `NAME|PATH|DESCRIPTION` of a logical name to a concrete local "
+            "absolute path with an optional description. Can be specified multiple times. The "
+            "mapping must include the pipe (`|`) separated name and absolute path components, but "
+            "the trailing pipe-separated description is optional. The mapping is used when "
+            "creating, and later reading, lock files to ensure the lock file created on one "
+            "machine can be used another with a potentially different realization of various paths "
```
Eric-Arellano, line 232: Did you mean this?

```suggestion
            "machine can be used on another with a potentially different realization of various paths "
```

Eric-Arellano (approved):
Great solution!

pex/resolve/lockfile/json_codec.py
jsirois: Urg. yeah - I hate non-local coupling like this. the JSON codec should have 0 knowledge of flags. Let me ruminate on being dirty vs maybe a structured exception type the CLI can trap and translate.

pex/resolve/lockfile/json_codec.py
jsirois: Its a bit different in that a very large percentage of locks will never use path mappings, but I guess that may not be any reason at all - I'll take a look.

pex/resolve/resolver_options.py
jsirois: Yup. Fixed.

jsirois:
> Great solution!

Well - I think its the only solution? It was also solved in August when I filed the issue.

pex/resolve/lockfile/json_codec.py
jsirois: Yeah, it was for some sort of misguided symmetry with the read side I think. On the read side I have to deal with the key not being there to be compatible with old locks. Symmetry is just beauty though and that's in the eye of the beholder and pretty much garbage for basing a decision on. Fixed!

pex/resolve/lockfile/json_codec.py
jsirois: OK, went with a structured exception type and a message mentioning the flag only in the file that defines the flag.

PR #1727, opened 2022-04-19
Fix Pex resolve checking.

jsirois:
Previously names were not always normalized per PEP-503 in version
compatibility checks.

Fixes #1726

PR #1732, opened 2022-04-20
Fix prerelease handling when checking resolves.

### LINKED ISSUE #1730: Pex resolve checking does not allow resolved pre-releases when `--no-pre`.
On the surface, this is exactly the behavior you'd want.

The issue though, is that Pip is smarter and when you hand it `--no-pre` (the default), it will still resolve prereleases either of these conditions is true:
+ Any of the requirements you supplied it are for a pre-release version explicitly; e.g.: `pantsbuild.pants==2.12.0.dev3`.
+ Any requirement, transitive or otherwise, is only satisfiable by a prerelease version.

As such, Pex currently has an expectation mis-match when it checks the resolve results since it does not implement these same smarts.

jsirois:
The fix added to resolve checking in #1727 led to a regression dealing
with resolved prereleases. Return to old behavior and add test coverage
for this behavior.

Fixes #1730

### DIFF
--- a/tests/integration/test_issue_1730.py
+++ b/tests/integration/test_issue_1730.py
@@ -1,0 +1,49 @@
+# Copyright 2022 Pants project contributors (see CONTRIBUTORS.md).
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+from __future__ import absolute_import
+
+import os
+import sys
+
+import pytest
+
+from pex.testing import run_pex_command
+from pex.typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from typing import Any
+
+
+@pytest.mark.skipif(
+    sys.version_info[:2] < (3, 7) or sys.version_info[:2] >= (3, 10),
+    reason="Pants 2.12.0.dev3 requires Python >=3.7,<3.10",
+)
+def test_check_install_issue_1730(
+    tmpdir,  # type: Any
+):
+    # type: (...) -> None
+
+    pex_root = os.path.join(str(tmpdir), "pex_root")
+    pex_args = [
+        "--pex-root",
+        pex_root,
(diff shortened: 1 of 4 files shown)


tests/integration/test_issue_1730.py
```diff
@@ -0,0 +1,49 @@
...
+
+from pex.testing import run_pex_command
+from pex.typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from typing import Any
+
+
+@pytest.mark.skipif(
+    sys.version_info[:2] < (3, 7) or sys.version_info[:2] >= (3, 10),
```
jsirois, line 19: This test does not run in CI, which just covers 2.7 and 3.10 in ITs. It does work though on `3.{7,8,9}` as tested locally and it is slow (~12s for me) since Pants is big; so it's for the better.

tests/test_resolver.py
```diff
@@ -626,3 +627,26 @@ def test_duplicate_requirements_issues_1550():
     distribution = installed_distribution.distribution
     assert "PyJWT" == distribution.project_name
     assert "1.7.1" == distribution.version
+
+
+def test_check_resolve_prerelease_transitive_dependencies_issue_1730(tmpdir):
+    # type: (Any) -> None
+
+    indirect_wheel = build_wheel(name="indirect", version="2.12.0.dev3")
```
jsirois, line 635: Because of the note above about the IT not actually being exercised in CI, I added a simulation here. This does fail prior to the fix with:
```
E           Unsatisfiable: Failed to resolve compatible distributions:
E           1: direct==2.12.0.dev3 requires indirect==2.12.0.dev3 but indirect 2.12.0.dev3 was resolved
```

Eric-Arellano (approved):

Issue #1735, opened 2022-04-25
Subcommand-specific -v flag shadows the global one


benjyw:
We register `-v` on subcommands (such as `lock create`), but we also register it globally. However argparse returns the global args and subcommand args in a single Namespace object, and in that object the value from the subcommand wins, and so the global value is hidden. 

In other words, in `pex -vvv lock create -vvvvv ....` the subcommand value (verbosity=5) is the one available in the Namespace object returned by `parse_args()`, and the global value (verbosity=3) is not available anywhere. This is true even if no explicit arg is provided for the subcommand. The subcommand arg's default (verbosity=0) overwrites the global value.

This is confusing to users, because `pex -vvv lock create` doesn't fail, it just silently ignores that value. Pants, in fact, made this mistake.

We work around this in Pants here: https://github.com/pantsbuild/pants/pull/15243 but should fix at the Pex level as well.

One option is to not register the verbosity flag globally. Another, possibly simpler, is to only register it globally and not register it on the subcommands. In either case, there is the following consequence: the flag cannot appear in that position. I.e., either `pex -vvv lock create ...` will fail or `pex lock create -vvv ...` will fail, respectively. 

A third, probably best, option is to register the global verbosity under a different dest. Then both will be available in the Namespace object. We then just have to decide what the effective verbosity should be if both are specified.

benjyw:
As you may have guessed, this affects all the global options. Basically registering them in global position, as we do today, has no practical effect and the args are silently ignored.

benjyw:
But since a subcommand is now required, possibly we can just not register global options in global position at all.

Or we can get clever and register them with alternate dests, and merge the values. This would allow them to be specified both before and after the subcommand. But I'm not sure that has much practical value?

benjyw:
See https://stackoverflow.com/questions/37933480/how-can-i-define-global-options-with-sub-parsers-in-python-argparse for more context

benjyw:
Addressed in #1736

benjyw:
That solution is simply to not register the global args in global position. It's simple and effective, but also does not allow users to specify the global flags before the subcommand. I think this is probably fine, but if we really want that we will have to contemplate more complicated solutions.

jsirois:
Hrm. Yeah, I was confused by subcommands in argparse and I think I still am. Trying to page this in, but the path you're pushing seems like the right way to fix things for now at the very least.

PR #1736, opened 2022-04-26
Don't register global args in global position.

benjyw:
They have no effect and are silently overwritten by
the same args (or their defaults) registered on each
subcommand.

Previously, `pex -v subcommand` would silently ignore the
arg in the global position because -v is also registered on
the subcommand (and similarly for all global args).

Now at least you get an error. With this change all args,
including global ones, must appear on the cmd-line after
the subcommand.

### DIFF
--- a/pex/commands/command.py
+++ b/pex/commands/command.py
@@ -284,7 +284,10 @@ def global_environment(options):
     if not hasattr(options, "rc_file"):
         # We don't register the global args on the root command (but do on every subcommand).
         # So if the user runs just `pex` with no subcommand we must not attempt to use those
+        # global args, including rc_file, which we check for here as a representative of the
         # global args.
+        # Note that we can't use command_type here because the legacy command line parser in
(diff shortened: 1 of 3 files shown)


benjyw:
See #1735

pex/commands/command.py
```diff
@@ -281,6 +281,13 @@ def global_environment(options):
     :yields: The configured global environment.
     :raises: :class:`GlobalConfigurationError` if invalid global option values were specified.
     """
+    if not hasattr(options, "rc_file"):
```
jsirois, line 284: It would probably be a bit nicer to use `command_type` as the trigger, but this definitely works.

jsirois (approved):

pex/commands/command.py
benjyw: You're right! Changed.

benjyw: Nope, turns out we can't use command_type because this code path is also used by the legacy command line parsing in pex/bin/pex.py, and that doesn't set command_type.

PR #1738, opened 2022-04-28
Fix universal lock creation for ICs.

### LINKED ISSUE #1734: The interpreter we run Pex with affects the resolve
```
python3.6 -m pex.cli lock create --style=universal --interpreter-constraint 'CPython<3.9,>=3.8' 'aioconsole' 
```

Fails with `ERROR: Package 'aioconsole' requires a different Python: 3.6.6 not in '>=3.7'` . 

Running with python3.7 succeeds.

Removing `--style=universal` succeeds.

Note: `aioconsole` only publishes an sdist.

jsirois:
Previously the current interpreter was used to perform the lock resolve
(which is OK) and the resolve post-processing (which was not OK). Now
we force finding a local interpreter that meets ICs up-front and use
that for both the resolve and post-processing.

Fixes #1734


jsirois:
Sorry about the too early review push. This new (correct) constraint on a compatible interpreter having to be present for universal locks to be created catches old tests out being sloppy.

Eric-Arellano (approved):

Issue #1745, opened 2022-05-01
PEX interpreters should support all underlying Python interpreter options.


jsirois:
Right now, when run as an interpreter, a PEX only handles `-c`, `-m`, `-` and a file positional argument (the "interface options": https://docs.python.org/3/using/cmdline.html#interface-options). It could handle the full array of Python interpreter arguments (like `-s`) and re-exec if any of these other Python interpreter arguments are present to ensure those options are passed to the Python interpreter instead of to the PEX; So this command line:
```
./interpreter.pex -u myscript.py -q foo bar
```
Would result in a re-exec of:
```
/usr/bin/python3.10 -u ~/.pex/unzipped_pexes/416f26c910ada914a16dd388ca842e151e685577/ -q foo bar
```

jsirois:
@BrandonTheBuilder has implemented this for the default ZIPAPP execution mode in #1746. All that remains is supporting `--venv` execution mode.

PR #1746, opened 2022-05-02
Re-exec with interpreter-options when set. (#1745)

BrandonTheBuilder:
Parse options args for interpreter options. If they are present re-exec passing the proper options to the interpreter. If they are not present run the code directly.

### DIFF
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -594,6 +594,20 @@ class PEX(object):  # noqa: T000
     def execute_interpreter(self):
         # type: () -> Any
         args = sys.argv[1:]
+        options = []
+        for index, arg in enumerate(args):
+            # Check if the arg is an expected startup arg
+            if arg.startswith("-") and arg not in {"-", "-c", "-m"}:
+                options.append(arg)
+                continue
+            else:
+                args = args[index:]
+                break
+
+        # The pex was called with interpreter options
+        if options:
+            return self.execute_with_options(options, args)
+
         if args:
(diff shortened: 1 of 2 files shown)


pex/pex.py
```diff
@@ -626,6 +640,27 @@ def execute_interpreter(self):
             code.interact()
             return None
 
+    def execute_with_options(self, options, args):
+        """
+        Restart the process passing the given options to the python interpreter
+        """
+        # Find the pex bootstrap location
+        # We expect this environment variable to be set by __main__.py
```
jsirois, line 648: This comment is a bit stale and needs adjustment.

pex/pex.py
```diff
@@ -626,6 +640,27 @@ def execute_interpreter(self):
...
+            cmdline = [interpreter] + options + [run_pex] + args
+            TRACER.log(
+                "Re-executing with interpreter options: "
+                "cmdline={cmdline!r}, ".format(
+                    cmdline=" ".join(cmdline),
+                )
+            )
+            os.execv(interpreter, cmdline)
+        else:
+            return "Unable to resolve pex path"
```
jsirois, line 662: This should be impossible; so its not very important except for the code reader, but it would be spot on if it said "Unable to resolve installed pex path".

jsirois (approved):
Thanks Loren. I really appreciate you stopping by at PyCon and chipping in. The comments are not worth addressing unless there is a natural opportunity like CI failures. I'll merge if all is green.

PR #1748, opened 2022-05-04
Complete support for PEX Python interpreter mode.

### LINKED ISSUE #1745: PEX interpreters should support all underlying Python interpreter options.

jsirois:
Apply @BrandonTheBuilder's work in #1746 to the `--venv` case and flesh
out complete support for Python interpreter emulation by adding `.` to
the `sys.path` when running as an interpreter and handling interpreter
execution of directories.

Fixes #1745

### DIFF
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -32,7 +32,20 @@ from pex.util import iter_pth_paths, named_temporary_file
 from pex.variables import ENV, Variables
 
 if TYPE_CHECKING:
-    from typing import Any, Dict, Iterable, Iterator, List, Mapping, Optional, Set, Tuple, TypeVar
+    from typing import (
+        Any,
+        Dict,
+        Iterable,
+        Iterator,
+        List,
+        Mapping,
+        NoReturn,
+        Optional,
+        Set,
+        Tuple,
+        TypeVar,
+        Union,
+    )
 
     _K = TypeVar("_K")
     _V = TypeVar("_V")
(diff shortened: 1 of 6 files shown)


pex/pex.py
```diff
@@ -640,26 +657,29 @@ def execute_interpreter(self):
...
-            interpreter = PythonInterpreter.get().binary
-            cmdline = [interpreter] + options + [run_pex] + args
-            TRACER.log(
-                "Re-executing with interpreter options: "
-                "cmdline={cmdline!r}, ".format(
-                    cmdline=" ".join(cmdline),
-                )
+        if not main or not main.__file__:
+            # N.B.: This should never happen.
+            return "Unable to resolve PEX __main__ module file: {}".format(main)
```
Eric-Arellano, line 673: How about raising `AssertionError`?

Eric-Arellano (approved):

pex/pex.py
jsirois: As the tests show, those can be turned off. This gives a clean non-zero exit with a non-backtrace mess message and is in line with all other exits here. What are you seeing as the advantage of an assertion over this? Collapsing a comment?

Eric-Arellano: That usually when you say "this shouldn't happen", using an assert is the way to ensure that doesn't happen. Sg though

jsirois: Well, an assert doesn't "ensure it doesn't happen" as the -O tests show, I think you must mean it expresses intent better, or at least more concisely. That I don't disagree with, but will leave as Loren established it.

PR #1754, opened 2022-05-13
Support credentials in `URLFetcher`.

### LINKED ISSUE #1753: pex3 lock create does not preserve URL credentials
As a result a lockfile can be generated that is not useable later.

For example, in one session:
```
$ PEX_TOOLS=1 pex cowsay --include-tools -- repository extract --serve
cowsay 4.0: Repacking wheel as /tmp/tmpyj_m1pdx/cowsay-4.0-py2.py3-none-any.whl...OK
Serving find-links repo of /tmp/tmpwmbfywvo via /tmp/tmpyj_m1pdx at http://localhost:40509
...
```

Then in another:
```
$ pex3 lock create --no-pypi --find-links http://joe:bob@localhost:40509 cowsay --indent 2
{
  "allow_builds": true,
  "allow_prereleases": false,
  "allow_wheels": true,
  "build_isolation": true,
  "constraints": [],
  "locked_resolves": [
    {
      "locked_requirements": [
        {
          "artifacts": [
            {
              "algorithm": "sha256",
              "hash": "b1b5ae7c2594ae9c1fe182a14f8a6961493b335746f921121c2699f8e025f26f",
              "url": "http://localhost:40509/cowsay-4.0-py2.py3-none-any.whl"
            }
          ],
          "project_name": "cowsay",
          "requires_dists": [],
          "requires_python": null,
          "version": "4"
        }
      ],
      "platform_tag": [
        "cp310",
        "cp310",
        "manylinux_2_35_x86_64"
      ]
    }
  ],
  "path_mappings": {},
  "pex_version": "2.1.85",
  "prefer_older_binary": false,
  "requirements": [
    "cowsay"
  ],
  "requires_python": [],
  "style": "strict",
  "transitive": true,
  "use_pep517": null
}
```
The lock URL is missing the `joe:bob@` plain text authentication; so if `http://localhost:40509` is actually authenticated, the lock will be unuseable.

The Pip log reveals:
```
$ PYTHONPATH=pex/vendor/_vendored/pip python -mpip download --no-index --find-links http://joe:bob@localhost:40509 cowsay --log pip.log
Looking in links: http://joe:****@localhost:40509
Collecting cowsay
  Downloading http://localhost:40509/cowsay-4.0-py2.py3-none-any.whl (42 kB)
     |████████████████████████████████| 42 kB 94.2 MB/s 
Saved ./cowsay-4.0-py2.py3-none-any.whl
Successfully downloaded cowsay
$ cat pip.log 
2022-05-11T10:53:33,726 Ignoring indexes: https://pypi.org/simple
2022-05-11T10:53:33,727 Created temporary directory: /tmp/pip-req-tracker-lcayp393
2022-05-11T10:53:33,727 Initialized build tracking at /tmp/pip-req-tracker-lcayp393
2022-05-11T10:53:33,727 Created build tracker: /tmp/pip-req-tracker-lcayp393
2022-05-11T10:53:33,727 Entered build tracker: /tmp/pip-req-tracker-lcayp393
2022-05-11T10:53:33,727 Created temporary directory: /tmp/pip-download-hafzr54p
2022-05-11T10:53:33,738 Looking in links: http://joe:****@localhost:40509
2022-05-11T10:53:33,738 1 location(s) to search for versions of cowsay:
2022-05-11T10:53:33,738 * http://joe:****@localhost:40509
2022-05-11T10:53:33,738 Skipping link: not a file: http://joe:****@localhost:40509 (from -f)
2022-05-11T10:53:33,738 Fetching project page and analyzing links: http://joe:****@localhost:40509
2022-05-11T10:53:33,738 Getting page http://joe:****@localhost:40509
2022-05-11T10:53:33,738 Found credentials in url for localhost:40509
2022-05-11T10:53:33,752   Found link http://localhost:40509/cowsay-4.0-py2.py3-none-any.whl (from http://localhost:40509/), version: 4.0
2022-05-11T10:53:33,752 Given no hashes to check 1 links for project 'cowsay': discarding no candidates
2022-05-11T10:53:33,753 Collecting cowsay
2022-05-11T10:53:33,753   Created temporary directory: /tmp/pip-unpack-ctwtz7ek
2022-05-11T10:53:33,755   Downloading http://localhost:40509/cowsay-4.0-py2.py3-none-any.whl (42 kB)
2022-05-11T10:53:33,755   Added cowsay from http://localhost:40509/cowsay-4.0-py2.py3-none-any.whl to build tracker '/tmp/pip-req-tracker-lcayp393'
2022-05-11T10:53:33,755   Removed cowsay from http://localhost:40509/cowsay-4.0-py2.py3-none-any.whl from build tracker '/tmp/pip-req-tracker-lcayp393'
2022-05-11T10:53:33,757 Created temporary directory: /tmp/pip-unpack-xfjn71js
2022-05-11T10:53:33,757 Saved ./cowsay-4.0-py2.py3-none-any.whl
2022-05-11T10:53:33,757 Successfully downloaded cowsay
2022-05-11T10:53:33,757 Removed build tracker: '/tmp/pip-req-tracker-lcayp393'
```

So there is probaly enough data there that is scrape-able to correlate `http://joe:bob@localhost:40509` with `http://localhost:40509` and either re-insert the auth in all URLs written down in the lock or else store it centrally much like the path mappings on a per-url-prefix basis.

jsirois:
This adds support for passing credentials when fetching URLs with
`URLFetcher` and wires up the lock resolve process to both grab any
credentials present in `~/.netrc` and any present as credentials
embdedded in custom index and find links URLs Pex is configured with.

Fixes #1753


jsirois:
I forked off #1757 to fix the unrelated vendor shard CI break.

pex/auth.py
```diff
@@ -0,0 +1,93 @@
...
+            for machine, (login, _, password) in netrc(netrc_path).hosts.items():
+                if password is None:
+                    continue
+
+                if machine == "default":
+                    # The 'default' entry is special and means just that; so we omit any qualifying
+                    # URI.
+                    yield PasswordEntry(username=login, password=password)
+
+                if urlparse.urlparse(machine).scheme:
```
benjyw, line 76: From my readings (e.g., https://www.labkey.org/Documentation/wiki-page.view?name=netrc, not sure how authoritative it is) it seems that in netrc a machine name is literally just a hostname, with no scheme or port. 

It also looks like the golang folks went through this exercise a couple of years ago (https://github.com/golang/go/issues/37130).  

So parsing a machine name as a URL seems like overkill? Although technically no harm. Just wondering if you'd encountered otherwise.  Perhaps worth documenting the thinking here (also see below for more perceived complications re ports).

pex/fetcher.py
```diff
@@ -50,11 +56,30 @@ def __init__(
...
+        )
         self._handlers = tuple(handlers)
 
     @contextmanager
     def get_body_stream(self, url):
         # type: (Text) -> Iterator[BinaryIO]
+
+        handlers = list(self._handlers)
+        if self._password_database.entries:
+            password_manager = HTTPPasswordMgrWithDefaultRealm()
```
benjyw, line 71: .netrc entries are keyed by machine name (without port) but are expected to be used for all auth to that machine, on any port. Or at least this is how curl behaves. 

So at first glance this seems broken, because the uri key we pass to add_password() doesn't include a port. We could inject the port in `uri_or_default()` since we know it there. But, from my reading of the urllib2 source code it looks like we don't need to, because that code strips any port information on both sides before comparing entries.

So I guess this comment is to confirm that this is also your understanding, and to say that perhaps this complication should be documented.

pex/auth.py
```diff
@@ -0,0 +1,93 @@
...
+        def iter_entries():
+            # type: () -> Iterator[PasswordEntry]
+            for machine, (login, _, password) in netrc(netrc_path).hosts.items():
+                if password is None:
+                    continue
+
+                if machine == "default":
+                    # The 'default' entry is special and means just that; so we omit any qualifying
+                    # URI.
+                    yield PasswordEntry(username=login, password=password)
```
benjyw, line 74: I get that this is part of how netrc works and so we should support it, but it seems bonkers to me to allow sending a password to every host. But I guess if that's how someone has configured their netrc, so be it.

pex/fetcher.py
benjyw: The urllib2 code also appears to ignore scheme information and will send a password associated with an https URL to an http URL on the same host. But 🤷

benjyw (approved):
LGTM with some questions to make sure ports are handled correctly, and possibly to ask that we document this quirk.

pex/auth.py
jsirois: Yeah, I'm not fighting how this ancient thing works - and if I don't work how it works I only break users who have made a choice.

pex/auth.py
jsirois: I did not encounter otherwise, I was just playing it safe. I'll document.

pex/fetcher.py
jsirois: > So at first glance this seems broken, because the uri key we pass to add_password() doesn't include a port. We could inject the port in uri_or_default() since we know it there.

It does include a port if the netrc entry happens to have a port. If it doesn't have a port (a traditional netrc entry), urllib actually creates a second entry that does include the port (80 or 443 as appropriate).

```
$ python2.7
Python 2.7.18 (default, Oct 10 2021, 22:29:32) 
[GCC 11.1.0] on linux2
>>> from urllib2 import HTTPPasswordMgrWithDefaultRealm
>>> manager = HTTPPasswordMgrWithDefaultRealm()
>>> manager.add_password(realm=None, uri="http://foo:9999", user="foo", passwd="bar")
>>> manager.add_password(realm=None, uri="http://bar", user="bar", passwd="baz")
>>> manager.passwd
{None: {(('foo:9999', '/'),): ('foo', 'bar'), (('bar', '/'),): ('bar', 'baz'), (('bar:80', '/'),): ('bar', 'baz')}}
```

IOW it's the opposite to what you suggest, urllib compares against the host name (possibly containing port) - it does not strip port.

I'll add some comment.

PR #1770, opened 2022-05-18
Select PEX runtime interpreter robustly.

### LINKED ISSUE #1020: Decouple PEX runtime interpreter selection from buildtime interpreter selection.

jsirois:
Previously two proxies for interpreter applicability were used:
1. The shebang selected interpreter or the explicit interpreter used
   to invoke the PEX.
2. Any embedded interpreter constraints.

This could lead to selecting an interpreter that was not actually able
to resolve all required distributions from within the PEX, which is the
only real criteria, and a failure to boot.

Fix the runtime interpreter resolution process to test an interpreter
can resolve the PEX before using it or re-execing to it.

In the use case, the resolve test performed is cached work and leads to
no extra overhead. In the re-exec case the resolve test can cost
O(100ms), but at the benefit of ensuring either the selected interpreter
will definitely work or no interpreters on the search path can work.

Fixes #1020

### DIFF
--- a/pex/environment.py
+++ b/pex/environment.py
@@ -27,12 +27,14 @@ from pex.typing import TYPE_CHECKING
 if TYPE_CHECKING:
     from typing import (
         DefaultDict,
+        Dict,
         FrozenSet,
         Iterable,
         Iterator,
         List,
         MutableMapping,
         Optional,
+        Tuple,
         Union,
     )
 
@@ -206,6 +208,8 @@ class _RequirementKey(ProjectName):
 
 
 class PEXEnvironment(object):
+    _CACHE = {}  # type: Dict[Tuple[str, str, Target], PEXEnvironment]
+
     @classmethod
     def mount(
         cls,
@@ -214,18 +218,24 @@ class PEXEnvironment(object):
         target=None,  # type: Optional[Target]
     ):
         # type: (...) -> PEXEnvironment
+        pex_file = os.path.realpath(pex)
         if not pex_info:
-            pex_info = PexInfo.from_pex(pex)
+            pex_info = PexInfo.from_pex(pex_file)
             pex_info.update(PexInfo.from_env())
         pex_hash = pex_info.pex_hash
         if pex_hash is None:
             raise AssertionError(
                 "There was no pex_hash stored in {} for {}.".format(PexInfo.PATH, pex)
             )
-        pex_root = pex_info.pex_root
-        pex = maybe_install(pex=pex, pex_root=pex_root, pex_hash=pex_hash) or pex
         target = target or targets.current()
-        return cls(pex=pex, pex_info=pex_info, target=target)
+        key = (pex_file, pex_hash, target)
+        mounted = cls._CACHE.get(key)
+        if mounted is None:
+            pex_root = pex_info.pex_root
+            pex = maybe_install(pex=pex, pex_root=pex_root, pex_hash=pex_hash) or pex
+            mounted = cls(pex=pex, pex_info=pex_info, target=target)
+            cls._CACHE[key] = mounted
+        return mounted
 
     def __init__(
         self,
--- a/pex/pex_bootstrapper.py
+++ b/pex/pex_bootstrapper.py
@@ -9,11 +9,13 @@ import sys
 
 from pex import pex_warnings
 from pex.common import atomic_directory, die, pluralize
+from pex.environment import ResolveError
 from pex.inherit_path import InheritPath
 from pex.interpreter import PythonInterpreter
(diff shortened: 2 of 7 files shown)


pex/pex_bootstrapper.py
```diff
@@ -36,6 +42,38 @@ def parse_path(path):
...
+        ):
+            from pex.environment import PEXEnvironment
+
+            pex_environment = PEXEnvironment.mount(
+                self.entry_point,
+                pex_info=self.pex_info,
+                target=LocalInterpreter.create(interpreter),
+            )
+            try:
+                pex_environment.resolve()
```
jsirois, line 71: FYI: This resolve test right here was what indirectly caused the need for #1768 and killing pkg_resources use. That used a bunch of pkg_resources.Requirement.parse and pkg_resources.Distribution finding and loading code and led to the boot-time vendor meta_path importer clash with Pex said same.

pex/environment.py
```diff
@@ -214,18 +218,24 @@ def mount(
...
         if pex_hash is None:
             raise AssertionError(
                 "There was no pex_hash stored in {} for {}.".format(PexInfo.PATH, pex)
             )
-        pex_root = pex_info.pex_root
-        pex = maybe_install(pex=pex, pex_root=pex_root, pex_hash=pex_hash) or pex
         target = target or targets.current()
-        return cls(pex=pex, pex_info=pex_info, target=target)
+        key = (pex_file, pex_hash, target)
+        mounted = cls._CACHE.get(key)
```
jsirois, line 232: This new `mount` caching combines with the pre-existing instance-level `resolve()` caching to avoid resolving twice in the happy path of the interpreter that tries to run a PEX 1st being a compatible one.

jsirois:
This change probably means shebang could be defaulted to `#!/usr/bin/env python`, but as discussed ad-nauseum that's still broken sometimes (for many Debian distributions that have `python3` but not `python`). The `--sh-boot` is probably the safest generic default, but as with that, changing how the shebang default works today (`#!/usr/bin/env pythonX.Y`) will break some users; so any change will have to wait for Pex 3. Pants though can do as it wishes with this new robustness!

Eric-Arellano:
> In the use case, the resolve test performed is cached work and leads to
no extra overhead. 

Which use case? I'm not following.

jsirois:
I'll follow-up here with the error message change since it's semi-significant. It won't go in the release notes but will go in the release.

benjyw (commented):
Late to the party, but looks great!

PR #1772, opened 2022-05-18
Fix runtime interpreter selection failure message.

jsirois:
The fix for #1020 did not add language for the new resolve check
failures.


Eric-Arellano (approved):
Thanks!

PR #1774, opened 2022-05-23
Ensure that lockfiles end in newlines.

benjyw:
I encountered a repo whose precommit checks choked on our lockfiles because they didn't end in newlines.
Since this is a reasonable expectation from a text file (indeed POSIX requires it, so various CLI tools assume it),
we oblige.

### DIFF
--- a/pex/cli/commands/lock.py
+++ b/pex/cli/commands/lock.py
@@ -300,6 +300,10 @@ class Lock(OutputMixin, JsonMixin, BuildTimeCommand):
         else:
             with self.output(self.options) as output:
                 dump(out=output)
+        # json.dump() does not write the newline terminating the last line, but some
+        # JSON linters, and line-based tools in general, expect it, and since these
+        # files are intended to be checked in to repos that may enforce this, we oblige.
+        output.write("\n")
 
     def _export(self):
         # type: () -> Result


Eric-Arellano (approved):

pex/cli/commands/lock.py
```diff
@@ -289,17 +288,23 @@ def _dump_lockfile(
...
-        dump = functools.partial(
-            self.dump_json,
-            self.options,
-            json_codec.as_json_data(lockfile=lock_file, path_mappings=path_mappings),
-            sort_keys=True,
-        )
+        def dump_with_terminating_newline(out):
+            # json.dump() does not write the newline terminating the last line, but some
+            # JSON linters, and line-based tools in general, expect it, and since these
+            # files are intended to be checked in to repos that may enforce this, we oblige.
```
jsirois, line 294: I'm fine with this, but no-one should be linting machine generated files! This is a human-read-only un-editable file.

jsirois (approved):

pex/cli/commands/lock.py
```diff
@@ -289,17 +288,23 @@ def _dump_lockfile(
     ):
         # type: (...) -> None
         path_mappings = resolver_options.get_path_mappings(self.options)
-        dump = functools.partial(
-            self.dump_json,
-            self.options,
-            json_codec.as_json_data(lockfile=lock_file, path_mappings=path_mappings),
-            sort_keys=True,
-        )
+        def dump_with_terminating_newline(out):
```
jsirois, line 291: CI is failing for lack of a newline between `path_mappings = ...` and `def ...`.

pex/cli/commands/lock.py
benjyw: At least the psf/black repo runs prettier on everything. I should probably make it ignore these. But there is still the issue of POSIX tools generally expecting text files to end with a newline. It may not be important enough to merge this though.

Also, is it strictly true that these are uneditable?I can imagine someone tweaking a single artifact in a lockfile for some rare reason.

jsirois: I'm fine with the merge, but black should probably also not be linting this as you say.

As far as POSIX tools that can't handle newlines - I'm sure there are some but I've encountered none. I will say that `cat`ing one of these does leave your terminal prompt odd; so I'll re-iterate I'm fine with this. It will "fix" that at the very least.

You're right, these are not strictly uneditable. You can certainly hack up a lock file at your own peril. I did this as a demonstration of a bugfix over here: https://github.com/pantsbuild/pants/issues/15614#issuecomment-1136548954 yesterday in fact. In spirit though, these things could be binary or toml or anything and no repo should care about formatting them.

PR #1785, opened 2022-05-27
Fix `--sh-boot` argv0.

### LINKED ISSUE #1782: Using `--sh-boot` results in `sys.argv[0]` based on hash
Building a PEX with the `--sh-boot` option results in the executable in `sys.argv[0]` set to a hash at runtime instead of the name of the PEX that was executed.  This often results in ugly help or usage output if using pex to build a console script.

Example using pex itself:

```
/tmp% pex -o pex-sh-boot.pex --sh-boot -c pex pex
/tmp% ./pex-sh-boot.pex -h | head -1
usage: 894c69f19720bd62c1be394be771eadfe7c6e8df [-o OUTPUT.PEX] [options] [-- arg1 arg2 ...]
```

But everything works as expected without `--sh-boot`:

```
/tmp% pex -o pex.pex -c pex pex
/tmp% ./pex.pex -h | head -1
usage: pex.pex [-o OUTPUT.PEX] [options] [-- arg1 arg2 ...]
```

Tested against pex 2.1.90.

jsirois:
Now `--sh-boot` mode PEXes have `sys.argv[0]` and the `PEX` environment
variable set consistently with all other modes of PEX execution.

Fixes #1782

### DIFF
--- a/pex/pex_builder.py
+++ b/pex/pex_builder.py
@@ -112,7 +112,7 @@ if __entry_point__ is None:
   sys.exit(2)
 
 __installed_from__ = os.environ.pop(__INSTALLED_FROM__, None)
-sys.argv[0] = __installed_from__ or sys.argv[0]
+sys.argv[0] = os.path.realpath(__installed_from__ or sys.argv[0])
 
 sys.path[0] = os.path.abspath(sys.path[0])
 sys.path.insert(0, os.path.abspath(os.path.join(__entry_point__, {bootstrap_dir!r})))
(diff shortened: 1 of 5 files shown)


benjyw (approved):

pex/pex_builder.py
```diff
@@ -112,7 +112,7 @@ def __maybe_run_venv__(pex, pex_root, pex_path):
   sys.exit(2)
 
 __installed_from__ = os.environ.pop(__INSTALLED_FROM__, None)
-sys.argv[0] = __installed_from__ or sys.argv[0]
+sys.argv[0] = os.path.realpath(__installed_from__ or sys.argv[0])
```
jsirois, line 115: This breaks the conscript use case of behaving differently based on symlink name.

jsirois: Probably the cause of #2006

PR #1787, opened 2022-05-28
Factor Pip patching into DownloadObservers.

jsirois:
This moves patching out of the Pip tool into dedicated foreign platform
and locker download observers that do their own patching. This allows
the notion of locking to be narrowed to just the lock create use of Pip
download.

This is part of a series of a refactors needed to remove dependency
cycles in support of #1696.


jsirois:
Reviewers, thanks in advance. I was able to split the ~2500 delta line change to support local project locking into many smaller patches, but this is the single largest one I couldn't reasonably break down further. The big thing to note is this one is all shuffling in the end - 0 functionality changes as indicated by the ~lack of test changes.

benjyw (approved):

PR #1791, opened 2022-05-30
Add PEP-517/518 sdist building support.

jsirois:
This will be needed to support getting a consistent fingerprint for
local projects when locking them.

Work towards #1696.


pex/build_system/pep_518.py
```diff
@@ -0,0 +1,183 @@
...
+    output_or_error = _execute_with_toml(
+        resolver,
+        dedent(
+            """\
+        import json
+        import sys
+
+        import toml
+
+        with open({pyproject_toml!r}) as fp:
```
benjyw, line 100: Presumably there is an obvious reason I'm not seeing, but why not simply parse toml in-process? Is it to avoid introducing a direct dep on the toml library? I ask because a lot of code here is just for running this parsing snippet in a toml pex, and the alternative of vendoring toml would be a lot simpler?

pex/build_system/pep_518.py
```diff
@@ -0,0 +1,183 @@
...
+                    "Pex distribution or source tree.".format(__name__)
+                )
+                contents = lock_file_contents.decode("utf-8")
+                source = os.path.join(os.path.dirname(__file__), "toml-lock.json")
+                lock = json_codec.loads(lockfile_contents=contents, source=source)
+
+                # Our toml lock was created against the default public PyPI index; so we can only
+                # use the lock if the resolve is also setup to use the default repos configuration
+                # as well.
+                if resolver.is_default_repos():
```
benjyw, line 56: IIUC if the PipConfiguration differs from the default in ways that could still allow lockfile use then we won't use the lockfile. That's fine but the one case I wonder about is repos_configuration including standard PyPI but also adding private repos. But that seems like a bridge we can cross when we come to it.

pex/build_system/pep_518.py
jsirois: I'm on a quest to get rid of all vendoring save for Pip. I've already ~killed setuptools and With the eventual drop of 2.7 & 3.5 support I'll be able to kill attrs. The setuptools and wheel needed to perform legacy builds are better resolved as ~latest. All that said, toml does vendor fine; so I'll do that for now.

pex/build_system/pep_518.py
jsirois: This is moot since I'll just vendor toml per above, but not using the lockfile even when we could've is only slower, not less correct, since the pinned version here is a dependency free dist. In other words, we just go through Pip hoops for it to eventually figure out there is only 1 project needed instead of directly fetching the locked dist URL. As such, instead of wading into somewhat titchy things - namely ensuring PyPI is the 1st index URL in the custom indexes list and double checking Pip uses indexes 1st than falls back to repos

pex/build_system/pep_518.py
benjyw: What's the reason? The vendoring seems pretty solid, and I've started playing around with vendoring portalocker in order to see what it would take to support Windows. Granted, that's not a lot of code and we could copy the salient parts into a single file, but still, what's the objection to vendoring?

jsirois: It barely works (doesn't work for some dists) and Red Baron died in 2019. Pip isn't vendored in the hard way (its imports are not re-written) and so it could just be dropped in w/o Red Baron.

On the portalocker front, we've had two contributors implement that bit without any library. The thing stopping all comers so far is my demand a Windows CI that is green for units & ITs comes along with the production fixes.

benjyw: If you feel that the long-term goal is to get rid of vendoring then I'm fine with this more complex solution, just wanted to verify that it is worth its weight.

jsirois: It's definitely a longer term goal, but I will just vendor for now. A long driving day just finished so I just haven't had time yet.

benjyw: SG just figured since you've already written it and I've already reviewed it, the alternative is also fine

jsirois: Now with toml vendored.

benjyw (approved):

PR #1792, opened 2022-05-31
Add support for local project locking.

### LINKED ISSUE #1696: Support local requirements with lockfiles

jsirois:
Closes #1696


tests/integration/cli/commands/test_local_project_lock.py
```diff
@@ -0,0 +1,133 @@
...
+        lock,
+        "--",
+        "-c",
+        "import colors; print(colors.__version__)",
+    ]
+    result = run_pex_command(args=print_colors_version_args)
+    result.assert_success()
+    assert "1.1.8" == result.output.strip()
+
+    # Running the test suite generates .pyc files which should not count against the project
```
jsirois, line 53: @benjyw thanks for all the prior reviews. All that work was to support this stability bit. That required building the local project into an sdist and hashing the sdist instead of hashing the local project dir directly. That, in turn, required resolving build system requirements if the local project was PEP-517/PEP-518, and that led to circularity issues with the standard resolver which led to all the circularity refactor PRs.

jsirois:
The test failure is unrelated and due to a traitlets release ~30 minutes ago: https://pypi.org/project/traitlets/5.2.2/#history. I've rolled the fix in to this change.

jsirois:
Urg, this of course breaks the release prep PR. I will break out the fix then.

pex/pip/local_project.py
```diff
@@ -0,0 +1,62 @@
...
+            sdist_or_error = pep_517.build_sdist(
+                project_directory=directory,
+                dist_dir=os.path.join(td, "dists"),
+                resolver=resolver,
+            )
+            if isinstance(sdist_or_error, Error):
+                return sdist_or_error
+            sdist = sdist_or_error
+
+            extract_dir = dest_dir or os.path.join(td, "extracted")
```
Eric-Arellano, line 52: When `dest_dir` is set, it means it's possible for this function to be called multiple times in parallel and to write to the same dest. What's the concurrency safety of this?

Eric-Arellano (approved):
Very exciting

tests/test_environment.py
```diff
@@ -138,7 +138,6 @@ def add_requirements(builder, cache):
         for installed_dist in resolver.resolve(
```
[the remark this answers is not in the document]
jsirois: That is correct.

pex/pip/local_project.py
jsirois: As you spotted, it's unsafe generally as an API but its safe as used.

There are 2 callers:
1. `fingerprint_local_project` above which does not pass a `dest`
2. `LocalProjectDownloadManager.save` which is given an exclusively locked atomic_directory by `DownloadManager.store` (downloads happen exactly-once).

jsirois: I'm going to merge, but I'm also not happy about this non-locally safe property. I'm thinking through ways to tighten this up so a local read shows this is always safe.

Eric-Arellano: Thanks for explaining. In the absence of a more elegant design...something like requiring the user to set `called_under_lock=True` if `dest_dir` is set could work. Like Rust's `unsafe!` declaration.

PR #1795, opened 2022-05-31
Fix test broken by traitlets 5.2.2 release.

jsirois:


jsirois:
Noticed in #1792 and #1794 which are both artificially red because of this.

benjyw (approved):

Eric-Arellano (approved):

PR #1797, opened 2022-06-01
Support adding new requirements in a lock update.

### LINKED ISSUE #1796: Add ability to update a lockfile without modifying existing entries
Sub-task of pantsbuild/pants#15704

When I add a new dependency to my project, I'd like to update my PEX lockfile with the dependency's version & transitive deps _without_ updating any of the existing entries in the lock. If it's not possible to do this (i.e. because the new dep has a transitive requirement that conflicts with something already in the lock), I'd like the command to fail (perhaps unless `-p the_transitive_dep` is passed to allow upgrading that one dependency).

Without this feature, I typically have to manually revert changes to my project's lockfile to remove "unrelated" changes. This is an error-prone / easy-to-forget step, so accidental upgrades sometimes sneak through to code review - from there they either cause confusion on the reviewer's part ("why are you upgrading X when the PR only talks about adding Y?"), or they pass through review and muddle our commit history.

jsirois:
If a `-p` targeted project update represents a new project requirement
and it can be resolved without conflicting with existing locked
projects, the lock update now succeeds.

Closes #1796


pex/cli/commands/lock.py
```diff
@@ -491,22 +486,37 @@ def _update(self):
                         file=output,
                     )
         if performed_update:
+            original_locked_project_names = {
+                locked_requirement.pin.project_name
+                for locked_resolve in lock_file.locked_resolves
+                for locked_requirement in locked_resolve.locked_requirements
+            }
+            new_requirements = OrderedSet(
```
jsirois, line 494: Erm, that's a bug when, for example, there was `-p foo` that was new, but it also introduces 10 other new transitive dependencies. Only the `foo` requirement should be added to the existing top-level requirements list.

jsirois: Actually, this is right. I'll push up a test that confirms.

Eric-Arellano (approved):
🎉

PR #1799, opened 2022-06-05
Add `pex3 lock update --dry-run check` mode.

jsirois:
This allows for clean automated checking if a lock needs updates.


pex/cli/commands/lock.py
```diff
@@ -485,38 +531,41 @@ def _update(self):
...
-                for update in updates
-                if update.project_name not in original_locked_project_names
-            )
-            constraints_by_project_name.update(
-                (constraint.project_name, constraint) for constraint in updates
+        if not version_updates:
+            return Ok()
+
+        if dry_run:
+            return Error(exit_code=len(version_updates)) if dry_run is DryRunStyle.CHECK else Ok()
```
benjyw, line 538: AFAICT nothing guarantees that the exit code will be <= 255, and if not, I believe the actual exit code will be this number mod 256, which could well be zero. 

Even if not, having an unpredictable exit code seems hard to work with and script around. 

So for those two reasons I think returning a fixed non-zero number makes more sense?

jsirois: True in the 1st. I don't agree on the second. Checking non-zero is straigt-forward. I'll go with the 1st justification, which could lead to issues and fix.

benjyw (approved):

PR #1800, opened 2022-06-05
Universal locks no longer record a `platform_tag`.

### LINKED ISSUE #1685: Remove platform_tag from Lockfile.
This used to be used for lock selection in multiplatform (non-universal) locks, but it's no longer needed since resolution uses resolve artifact ranking to find the best match.

jsirois:
The `platform_tag` of the interpreter used to perform a
`--style universal` lock is not relevant to the generated lock.
Recording it only led to lock diff noise; so, remove it for that style
of lock.

Fixes #1685


pex/resolver.py
```diff
@@ -47,46 +44,6 @@
...
-                max_parallel_jobs=self.max_parallel_jobs,
-            )
-            dist_metadatas = tuple(
-                DistMetadata.load(install_request.wheel_path)
-                for install_request in itertools.chain(
-                    tuple(self.download_result.install_requests()),
-                    itertools.chain.from_iterable(build_results.values()),
-                )
-            )
-        locked_resolve = LockedResolve.create(
```
jsirois, line 81: The removals in this file were incidental / should have happened in a prior refactor, Checking uses of `LockedResolve.create`, though, to see who passed platform_tag, led to discovering this dead use.

benjyw (approved):

PR #1823, opened 2022-07-03
Support restricting universal lock target os.

### LINKED ISSUE #1821: Lock creation should skip Windows-only requirements and / or allow selecting target platforms (OS classes).
Today, certain locks cannot be constructed due to problems with Windows only requirements. For example:
```
pex3 lock create --style universal --resolver-version pip-2020-resolver --interpreter-constraint "CPython>=3.10,<4" cryptography==36.0.2 docker==5.0.3
pid 20175 -> /home/jsirois/.pex/venvs/37a3fee58284b95ceb5b687f7c64e015c52553e2/764c26d931ec35862536cb560c7e8fb492d10e6d/pex --disable-pip-version-check --no-python-version-warning --exists-action a --no-input --isolated -q --cache-dir /home/jsirois/.pex/pip_cache --log /tmp/pex-pip-log.ln08x0ca/pip.log download --dest /tmp/tmpej6_nofx/home.jsirois..venv.pex.bin.python3 cryptography==36.0.2 docker==5.0.3 --index-url https://pypi.org/simple --retries 5 --timeout 15 exited with 1 and STDERR:
ERROR: Could not find a version that satisfies the requirement pywin32==227; sys_platform == "win32" (from docker)
ERROR: No matching distribution found for pywin32==227; sys_platform == "win32"
```

Since Pex does not currently support Windows, this seems like a bug. Windows aside, a similar issue could prevent an intended Mac-only or Linux-only universal lock from being created which suggests the feature of being able to select the target platform (OS class). N.B.: That this is different from selecting the Pex notion of `--platform` which is more specific and can be achieved directly with `--style strict` and `--style sources`.

jsirois:
Add the new `pex3 lock create --target-system {linux,mac,windows} ...`
option to restrict `--style universal` lock creation to dependencies
and artifacts compatible with the given target operating systems.

Fixes #1821

### DIFF
--- a/pex/cli/commands/lock.py
+++ b/pex/cli/commands/lock.py
@@ -15,7 +15,7 @@ from pex.dist_metadata import Requirement, RequirementParseError
 from pex.enum import Enum
 from pex.orderedset import OrderedSet
 from pex.resolve import requirement_options, resolver_options, target_options
-from pex.resolve.locked_resolve import LockConfiguration, LockedResolve, LockStyle
+from pex.resolve.locked_resolve import LockConfiguration, LockedResolve, LockStyle, TargetSystem
 from pex.resolve.lockfile import json_codec
 from pex.resolve.lockfile.model import Lockfile
 from pex.resolve.lockfile.operations import create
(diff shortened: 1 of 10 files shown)


pex/cli/commands/lock.py
```diff
@@ -584,7 +584,6 @@ def _update(self):
                 lock_file=attr.evolve(
                     lock_file,
                     pex_version=__version__,
-                    target_systems=SortedTuple(lock_updater.lock_configuration.target_systems),
```
jsirois: Originally I was gunning for `lock update --target-system ...` being able to amend the original target systems much like the recent change to make `lock update -p ...` accept adding new requirements and not just updating existing ones. So, you could start with a lock for Linux and Mac and years later add Windows support to the lock. That requires a bit of work though; so I deferred but had not fully cleaned up correctly.

benjyw (approved):

PR #1824, opened 2022-07-04
Check lock can resolve platforms at creation time.

### LINKED ISSUE #1595: Feature request: "universal" lock mode can reject unsupported platforms
See https://github.com/pantsbuild/pants/issues/12612#issuecomment-1023579158. The lockfile would still have every possible distribution like a normal "universal" lock mode. 

But it will validate that each dependency also has a prebuilt wheel available for the specified platforms. This allows eagerly failing to lock, rather than waiting to fail until building a PEX from that lock.

jsirois:
When a lock is created or updated, support specifying `--platform` or
`--complete-platform` to enable up-front checks that the resulting lock
will resolve for the specified platforms before materializing the lock.

Fixes #1595

### DIFF
--- a/pex/cli/commands/lock.py
+++ b/pex/cli/commands/lock.py
@@ -301,8 +301,24 @@ class Lock(OutputMixin, JsonMixin, BuildTimeCommand):
         if style is not LockStyle.UNIVERSAL:
             return target_config.resolve_targets()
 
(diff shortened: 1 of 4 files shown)


Eric-Arellano (approved):
Neat!

PR #1721, opened 2022-04-14
Support booting via `/bin/sh` with `--sh-boot`.

jsirois:
Allow users to choose `sh` as the boot mechanism for their PEXes. Not
only is `/bin/sh` probably more widely available than any given Python
shebang, but its also much faster.

PR #1827, opened 2022-07-06
Work around linux 128 byte shebang limit in venv

hrfuller:
I want to be able to run pex venvs on linux in environments where the $HOME is nested under a long path. Usually ~450 bytes including the short venv path parts.

### DIFF
--- a/pex/venv/pex.py
+++ b/pex/venv/pex.py
@@ -106,16 +106,17 @@ def _create_shebang(
     # This multiline shebang locates the interpreter in the venv relative to the file this shebang is placed in.
     # For example the script is usually ~/.pex/venvs/<sha>/<sha>/pex and the venv bin dir is ~/.pex/venvs/<sha>/<sha>/bin/.
     # The exec command re-execs the venv pex script with the venv interpreter. The unfortunate
-    # trick here is that the shell exec command and its arguemetns are wrapped in quotes so the python
+    # trick here is that the shell exec command and its arguments are wrapped in quotes so the python
     # interpreter treats it as a static string when executing the script, essentially ignoring it.
-    long_shebang_exec = '#!{sh_path}\n"exec" "`dirname $0`/bin/{interpreter_basename}" "-sE" "$0" "$@"'
+    long_shebang_exec = '#!{sh_path}\n        "exec" "`dirname $0`/bin/{interpreter_basename}" "-sE" "$0" "$@"'
 
     shebang = "#!{} -sE".format(venv_python)
     if len(shebang) > 128:
         sh_path = None
         # shutil.which only exists in CPython >= 3.3
         if hasattr(shutil, "which"):
-            sh_path = shutil.which("sh")
+            # ignore mypy because we are already checking for this attr above.
+            sh_path = shutil.which("sh") # type: ignore[attr-defined]
         elif os.path.exists("/bin/sh"):
             # fallback to /bin/sh if running python2.
             sh_path = "/bin/sh"
@@ -123,7 +124,7 @@ def _create_shebang(
         if sh_path is not None:
             shebang = long_shebang_exec.format(sh_path=sh_path, interpreter_basename=os.path.basename(venv_python))
     return shebang
-        
+
 
 class CollisionError(Exception):
     """Indicates multiple distributions provided the same file when merging a PEX into a venv."""
--- a/pex/venv/pex.py
+++ b/pex/venv/pex.py
@@ -91,6 +91,40 @@ def _copytree(
             return
 
 
+def _create_shebang(
+    venv_python, # type: str
+):
+    # type: (...) -> str
+    """Return the shebang to be used in the venv pex invoker script.
+
+    When the shebang is longer than the common linux length limit of 128 Bytes we try
+    to replace the shebang with a snippet of shell script which will invoke the correct
+    python interpreter from a path relative to the script itself, which avoids the need
+    to encode the long interpreter as a shebang.
+    """
+
+    # This multiline shebang locates the interpreter in the venv relative to the file this shebang is placed in.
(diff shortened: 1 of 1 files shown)


jsirois:
@hrfuller does `--sh-boot` work for you as an alternative?: https://github.com/pantsbuild/pex/pull/1721

pex/venv/pex.py
```diff
@@ -91,6 +91,40 @@ def _copytree(
...
+    # trick here is that the shell exec command and its arguemetns are wrapped in quotes so the python
+    # interpreter treats it as a static string when executing the script, essentially ignoring it.
+    long_shebang_exec = '#!{sh_path}\n"exec" "`dirname $0`/bin/{interpreter_basename}" "-sE" "$0" "$@"'
+
+    shebang = "#!{} -sE".format(venv_python)
+    if len(shebang) > 128:
+        sh_path = None
+        # shutil.which only exists in CPython >= 3.3
+        if hasattr(shutil, "which"):
+            sh_path = shutil.which("sh")
```
jsirois, line 118: This isn't generally valid. You may be building the PEX on a machine with a different answer than the (or one of the) deployment targets. That said, `/bin/sh` is very, very universally available.

pex/venv/pex.py
```diff
@@ -91,6 +91,40 @@ def _copytree(
...
+
+    # This multiline shebang locates the interpreter in the venv relative to the file this shebang is placed in.
+    # For example the script is usually ~/.pex/venvs/<sha>/<sha>/pex and the venv bin dir is ~/.pex/venvs/<sha>/<sha>/bin/.
+    # The exec command re-execs the venv pex script with the venv interpreter. The unfortunate
+    # trick here is that the shell exec command and its arguemetns are wrapped in quotes so the python
+    # interpreter treats it as a static string when executing the script, essentially ignoring it.
+    long_shebang_exec = '#!{sh_path}\n"exec" "`dirname $0`/bin/{interpreter_basename}" "-sE" "$0" "$@"'
+
+    shebang = "#!{} -sE".format(venv_python)
+    if len(shebang) > 128:
```
jsirois, line 114: And this is even less generally valid.

pex/venv/pex.py
hrfuller: My understanding is that this piece of code would always execute on the runtime host during the venv extraction, so the sh_path would be valid on that particular machine. Maybe I'm misunderstand your comment though.

pex/venv/pex.py
hrfuller: Yeah, it's not always the limit. Could check for darwin platform here, but I figured might as well no throw in platform logic.

hrfuller:
Hey @jsirois. Let me check on --sh-boot. Thanks.

pex/venv/pex.py
jsirois: My point is more about the fact you're checking on the wrong machine. This PEX build machine is generally not the deployment machine. IOW this may work for your case (say PEXes are built in CI on machines / images that match prod deployment machines / images), but it does not generally work and the feature is not behind a flag - you're subjecting every Pex user to it.

hrfuller:
I'm not sure if --sh-boot will work for me because the issue is the shebang in the venv invoker is too long. The one in the zipped pex is fine.

pex/venv/pex.py
jsirois: Ah, no, you're right. I got confused. You can disregard my wrong machine comments. That said, the fact that `--sh-boot` has lower latency anyhow will likely trump unless you find some issue with using that in your environment.

jsirois:
Let me take a look, but I think the --sh-boot script can be modified to execute `<venv python> <venv pex script>`.

pex/venv/pex.py
hrfuller: For the particular situation where I ran into this we really want to execute from a pex built venv not an unzipped pex so I don't think --sh-boot will be enough, though I do know we have folks at twtr that want to use --sh-boot in different contexts :)

jsirois: --sh-boot works for both --venv and ZIPAPP PEXes. I'm confused. Are you manually pre-building venvs with `PEX_TOOLS=1 my.pex venv create/venv/here`? If so you can just `create/venv/here/bin/python create/venv/here/pex`

hrfuller: You're right that --sh-boot does work with venv. sorry for the confusion. However it looks like it relies on the same [venv script shebang](https://github.com/pantsbuild/pex/blob/c15508ad1a2d48bdef9fbac677dbfa32927e6625/pex/bin/sh_boot.py#L182-L187), so I think in its current form this problem will still exist with --sh-boot. 

We aren't manually creating the venvs, we're relying on the venvs to bootstrap themselves so they can be distributed as a single executable.

jsirois: Ok. Let me finish my test of modifying `--sh-boot`. If that works, it will be more robust and faster.

jsirois: Alright - it looks like that works just fine. I'll post a PR since that was a bug anyhow since the shebang length limit was avoided for the other cases in that script, just not venv. I'll include a test that sets up a 10k venv path or some such.

jsirois: Ok, #1828. I have a robust test there for Linux. Seeing if my technique, which relies on a few errno symbolic constants being the same errnos for shebang too long and file name too long, works on Mac. It does work on Linux.

jsirois: Alright, took a while to get the generic shebang too long testing code working on Mac and Linux, but that's sorted. I'll polish up #1828 for review and I think that should satisfy your concerns++. It actually directly fixes the non --sh-boot case as well, but you'll probably want to switch to --sh-boot if your deployment machines all have `/bin/sh` since it will shave ~50ms of startup overhead and give you native venv speed + ~1ms.

hrfuller:
Closing in favor of #1828

jsirois:
Thanks for pushing on this @hrfuller - its nice to have this corner case for the most part definitively solved now. I'll get the 2.1.95 release out with it tomorrow.

PR #1828, opened 2022-07-07
Avoid ENOEXEC for --venv shebangs.

jsirois:
Re-work both ZIPAPP and --sh-boot execution of --venv PEXes to avoid shebang
length limits.

### DIFF
--- a/pex/pex_bootstrapper.py
+++ b/pex/pex_bootstrapper.py
@@ -559,7 +559,8 @@ def bootstrap_pex(entry_point):
                 venv_pex = ensure_venv(pex.PEX(entry_point, interpreter=target))
             except ValueError as e:
                 die(str(e))
-            os.execv(venv_pex, [venv_pex] + sys.argv[1:])
+            venv_python = os.path.join(os.path.dirname(venv_pex), "bin", "python")
+            os.execv(venv_python, [venv_python, venv_pex] + sys.argv[1:])
         else:
             maybe_reexec_pex(interpreter_test=interpreter_test)
             from . import pex
--- a/pex/pex_bootstrapper.py
+++ b/pex/pex_bootstrapper.py
@@ -560,7 +560,7 @@ def bootstrap_pex(entry_point):
             except ValueError as e:
                 die(str(e))
             venv_python = os.path.join(os.path.dirname(venv_pex), "bin", "python")
-            os.execv(venv_python, [venv_python, venv_pex] + sys.argv[1:])
+            os.execv(venv_python, [venv_python, "-sE", venv_pex] + sys.argv[1:])
         else:
             maybe_reexec_pex(interpreter_test=interpreter_test)
             from . import pex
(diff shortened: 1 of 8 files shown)


pex/pex_builder.py
```diff
@@ -94,7 +94,8 @@ def __maybe_run_venv__(pex, pex_root, pex_path):
     return
 
   TRACER.log('Executing venv PEX for {{}} at {{}}'.format(pex, venv_pex))
-  __re_exec__(venv_pex)
+  venv_python = os.path.join(venv_home, 'bin', 'python')
```
jsirois, line 97: Facepalm. I'm pretty sure I never really needed to invent the short venv symlinks - this pretty much covers things.

jsirois: Well, I guess it doesn't. The venv `bin/` console scripts are still used by folks building PEXes with `--venv {prepend,append}`. Those folks leverage the fact the console scripts are put on the PATH by Pex to call them by name (and not full path), and that utilizes the short venv symlink shebang we write into those console scripts. So that still has value.

pex/pex_bootstrapper.py
```diff
@@ -559,7 +559,8 @@ def bootstrap_pex(entry_point):
                 venv_pex = ensure_venv(pex.PEX(entry_point, interpreter=target))
             except ValueError as e:
                 die(str(e))
-            os.execv(venv_pex, [venv_pex] + sys.argv[1:])
+            venv_python = os.path.join(os.path.dirname(venv_pex), "bin", "python")
```
hrfuller, line 562: Nice! Is the non-version specific symlink guaranteed by virtualenv to exist?

tests/integration/test_sh_boot.py
```diff
@@ -196,3 +197,194 @@ def main():
...
+        assert (
+            subprocess.check_output(args=[pex_file] + test_pex_args)
+            .decode("utf8")
+            .startswith(pex_root)
+        )
+
+    if "--venv" in execution_mode_args:
+        # Running the venv pex directly should fail since the shebang length is too long.
+        with pytest.raises(OSError) as exc_info:
+            subprocess.check_call(args=[seeded_pex] + test_pex_args)
```
hrfuller, line 385: Is it worth allowing the /pex script be be executed directly anymore?

hrfuller (approved):
Thanks @jsirois. I think this should be sufficient to solve my problem.

tests/integration/test_sh_boot.py
```diff
@@ -196,3 +197,194 @@ def main():
...
+
+@pytest.mark.parametrize("execution_mode_args", EXECUTION_MODE_ARGS_PERMUTATIONS)
+def test_shebang_length_limit(
+    tmpdir,  # type: Any
+    execution_mode_args,  # type: List[str]
+    shebang_length_limit,  # type: int
+):
+    # type: (...) -> None
+
+    long_dir_name = "x" * shebang_length_limit
```
hrfuller, line 353: Looking at the failing test is the shebang length limit computed the max allowed or the shortest that will fail?
```suggestion
    long_dir_name = "x" * shebang_length_limit - 1
```

pex/pex_bootstrapper.py
jsirois: Yup.

jsirois:
Thanks for reviewing @hrfuller!

PR #1829, opened 2022-07-07
Allow pex to read pants style pex lock files

asherf:
Pants add comments to the lock files, we want pex to be able to deal with that without running any commands to clean those up.


jsirois (commented):
I don't think this is the right thing to do. Pants should not be adding content to an opaque format. I argued this point extensively in Slack. The issue that summarizes some of this is here: https://github.com/pantsbuild/pants/issues/14281

Pants should probably do the stripping instead since it added the headers in the 1st place. It seems like `./pants export --lockfile ...` would be an appropriate place / new feature.

asherf:
> I don't think this is the right thing to do. Pants should not be adding content to an opaque format. I argued this point extensively in Slack. The issue that summarizes some of this is here: [pantsbuild/pants#14281](https://github.com/pantsbuild/pants/issues/14281)
> 
> Pants should probably do the stripping instead since it added the headers in the 1st place. It seems like `./pants export --lockfile ...` would be an appropriate place / new feature.



> I don't think this is the right thing to do. Pants should not be adding content to an opaque format. I argued this point extensively in Slack. The issue that summarizes some of this is here: [pantsbuild/pants#14281](https://github.com/pantsbuild/pants/issues/14281)
> 
> Pants should probably do the stripping instead since it added the headers in the 1st place. It seems like `./pants export --lockfile ...` would be an appropriate place / new feature.

I did add https://github.com/pantsbuild/pants/issues/15940 but I don't think it is going to get implemented any time soon. also it seems to be a little heavy handed for my use case which is that I need a lock file I can run [pip-audit](https://github.com/trailofbits/pip-audit) on to make sure we don't have any issues with the packages we use.

jsirois:
That does sound heavy handed for your use case. `./pants export --lockfile` sounds more on point. That would export all lockfiles, Poetry or Pex format, to Pip format. You'd get too many lockfiles until your export selection issue is taken up, but it would be fast still.

PR #1832, opened 2022-07-08
Fix `PEX_EXTRA_SYS_PATH` propagation.

### LINKED ISSUE #1825: Support exporting PYTHONPATH before running user code
As discussed in Slack, it'd be nice to use `PYTHONPATH` in Pants (and not have it be scrubbed) when invoking user PEX script so that the user code can use `subprocess.run([sys.executable, ...], ...)` and not have to pass a constructed `PYTHONPATH` in the `env=` argument.

How this looks is dealer's choice, as it's hidden from me, the client, through Pants.

jsirois:
Previously PEXes were not behaving like a Python interpreter when
`PEX_EXTRA_SYS_PATH` was used to add entries to the `sys.path`. In a
traditional Python interpreter those would be added with `PYTHONPATH`
which subprocesses forked from the Python interpreter would inherit by
default.

Fixes #1825

### DIFF
--- a/pex/interpreter.py
+++ b/pex/interpreter.py
@@ -124,7 +124,7 @@ class PythonIdentity(object):
         # Pex identifies interpreters using a bit of Pex code injected via an extraction of that
         # code under the `PEX_ROOT` adjoined to `sys.path` via `PYTHONPATH`. We ignore such adjoined
         # `sys.path` entries to discover the true base interpreter `sys.path`.
-        pythonpath = frozenset(os.environ.get("PYTHONPATH", "").split(os.pathsep))
+        pythonpath = frozenset(os.environ.get("PYTHONPATH", "").split(":"))
         sys_path = [item for item in sys.path if item and item not in pythonpath]
 
         return cls(
--- a/pex/interpreter.py
+++ b/pex/interpreter.py
@@ -124,7 +124,7 @@ class PythonIdentity(object):
         # Pex identifies interpreters using a bit of Pex code injected via an extraction of that
         # code under the `PEX_ROOT` adjoined to `sys.path` via `PYTHONPATH`. We ignore such adjoined
         # `sys.path` entries to discover the true base interpreter `sys.path`.
-        pythonpath = frozenset(os.environ.get("PYTHONPATH", "").split(":"))
+        pythonpath = frozenset(os.environ.get("PYTHONPATH", "").split(os.pathsep))
         sys_path = [item for item in sys.path if item and item not in pythonpath]
 
         return cls(
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -327,26 +327,35 @@ class PEX(object):  # noqa: T000
 
         pythonpath = cls.unstash_pythonpath()
         if pythonpath is not None:
-            original_pythonpath = pythonpath.split(os.pathsep)
-            user_pythonpath = list(OrderedSet(original_pythonpath) - set(sys.path))
-            if original_pythonpath == user_pythonpath:
+            original_pythonpath = pythonpath.split(":")
+            user_pythonpath_entries = list(OrderedSet(original_pythonpath) - set(sys.path))
+            user_pythonpath = ":".join(user_pythonpath_entries)
+            if original_pythonpath == user_pythonpath_entries:
                 TRACER.log("Unstashed PYTHONPATH of %s" % pythonpath, V=2)
             else:
                 TRACER.log(
-                    "Extracted user PYTHONPATH of %s from unstashed PYTHONPATH of %s"
-                    % (os.pathsep.join(user_pythonpath), pythonpath),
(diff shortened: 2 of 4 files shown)


pex/interpreter.py
```diff
@@ -124,7 +124,7 @@ def get(cls, binary=None):
         # Pex identifies interpreters using a bit of Pex code injected via an extraction of that
         # code under the `PEX_ROOT` adjoined to `sys.path` via `PYTHONPATH`. We ignore such adjoined
         # `sys.path` entries to discover the true base interpreter `sys.path`.
-        pythonpath = frozenset(os.environ.get("PYTHONPATH", "").split(os.pathsep))
+        pythonpath = frozenset(os.environ.get("PYTHONPATH", "").split(":"))
```
jsirois, line 127: Wholesale fixed these. Python defines the correct character under all OSes as ":".

jsirois:
OK, this is now robust for all permutations of relevant Pex inheritance settings and execution modes and is ready for review.

jsirois:
Sorry for the thrash on `":"` vs `os.pathsep`. The relevant code is here and it uses a system-specific `DELIM` value: https://github.com/python/cpython/blob/4bed0db7c222f8df1b4e31107c0305214caf3f56/Modules/getpath.py#L637-L640

This was particularly confusing since our `PEX_EXTRA_SYS_PATH` specifies a fixed value of `":"`. I guess when Windows support is added `PEX_EXTRA_SYS_PATH` can be retroactively made to work like `PYTHONPATH` and use `";"` (`os.pathsep`)on Windows without breaking pre-existing Pex users.

pex/pex.py
```diff
@@ -580,7 +592,7 @@ def execute_interpreter(self):
         # type: () -> Any
 
         # A Python interpreter always inserts the CWD at the head of the sys.path.
-        sys.path.insert(0, ".")
+        sys.path.insert(0, "")
```
benjyw, line 595: Did "." never work? Or is "" just better spelling? I notice that in a vanilla python interpreter `sys.path[0]` is indeed `""`, so this definitely makes sense to me as a change. Just curious about why now.

benjyw (approved):

pex/pex.py
jsirois: Just better spelling. At least consistent spelling with CPython and PyPy as you found out.

jsirois:
I'll submit shortly to get out the 2.1.95 release. @thejcannon hopefully the unit test directly addresses your use case.

thejcannon:
Yeah sorry haven't had time to test it out. You're always very thorough with your changes and tests, so I'm not worried 😉

jsirois:
Ah, I was just hoping for a code review of the code / test case. Actually testing against this Pex code is above and beyond. I'll go ahead and submit then.

pex/pex.py
```diff
@@ -413,7 +413,19 @@ def patch_all(path, path_importer_cache, modules):
...
+
+            # Let Python subprocesses see the same sys.path additions we see. If those Python
+            # subprocesses are PEX subprocesses, they'll do their own (re-)scrubbing as needed.
+            if inherit_path is InheritPath.FALSE:
+                pythonpath_entries = extra_sys_path
+            else:
+                raw_pythonpath = os.environ.get(self._PYTHONPATH)
+                pythonpath = raw_pythonpath.split(os.pathsep) if raw_pythonpath else []
+                pythonpath_entries = pythonpath + extra_sys_path
+            os.environ[self._PYTHONPATH] = os.pathsep.join(pythonpath_entries)
```
jsirois, line 427: This mechanism is broken for some use cases. The fundamental issue is the mechanism of `PEX_EXTRA_SYS_PATH` propagation (`PYTHONPATH`) is not faithful. `PEX_EXTRA_SYS_PATH` appends to the `sys.path` but `PYTHONPATH` prepends. This will need to be reverted or re-worked to produce a faithful representation.

pex/venv/pex.py
```diff
@@ -417,7 +417,17 @@ def sys_executable_paths():
                 )
 
             os.environ["VIRTUAL_ENV"] = venv_dir
-            sys.path.extend(os.environ.get("PEX_EXTRA_SYS_PATH", "").split(os.pathsep))
+
+            # A Python interpreter always inserts the CWD at the head of the sys.path.
+            sys.path.insert(0, "")
```
jsirois, line 422: This is almost certainly a bug. Inserting CWD here before determining below if we're actually running as an interpreter. The equivalent PEX zipapp code only triggers when running as an interpreter.

thejcannon: https://docs.python.org/3/library/sys.html#sys.path lists the cases for which paths get prepended (including `""`)

thejcannon: https://github.com/pantsbuild/pex/issues/1983

PR #1833, opened 2022-07-08
Change user agent of vendored pip version

asherf:
this will allow distinguishing (in the pypi BigQuery data set, for example) between pip downloads of pex and pex using the vendored pip to download packages.

### DIFF
--- a/pex/vendor/_vendored/pip/pip/_internal/network/session.py
+++ b/pex/vendor/_vendored/pip/pip/_internal/network/session.py
@@ -100,7 +100,7 @@ def user_agent():
     Return a string representing the user agent.
     """
     data = {
-        "installer": {"name": "pip", "version": __version__},
+        "installer": {"name": "pex-pip", "version": __version__},
         "python": platform.python_version(),
         "implementation": {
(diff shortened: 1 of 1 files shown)


pex/vendor/_vendored/pip/pip/_internal/network/session.py
```diff
@@ -100,7 +100,7 @@ def user_agent():
     Return a string representing the user agent.
     """
     data = {
-        "installer": {"name": "pip", "version": __version__},
+        "installer": {"name": "pex-pip", "version": __version__},
```
jsirois, line 103: If you do re-open using a patch applied post-vendoring, which is the way this should be done, it would be better to patch the version than the name, maybe just add a suffix of `" (pex)"`. If there are user-agent limits imposed or relaxed for Pip, Pex should probably suffer or enjoy those same which is probably keyed of name.

asherf: Yes, I like your approach better and I agree with the rule that the CI enforces (no changes to vendored code)
Will give it a try.

PR #1837, opened 2022-07-11
Fix `--venv` PEX `PEX_EXTRA_SYS_PATH` propagation.

### LINKED ISSUE #1836: `PEX_EXTRA_SYS_PATH` propagation can break subprocesses run against other venvs.
The propagation was introduced in #1832 released in 2.1.95 and the first known instance of the break was seen over in Pants and led to a revert of Pex 2.1.95 here: https://github.com/pantsbuild/pants/pull/16125

For the zipapp case, `PYTHONPATH` is the only way to get `PEX_EXTRA_SYS_PATH` propagated; so that scheme will always work as you expect for subprocesses against sys.executable but may fail for subprocesses against another Python. For the venv case, we could do better and use the `.pth` path extension mechanism to effect `sys.path` extension for the venv interpreter without polluting other interpreters via `PYTHONPATH`. The `.pth` mechanism only works in a `site-packages` dir, but it's exactly in the venv case that Pex controls creation of `site-packages` and can inject a `.pth` shim that looks at `PEX_EXTRA_SYS_PATH` and does extension there. That way, the `sys.path` extension is both pluggable via env var and native to the venv itself and any execution of its interpreter.

jsirois:
Previously this used `PYTHONPATH` for propagation, which could lead to
issues executing foreign Pythons in subprocesses. Switch to using a
`.pth` file. Although the problem of polluting foreign Pythons in
subprocesses remains for zipapp PEXes, it is the case that `--venv` is
Pex's maximum compatibility mode and this can be viewed as yet another
case where a Pex feature breaking a Python app should be solved by using
`--venv` mode.

Fixes #1836

### DIFF
--- a/pex/venv/pex.py
+++ b/pex/venv/pex.py
@@ -303,6 +303,17 @@ def _populate_sources(
     ):
         yield src, dst
 
+    with open(os.path.join(venv.site_packages_dir, "PEX_EXTRA_SYS_PATH.pth"), "w") as fp:
+        # N.B.: .pth import lines must be single lines: https://docs.python.org/3/library/site.html
+        for env_var in "PEX_EXTRA_SYS_PATH", "__PEX_EXTRA_SYS_PATH__":
+            print(
+                "import os, sys; "
+                "sys.path.extend("
+                "entry for entry in os.environ.get('{env_var}', '').split(':') if entry"
+                ")".format(env_var=env_var),
+                file=fp,
+            )
+
     with open(os.path.join(venv.venv_dir, pex_info.PATH), "w") as fp:
         fp.write(pex_info.dump())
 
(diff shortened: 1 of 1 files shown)


jsirois:
N.B.: The pre-existing tests from #989 and #1832 vet this. In particular, the following test added in #1832 brought to light the issue with `PEX*` scrubbing / the use of `__PEX_EXTRA_SYS_PATH__` in that case:
https://github.com/pantsbuild/pex/blob/4aecc86263ccf8d27630ef761335591ab5100759/tests/integration/test_issue_1825.py#L114-L133

This test failed in `assert_inherited` in VENV mode.

jsirois:
I did verify that this Pex causes `./build-support/bin/release.sh build-wheels` to go green over in https://github/pantsbuild/pants.

benjyw (approved):

PR #1863, opened 2022-07-25
Make wheel install site-packages detection robust.

jsirois:
We already had robust and well tested code for this; so just re-plumb a
bit and use it.

Fixes #1861

PR #1865, opened 2022-07-25
Inline `pex.common.find_site_packages`.

jsirois:
The function is only used by the vendoring script where its
non-robustness in the general case is fine and where its use instead of
`pex.venv.virtualenv.find_site_packages_dir` is necessitated by the
bootstrapping issues inherent in using vendored code while vendoring it.

Follow up to #1863

### DIFF
--- a/pex/vendor/__main__.py
+++ b/pex/vendor/__main__.py
@@ -14,14 +14,7 @@ from collections import OrderedDict, defaultdict
 from colors import bold, green, yellow
 from redbaron import CommentNode, LiteralyEvaluable, NameNode, RedBaron
 
-from pex.common import (
-    find_site_packages,
-    safe_delete,
-    safe_mkdir,
-    safe_mkdtemp,
-    safe_open,
-    safe_rmtree,
-)
+from pex.common import safe_delete, safe_mkdir, safe_mkdtemp, safe_open, safe_rmtree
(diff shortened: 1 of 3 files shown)


jsirois:
I tested this with `rm -rf pex/vendor/_vendored/ && tox -evendor` which works for this inlining but does not work for using `pex.venv.virtualenv.find_site_packages_dir` instead which indirectly relies on vendored code served up through `pex.third_party.{attr,packaging}`.

pex/vendor/__main__.py
```diff
@@ -201,6 +194,18 @@ class VendorizeError(Exception):
     """Indicates an error was encountered updating vendored libraries."""
 
 
+def find_site_packages(prefix_dir):
```
benjyw, line 197: A comment explaining to the naive reader why we can't use `find_site_packages_dir` here would be valuable.

benjyw (approved):

pex/vendor/__main__.py
jsirois: Like this?: https://github.com/pantsbuild/pex/blob/e38fdc5047e7309b3b78d71ee95dfa73f457f471/pex/vendor/__main__.py#L210-L216

benjyw: I guess that is sufficient. Plus if anyone tries to change this, tests will break.

PR #1878, opened 2022-08-10
Upgrade to Python 3.11.0-rc.1.

jsirois:


benjyw (approved):

benjyw (approved):

benjyw (approved):
Let me know if you want me to push on the Mac side since I have one

jsirois:
> Let me know if you want me to push on the Mac side since I have one

Thanks. The macstadium Macs are handy for this - although the non-M1 mac has no access info yet in the vault. I took another approach and questioned why Pillow - it turns out Pillow wasn't the point at all; so I replaced it with a universal wheel dist instead.

benjyw (approved):

Issue #1805, opened 2022-06-17
Pull in Pip fixes.


jsirois:
Some Pex users are encountering resolve issues fixed in newer versions of Pip. Pex has been stuck on the last version of Pip to support Python 2.7 (20.3.4) in order to continue to support Python 2.7 customers. Either Pex has to cut those customers loose or it has to support both old and new Pips.

The basic scenarios:
1. Cut loose - fend for yourself: Pex simply bumps its major version and moves on; Python 2.7 users get no bug fixes or features and are stuck on the last version of Pex to vendor Pip 20.3.4.
2. Bifurcate with support: Like 1, except Pex cuts a long-term branch where fixes and releases for the Pex 2.1.x / Pip 20.3.4 series go on in parallel to Pex 3.x / Pip ~latest development and releases off of main.
3. Support two versions of Pip in the same Pex release: This could keep the 20.3.4 Pip vendoring in the released Pex but just-in-time resolve a newer version of Pip using the older vendored Pip. The use of the newer Pip could either be controlled via flag or via auto-detect from interpreter constraints / interpreters used / platforms.

Obviously, 1 is easiest from a maintainer perspective, but it violates an implicit contract with users. It would be better to phase out Python 2.7 / 3.5 / 3.6 / etc. support in a controlled way independent of needing to upgrade Pip to get Pip bug-fixes. As the current ~lone maintainer, I'm partial to 3 to both meet users needs and have a compact codebase to maintain. I fear 2 would lead to more complicated maintenance and missed / partial fixes.

jsirois:
The motivating slow lock resolves were reported variously through #1798 on behalf of Pants users.

jsirois:
This upgrade looks like it will close #1526, #1527 and #1528.

jsirois:
The only apparent differences found in Pip 22.2.2 are:
1. An extra Wheel model method that needs to be runtime patched for --style universal locks. This patch is a noop against 20.3.4-patched; so this works out well.
2. Pip now attempts to use the json API defined in https://peps.python.org/pep-0691/. This API is implemented by PyPI; so Pip logs from PyPI downloads no longer contain hashes, they are hidden in a fetch of `https://pypi.org/simple/<project name>/` with `Accept: application/vnd.pypi.simple.v1+json`.

[...]

PR #1880, opened 2022-08-15
Remove unused Pip log scraping.

jsirois:
Fixes #facepalm.

Work towards #1805

### DIFF
--- a/pex/resolve/locker.py
+++ b/pex/resolve/locker.py
@@ -273,19 +273,6 @@ class Locker(LogAnalyzer):
                 self._links[project_name_and_version].add(partial_artifact)
                 return self.Continue()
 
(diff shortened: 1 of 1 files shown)


Eric-Arellano (approved):

jsirois:
This was un-needed when introduced in #1579. Proof? Well tests still pass, but the log line format does not even exist in the vendored Pip, only in more modern Pip. Clearly I had experimented with Pip in a venv and did not use the right version, added scraping, wondered why it didn't work, went on to enhance runtime patches, then never cleaned up. The runtime patches ended up being needed more widely to support `--target-systems` and proper `--interpreter-constraints` handling, but facepalm nonetheless.

PR #1882, opened 2022-08-19
Add support for PEP-691.

jsirois:
This is needed to support newer Pip where fingerprints are hidden from the
verbose logs Pex scrapes to build locks. The Pex fallback of downloading
every artifact locked and hashing it works but is just too expensive to
be the normal case.

Work towards #1805

### DIFF
--- a/pex/resolve/pep_691/api.py
+++ b/pex/resolve/pep_691/api.py
@@ -3,6 +3,7 @@
 
 from __future__ import absolute_import
 
+import io
 import json
 
 from pex.compatibility import HTTPError, text, urlparse
@@ -70,7 +71,9 @@ class Client(object):
                 with self._url_fetcher.get_body_stream(
                     endpoint.url, extra_headers={"Accept": endpoint.content_type}
                 ) as fp:
-                    data = json.load(fp)
+                    # JSON exchanged between systems must be UTF-8:
+                    # https://www.rfc-editor.org/rfc/rfc8259#section-8.1
+                    data = json.load(io.TextIOWrapper(fp, encoding="utf-8"))
         except (IOError, OSError, HTTPError) as e:
             raise request_error("failed: {err}".format(err=e))
         except ValueError as e:
--- a/pex/resolve/pep_691/api.py
+++ b/pex/resolve/pep_691/api.py
@@ -80,8 +80,8 @@ class Client(object):
                         if sys.version_info[:2] == (3, 5)
                         else fp
                     )
-                    # The above is tested to work with PyPy 27,3.{6,7,8,9} and
-                    # CPython 2.7,3.{5,6,7,8,9,10,11}, so the type suppression is well vetted here.
+                    # The above is tested to work with PyPy 2.7, PyPy 3.{6,7,8,9}, CPython 2.7 and
+                    # CPython 3.{5,6,7,8,9,10,11}; so the type suppression is well vetted here.
                     data = json.load(stream)  # type: ignore[arg-type]
         except (IOError, OSError, HTTPError) as e:
             raise request_error("failed: {err}".format(err=e))
--- a/pex/resolve/pep_691/api.py
+++ b/pex/resolve/pep_691/api.py
@@ -5,6 +5,7 @@ from __future__ import absolute_import
 
 import io
 import json
+import sys
 
 from pex.compatibility import HTTPError, text, urlparse
 from pex.fetcher import URLFetcher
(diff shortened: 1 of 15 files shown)


pex/resolve/pep_691/model.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+@attr.s(frozen=True)
+class File(object):
+    _GUARANTEED_HASH_ALGORITHM_DIGEST_SIZES = {
+        algorithm: hashlib.new(algorithm).digest_size for algorithm in hashlib.algorithms_guaranteed
+    }
+
+    @classmethod
+    def _select_algorithm(cls, algorithms):
```
jsirois, line 29: This gets tested in `test_fingerprint_service.test_mixed`, but only indirectly. I could add a dedicated unit test, but I already feel bad about the size of this thing.

jsirois:
Reviewers, feel free to beg off or else accept my condolences. Pretty huge again, although its just 2 new production types api.Client and FingerprintService. The former gets the hash data from PyPI, the latter caches it. The rest is a bit of shuffling in seperate commits and then tests for those two new types.

Here's a reminder link: https://peps.python.org/pep-0691/

pex/resolve/pep_691/api.py
```diff
@@ -0,0 +1,185 @@
...
+            )
+
+        try:
+            with TRACER.timed(
+                "Fetching PEP-691 index metadata from {url} for {content_type}".format(
+                    url=endpoint.url, content_type=endpoint.content_type
+                )
+            ):
+                with self._url_fetcher.get_body_stream(
+                    endpoint.url, extra_headers={"Accept": endpoint.content_type}
```
jsirois, line 71: FWIW these requests average ~200ms on my CenturyLink mildly crappy home DSL. I get a parallelism of 10 and so for pantsbuild.pants with no cache I see ~1.2s to fetch all the fingerprints. Cached is sub-ms.

jsirois:
This will be interesting, all green on 2.7.18 (and all other Pythons) on my machine: https://github.com/pantsbuild/pex/runs/7910216983?check_suite_focus=true#step:7:1236

jsirois:
Alright - this should be all green now and good to go. A bit of SQLite dialect massaging to be compatible across variants and json.load massaging to work with the lone outlier in Python 3.5.

pex/resolve/pep_691/fingerprint_service.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+@attr.s(frozen=True)
+class _FingerprintedURL(object):
+    url = attr.ib()  # type: str
+    fingerprint = attr.ib()  # type: Fingerprint
+
+
+@attr.s(frozen=True)
+class FingerprintService(object):
```
Eric-Arellano, line 39: Docstring may help future readers explaining why this exists.

pex/resolve/pep_691/fingerprint_service.py
```diff
@@ -0,0 +1,217 @@
...
+
+        :param artifacts: The artifacts to fill in missing fingerprints for.
+        :return: An iterator over all the artifacts given, where some returned artifacts may have
+                 previously missing fingerprints filled in (but not verified).
+        """
+        # N.B.: In the absence of having an advertised fingerprint for an artifact, Pex will
+        # download the artifact and hash it at higher layers; so simply warning about the error is
+        # enough; we need not fail the whole endeavor. The partial artifact just remains partial
+        # and becomes more expensive to lock. The warning is likely un-actionable by the user but
+        # also rare; so it can be reported to maintainers in case the 5s default sqlite3 timeout is
```
Eric-Arellano, line 131: Would it make sense to encourage users to post issues to pex repo when they encounter this?

pex/resolve/pep_691/model.py
Eric-Arellano: > I already feel bad about the size of this thing.

It's okay. As typical, your code is well documented and tested, and easy to follow along :)

I think a unit test would be valuable, if you're willing

pex/resolve/pep_691/api.py
```diff
@@ -0,0 +1,197 @@
...
+
+            files.append(
+                File(
+                    filename=get("filename", text, obj=file, path=path),
+                    # N.B.: All URLs in the PEP-691 API are allowed to be relative, see:
+                    #   https://peps.python.org/pep-0691/#json-serialization
+                    # The `urljoin` functions does the right thing here and creates an absolute URL
+                    # only when needed (never for the current PyPI scheme, potentially for other
+                    # indexes though).
+                    url=urlparse.urljoin(endpoint.url, get("url", text, obj=file, path=path)),
```
Eric-Arellano, line 192: Maybe worth a dedicated unit test? This seems niche and unlikely to be encountered frequently in the wild, so it would be easy to regress here without realizing.

Eric-Arellano (approved):

pex/resolve/pep_691/api.py
jsirois: Well, it won't be easy to regress since it is tested already, but I think breaking out the test to a dedicated unit test makes sense for other reasons.

pex/resolve/pep_691/fingerprint_service.py
jsirois: No, but yes if they encounter it often. I'll add to the warning with more verbage.

pex/resolve/pep_691/fingerprint_service.py
jsirois: It's only got 1 (real) public method and that _is_ documented so I think this is already well covered. Doubly so as soon as it gets its 1st non-test use here forthcoming.

Eric-Arellano: Fair enough -- thanks

pex/resolve/pep_691/model.py
jsirois: Done.

pex/resolve/pep_691/api.py
jsirois: Done.

Eric-Arellano (approved):

PR #1884, opened 2022-08-21
Make lock update sensitive to artifacts.

jsirois:
Previously lock updates were sensitive to changes in pins but not to
changes in the set of artifacts when pins remained unchanged.


jsirois:
OMFG

PR #1886, opened 2022-08-23
Ensure locally built wheel is consumable locally.

jsirois:
Previously there was ambiguity in the key used to store built wheels when the
local OS (platform tag) could take on different values as is the case in
macOS Big Sur where 10.16 and 11.0 are synonyms.


jsirois:
No issue reported here, I just happened to hit this in a macOS Python 2.7 CI shard in an otherwise unrelated change. It seems clear this could also happen in slightly more contrived scenarios where a user mounts the PEX_ROOT into a vm or container that has a differing platform tag too; so I let the fix be general instead of macOS-specific.

Eric-Arellano (approved):

PR #1887, opened 2022-08-23
Make lock update sensitive to artifacts.

jsirois:
Previously lock updates were sensitive to changes in pins but not to
changes in the set of artifacts when pins remained unchanged.


jsirois:
The second commit is needed for green CI but is off for seperate review in #1886.

jsirois:
This shortcoming in functionality was reported by @zmanji in Slack - no associated issue.

pex/cli/commands/lock.py
```diff
@@ -32,13 +42,22 @@
...
     import attr  # vendor:skip
 else:
     from pex.third_party import attr
 
 
+class FingerprintMismatch(Enum["FingerprintMismatch.Value"]):
+    class Value(Enum.Value):
+        pass
+
+    IGNORE = Value("ignore")
```
stuhood, line 56: Does ignoring in this case amount to adjusting the lockfile to use a new fingerprint? Or will it keep the old fingerprint? Should this have a different name that clarifies that?

pex/resolve/lockfile/updater.py
```diff
@@ -110,9 +232,13 @@ def create(
     pip_configuration = attr.ib()  # type: PipConfiguration
 
     @contextmanager
-    def _calculate_requirement_configuration(self, locked_resolve):
-        # type: (LockedResolve) -> Iterator[RequirementConfiguration]
-        if not self.update_constraints_by_project_name:
+    def _calculate_requirement_configuration(
+        self,
+        locked_resolve,  # type: LockedResolve
+        pin_all=False,  # type: bool
```
stuhood, line 238: Given that this is `pin_all` (implying that a different scope of pinning might be necessary/useful), should the `--pin` flag maybe start as either an enum, or perhaps a list option with an initial special-case value of e.g. `:all:`?

stuhood (approved):

pex/cli/commands/lock.py
jsirois: When you ignore the mismatch, the new fingerprint is used. I expanded the corresponding option help to document what each value does.

pex/resolve/lockfile/updater.py
jsirois: I don't think so. The two existing modes of lock update are:
+ no args: Update the lock given its original input root requirements (these are stored in a lockfile)
+ `-p one -p "two<2" -p three==1.0` ...: Update the lock pinning everything else and just updating (or adding) these projects if possible.

So this just offers the only missing mode as far as I can tell - update the lock keeping everything `--pin`ned.

PR #1890, opened 2022-08-25
Avoid in-flight .pyc compiles of Pex bootstrap code.

### LINKED ISSUE #1889: PEXBuilder is not robust to ephemeral .pyc compiles.
The current code ignores the final `.pyc`, but not the intermediate form:
https://github.com/pantsbuild/pex/blob/d7890631620b231ad68a15d8bff23abf4884c185/pex/pex_builder.py#L589-L591

The code to handle this correctly is here:
https://github.com/pantsbuild/pex/blob/9901a05d0ec8aee9b8a6e05c6f2a00999df3bab6/pex/common.py#L60-L76

jsirois:
Fixes #1889

### DIFF
--- a/pex/pex_builder.py
+++ b/pex/pex_builder.py
@@ -13,6 +13,7 @@ from pex.common import (
     Chroot,
     atomic_directory,
     chmod_plus_x,
+    filter_pyc_files,
     is_pyc_temporary_file,
     safe_copy,
     safe_delete,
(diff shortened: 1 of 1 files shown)


jsirois:
A grep over the codebase confirms this is the last spot that was bespoke; now everything uses this utility function.

Eric-Arellano (approved):

PR #1891, opened 2022-08-25
Make vcs zip handling more robust.

jsirois:
This change works with both current vendored Pip and Pip 22.2.2.

Work towards #1805

### DIFF
--- a/pex/pip/vcs.py
+++ b/pex/pip/vcs.py
@@ -5,16 +5,54 @@ from __future__ import absolute_import
 
 import hashlib
 import os
+import re
 
 from pex import hashing
 from pex.common import filter_pyc_dirs, filter_pyc_files, open_zip, temporary_dir
+from pex.pep_440 import Version
+from pex.pep_503 import ProjectName
 from pex.requirements import VCS
 from pex.resolve.resolved_requirement import Fingerprint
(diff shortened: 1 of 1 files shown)


jsirois:
This is part 1 of 3 in breaking the final ~1k delta lines into chunks for review.

pex/pip/vcs.py
```diff
@@ -50,7 +84,7 @@ def digest_vcs_archive(
     # `pip._internal.req.req_install.InstallRequirement.archive`.
     with TRACER.timed(
         "Digesting {archive} {vcs} archive".format(archive=os.path.basename(archive_path), vcs=vcs)
-    ), temporary_dir(cleanup=False) as chroot, open_zip(archive_path) as archive:
+    ), temporary_dir() as chroot, open_zip(archive_path) as archive:
```
Eric-Arellano, line 87: Why this change?

Eric-Arellano (approved):

pex/pip/vcs.py
jsirois: Unrelated to this change to be sure, but a long standing lack of cleanup from some bug hunt in the past that I spent 15 minutes scrutinizing for, in the end, no reason. No more!

PR #1893, opened 2022-08-25
Add support for Pip 22.2.2.

### LINKED ISSUE #1805: Pull in Pip fixes.

jsirois:
Pex can now support multiple Pip versions and be configured to fallback
as needed when newer Pip does not support a specified interpreter.

Fixes #1805

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,54 +22,38 @@ jobs:
       - name: Noop
         run: "true"
   checks:
-    name:  TOXENV=${{ matrix.tox-env }}
+    name:  TOXENV=format-check,typecheck,vendor-check,package -- --additional-format sdist --additional-format wheel
     needs: org-check
     runs-on: ubuntu-20.04
-    strategy:
-      matrix:
-        include:
-          - check-name: Formatting
-            python-version: "3.10"
-            tox-env: format-check
-            fetch-depth: 1
-          - check-name: Types
-            # N.B.: Python 3.10.6 introduces support for parenthesized context managers which
-            # breaks typechecking against --python-version 2.7 and so we pin low.
-            python-version: "3.10.5"
-            tox-env: typecheck
-            fetch-depth: 1
-          - check-name: Vendoring
-            python-version: "3.8"
-            tox-env: vendor-check
-            fetch-depth: 1
-          - check-name: Packaging
-            python-version: "3.10"
-            tox-env: package -- --additional-format sdist --additional-format wheel
-            # We need branches and tags since package leans on `git describe`. Passing 0 gets us
-            # complete history.
-            fetch-depth: 0
     steps:
       - name: Checkout Pex
         uses: actions/checkout@v2
         with:
-          fetch-depth: ${{ matrix.fetch-depth }}
-      - name: Setup Python ${{ matrix.python-version }}
+          # We need branches and tags since package leans on `git describe`. Passing 0 gets us
+          # complete history.
+          fetch-depth: 0
+      - name: Setup Python 3.8
         uses: actions/setup-python@v3
         with:
-          python-version: "${{ matrix.python-version }}"
-      - name: Check ${{ matrix.check-name }}
+          # We need to keep Python 3.8 for consistent vendoring with tox.
+          python-version: "3.8"
+      - name: Check Formatting, Types, Vendoring and Packaging
         uses: pantsbuild/actions/run-tox@95209b287c817c78a765962d40ac6cea790fc511
         with:
-          tox-env: ${{ matrix.tox-env }}
+          tox-env: format-check,typecheck,vendor-check,package -- --additional-format sdist --additional-format wheel
   cpython-unit-tests:
-    name: (${{ matrix.os }}) TOXENV=py${{ matrix.python-version[0] }}${{ matrix.python-version[1] }}
+    name: (${{ matrix.os }}) Pip ${{ matrix.pip-version }} TOXENV=py${{ matrix.python-version[0] }}${{ matrix.python-version[1] }}
     needs: org-check
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         python-version: [[2, 7], [3, 5], [3, 6], [3, 7], [3, 8], [3, 9], [3, 10], [3, 11, "0-rc.1"]]
         os: [ubuntu-20.04, macos-11]
+        pip-version: ["20", "22"]
         exclude:
+          - os: macos-11
+            python-version: [2, 7]
+            pip-version: "22"
           - os: macos-11
             python-version: [3, 5]
           - os: macos-11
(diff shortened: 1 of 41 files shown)


jsirois:
Lots of plumbing in this one that I found no obvious way to break up. I'll add pointers to the non-plumbing bits of interest.

pex/resolve/locker.py
```diff
@@ -201,6 +211,20 @@ def analyze(self, line):
                 self._saved.add(pin)
             return self.Continue()
 
+        match = re.search(
```
jsirois, line 214: This is the new PEP-691 interaction needed for Pip 22.2.2 since it no longer logs hashes along with URLs. The vendored Pip never triggers this code path because it does not log this "Fetched page ..." line.

pex/resolve/locker_patches.py
```diff
@@ -154,6 +154,11 @@ def supported_platform_tag(self, *_args, **_kwargs):
         check(*args, **kwargs) for check in supported_checks
     )
 
+    # N.B.: This patch is a noop for the 20.3.4-patched Pip but is required in newer Pip.
```
jsirois, line 157: This was the only change actually required to get Pip 22.2.2 working. The PEP-691 work was just needed to make locking comparably fast with vendored Pip (without PEP-691 each artifact is downloaded just so it can be hashed locally using 22.2.2).

jsirois:
A final apology in advance. This is largish and sprawling. The plumbing tendrils for the new pip-version go pretty wide and deep.

tox.ini
```diff
@@ -48,19 +58,13 @@ whitelist_externals =
...
 
-[testenv:py{py27-subprocess,py27,py35,py36,py37,py38,py39,27,35,36,37,38,39,310,311}-integration]
+[testenv:py{py27-subprocess,py27,py35,py36,py37,py38,py39,27,35,36,37,38,39,310,311}-{,pip20-,pip22-}integration]
 deps =
     pytest-xdist==1.34.0
     {[testenv]deps}
 commands =
+    {[_printenv]commands}
     pytest -n auto tests/integration {posargs:-vvs}
-setenv =
```
jsirois, line 57: A lesson we already know over in Pants. Users may use the build tool but they probably don't understand it. When I took up work on Pex after Wickman / Wilson I did not know tox. I still do not know tox as well as I should. I did enough to get by. I now, however, understand the tox "generative names" concept better and realized this was all un-needed dup.

.github/workflows/ci.yml
```diff
@@ -22,54 +22,38 @@ jobs:
       - name: Noop
         run: "true"
   checks:
-    name:  TOXENV=${{ matrix.tox-env }}
+    name:  TOXENV=format-check,typecheck,vendor-check,package -- --additional-format sdist --additional-format wheel
```
jsirois, line 25: Just helping clear space. All Combined this is ~5 minutes whereas the fastest test shards are 8. With the 3 job headspace cleared, the overall job count went from 23 -> 27. Previously 4 mac shards, now 5.

jsirois:
And I'll respond to any feedback from Windows. Wish me luck.

pex/resolve/downloads.py
```diff
@@ -45,10 +48,20 @@ def get_downloads_dir(pex_root=None):
...
     resolver = attr.ib()  # type: Resolver
-    pip = attr.ib(factory=get_pip)  # type: Pip
+    target = attr.ib(default=LocalInterpreter.create())  # type: Target
     package_index_configuration = attr.ib(
         default=PackageIndexConfiguration.create()
     )  # type: PackageIndexConfiguration
+    max_parallel_jobs = attr.ib(default=None)  # type: Optional[int]
+    pip = attr.ib(init=False)  # type: Pip
+
+    @pip.default
```
benjyw, line 58: Took me a little spelunking of the attrs docs to grok this, and particularly the use of the partially-initialized self, but it's neat!

pex/resolve/lock_resolver.py
```diff
@@ -20,6 +20,7 @@
 from pex.pip.local_project import digest_local_project
```
benjyw, line 20: The get_pip import above this line is not used

benjyw: (My IDE pointed it out, that is not some Jedi-level code grokking on my part...)

pex/resolve/lock_resolver.py
```diff
@@ -258,24 +269,25 @@ def resolve_from_lock(
     # errors under the right interleaving of processes and threads and download artifact targets.
     file_lock_style = FileLockStyle.BSD
 
-    # We eagerly initialize a Pip tool for reasons alluded to above, creation of the Pip tool venv
```
benjyw, line 261: I don't quite grok why this is no longer required as a result of this change. Perhaps it wasn't required even before this change? AFAICT the pip venv now gets created lazily in the ArtifactDownloader.pip factory, as it would have before?

benjyw (approved):
Big change, but the meaty part is relatively concentrated and most of the other files are just tentacles, so not too bad to review.

pex/resolve/downloads.py
jsirois: Yes, attrs is quite a bit nicer than dataclasses.

pex/resolve/lock_resolver.py
jsirois: Exactly, it was not required before. I was attempting to substitute clarity when what i really want is some more sane way to deal with this. The only requirement is that the Pip venv is built before the thread pool is used below and that was satisfied either by passing in a pre-built Pip in the obvious way here or via the default / factory used in `ArtifactDownloader`. What this is all hammering on about is that `ArtifactDownloader` used to just use `get_pip()` lazily in-line where it needed it; that was the real problem. The deeper problem is POSIX locks vs BSD. I could switch AtomicDirectory to just always use BSD and then the issue goes away, but as things stand I try to maintain compatibility with pre-existing users of PEX + NFS in all codepaths possible.

pex/resolve/lock_resolver.py
jsirois: Yeah - I generally hire the same ninja. I run IDEA's Optimize Imports over the whole change then `tox -eformat`. Missed it this time!

PR #1902, opened 2022-09-10
Fix PATH-style inputs to use `os.pathsep`.

jsirois:
In some places we assumed ':' which is hostile to bringing in Windows support with the ever-present 'C:\\...' paths.


benjyw (approved):
This is exactly the kind of change I'd expect when adding Windows support... I look forward to the one that replaces our assumptions about `/` vs `\`...

Changing the type of an existing var always terrifies me a little, without a compiler to check my work, but we have enough typechecking and testing that I assume all corner cases are covered.

jsirois:
Unfortunately, the types of changes you expect like this are just the easy ones!

Yeah, type checking is gravy and the thorough integration testing is the meat. Lots of test failures even with type check green, although these were mainly over-zealous stderr error message checking.

PR #1906, opened 2022-09-15
Use raw_pex_root when constructing sh_boot pexes.

### LINKED ISSUE #1881: sh_boot script has wrong $INSTALLED_PEX path in certain circumstances.
Problem:
Value of pex_root inside PEX-INFO of sh_boot-ed pex can vary from the value used in the sh_boot header script. This happens because the pex needs to be frozen prior to the sh_boot script being generated (so PexInfo.pex_hash is set, and the PEX-INFO file is written).

1. If the value of PexInfo.raw_pex_root is *not writable* during build time (common in a sandboxed build environment) the sh_boot script will change the value of raw_pex_root to a temp directory, and assume that a venv pex will be installed on the runtime host relative to this tmp_dir. But when the sh_boot-ed script runs for the first time the value of pex_root in PEX-INFO will be used as the pex_root meaning the low startup overhead of sh_boot cannot be accomplished because the sh_boot script will never see the INSTALLED_PEX exist.
2. The sh_boot-ed pex is run in an environment with a different $HOME value than build time. The user expansion of the pex_root at build time is different than the location of the pex_root at runtime. The mechanics of this discrepancy are the same as above when the $INSTALLED_PEX location is never created.

Possible Solution:
Don't try to infer the tilde-expansion of INSTALLED_PEX at build time. Embed the $INSTALLED_PEX location relative to the PexInfo.raw_pex_root. Trying to use the fully qualified PexInfo.pex_root property at build time is what causes the discrepancy between the sh boot header and PEX-INFO.

hrfuller:
Fixes #1881

Use of raw pex root means that `runtime_pex_root` and other pex_info
values are not mutated or expected to exist at build time. This makes sh_boot pexes
(zipapp and venv modes) portable across environments where the
`runtime_pex_root` may not exist in the build environment. We are
not taking relative paths to INSTALLED_PEX based on the user expanded
pex_root either. `pex_info.pex_root` is expanded only at runtime.
    
This also fixes an issue where `PEX(<pex_path>).pex_info()` would override
runtime_pex_root with the defaulted ENV.PEX_ROOT property.   By encapsulating the
sh_boot creation method in the PEXBuilder and using pex_info from
the PEXBuilder itself this no longer happens


pex/bin/pex.py
```diff
@@ -872,6 +868,9 @@ def seed_cache(
 
     pex_path = cast(str, options.pex_name)
     with TRACER.timed("Seeding local caches for {}".format(pex_path)):
+        # NB: This pattern causes the defaulted PEX_ROOT property to overwrite any runtime_pex_root in the written PEX-INFO.
```
jsirois, line 871: Seeding is necessarily a local operation / optimization. As far as I know, only Pants uses this to pre-seed PEX venvs as the PEXes are built for local venvs it uses to execute Python tools. The whole mechanism was exposed before `--sh-boot` and `--sh-boot` can proably replace it now.

With that background - most importantly the local bit - can your NB turn into a bug with an issue or should it go away entirely? I don't like the idea of an NB staying in the codebase that, by language, appears to point out a maybe bug. I'd like to complete the analysis and file a bug or realize it's not if possible.

pex/pex_builder.py
```diff
@@ -9,6 +9,7 @@
 import shutil
 
 from pex import pex_warnings
+from pex.bin.sh_boot import create_sh_boot_script
```
jsirois, line 12: `pex.pex.builder` -> `pex.bin` is not great. You'd generally want `pex.bin` (a main) dependending on library code but not vice-versa. Perhaps just a follow-up that just moves `sh_boot.py` into the main `pex` package.

pex/pex_builder.py
```diff
@@ -675,6 +677,28 @@ def build(
             safe_delete(path)
         os.rename(tmp_pex, path)
 
+    def set_sh_boot_script(
+        self,
+        pex_name,  # type: str
+        targets,  # type: Targets
+        python_shebang,  # type: Optional[str]
+    ):
+        if not self._frozen:
```
jsirois, line 686: Erm. Hopefully this warning never fires in Pex code today; if so I'd rather it be an assert / raise. Pex is not intended for use as a library, just as a stable CLI API. Of course I'd like to circle back and cleanup many APIs internally any how, like making this builder undego state tranitions by returning new types that expose the newly valid methods only when valid, but that's a 10% benefit thing and doesn't have the pull currently.

jsirois (approved):
Thanks for identifying and working through this Henry. The build-time vs runtime aspects of Pex are it most confusing aspects (for me anyhow), and the fact the code is not arranged to separate the two when possible exacerbates this.

A few not that important comments below that I'll leave as is and just merge to get a release out for this fix.

jsirois:
> A few not that important comments below that I'll leave as is and just merge to get a release out for this fix.

I just saw this was marked draft still so I'll hold off on a merge until we talk.

hrfuller:
Thanks @jsirois I'm happy to fix up based on the comments if I get to them before you push the button.

pex/bin/pex.py
hrfuller: Makes sense. I didn't understand the ramifications of changing seed pex / didn't want to fix it in this change. I can write up a bug and change to a TODO.

jsirois: Ok, if you're sure its actually a bug for the local case. Seeding is only useful if the local pex root is writeable.

pex/pex_builder.py
hrfuller: Sgtm.

pex/pex_builder.py
hrfuller: Will do. Fwiw I do use pex as a library and extend the pex builder. Generally it has been fine adapting to the fact that pex builder itself isn't guaranteed stable during updates.

jsirois: Aha - OK. You are warned at any rate. I definitely reserve the right to freely refactor all except the CLI.

jsirois: I would be interested in why the CLI isn't good enough though! If you want to relay some of that information via issues or offline channels, I'm definitely interested if its somethng more fundamental than legacy reasons.

pex/bin/pex.py
hrfuller: In that case perhaps not a bug.

pex/pex_builder.py
hrfuller: Sure, I would be happy to. Generally we use PEXBuilder to have fine grained control over the chroot path of source files. Adding whole directories isn't enough to handle all the subtleties of source roots for us.

hrfuller: opened #1907

jsirois (approved):

tests/integration/build_system/test_pep_517.py
```diff
@@ -25,9 +25,9 @@ def test_build_sdist_pyproject_toml(
     # The Poetry backend is important to supprt and the Poetry project dogfoods itself in its build.
     project_dir = clone(
         "https://github.com/python-poetry/poetry",
-        "8cb3aab3d0eaf5a25b3cf57e0cfc633231774524",
+        "7cd761a6a0bdec7fab76cf742ee4109db854ab76",
```
jsirois, line 28: I broke this fix out in #1908 to start a commit train.

PR #1914, opened 2022-09-20
Migrate PY27 pyenv to PY39.

jsirois:
This brings the set of canned pyenv interpreters to 2.7 and 3.{8,9,10},
paving the way for pyenv-win. PY27 use is minimized and made explicit
and lazy such that appropriate `pytest.skip()`s can be added in the
Windows port to skip over the handful of ramining integration tests that
truly mean to test against a Python 2.7 interpreter.

### DIFF
--- a/pex/testing.py
+++ b/pex/testing.py
@@ -463,11 +463,15 @@ def bootstrap_python_installer(dest):
 # N.B.: Make sure to stick to versions that have binary releases for all supported platforms to
 # support use of pyenv-win which does not build from source, just running released installers
 # robotically instead.
+PY27 = "2.7.18"
 PY38 = "3.8.10"
 PY39 = "3.9.13"
 PY310 = "3.10.7"
 
-ALL_PY_VERSIONS = (PY38, PY39, PY310)
+ALL_PY_VERSIONS = (PY27, PY38, PY39, PY310)
+_ALL_PY_VERSIONS_TO_VERSION_INFO = {
+    version: tuple(map(int, version.split("."))) for version in ALL_PY_VERSIONS
+}
 
 
 def ensure_python_distribution(version):
@@ -541,10 +545,17 @@ def ensure_python_venv(
     # type: (...) -> Tuple[str, str]
     _, python, pip, _ = ensure_python_distribution(version)
     venv = safe_mkdtemp()
-    args = [python, "-m", "venv", venv]
-    if system_site_packages:
-        args.append("--system-site-packages")
-    subprocess.check_call(args=args)
+    if _ALL_PY_VERSIONS_TO_VERSION_INFO[version][0] == 3:
+        args = [python, "-m", "venv", venv]
+        if system_site_packages:
+            args.append("--system-site-packages")
+        subprocess.check_call(args=args)
+    else:
+        subprocess.check_call(args=[pip, "install", "virtualenv==16.7.10"])
+        args = [python, "-m", "virtualenv", venv, "-q"]
+        if system_site_packages:
+            args.append("--system-site-packages")
+        subprocess.check_call(args=args)
     python, pip = tuple(os.path.join(venv, "bin", exe) for exe in ("python", "pip"))
     if latest_pip:
         subprocess.check_call(args=[pip, "install", "-U", "pip<22.1"])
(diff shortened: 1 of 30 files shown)


pex/venv/virtualenv.py
```diff
@@ -84,12 +84,19 @@ def find_site_packages_dir(
...
-    for path in OrderedSet(os.path.realpath(entry) for entry in interpreter.sys_path):
-        _, tail = os.path.split(path)
-        if "site-packages" == tail and os.path.isdir(path):
-            site_packages_dirs.append(path)
+    for entry in interpreter.sys_path:
+        real_entry_path = os.path.realpath(entry)
+        if os.path.commonprefix((real_venv_dir, real_entry_path)) != real_venv_dir:
+            # This ignores system site packages when the venv is built with --system-site-packages.
+            continue
+        if "site-packages" == os.path.basename(real_entry_path) and os.path.isdir(real_entry_path):
```
jsirois, line 97: @hrfuller the promised `os.path.split` -> `os.path.basename`.

pex/testing.py
```diff
@@ -551,6 +557,122 @@ def ensure_python_interpreter(version):
...
+    CPython = Value("CPython")
+    PyPy = Value("PyPy")
+
+
+def find_python_interpreter(
+    version=(),  # type: Tuple[int, ...]
+    implementation=InterpreterImplementation.CPython,  # type: InterpreterImplementation.Value
+):
+    # type: (...) -> Optional[str]
+    for interpreter in PythonInterpreter.iter():
```
jsirois, line 573: The Windows support branch will add a skip here for `version < (3, 8)` since Windows Python 3.7 doesn't have long path support enabled in its manifest and lower versions don't have symlink support. This will get Windows running all the exisitng ITs it can that do not require Python 2.7.

pex/venv/virtualenv.py
```diff
@@ -84,12 +84,19 @@ def find_site_packages_dir(
...
     interpreter = interpreter or PythonInterpreter.get()
-    site_packages_dirs = []
-    for path in OrderedSet(os.path.realpath(entry) for entry in interpreter.sys_path):
-        _, tail = os.path.split(path)
-        if "site-packages" == tail and os.path.isdir(path):
-            site_packages_dirs.append(path)
+    for entry in interpreter.sys_path:
+        real_entry_path = os.path.realpath(entry)
+        if os.path.commonprefix((real_venv_dir, real_entry_path)) != real_venv_dir:
+            # This ignores system site packages when the venv is built with --system-site-packages.
```
hrfuller, line 95: nice! Alternatively `if not real_entry_path.startswith(real_venv_dir)` is a bit more readable, but maybe has some edge cases that the os.path modules would handle.

pex/venv/virtualenv.py
```diff
@@ -130,6 +137,7 @@ def create(
         interpreter=None,  # type: Optional[PythonInterpreter]
         force=False,  # type: bool
         copies=False,  # type: bool
+        system_site_packages=False,  # type: bool
```
hrfuller, line 140: When are system_site_packages relevant if we are ignoring them above in find_site_packages_dir?

pex/testing.py
```diff
@@ -525,20 +533,18 @@ def run_pyenv(args):
...
-        args = [python, "-m", "venv", venv]
-        if system_site_packages:
-            args.append("--system-site-packages")
-        subprocess.check_call(args=args)
-    else:
-        subprocess.check_call(args=[pip, "install", "virtualenv==16.7.10"])
-        args = [python, "-m", "virtualenv", venv, "-q"]
-        if system_site_packages:
-            args.append("--system-site-packages")
-        subprocess.check_call(args=args)
```
hrfuller, line 541: Does removing this mean python27 doesn't have venv support?

hrfuller (approved):
LGTM.

pex/venv/virtualenv.py
jsirois: Well, you hit something that has always bothered me. What you suggested is not good since `a/bad/match` starts with `a/b` but the real common prefix is `a`. That said - what I'm using does this same badness!:
```
jsirois@Gill-Windows:~/dev/pantsbuild/jsirois-pex (main) $ python2.7 -c 'import os; print(os.path.commonprefix(["a/b", "a/bad/match"]))'
a/b
jsirois@Gill-Windows:~/dev/pantsbuild/jsirois-pex (main) $ python3.11 -c 'import os; print(os.path.commonprefix(["a/b", "a/bad/match"]))'
a/b
```

I just sort of obstinately use it since some from `os.path` should really do the right thing. What I really should do is write a utility and re-use that throughout Pex: https://github.com/pantsbuild/pex/issues/1916

pex/venv/virtualenv.py
jsirois: There is a pre-existing test that uses this new functionality (previously a pyenv-based utility in `pex/testing.py` was used to builr venvs with `--system-site-packages`). That test is an important test that fixed a real bug found in the wild. So just to support that - its not used by production code and I think never will be. Pex is all about hermetic and `--system-site-packages` is all about not!

pex/testing.py
jsirois: Nope, it still does, but the Python 2.7 has to be found on your machine pre-installed. That's these additions: https://github.com/pantsbuild/pex/pull/1914/commits/24ca825fa5f59e2564d393c1dbb88588138bcc2f#diff-79ebd789ddd6317a85b1ccbcddcbdefc0ce8b07ed30066d9bc58805d1c22f83bR596-R622

pex/venv/virtualenv.py
jsirois: It looks like the fix came in 3.5: https://docs.python.org/3/library/os.path.html#os.path.commonpath

PR #1926, opened 2022-10-03
Add a Zip utility for manipulating zip headers.

jsirois:
This is not needed by Pex mainline but is needed to support Windows
console scripts which are realized as a sandwich of a windows PE stub,
a shebang line and a zipapp trailer. In this arrangement, the stub
executable header reads the zip trailer to find the sandwiched shebang
and then uses that shebang to pick a Python interpreter to run the
sandwich with. Pex needs to customize the sandwiched shebang just like
it needs to customize unix console script header shebangs so that the
venv interpreter is used to run the console script.

### DIFF
--- a/pex/common.py
+++ b/pex/common.py
@@ -227,6 +227,11 @@ def open_zip(path, *args, **kwargs):
 
     Passes through positional and kwargs to zipfile.ZipFile.
     """
+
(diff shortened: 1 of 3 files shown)


jsirois:
All right, done fiddling, good to go.

benjyw (approved):
It's been a while since I've encountered the internals of zip...

PR #1937, opened 2022-10-05
Vendor latest packaging; support musllinux wheels.

### LINKED ISSUE #1933: pex does not support musllinux wheels
[PEP 656](https://peps.python.org/pep-0656/) introduced the `musllinux` tag to support Alpine Linux and other cases where CPython does not use glibc.

I have confirmed that using python3 and pip in a venv a `musllinux` wheel is downloaded.
```
pip install 'psycopg2-binary'
Collecting psycopg2-binary
  Downloading psycopg2_binary-2.9.3-cp310-cp310-musllinux_1_1_x86_64.whl (1.9 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 1.9/1.9 MB 8.5 MB/s eta 0:00:00
Installing collected packages: psycopg2-binary
Successfully installed psycopg2-binary-2.9.3
WARNING: You are using pip version 22.0.4; however, version 22.2.2 is available.
You should consider upgrading via the '/root/venv/bin/python3 -m pip install --upgrade pip' command.
```

However using pex results in:
```
pex --pip-version 22.2.2 --resolver-version pip-2020-resolver --no-build 'psycopg2-binary'
Failed to find compatible interpreter on path /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin.

1.) /usr/bin/python3.10 CPython==3.10.5

  A distribution for psycopg2-binary could not be resolved for /usr/bin/python3.10.
  Found 1 distribution for psycopg2-binary that do not apply:
  1.) The wheel tags for psycopg2-binary 2.9.3 are cp310-cp310-musllinux_1_1_x86_64 which do not match the supported tags of /usr/bin/python3.10:
  cp310-cp310-linux_x86_64
  ... 34 more ...
```

To reproduce the above run the following to get alpine linux with python3 in docker.
```
$ docker run --rm -it alpine:3.16
# apk update
# apk add python3
```

I suspect that this is related to how the vendored packaging library is `20.9` but the project added support for musllinux wheels in `21.0`. Unfortunately that release dropped support for Python 2 as well.

A possible fix would be to have `--pip-version 22.2.2` also imply a newer version of the `packaging` library.

jsirois:
Pex now vendors two versions of packaging and the correct one is chosen
at runtime by the `pex.third_party` vendor importer. This allows newer
Pythons to resolve the latest packaging specifications.

Fixes #1933.


pex/pex_builder.py
```diff
@@ -495,17 +495,15 @@ def add_dist_location(
...
             for label in ("source", "executable", "main", "bootstrap")
             for path in self._chroot.filesets.get(label, ())
             if path.endswith(".py")
-            # N.B.: This file if Python 3.6+ only and will not compile under Python 2.7 or
-            # Python 3.5. Since we don't actually use it we just skip compiling it.
-            and path
-            != os.path.join(
-                self._pex_info.bootstrap, "pex/vendor/_vendored/attrs/attr/_next_gen.py"
-            )
+            # N.B.: Some of our vendored code does not work with all versions of Python we support;
```
benjyw, line 504: Previously we were excluding a single file, now we're excluding all the vendored boostrap code en masse. Will this not be a perf hit that belies the promise of bytecode_compile=True? 

Would it make sense (in a followup, if at all) to best-effort this and ignore any errors? Or is the performance hit all in the noise anyway? The code in _COMPILER_MAIN already splits out successes and errors, and it would just mean propagating that to the invoker.

benjyw (approved):

pex/pex_builder.py
jsirois: It had become irrelevant. The only time the bootstrap code is executed from within a zipped PEX is on 1st cold start, after that the PEX runs from either a PEX_ROOT venv or a PEX_ROOT unzipped_pexes directory. In either of those cases pycs are compiled and stick around. As such microoptimizing the known slow case of a cold start just didn't seem close to worth it.

benjyw: Makes sense!

jsirois: Ok, the colde-start is surprisingly bad - I'll dig in to see why before landing this:
```
$ pex -oold-empty.pex
$ pex --compile -oold-empty-compiled.pex
$ python -mpex -onew-empty.pex
$ python -mpex --compile -onew-empty-compiled.pex

$ hyperfine -p 'rm -rf ~/.pex' './old-empty.pex -c ""' './old-empty-compiled.pex -c ""' './new-empty.pex -c ""' './new-empty-compiled.pex -c ""'
Benchmark 1: ./old-empty.pex -c ""
  Time (mean ± σ):     259.5 ms ±   3.5 ms    [User: 235.1 ms, System: 25.6 ms]
  Range (min … max):   253.3 ms … 262.7 ms    10 runs

Benchmark 2: ./old-empty-compiled.pex -c ""
  Time (mean ± σ):     277.8 ms ±  15.7 ms    [User: 249.7 ms, System: 27.8 ms]
  Range (min … max):   265.4 ms … 317.5 ms    11 runs

Benchmark 3: ./new-empty.pex -c ""
  Time (mean ± σ):     411.5 ms ±   8.3 ms    [User: 381.9 ms, System: 30.7 ms]
  Range (min … max):   403.3 ms … 431.6 ms    10 runs

Benchmark 4: ./new-empty-compiled.pex -c ""
  Time (mean ± σ):     423.0 ms ±   6.5 ms    [User: 395.0 ms, System: 29.0 ms]
  Range (min … max):   413.9 ms … 433.5 ms    10 runs

Summary
  './old-empty.pex -c ""' ran
    1.07 ± 0.06 times faster than './old-empty-compiled.pex -c ""'
    1.59 ± 0.04 times faster than './new-empty.pex -c ""'
    1.63 ± 0.03 times faster than './new-empty-compiled.pex -c ""'

$ hyperfine -w 2 './old-empty.pex -c ""' './old-empty-compiled.pex -c ""' './new-empty.pex -c ""' './new-empty-compiled.pex -c ""'
Benchmark 1: ./old-empty.pex -c ""
  Time (mean ± σ):     170.6 ms ±   3.6 ms    [User: 158.6 ms, System: 13.3 ms]
  Range (min … max):   164.7 ms … 175.6 ms    17 runs

Benchmark 2: ./old-empty-compiled.pex -c ""
  Time (mean ± σ):     172.1 ms ±   3.3 ms    [User: 157.8 ms, System: 15.6 ms]
  Range (min … max):   167.0 ms … 179.5 ms    17 runs

Benchmark 3: ./new-empty.pex -c ""
  Time (mean ± σ):     170.8 ms ±   3.6 ms    [User: 158.3 ms, System: 13.8 ms]
  Range (min … max):   166.2 ms … 176.5 ms    17 runs

Benchmark 4: ./new-empty-compiled.pex -c ""
  Time (mean ± σ):     174.2 ms ±   3.7 ms    [User: 157.6 ms, System: 17.9 ms]
  Range (min … max):   167.8 ms … 178.6 ms    17 runs

Summary
  './old-empty.pex -c ""' ran
    1.00 ± 0.03 times faster than './new-empty.pex -c ""'
    1.01 ± 0.03 times faster than './old-empty-compiled.pex -c ""'
    1.02 ± 0.03 times faster than './new-empty-compiled.pex -c ""'
```

zmanji:
Should this diff delete `_vendored/packaging` as well?

tests/integration/test_issue_1933.py
```diff
@@ -0,0 +1,63 @@
...
+from pex.testing import run_pex_command
+from pex.typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from typing import Any
+
+
+@pytest.mark.skipif(
+    not any(
+        is_exe(os.path.join(entry, "docker"))
```
hrfuller, line 19: `shutil.which` would be easier to read. It was introduced in cp3.3, though it looks like we don't need to run this test with 2.7

jsirois: It may be, but actually using means a conditional import / it would be cleanest to provide a `pex.compatibility.which` shim, but that's for prod and this is only done in tests. At the end of the day, since Pex supports 2.7 the right thing to do is accept that and write in 2.7. I do this daily and flip the Python stdlib webdite api docs drop-down back to 2.7 to make sure I'm using APIs that I can actually use safely.

pex/interpreter.py
```diff
@@ -152,7 +154,7 @@ def get(cls, binary=None):
     def decode(cls, encoded):
         TRACER.log("creating PythonIdentity from encoded: %s" % encoded, V=9)
         values = json.loads(encoded)
-        if len(values) != 11:
+        if len(values) != 12:
```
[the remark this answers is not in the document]
jsirois: It would take some fancy introspection of the fields -> field count of PythonIdentity. That would be easy it it were `@attr.s`'d, but its not yet. If PythonIdentity / PythonInterpreter can be `@attr.s`'d without harming PEX bootstrap time perf, this would be a good idea. I'm not sure about the perf aspect though - I assume that's why I have not done this yet.

tests/integration/test_issue_1933.py
hrfuller: I'll stop beating this drum :)

PR #1938, opened 2022-10-05
Don't treat `PEX_PATH=` as `.` like other PATHS.

### LINKED ISSUE #1936: Empty string PEX_PATH="" env var causes CWD (.) to be added bootstrapped pex_path
The CWD doesn't usually contain a pex so the the pex being executed fails to bootstrap with a pex_info load error looking for a file `./PEX_INFO`.

Ran into this issue when upgrading from 2.1.97 -> 2.1.105. Was able to work around it by unsetting PEX_PATH entirely, but if this is a regression I can look into the root cause. Is this an intentional change?

jsirois:
This fixes a regression reading the `PEX_PATH` variable introduced by
changes in #1902 that ended up attempting to add CWD as a PEX when 
`PEX_PATH` was set to the empty string in the environment.

Since `PEX_PATH` is closest in spirit to a `PATH` of files instead of a
`PATH` of directories like `PATH` and `PYTHONPATH`, the old relied-upon
behavior makes sense generally.

Fixes #1936

### DIFF
--- a/pex/variables.py
+++ b/pex/variables.py
@@ -301,10 +301,16 @@ class Variables(object):
                 % (variable, self._environ[variable])
             )
 
-    def _maybe_get_path_tuple(self, variable):
-        # type: (str) -> Optional[Tuple[str, ...]]
+    def _maybe_get_path_tuple(
+        self,
+        variable,  # type: str
+        empty_string_is_cwd=True,  # type: bool
+    ):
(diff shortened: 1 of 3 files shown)


jsirois:
Both new tests fail with the fix reverted (new `bool` flipped).

hrfuller (approved):
Thanks for the quick fix!

PR #1951, opened 2022-10-18
Fix lock resolution to handle arbitrary equality.

### LINKED ISSUE #1949: Pex fails to resolve arbitrary equality requirements when candidate package uses unnormalized version number
I'm trying to deal with a project that uses an internal fork of a package with version number 2.49.0a1. Unfortunately, this is not a canonicalized version number. The canonical form is 2.49a1. As a result, Pex fails to deal with the lockfile generated by Pants because it tries to match the canonical form 2.49a1 against the exact specifier `===2.49.0a1`.

Although the best solution is probably to avoid adding package versions not in canonical form, it wouldn't hurt if Pex could handle this use case. A couple options:

1. Convert every specifier to canonical form. I'm not sure if this is viable in general beyond exact equality specifiers.
2. Test only the raw version string against the specifier to check for a match.
3. Test both the normalized and raw version strings against the specifier to check for a match.

The latter solution would look like this

```diff
diff --git a/pex/resolve/locked_resolve.py b/pex/resolve/locked_resolve.py
index 8064058..669f76d 100644
--- a/pex/resolve/locked_resolve.py
+++ b/pex/resolve/locked_resolve.py
@@ -501,7 +501,6 @@ class LockedResolve(object):
         include_all_matches=False,  # type: bool
     ):
         # type: (...) -> Union[Resolved, Error]
-
         is_local_interpreter = isinstance(target, LocalInterpreter)
         if not use_wheel:
             if not build:
@@ -605,9 +604,11 @@ class LockedResolve(object):
                     # https://peps.python.org/pep-0440/#handling-of-pre-releases during the lock
                     # resolve; so we trust that resolve's conclusion about prereleases and are
                     # permissive here.
-                    if not resolve_request.requirement.specifier.contains(
-                        str(locked_requirement.pin.version), prereleases=True
-                    ):
+                    def specifier_contains(version_str):
+                        # type: (str) -> bool
+                        return resolve_request.requirement.specifier.contains(version_str, prereleases=True)
+                    version = locked_requirement.pin.version
+                    if not specifier_contains(version.normalized) and not specifier_contains(version.raw):
                         version_mismatches.append(
                             "{specifier} ({via})".format(
                                 specifier=resolve_request.requirement.specifier,
```

jsirois:
Previously, arbitrary equality could not be used in requirements when resolving against a lockfile when the project being resolved had a PEP-440 compatible version that was normalized.

Fixes #1949

### DIFF
--- a/tests/integration/test_issue_1949.py
+++ b/tests/integration/test_issue_1949.py
@@ -1,0 +1,107 @@
+# Copyright 2022 Pants project contributors (see CONTRIBUTORS.md).
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+import os.path
+
+import pytest
+
+from pex.cli.testing import run_pex3
+from pex.compatibility import commonpath
+from pex.testing import IntegResults, built_wheel, run_pex_command
+from pex.typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from typing import Any, Iterator
+
+
+@pytest.fixture
+def wheel():
+    # type: () -> Iterator[str]
+    with built_wheel(name="not_boto", version="2.49.0a1") as whl:
+        yield whl
+
+
+def test_resolve_arbitrary_equality(
+    tmpdir,  # type: Any
+    wheel,  # type: str
+):
+    # type: (...) -> None
+
+    lock = os.path.join(str(tmpdir), "lock.json")
+    run_pex3(
+        "lock",
+        "create",
+        "--no-pypi",
+        "--find-links",
+        os.path.dirname(wheel),
+        "not_boto===2.49.0a1",
+        "--indent",
+        "2",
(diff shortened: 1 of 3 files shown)


tests/integration/test_issue_1949.py
```diff
@@ -0,0 +1,107 @@
...
+                "--runtime-pex-root",
+                pex_root,
+                "--lock",
+                lock,
+            ]
+            + list(additional_args)
+            + ["--", "-c", "import not_boto; print(not_boto.__file__)"]
+        )
+
+    result = create_pex_from_lock()
```
jsirois, line 60: This test and the next fail before the fix in the same way as reported in #1949.

benjyw (approved):

PR #1961, opened 2022-10-20
Re-work atomic_directory locking for faster / clearer failures.

jsirois:
Change `atomic_directory` to always grab an exclusive lock and use a
stable work directory per target directory to surface multiple lock
owners as up-front errors instead of possibly slow corruptions.

### DIFF
--- a/pex/atomic_directory.py
+++ b/pex/atomic_directory.py
@@ -1,0 +1,164 @@
+# Copyright 2022 Pants project contributors (see CONTRIBUTORS.md).
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+from __future__ import absolute_import
+
+import errno
+import fcntl
+import os
+from contextlib import contextmanager
+from uuid import uuid4
+
+from pex.common import safe_mkdir, safe_rmtree
+from pex.enum import Enum
+from pex.typing import TYPE_CHECKING, cast
+
+if TYPE_CHECKING:
+    from typing import Callable, Iterator, Optional, Union
+
+
+class AtomicDirectory(object):
+    def __init__(self, target_dir):
+        # type: (str) -> None
+        self._target_dir = target_dir
+        self._work_dir = "{}.{}".format(target_dir, uuid4().hex)
+
+    @property
+    def work_dir(self):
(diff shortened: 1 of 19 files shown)


jsirois:
The 1st commit is a pure re-name such that only the subsequent commits need be looked at to see changes to the locking code and call sites.

jsirois:
I tested this over in Pants with `./pants fmt lint check test --force src/python:: tests/python::` after removing the named caches dir. For what that's worth with the elusive-to-repro missing files bugs, I hit no errors of that sort.

benjyw (approved):
Does this mean we can get rid of `FileLockStyle.BSD` ?

jsirois:
> Does this mean we can get rid of `FileLockStyle.BSD` ?

No. That's still needed in two cases where thread pools are used. Posix locks do not work on their own in the presence of threads  for several reasons.

jsirois:
If it's not clear, this PR fixes nothing and O can still spot no bugs before or after this change. I'm only aiming to make what was, afaict, correct code, simpler. The additional change of not using a uuid suffix for the workdir just doubles down on the correctness assertion asking for a clear error up front to prove something is somehow wrong with the code.

pex/atomic_directory.py
```diff
@@ -85,80 +84,80 @@ class Value(Enum.Value):
...
+        try:
+            yield atomic_dir
+        finally:
+            unlock()
+        return
 
+    # If there is an error making the work_dir that means file-locking guarantees have failed
+    # somehow and another process has the lock and has made the work_dir already. We let the error
+    # from os.mkdir propagate in that case.
+    os.mkdir(atomic_dir.work_dir)
```
stuhood, line 154: It could also mean that something segfaulted or was sigkilled, such that it exited uncleanly without tearing down the working directory?

jsirois: Yes, that's true. I'll step back and think about this a bit.

jsirois: Ok, I just warn now to ensure we get stderr data but clean up stale workdirs to prevent a stop-in-your-tracks bug that requires manual intervention.

Issue #1964, opened 2022-10-24
Allow to set multiple folders for PEX_ROOT


manuzhang:
It will be helpful to set multiple folders for PEX_ROOT in case one folder is not writable due to permission or disk full.

jsirois:
Can you explain a bit more? Do you have situations where building the PEX file with `--runtime-pex-root X` is not enough? If so, can you provide more details about your deployment targets that lead to this need?

manuzhang:
On our cluster, we usually have two disks on a node and we won't decommission the node if only one disk is full or unhealthy. Say the two disks are mounted at `/disk1` and `/disk2`, it will be great if we can set `PEX_ROOT` to `/disk1,/disk2` and pex will find the first writable folder.

jsirois:
Aha, ok. So the PEX runtime tries `os.mkdir` on `/disk1` which nets an `OSError` (or `IOError`) due to a disk full condition and then falls back to `/disk2`. That makes sense to me. This is all self contained in `pex.pex-info.PexInfo` class except for the `--runtime-pex-root` plumbing which is in `pex/bin/pex.py`.

@manuzhang are you willing to contribute this feature?

manuzhang:
@jsirois thanks, I've submitted https://github.com/pantsbuild/pex/pull/1965. Please help approve to run workflow and review.

manuzhang:
@jsirois Our usage is for PySpark applications(https://spark.apache.org/docs/3.1.1/api/python/user_guide/python_packaging.html#using-pex) and pex root is set like `spark.executorEnv.PEX_ROOT=/disk1/pex`. `/disk1` is shared by many applications besides pex and can become full. For other applications, the node is managed by YARN, which has a [health checker](https://hadoop.apache.org/docs/r2.7.3/hadoop-yarn/hadoop-yarn-site/NodeManager.html) monitoring the disk state. It's configured with two disks and won't assign application to the node when both disks are unhealthy.

I'm thinking about adding similar feature to pex.

jsirois:
OK, thanks for the background. The health checker approach is the right approach! You need a daemon or a cron job for that. Pex is neither, it's a one-shot job. Its very hard to solve this with a one-shot job is what I'm trying to point out. That's not the right layer to solve this problem. The best way I can think of to solve this problem for a one shot job is to make sure it uses a wrapper that does something like (psuedo code - this could be a bash wrapper or even a python wrapper):
```
pex_roots = [...]
[...]

PR #1965, opened 2022-10-25
Allow to set multiple dirs for pex root

manuzhang:


pex/pex_info.py
```diff
@@ -451,11 +451,18 @@ def raw_pex_root(self):
...
+        for path in paths:
+            if can_write_dir(path):
+                pex_root = path
+                writable = True
+                break
+        if not writable:
             tmp_root = safe_mkdtemp()
             pex_warnings.warn(
-                "PEX_ROOT is configured as {pex_root} but that path is un-writeable, "
+                "PEX_ROOT is configured as {pex_root} but all paths are un-writeable, "
```
Eric-Arellano, line 465: This might have `pex_root` as `None`. Maybe it should be the `raw_pex_root`?

pex/pex_info.py
```diff
@@ -451,11 +451,18 @@ def raw_pex_root(self):
     @property
     def pex_root(self):
         # type: () -> str
-        pex_root = os.path.expanduser(self.raw_pex_root)
-        if not can_write_dir(pex_root):
+        paths = os.path.expanduser(self.raw_pex_root).split(",")
```
Eric-Arellano, line 454: I think you probably want to first `split`, then call `os.path.expanduser` on each element?

Eric-Arellano (commented):
Thanks!

pex/pex_info.py
```diff
@@ -451,11 +451,18 @@ def raw_pex_root(self):
...
+        pex_root = None
+        writable = False
+        for path in paths:
+            if can_write_dir(path):
```
jsirois, line 458: Your use case specifically mention full or unhealthy disks as the reason for wanting fallback PEX_ROOT values. This test tests neither of those things. The implication of this PR is that you manually - or via some automation on your nodes - change filesystem permissions on disks that have been determined to disk full or have gone unhealthy. My expecatation was that you'd be changing the test here to actually attempt a mkdir to detect disk full.

If you are in fact doing the permission change on your nodes then you could just be using a PEX rc file on those nodes and changing that file too. This [documentation ](https://pex.readthedocs.io/en/v2.1.112/buildingpex.html?highlight=pexrc#tailoring-pex-execution-at-build-time) is only partly correct but the CLI help, which should always be your primary reading material, is accurate:
```
$ pex --help | grep -A5 pexrc
  --rcfile RC_FILE      An additional path to a pexrc file to read during
                        configuration parsing, in addition to reading
                        `/etc/pexrc` and `~/.pexrc`. If
                        `PEX_IGNORE_RCFILES=true`, then all rc files will be
                        ignored. (default: None)
```

So you could just drop in an `/etc/pexrc ` ith contents:
```
PEX_ROOT=/disk1/.pex
# PEX_ROOT=/disk2/.pex
```
And edit the file when `/disk1` fails.

pex/pex_info.py
jsirois: If `/etc/pexrc` (see comment below) is not the answer here, the option really out to be changed to a list option up in `pex/bin/pex.py` and stored as a list here in `PexInfo` too. The, only on read of the value is the list whittled down to one value or a tmp dir if all else fails. There is no reason to court escaping issues when you don't have to in the first place: paths can validly contain commas in them on some systems.

pex/pex_info.py
manuzhang: We are not doing permission change. All user can configure is `PEX_ROOT` and currently user application would fail immediately if there is a disk issue. As I explained in https://github.com/pantsbuild/pex/issues/1964, we won't fix the node unless all disk are unhealthy.

pex/pex_info.py
manuzhang: Sure. Let me try this way.

pex/pex_info.py
jsirois: @manuzhang I'm confused. I have not tested this, but if a disk is full and it has a directory you have write permissions to that was established prior to the disk full condition, I'm pretty sure a permissions check will pass; but later, when you go to use that directory to write files into, those writes will fail - not due to a permissions issue, but due to a disk full. If that's right, that's why I suggest this PR is confusing, since to test for both write permissions and not disk full, you'd have to change the test to an `os.mkdir`. Even after doing that though - since the directory may have been created long before disk full, the `os.mkdir` will no-op successfully and you will still not have tested the disk full condition!

pex/pex_info.py
jsirois: If you mean the environment variable (`Variables.PEX_ROOT`), I think that can remain singular. If you're exporting an environment variable you're exporting it on a given host. If that host has a disk full condition, you'd just change that environment variable manually. This is what you can do via `/etc/pexrc`.

When you think about the details here - unless I'm missing something, a disk can go full at any time, even in between when Pex 1st checks for the condition and later when it goes to write files during a single run of a PEX. You can never fully avoid that race condition. You can check early in a PEX run whether a disk is _already_ full - which I think is what you're trying to do - but that's still best effort if the answer is no. Now as far as getting an answer yes - the disk is full - how do you propose that is done? I have no clue how you do this except by something hacky like: assume the system has the `df` binary on the path and use it to get the current full state of the disks on the system and then map the given pex root to one of those disks by checking the mount prefix that contains it. That seems super hacky. Do you have some other idea?

pex/pex_info.py
manuzhang: @jsirois Permission issue and disk full are two cases. I didn't check the details of `can_write_dir` and thought it returned false whenever the disk couldn't be written to.  Thanks for patiently explain the race condition. Now I understand it and think it would be better to fall back on exceptions. WDYT?

pex/pex_info.py
manuzhang: How about adding a variable like `PEX_ROOT_FALLBACK` that we can fall back to on exceptions?

pex/pex_info.py
jsirois: I think you still need to spell out some details. How do you propose you trigger those exceptions when calculating which pex root to use from the list of pex roots? Do you try to write a 1 byte file and then delete it if successful and proceed? Do you try to write a file that's bigger than that to attempt to ensure the whole PEX run will succeed - if so, what size do you guess at? Say the guess is 100MB. That's pretty big so writing a file full of 0s - say - and then deleting it will take some time. I wouldn't want to force that time - say its 50ms on all users. Pants for one would certainly not like this at all!

You have a very thorny problem here in my opinion to attempt to solve automatically. If this were my problem, from the information you've given so far, I'd personally write a cron job that pre-emptively checked for disk health / impending disk full - lets say every 10 minutes - and I'd have the job re-write `/etc/pexrc` if it determined a disk was going bad or getting within some threshold of disk full.

pex/pex_info.py
jsirois: The problem is the pex root is accessed all over the place in code. You'd need to wrap every one of those in a try / except. Like I said, the condition of a disk is racy. It can go bad or full at any time, even in the middle of a Pex run.

pex/pex_info.py
manuzhang: We don't trigger those exceptions or detect disk issues proactively. I mean we fall back on user exceptions.

jsirois: I understand that you want to do that, I'm suggesting its nearly impossible. Pro-actively managing a machine is not. That you can do with a cron job like I suggested or via other means. This is a very standard thing to do! Many machines have /tmp cleaner cron jobs for example to clear up disk (or memory) space used by old temp files for just this sort of reason.

jsirois: Maybe it makes sense for you to really detail the exact use case fully on the ticket. Whose machines are these? Can you install cron jobs or can you not for some reason? Why do you encounter disk fulls? How often do you encounter bad disks? Etc...

manuzhang: I've added more details in the issue comments.

jsirois:
I've closed #1964 since it seems like you've moved on to using a higher level wrapper around your PEX to detect disk full conditions in the Spark cluster. As such, I'll close this PR as well.

PR #1972, opened 2022-11-02
Map pex python path interpreter to realpath when creating venv dir hash.

### LINKED ISSUE #1885: venv_dir calculation doesn't correctly handle PEX_PYTHON_PATH with symlinks.
I noticed a bug in the variables.venv_dir hash computation. If PEX_PYTHON_PATH has the path of a symlinked interpreter that interpreter value isn't added to the hash of the venv_dir. This can cause collisions in which interpreter is used to resolve dependencies via resolver.resolve. This happens because the canonical path to the interpreter is computed but the PEX_PYTHON_PATH is assumed to have canonical paths. [line in  question](https://github.com/pantsbuild/pex/blob/8c14e563493b8f4a85b4a3c735596dd55586a52b/pex/variables.py#L748)

I think it would be reasonable to map the PEX_PYTHON_PATH entries to their realpath's before comparing with the requested venv interpreter.

hrfuller:
fixes #1885 

Problem:
When the PEX_PYTHON_PATH contains paths to allowed interpreters that are symlinks to the real interpreter the venv_dir hash value doesn't end up including a valid interpreter. The is a particular problem for the pip.pex venvs that are created by `resolver.resolve` because those pip.pex have no interpreter constraints. they rely on the venv interpreter being the correct one matching the resolve target. If an existing pip.pex venv exists with the wrong venv interpterer the venv_dir paths can collide resulting in the `ensure_venv` returning a path to a pip.pex with the incorrect interpreter, and thus with different resolve behavior.

Solution:
Make sure that when computing the venv_dir if the PEX_PYTHON_PATH contains symlinks to interpreters those are resolved down to their realpath ensuring that the venv_dir path doesn't collide between different interpreters on the PEX_PYTHON_PATH.

### DIFF
--- a/pex/vendor/_vendored/attrs/attrs-21.5.0.dev0.dist-info/WHEEL
+++ b/pex/vendor/_vendored/attrs/attrs-21.5.0.dev0.dist-info/WHEEL
@@ -1,5 +1,5 @@
 Wheel-Version: 1.0
-Generator: bdist_wheel (0.37.1)
+Generator: bdist_wheel (0.38.3)
 Root-Is-Purelib: true
 Tag: py2-none-any
 Tag: py3-none-any
--- a/pex/variables.py
+++ b/pex/variables.py
@@ -773,7 +773,15 @@ def venv_dir(
     if interpreter_path:
         venv_contents["interpreter"] = os.path.realpath(interpreter_path)
 
-    print("VENV CONTENTS with PEX_PYTHON_PATH={}: {}".format(ENV.PEX_PYTHON_PATH, json.dumps(venv_contents, sort_keys=True)))
+    real_stdout = sys.stdout
+    sys.stdout = sys.stderr
+    print(
+        "VENV CONTENTS with PEX_PYTHON_PATH={}: {}".format(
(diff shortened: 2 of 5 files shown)


jsirois (approved):
Thanks for figuring this one out Henry.

pex/vendor/_vendored/attrs/attrs-21.5.0.dev0.dist-info/WHEEL
```diff
@@ -1,5 +1,5 @@
 Wheel-Version: 1.0
-Generator: bdist_wheel (0.37.1)
+Generator: bdist_wheel (0.38.3)
```
jsirois, line 2: Hrm, its a bug that these changed. I'll circle back to file and fix. The vendoring should not depend on who does the vendoring. I need to pin wheel somewhere.

jsirois: Ok fix here: https://github.com/pantsbuild/pex/pull/1979

PR #1661, opened 2022-03-10
Switch Pex installed wheels to `--prefix` scheme.

jsirois:
Previously Pex used Pip's `--target` scheme which had both known bugs
(https://github.com/pypa/pip/issues/7658) and unknown quirks that Pex
was failing to fully be able to work around. Switch to the `--prefix`
scheme which exactly mirrors the scheme venvs use so that venvs can be
created with content of all sorts placed where it belongs.

This removes fragile parsing and interpretation of the RECORD; now Pex
only creates a RECORD, which is much more straight forward, when
building a venv.

Partially addresses #1655 by switching to sha256 for all external
artifact hashing. Only internal hashing remains for:
1. `interpreters` / INTERP-INFO
2. `venvs` and `unzipped_pexes` / PEX-INFO pex_hash (but this is a hash
   that includes all distributions' sha256 hashes).

Fixes #1656
Closes #1662

PR #1979, opened 2022-11-10
Fix vendoring wheel dependency drift.

jsirois:
This was all accounted for with `--no-build-isolation` and tox venv
pinning with an extensive comment about all this to boot. For some
reason unclear to me, I undid `--no-build-isolation` in
https://github.com/pantsbuild/pex/pull/1661 though.

This just re-instates that flag and re-runs vendoring to get back to
stable wheel versions.


Eric-Arellano (approved):

hrfuller (approved):
Thanks for sorting this out!

PR #1980, opened 2022-11-11
Add '--system-site-package' option to pex-tools venv

PhoenixHe-NV:


tests/tools/commands/test_venv.py
```diff
@@ -798,3 +798,45 @@ def create_venv_pex():
...
+    assert not os.path.exists(venv_dir)
+
+    venv_pex = create_venv_pex()
+    subprocess.check_call(args=[venv_pex, "venv", venv_dir], env=make_env(PEX_TOOLS=True))
+    assert os.path.exists(venv_dir)
+    assert os.path.exists(venv_pex)
+    assert os.path.exists(pex_root)
+
+    pyvenv_content = open(os.path.join(venv_dir, "pyvenv.cfg")).read()
+    print(pyvenv_content)
```
jsirois, line 839: These are not great tests :)

How about something like this:
```python
venv = Virtualenv.enclosing(os.path.join(venv_dir, "bin", "python"))
venv_interpreter = venv.interpreter
venv_base_interpreter = venv_interpreter.resolve_base_interpreter()
assert venv_interpreter != venv_base_interpreter, (
    "The venv base interpreter should be the system interpreter; not the venv interpreter itself."
)

venv_sys_path = venv.sys_path
venv_interpreter_site_packages = venv_interpreter.site_packages
system_site_packages = venv_base_interpreter.site_packages
assert not set(venv_interpreter_site_packages).intersection(system_site_packages)
assert system_site_packages in venv_sys_path
```

You may need to experiment a bit to get the set math right, but this sort of thing gives you a real test that will succeed or fail depending on whether system site packages are actually included. Currently the test requires a human to go look at the output you print, which is not great. The test can break and no-one may know because they don't read the print output.

jsirois (commented):
Thanks for taking this up @PhoenixHe-TSP. The production change looks good but the test could use some work so that it actually fails when you hard code line 189 in pex/tools/commands/venv.py to either True or False.

tests/tools/commands/test_venv.py
PhoenixHe-TSP: Sorry currently it's WIP. I forgot to mark it as draft. Thanks for your code snippet :)

PhoenixHe-TSP: Now it's ready for review!

tests/tools/commands/test_venv.py
```diff
@@ -798,3 +799,65 @@ def create_venv_pex():
...
+    if enable_system_site_package:
+        create_venv_args.append("--system-site-packages")
+    subprocess.check_call(args=create_venv_args, env=make_env(PEX_TOOLS=True))
+
+    assert os.path.exists(venv_dir)
+    assert os.path.exists(venv_pex)
+    assert os.path.exists(pex_root)
+
+    # Check pyvenv.cfg
+    pyvenv_cfg_lines = open(os.path.join(venv_dir, "pyvenv.cfg")).readlines()
```
jsirois, line 839: This test is failing for Python 2.7 / Virtualenv . Since its besides the point, you only really care about what the tests below prove, how about eliminate this test and just keep the set math tests.

PhoenixHe-TSP: Python 2.7 was deprecated, why do we need to test for python 2.7? If that test is required, I'm afraid "pyvenv.cfg" magic doesn't apply to python 2.7, so maybe we should skip this test on python < 3. Let me check the python doc.

jsirois: Pex supports Python 2.7 and 3.5+

Please just remove that part of the test. Like I said, it doesn't test anything you actually care about anyway. What you care about is the site packages set math.

PhoenixHe-TSP: Python 2.7 doesn't support "pyvenv.cfg" according to https://docs.python.org/2.7/library/site.html .

PhoenixHe-TSP: OK I removed pyvenv.cfg check. I'm afraid the test will fail on py2.7. Could you please let the CI run?

tests/tools/commands/test_venv.py
```diff
@@ -801,6 +801,7 @@ def create_venv_pex():
     assert not os.path.exists(pex_root)
 
 
+@pytest.mark.skipif(sys.version_info < (3,5), reason="py2 doesn't support system-site-packages")
```
jsirois: Please do not skip the test for Python 2.7. Please instead remove the portion of the test that checks for pyvenv.cfg. Python 2.7 _does_ support supprt system site packages.

I feel like I'm not making the key point clear. That is, you don't actually care how system-site-packages get wired. The pyvenv.cfg file is a detail. ALl you care about is that system site packages get _used_ in actuality. The rest of your test test that and that's good.

PhoenixHe-TSP: OK. I hope you are right about py2.7 support system site packages.

PhoenixHe-TSP: Could you please rerun CI? There are some network errors I suppose.

jsirois: Done. You may want to use `./dtox.sh -epy27 -- test_venv` to cycle on this more quickly on your own machine. That will set up (one time) a docker container with Python 2.7 in it and run tox from inside there against Pythhon 2.7.

PhoenixHe-TSP: Command failed
```
$ ./dtox.sh -epy27 -- test_venv
[+] Building 2.1s (4/4) FINISHED                                                                          
 => [internal] load build definition from Dockerfile                                                 0.0s
 => => transferring dockerfile: 522B                                                                 0.0s
 => [internal] load .dockerignore                                                                    0.0s
 => => transferring context: 2B                                                                      0.0s
 => ERROR [internal] load metadata for docker.io/library/a7f7e4652237:latest                         2.0s
 => [auth] library/a7f7e4652237:pull token for registry-1.docker.io                                  0.0s
------
 > [internal] load metadata for docker.io/library/a7f7e4652237:latest:
------
failed to solve with frontend dockerfile.v0: failed to create LLB definition: pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed
```

Anyway, from the error log I believe py2.7 site doesn't have support for system-site-packages.

jsirois: It absolutely does.

Using your branch, I first make 1 pex and then 2 venvs from it - one using your new `--system-site-packages` option and one not:
```
$ python2.7 -mpex cowsay -ccowsay -o cowsay.pex --include-tools
$ PEX_TOOLS=1 ./cowsay.pex venv --system-site-packages venv.ssp
$ PEX_TOOLS=1 ./cowsay.pex venv venv.no-ssp
```

I then test the sys.path of each:
```
$ PEX_INTERPRETER=1 venv.ssp/pex -c 'import sys; print("\n".join(sys.path))'

/home/jsirois/dev/pantsbuild/jsirois-pex/venv.ssp
/home/jsirois/dev/pantsbuild/jsirois-pex/venv.ssp/lib/python2.7
/home/jsirois/dev/pantsbuild/jsirois-pex/venv.ssp/lib/python2.7/plat-x86_64-linux-gnu
/home/jsirois/dev/pantsbuild/jsirois-pex/venv.ssp/lib/python2.7/lib-tk
/home/jsirois/dev/pantsbuild/jsirois-pex/venv.ssp/lib/python2.7/lib-old
/home/jsirois/dev/pantsbuild/jsirois-pex/venv.ssp/lib/python2.7/lib-dynload
/usr/lib/python2.7
/usr/lib/python2.7/plat-x86_64-linux-gnu
/usr/lib/python2.7/lib-tk
/home/jsirois/dev/pantsbuild/jsirois-pex/venv.ssp/local/lib/python2.7/site-packages
/home/jsirois/dev/pantsbuild/jsirois-pex/venv.ssp/lib/python2.7/site-packages
/usr/local/lib/python2.7/site-packages
/usr/local/lib/python2.7/dist-packages
/usr/lib/python2.7/dist-packages
$ PEX_INTERPRETER=1 venv.no-ssp/pex -c 'import sys; print("\n".join(sys.path))'

/home/jsirois/dev/pantsbuild/jsirois-pex/venv.no-ssp
/home/jsirois/dev/pantsbuild/jsirois-pex/venv.no-ssp/lib/python2.7
/home/jsirois/dev/pantsbuild/jsirois-pex/venv.no-ssp/lib/python2.7/plat-x86_64-linux-gnu
/home/jsirois/dev/pantsbuild/jsirois-pex/venv.no-ssp/lib/python2.7/lib-tk
/home/jsirois/dev/pantsbuild/jsirois-pex/venv.no-ssp/lib/python2.7/lib-old
/home/jsirois/dev/pantsbuild/jsirois-pex/venv.no-ssp/lib/python2.7/lib-dynload
/usr/lib/python2.7
/usr/lib/python2.7/plat-x86_64-linux-gnu
/usr/lib/python2.7/lib-tk
/home/jsirois/dev/pantsbuild/jsirois-pex/venv.no-ssp/local/lib/python2.7/site-packages
/home/jsirois/dev/pantsbuild/jsirois-pex/venv.no-ssp/lib/python2.7/site-packages
```

That's a difference of:
```
$ diff <(PEX_INTERPRETER=1 venv.ssp/pex -c 'import sys; print("\n".join(sys.path))' | sed 's|/home/jsirois/dev/pantsbuild/jsirois-pex/venv.ssp||') <(PEX_INTERPRETER=1 venv.no-ssp/pex -c 'import sys; print("\n".join(sys.path))' | sed 's|/home/jsirois/dev/pantsbuild/jsirois-pex/venv.no-ssp||')
13,15d12
< /usr/local/lib/python2.7/site-packages
< /usr/local/lib/python2.7/dist-packages
< /usr/lib/python2.7/dist-packages
```

And that difference is, in fact, the python2.7 (system) site-packages.

PhoenixHe-TSP: Seems like pex will inject customized virtual env for python 2.7, and `site.py` is manipulated to detect `no-global-site-packages.txt`.

PhoenixHe-TSP: Original test would fail because for the injected venv, its venv_interpreter.site_packages seems broken.

jsirois: @PhoenixHe-TSP let me know if you need any help here, but I'll assume you've used [pyenv](https://github.com/pyenv/pyenv) or similar to get a local test / edit feedback loop going where you use `tox -epy27 -- test_venv`.

jsirois:
@PhoenixHe-NV thanks for your efforts on this in the past. I circled back around and knocked this out in #2500 released in Pex 2.16.0: https://github.com/pex-tool/pex/releases/tag/v2.16.0

Issue #1983, opened 2022-11-17
Don't unconditionally prepend `""` to a venv PEX's `sys.path`


thejcannon:
https://github.com/pantsbuild/pex/pull/1832#discussion_r1025363133 captures the location of the bug.

`sys.path` behavior is documented https://docs.python.org/3/library/sys.html#sys.path which shows the prepending to `sys.path` is dependent on how `python` is invoked.

And to further prove the point:

```
$ cat foo.py
print(__import__("sys").path)

$ python -m venv venv
...

$ venv/bin/python foo.py
['/home/josh/work/pants', '/usr/lib/python38.zip', '/usr/lib/python3.8', '/usr/lib/python3.8/lib-dynload', '/home/josh/work/pants/venv/lib/python3.8/site-packages']

$ venv/bin/python -m foo
['/home/josh/work/pants', '/usr/lib/python38.zip', '/usr/lib/python3.8', '/usr/lib/python3.8/lib-dynload', '/home/josh/work/pants/venv/lib/python3.8/site-packages']

$ MYCODE=$(cat foo.py) &&  venv/bin/python -c "$MYCODE"
['', '/usr/lib/python38.zip', '/usr/lib/python3.8', '/usr/lib/python3.8/lib-dynload', '/home/josh/work/pants/venv/lib/python3.8/site-packages']
```

benjyw:
Related: https://github.com/pantsbuild/pants/issues/17531

jsirois:
@thejcannon, your example shell session is useful and is also true all the way back to Python 2.7 - almost:
```
$ cat foo.py
import sys; print("\n".join(sys.path))
$ python2.7 foo.py
/home/jsirois/dev/pantsbuild/jsirois-pex
/usr/lib/python2.7
/usr/lib/python2.7/plat-x86_64-linux-gnu
/usr/lib/python2.7/lib-tk
/usr/lib/python2.7/lib-old
/usr/lib/python2.7/lib-dynload
/usr/local/lib/python2.7/dist-packages
/usr/lib/python2.7/dist-packages
$ python2.7 -m foo

/usr/lib/python2.7
/usr/lib/python2.7/plat-x86_64-linux-gnu
/usr/lib/python2.7/lib-tk
/usr/lib/python2.7/lib-old
/usr/lib/python2.7/lib-dynload
/usr/local/lib/python2.7/dist-packages
/usr/lib/python2.7/dist-packages
$ python2.7 -c "$(cat foo.py)"

[...]

Issue #1982, opened 2022-11-17
Should `pex ... -e module` be prepending CWD?


thejcannon:
According to https://docs.python.org/3/library/sys.html#sys.path `python -m module` prepends CWD to `sys.path`.

```
$ cat code/foo.py 
import sys
import os

def main():
    print(os.getcwd())
    print(sys.path)


if __name__ == "__main__":
    main()
```

```
$ python -m code.foo
/home/josh/work/pants
['/home/josh/work/pants', '/usr/lib/python38.zip', '/usr/lib/python3.8', '/usr/lib/python3.8/lib-dynload', '/home/josh/.local/lib/python3.8/site-packages', '/usr/local/lib/python3.8/dist-packages', '/usr/lib/python3/dist-packages']
```

---

PEX help says:
```
-m MODULE[:SYMBOL], -e MODULE[:SYMBOL], --entry-point MODULE[:SYMBOL]
                        Set the entry point to module or module:symbol. If just specifying module, pex behaves like python -m, e.g. python -m SimpleHTTPServer. If specifying
                        module:symbol, pex assume symbol is a n0-arg callable and imports that symbol and invokes it as if via `sys.exit(symbol())`. (default: None)
```

So presumably using `pex -D code -e foo` would act like `python -m code.foo`, however:

```
$ pex -D=code -e foo
/home/josh/work/pants
['/tmp/tmpnu421wmb', '/usr/lib/python38.zip', '/usr/lib/python3.8', '/usr/lib/python3.8/lib-dynload', '/tmp/tmpnu421wmb/.bootstrap']
```
even if I twiddle with `PEX_INHERIT_PATH:
```
$ PEX_INHERIT_PATH=fallback pex -D=code -e foo
/home/josh/work/pants
['/tmp/tmpfauv3vos', '/usr/lib/python38.zip', '/usr/lib/python3.8', '/usr/lib/python3.8/lib-dynload', '/usr/local/lib/python3.8/dist-packages', '/usr/lib/python3/dist-packages', '/tmp/tmpfauv3vos/.bootstrap']
```
```
$ PEX_INHERIT_PATH=prefer pex -D=code -e foo
/home/josh/work/pants
[...]

PR #1984, opened 2022-11-17
Only insert `""` to head of `sys.path` if a venv PEX runs in interpreter mode

thejcannon:
Fixes https://github.com/pantsbuild/pex/issues/1983.

Note that this PR doesn't revert the code to use the prior code (`sys.path.extend(os.environ.get("PEX_EXTRA_SYS_PATH", "").split(os.pathsep))`) as that might still add `""` when `PEX_EXTRA_SYS_PATH` is not in the environment.

Also note that this doesn't change anything related to https://github.com/pantsbuild/pex/issues/1982 (whose outcome might be that `""` is in `sys.path`, but for other reasons and in unrelated cases).

### DIFF
--- a/pex/venv/pex.py
+++ b/pex/venv/pex.py
@@ -489,18 +489,18 @@ def _populate_sources(
                 os.execv(script_path, [script_path] + sys.argv[1:])
 
             pex_interpreter = pex_overrides.get("PEX_INTERPRETER", "").lower() in ("1", "true")
-
-            if pex_interpreter:
-                # A Python interpreter always inserts the CWD at the head of the sys.path.
-                # See https://docs.python.org/3/library/sys.html#sys.path
-                sys.path.insert(0, "")
-
             PEX_INTERPRETER_ENTRYPOINT = "code:interact"
             entry_point = (
                 PEX_INTERPRETER_ENTRYPOINT
                 if pex_interpreter
                 else pex_overrides.get("PEX_MODULE", {entry_point!r} or PEX_INTERPRETER_ENTRYPOINT)
             )
+
+            if entry_point == PEX_INTERPRETER_ENTRYPOINT:
+                # A Python interpreter always inserts the CWD at the head of the sys.path.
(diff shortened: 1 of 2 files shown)


pex/venv/pex.py
```diff
@@ -440,9 +440,6 @@ def sys_executable_paths():
 
             os.environ["VIRTUAL_ENV"] = venv_dir
 
-            # A Python interpreter always inserts the CWD at the head of the sys.path.
-            sys.path.insert(0, "")
```
jsirois, line 444: I need to think through what PEX should be aiming for, but this is certainly not right no matter how that shakes out. In the zipapp execution mode, for example, "" is inserted, but only when the PEX runs in interpreter mode:
https://github.com/pantsbuild/pex/blob/67fc7e05315fc347bcf269712814f160c6035efe/pex/pex.py#L627-L631

Here in venv execution mode you've just wiped out "" insertion in all situations; so they don't behave the same.

thejcannon: Ah I see, thanks for the code link. I'll dive further.

pex/venv/pex.py
```diff
@@ -489,6 +489,12 @@ def sys_executable_paths():
                 os.execv(script_path, [script_path] + sys.argv[1:])
 
             pex_interpreter = pex_overrides.get("PEX_INTERPRETER", "").lower() in ("1", "true")
+
+            if pex_interpreter:
```
jsirois, line 493: I think this needs to actually test `entry_point == PEX_INTERPRETER_ENTRYPOINT` down further. If the venv pex was built with no entrypoint, then `{entry_point!r}` will evaluate to the `None` literal and the `entry_point` will be assigned `code:interact` which you'll have missed as things stand currently.

thejcannon: Ah right, I forgot about the non-env-var case.

jsirois:
CI fix here: https://github.com/pantsbuild/pex/pull/1985

jsirois (approved):

PR #1987, opened 2022-11-21
Prepare the 2.1.114 release.

### LINKED ISSUE #1975: Release 2.1.214
Backlog:
+ [ ] Cleanup sys.path after __pex__ is imported #1959
+ [ ] importing a __pex__ leaks bootstrap vendored code #1954 
+ [ ] The --compile option is only useful for Python 2.7 #1939
+ [ ] Universal lock creation failure for disjoint IC and Requires-Python is correct but the error message from Pip / Pex is confusing. #1927
+ [ ] The `__pex__` import hook does not handle registering top level PEP-420 packages #1910 
+ [ ] The `--resolve-local-platforms` option does not work with `--complete-platforms` #1899 
+ [ ] Expose setuptools and wheel version flags for use with non-vendored Pips. #1895 
+ [ ] Plumb `--pip-version` to Platform tag calculation. #1894
+ [ ] venv_dir calculation doesn't correctly handle PEX_PYTHON_PATH with symlinks. #1885 
---
+ [ ] PEX_INHERIT_PATH+default shebangs don't interact well with virtualenvs #1344
+ [ ] Pex interpreter cache is not upgrade / downgrade safe. #1176
+ [ ] Provide a better resolve error message for a Platform DistributionTarget when `--resolve-local-platforms`. #1070
+ [ ] Remove `PEX_PYTHON_PATH` from build time interpreter selection. #1075

jsirois:
Closes #1975

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,17 @@
 Release Notes
 =============
 
+2.1.114
+-------
+
+This release brings two fixes for ``--venv`` mode PEXes.
+
+* Only insert "" to head of sys.path if a venv PEX runs in interpreter mode (#1984)
+  `PR #1984 <https://github.com/pantsbuild/pex/pull/1984>`_
+
+* Map pex python path interpreter to realpath when creating venv dir hash. (#1972)
+  `PR #1972 <https://github.com/pantsbuild/pex/pull/1972>`_
+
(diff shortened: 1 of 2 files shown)


hrfuller (approved):
LGTM, thanks John.

Issue #1989, opened 2022-11-22
Exported `requirements.txt` using different (abbreviated) version from `lock.json`


thejcannon:
Lock with PEX:
```
$ pex3 lock create --indent 2 absl-py~=0.10.0 | tee lock.json | grep 0.10.0
              "url": "https://files.pythonhosted.org/packages/b9/07/f69dd3367368ad69f174bfe426a973651412ec11d48ec05c000f19fe0561/absl_py-0.10.0-py3-none-any.whl"
          "version": "0.10.0"
    "absl-py~=0.10.0"
```

Then export the lockfile:
```
$ pex3 lock export lock.json 
absl-py==0.10 \
  --hash=sha256:ea07d7d437798bffc14f39fccec3909d251a1e76e233205ded72b71c267e0178
six==1.16 \
  --hash=sha256:8abb2f1d86890a2dfb989f9a77cfcfd3e47c2a354b01111771326f8aa26e0254
```

and note `absl-py` and `six` has been truncated from `0.10.0` to `0.10`, and `1.16.0` to `1.16`.

`pip`, of course doesn't care. But migrating from `pip-tools` (specifically `pip-compile`) created `requirements.txt` to `pex`'s exported one, I get a noisy diff due to the change.

jsirois:
The issue is here:
https://github.com/pantsbuild/pex/blob/004d579be4f31aca33da49da067afe3bf35e6105/pex/cli/commands/lock.py#L544-L556

The Pin has both a project_name and a version that are default normalized. They each have the raw version still available though:
https://github.com/pantsbuild/pex/blob/main/pex/pep_503.py#L32
https://github.com/pantsbuild/pex/blob/main/pex/pep_440.py#L34

thejcannon:
Is there a reason to prefer the normalized value to the raw one? Just trying to think through this.

jsirois:
Not that I can see. Normalization is only for taming the wild west when making comparisons in code. One project might depend on Markdown==3.0 another Markdown>=3 and doing code comparisons those should be both be treated as satisfied by Markdown 3.0.0. In the final output of a lockfile, in whatever format, when all comparisons are done, using the raw value of the version from the source artifact is fine. There is no loss of information and catering to human eyeballs seems appropriate. Afaict you are the 1st to raise and so you get to stake ground here. Someone may come along with a use case that begs for the opposite of what you'd like here and that battle is for another day.

thejcannon:
[...]

PR #1990, opened 2022-11-22
Use raw version when exporting

thejcannon:
Fixes https://github.com/pantsbuild/pex/issues/1989 and matches `pip-compile`'s behavior.

Test fails before change, passes after.

### DIFF
--- a/pex/cli/commands/lock.py
+++ b/pex/cli/commands/lock.py
@@ -545,7 +545,7 @@ class Lock(OutputMixin, JsonMixin, BuildTimeCommand):
                     "{project_name}=={version} \\\n"
                     "  {hashes}\n".format(
                         project_name=pin.project_name,
-                        version=pin.version,
+                        version=pin.version.raw,
                         hashes=" \\\n  ".join(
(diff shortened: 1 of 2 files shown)


pex/cli/commands/lock.py
```diff
@@ -545,7 +545,7 @@ def _export(self):
                     "{project_name}=={version} \\\n"
                     "  {hashes}\n".format(
                         project_name=pin.project_name,
```
jsirois, line 547: Are you sure normalized project names aren't also a problem for pip-tools comparisons? Circling back is fine too, but it would be nice to settle all the pip-tools driven items in 1 release if possible.

jsirois (approved):

pex/cli/commands/lock.py
thejcannon: If I read the code correctly they do "canonicalize" the requirement name: https://github.com/jazzband/pip-tools/blob/a524d983def05802d98e70e08c75c743894562bf/piptools/utils.py#L112 (which "normalizes" it. Good name y'all).

Issue #1988, opened 2022-11-21
Allow for sorting the requirements when exporting `requirements.txt`


thejcannon:
This allows for much easier diffing (migrating from `pip-tools`) and grokking (trying to scroll and find)

thejcannon:
Currently I have the following:

```
sed -i -z "s/\\\\\n  //g" requirements.txt
sort -o requirements.txt requirements.txt
sed -i "s/ --hash=\(.*\)/ \\\\\\n  --hash=\1/" requirements.txt
```

jsirois:
This seems fine, if ~arbitrary. That said, its hard to imagine too many more axes beyond alpha-sort an pip-tools is a thing.

PR #1992, opened 2022-11-23
Sort requirements based on nromalized project name when exporting

thejcannon:
Fixes https://github.com/pantsbuild/pex/issues/1988

I couldn't find the right twiddling in the tests to make this fail before my changes. I tried ordering `requirements` and ordering `locked_requirements` but to no avail.

### DIFF
--- a/pex/cli/commands/lock.py
+++ b/pex/cli/commands/lock.py
@@ -541,7 +541,7 @@ class Lock(OutputMixin, JsonMixin, BuildTimeCommand):
             )
 
         with self.output(self.options) as output:
-            for pin in sorted(fingerprints_by_pin.keys(), key=attrgetter("project_name")):
+            for pin in sorted(fingerprints_by_pin.keys(), key=attrgetter("project_name.normalized")):
                 fingerprints = fingerprints_by_pin[pin]
                 output.write(
                     "{project_name}=={version} \\\n"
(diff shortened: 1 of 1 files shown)


thejcannon:
This is weird. I run `tox -e format-run` and it passes, but CI is failing.

thejcannon:
Ah but running it a second time does reformat? Hmm. Probably something I'm doing wrong.

pex/cli/commands/lock.py
```diff
@@ -540,7 +541,10 @@ def _export(self):
             )
 
         with self.output(self.options) as output:
-            for pin, fingerprints in fingerprints_by_pin.items():
+            for pin in sorted(
```
jsirois, line 544: This should be gated by a new CLI option. You just gave someone who is diffing these exports vs themselves your same bad day.

jsirois (commented):
The lack of success testing would give me pause, especially because the lock file format these exports derive from go to lengths with SortedTuple. It would be good to get to the bottom of that.

pex/cli/commands/lock.py
thejcannon: Ah yes, good point!

thejcannon:
Ah I see it's sorted based on specificity. I have a new test that fails without the change.

pex/cli/commands/lock.py
thejcannon: You have a preference on the name? Otherwise it'll be `--sorted`

thejcannon: Alternatively, we could make it an enum `--sort-by` with values `"specificity"` (the default) and `project-name`.

PR #1998, opened 2022-11-27
Fix interpreter constraint parsing.

### LINKED ISSUE #1995: Published pex on github no longer works with PyPy since 2.1.109
If I download the pex on github for 2.1.108:
```
$ wget https://github.com/pantsbuild/pex/releases/download/v2.1.108/pex
$ PEX_PYTHON=/usr/bin/pypy3.8 /usr/bin/pypy3.8 ./pex -V              
2.1.108
```

For 2.1.109 or newer:
```
$ wget https://github.com/pantsbuild/pex/releases/download/v2.1.109/pex
$ PEX_PYTHON=/usr/bin/pypy3.8 /usr/bin/pypy3.8 ./pex -V
Failed to find a compatible PEX_PYTHON=/usr/bin/pypy3.8.

1.) /usr/bin/pypy3.8 PyPy==3.8.13

  Version matches CPython!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*,<3.12,>=2.7
```

Not a big deal because one can still create their own pex that doesn't have the interpreter constraint.

jsirois:
Previously a bare specifier assumed CPython; now it does not.

Fixes #1995

### DIFF
--- a/pex/interpreter_constraints.py
+++ b/pex/interpreter_constraints.py
@@ -31,20 +31,22 @@ class InterpreterConstraint(object):
     def parse(
         cls,
         constraint,  # type: str
-        default_interpreter="CPython",  # type: str
+        default_interpreter=None,  # type: Optional[str]
     ):
         # type: (...) -> InterpreterConstraint
         try:
-            return cls(Requirement.parse(constraint))
+            requirement = Requirement.parse(constraint)
+            return cls(specifier=requirement.specifier, name=requirement.name)
         except RequirementParseError:
-            try:
-                return cls(
-                    Requirement.parse(
-                        "{interpreter}{specifier}".format(
-                            interpreter=default_interpreter, specifier=constraint
-                        )
-                    )
+            if default_interpreter is not None:
+                return cls.parse(
+                    constraint="{interpreter}{specifier}".format(
+                        interpreter=default_interpreter, specifier=constraint
+                    ),
+                    default_interpreter=None,
                 )
+            try:
+                return cls(specifier=SpecifierSet(constraint))
             except RequirementParseError as e:
                 raise ValueError(
                     "Unparseable interpreter constraint {constraint}: {err}".format(
@@ -70,7 +72,8 @@ class InterpreterConstraint(object):
             )
         )
 
-    requirement = attr.ib()  # type: Requirement
+    specifier = attr.ib()  # type: SpecifierSet
+    name = attr.ib(default=None)  # type: Optional[str]
 
     def iter_matching(self, paths=None):
         # type: (Optional[Iterable[str]]) -> Iterator[PythonInterpreter]
@@ -81,21 +84,18 @@ class InterpreterConstraint(object):
     @property
     def requires_python(self):
         # type: () -> SpecifierSet
-        return self.requirement.specifier
+        return self.specifier
 
     def __str__(self):
         # type: () -> str
-        return str(self.requirement)
+        return "{name}{specifier}".format(name=self.name or "", specifier=self.specifier)
 
     def __contains__(self, interpreter):
         # type: (PythonInterpreter) -> bool
         python_identity = interpreter.identity
-        return (
-            ProjectNameAndVersion(
-                project_name=python_identity.interpreter, version=python_identity.version_str
-            )
-            in self.requirement
-        )
+        if self.name and self.name != python_identity.interpreter:
+            return False
+        return python_identity.version_str in self.specifier
 
 
 @attr.s(frozen=True)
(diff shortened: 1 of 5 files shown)


jsirois:
@zmanji thanks for finding this. I invited you to the triage role, and would like to add you as a reviewer if possible.

benjyw (approved):
Code looks good. 

I couldn't tell by eyeballing https://github.com/pantsbuild/pex/commit/1a1b9aa63f2b5464d1b1ce8d2479eafd0006d209 exactly how this regression was introduced (if it was indeed in that commit).  I assume we're certain this fix is not breaking any expected behavior?

zmanji:
> Code looks good.
> 
> I couldn't tell by eyeballing [1a1b9aa](https://github.com/pantsbuild/pex/commit/1a1b9aa63f2b5464d1b1ce8d2479eafd0006d209) exactly how this regression was introduced (if it was indeed in that commit). I assume we're certain this fix is not breaking any expected behavior?

The regression is in the introduced `InterpreterConstraint` type. It has a default interpreter of `CPython` which is different than the behaviour that was previously there which was not having any interpreter. 

You can see this regression in play when looking at the PEX-INFO of the released pex files on GitHub.

For 2.1.109
```
 PEX_TOOLS=1 python3 ./pex info | jq .interpreter_constraints
[
  "CPython!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*,<3.12,>=2.7"
]
```

For 2.1.108
```
 PEX_TOOLS=1 python3 ./pex info | jq .interpreter_constraints
[
  ">=2.7,<3.12,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*"
]
```

The regression adds a default of `CPython` when the interpreter is not specified which means the published pex will not work for PyPy. 


This code LGTM. The regression test ensures we are good going forward.

benjyw:
AFAICT that default was present previously in the `parse_requirement()` signature: https://github.com/pantsbuild/pex/commit/1a1b9aa63f2b5464d1b1ce8d2479eafd0006d209#diff-b1297a55702e854b6233796710a712f1f3fd47905ecc1fabd9f03c63ac980b43L335 ?

zmanji:
> AFAICT that default was present previously in the `parse_requirement()` signature: [1a1b9aa#diff-b1297a55702e854b6233796710a712f1f3fd47905ecc1fabd9f03c63ac980b43L335](https://github.com/pantsbuild/pex/commit/1a1b9aa63f2b5464d1b1ce8d2479eafd0006d209#diff-b1297a55702e854b6233796710a712f1f3fd47905ecc1fabd9f03c63ac980b43L335) ?

That code was never used in the pex builder code path.

https://github.com/pantsbuild/pex/commit/1a1b9aa63f2b5464d1b1ce8d2479eafd0006d209#diff-8018b797310f9896a6574a746376acaf0730b409d6fe03ab73887f823240505c

jsirois:
There was both a regression here and a long standing bug. Both are fixed. It used to be the case that a bare IC always assumed CPython. You could be explicit about this or explicit about PyPy, but you could not say you supported both. Now you can in the obvious way, leave off the interpreter name. This is what Pex itself supports, either CPython or PyPy for all the interpreter versions it explicitly supports.

Thinking more, this can present a break for users previously relying on the default of CPython. I need to think more on this aspect. Perhaps a flag is needed in the CLI that defaults to CPython (old behavior) but can be set to default empty. Certainly awkward.

jsirois:
Ok, reading through usages I think the negative fallout is contained to, for example `--interpreter-constraint "==3.8.*"` now selecting PyPy 3.8 interpreters if installed on your system in addition to CPython 3.8 interpreters. This would mean that a Pex build could fail now that did not before if the dependencies involved were not available for PyPy or compilable for PyPy. The remedy would be adjusting the IC to `--interpreter-constraint "CPython==3.8.*"`. Since there have been 0 reports of issues due to this in Pex 2.1.109 through 2.1.116 save for @zmanji's I think it makes sense to proceed as is - there is a logical corrective for any breaks, be more specific in your IC.

zmanji:
> Ok, reading through usages I think the negative fallout is contained to, for example `--interpreter-constraint "==3.8.*"` now selecting PyPy 3.8 interpreters if installed on your system in addition to CPython 3.8 interpreters. This would mean that a Pex build could fail now that did not before if the dependencies involved were not available for PyPy or compilable for PyPy. The remedy would be adjusting the IC to `--interpreter-constraint "CPython==3.8.*"`. Since there have been 0 reports of issues due to this in Pex 2.1.109 through 2.1.116 save for @zmanji's I think it makes sense to proceed as is - there is a logical corrective for any breaks, be more specific in your IC.

Sounds fine to me. My issue was rooted in having downloading the officially published pex fail to work on PyPy systems. I think this patch corrects that.

benjyw:
Makes sense to me

Issue #2001, opened 2022-12-01
The PEX_TOOLS venv `--compile` option should ignore compile errors.


jsirois:
If there is a compile error, it may be irrelevant as in this case https://github.com/ShantanuKumar/pants-multi-poetry.

There the PEX produced via `./pants package src/package-a/package_a:pex_package_a` runs into this issue:
```
$ PEX_TOOLS=1 python3.9 dist/src.package-a.package_a/pex_package_a.pex venv --scope=deps --compile ./pex-deps-compiled 2>&1 | grep -C5 SyntaxError
Compiling './pex-deps-compiled/lib/python3.9/site-packages/aenum/__init__.py'...
Compiling './pex-deps-compiled/lib/python3.9/site-packages/aenum/_py2.py'...
***   File "./pex-deps-compiled/lib/python3.9/site-packages/aenum/_py2.py", line 5
    raise exc, None, tb
             ^
SyntaxError: invalid syntax

Compiling './pex-deps-compiled/lib/python3.9/site-packages/aenum/_py3.py'...
Listing './pex-deps-compiled/lib/python3.9/site-packages/aenum/doc'...
Compiling './pex-deps-compiled/lib/python3.9/site-packages/aenum/test.py'...
Compiling './pex-deps-compiled/lib/python3.9/site-packages/aenum/test_v3.py'...
```

The `aenum/_py2.py` file will never get imported by the Python 3.9 interpreter; so this compile failure is not relevant.

If the compile error is real, the application will encounter it at runtime and fail anyhow. On balance, it seems a later warning is by far the lesser of two evils here. Currently failing the venv creation due to failing compilation of a file stops things in their tracks with no recourse.

jsirois:
The failing line in question is here:
https://github.com/pantsbuild/pex/blob/af6f74da59d6053cf0b55669449328af3fc79cd4/pex/tools/commands/venv.py#L209

ShantanuKumar:
Should we then just ignore this error using `try/except`?

ShantanuKumar:
By the way do I need to install all the tox envs to run tests?

jsirois:
That should do. Note that the exception raised is `pex.executor.Executor.NonZeroExit`:
https://github.com/pantsbuild/pex/blob/61c150d3635b3603ee9a0e8cf9bc748c415a8660/pex/executor.py#L40-L49

[...]

PR #2002, opened 2022-12-01
ignore compile error for PEX_TOOLS=1

ShantanuKumar:
Fixes https://github.com/pantsbuild/pex/issues/2001

### DIFF
--- a/pex/tools/commands/venv.py
+++ b/pex/tools/commands/venv.py
@@ -8,8 +8,10 @@ import logging
 import os
 from argparse import ArgumentParser
 
+from pex import pex_warnings
 from pex.common import safe_delete, safe_rmtree
 from pex.enum import Enum
+from pex.executor import Executor
 from pex.pex import PEX
 from pex.result import Error, Ok, Result
 from pex.tools.command import PEXCommand
@@ -206,7 +208,11 @@ class Venv(PEXCommand):
                     "installed:\n{}".format(e)
                 )
         if self.options.compile:
-            pex.interpreter.execute(["-m", "compileall", venv_dir])
+            try:
+                pex.interpreter.execute(["-m", "compileall", venv_dir])
+            except Executor.NonZeroExit as non_zero_exit:
+                pex_warnings.warn(non_zero_exit.stderr)
+
         if self.options.remove is not None:
             if os.path.isdir(pex.path()):
                 safe_rmtree(pex.path())
--- a/pex/tools/commands/venv.py
+++ b/pex/tools/commands/venv.py
@@ -222,7 +222,7 @@ class Venv(PEXCommand):
             try:
                 pex.interpreter.execute(["-m", "compileall", venv_dir])
             except Executor.NonZeroExit as non_zero_exit:
-                pex_warnings.warn(non_zero_exit.stderr)
+                pex_warnings.warn("ignoring compile error {}".format(repr(non_zero_exit)))
 
         if self.options.remove is not None:
             if os.path.isdir(pex.path()):
--- a/tests/integration/test_issue_2001.py
+++ b/tests/integration/test_issue_2001.py
@@ -17,7 +17,7 @@ def test_compile_error_as_warning(
     run_pex_command(
         args=[
             "--include-tools",
-            "aenum",
+            "aenum==3.1.11",
             "-o",
             pex,
         ],
@@ -34,4 +34,6 @@ def test_compile_error_as_warning(
         stderr=subprocess.PIPE,
         text=True,
     )
-    assert "PEXWarning" in output.stderr, "PEXWarning should be generated."
+    assert (
+        "PEXWarning: ignoring compile error NonZeroExit" in output.stderr
+    ), "PEXWarning should be generated."
--- a/tests/integration/test_issue_2001.py
+++ b/tests/integration/test_issue_2001.py
@@ -1,9 +1,8 @@
 import os.path
-import pytest
 import subprocess
 import sys
 
-from pex.testing import PY38, ensure_python_interpreter, make_env, run_pex_command
+from pex.testing import make_env, run_pex_command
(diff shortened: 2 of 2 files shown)


pex/tools/commands/venv.py
```diff
@@ -206,7 +208,11 @@ def run(self, pex):
                     "installed:\n{}".format(e)
                 )
         if self.options.compile:
-            pex.interpreter.execute(["-m", "compileall", venv_dir])
+            try:
+                pex.interpreter.execute(["-m", "compileall", venv_dir])
+            except Executor.NonZeroExit as non_zero_exit:
+                pex_warnings.warn(non_zero_exit.stderr)
```
jsirois, line 214: Have you tested this includes the relevant compile failure information? It would be great if you could add an integration test that uses your case. You could trim your case down to just the problematic [aenum](https://pypi.org/project/aenum/3.1.11/) case.

Integration tests live here: https://github.com/pantsbuild/pex/tree/main/tests/integration
You could use `pex/tests/integration/test_issue_2001.py` for the new test file. Really the test just needs to confirm a venv can be created and check the compilation error shows up in stderr if PEX_VERBOSE=1 (which turns on warnings). Here is a similar enough IT: https://github.com/pantsbuild/pex/blob/31020410809e94fd2a8565294d26b5346164800e/tests/integration/test_issue_1995.py

The key bits are:
+ tmpdir: for making up paths for the test.
+ run_pex_command(...).assert_success(): For running the pex CLI and making sure it works
+ make_env: for creating an env dict that includes the current env + your additions.

ShantanuKumar: Thank you John for all the instructions. Yeah I was thinking of adding test but didn't have very good understanding of this repo. I will try to add it soon.

tests/integration/test_issue_2001.py
```diff
@@ -0,0 +1,37 @@
...
+    # type: (...) -> None
+    pex = os.path.join(str(tmpdir), "pex.pex")
+    run_pex_command(
+        args=[
+            "--include-tools",
+            "aenum",
+            "-o",
+            pex,
+        ],
+        # The package script requires Python 3.
```
jsirois, line 24: Is that true? I don't see any 3-isms here: https://github.com/ethanfurman/aenum/blob/master/setup.py (they don't use tags or branches for releases so I'm just guessing really!).

tests/integration/test_issue_2001.py
```diff
@@ -0,0 +1,37 @@
...
+
+def test_compile_error_as_warning(
+    tmpdir,  # type: Any
+):
+    # type: (...) -> None
+    pex = os.path.join(str(tmpdir), "pex.pex")
+    run_pex_command(
+        args=[
+            "--include-tools",
+            "aenum",
```
jsirois, line 20: Pleas pin to the current latest release with "aenum==3.1.11" so that this test doesn't break when they publish new releases.

tests/integration/test_issue_2001.py
```diff
@@ -0,0 +1,37 @@
...
+    # interpreter.
+    compiled_deps_dir = str(tmpdir)
+    output = subprocess.run(
+        args=[sys.executable, pex, "venv", "--scope", "deps", "--compile", compiled_deps_dir],
+        env=make_env(PEX_PYTHON=sys.executable, PEX_TOOLS=1),
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        text=True,
+    )
+    assert "PEXWarning" in output.stderr, "PEXWarning should be generated."
```
jsirois, line 37: Can you match against slightly more text to catch the compile error? There are other PEXWarnings this could be. The less tight the match the better, but enough to prove compile issues.

jsirois: Perhaps just add a prefix message to `pex_warnings.warn(non_zero_exit.stderr)` that makes it clear a compile error was hit but being ignored.

jsirois (approved):
Thanks @ShantanuKumar the last bits are small, but the aenum pinning is important.

tests/integration/test_issue_2001.py
```diff
@@ -0,0 +1,37 @@
...
+            "-o",
+            pex,
+        ],
+        # The package script requires Python 3.
+        python=sys.executable if sys.version_info[0] >= 3 else ensure_python_interpreter(PY38),
+    ).assert_success()
+    # The packaged Pex PEX should work with all Pythons we support, including the current test
+    # interpreter.
+    compiled_deps_dir = str(tmpdir)
+    output = subprocess.run(
```
jsirois, line 30: CI caught a real issue here actually: https://github.com/pantsbuild/pex/actions/runs/3688289216/jobs/6242918161#step:4:149

Pex supports back to Python 2.7 so the code must be 2.7 compatible. Just use:
```python
process = subprocess.check_call(args=..., env=..., stderr=...)
_, stderr_bytes = process.communicate()
...
```

tests/integration/test_issue_2001.py
ShantanuKumar: This is just me doing copy paste.

tests/integration/test_issue_2001.py
ShantanuKumar: So it looks like this error isn't generated for python2.7
I even installed python2.7 locally to check and I don't see any warning being thrown

jsirois: Well, the issue was compiling Python 2.7 syntax under Python 3; not the other way around:
```
***   File "./pex-deps-compiled/lib/python3.9/site-packages/aenum/_py2.py", line 5
    raise exc, None, tb
             ^
SyntaxError: invalid syntax
```

So the test needs to check for no warning under Python 2.7 and warning otherwise.

jsirois:
Thanks @ShantanuKumar - this will go out in 2.1.119 later today.

PR #2010, opened 2022-12-08
Add `--non-hermetic-scripts` option to `venv` tool.

### LINKED ISSUE #2004: Support for generating non-hermetic console scripts in `venv`
The `venv` command rewrites console script shebangs in the generated virtualenv [here](https://github.com/pantsbuild/pex/blob/632a6ba3867d520e2721bf70fa1e7f114050c78b/pex/venv/pex.py#L276-L278), adding `-sE` in the process. Those added flags make it more difficult to use scripts that rely on `PYTHONPATH` (i.e. `pylint` trying to load in-repo plugins). It would be nice if `venv` supported an option to disable the behavior.

danxmoran:
Closes #2004 

When the option is set, console scripts in the generated virtualenv will not pass `-sE` to `python` in their shebangs. This is needed to allow customization of `PYTHONPATH` when running tools like `pylint`.

### DIFF
--- a/pex/tools/commands/venv.py
+++ b/pex/tools/commands/venv.py
@@ -164,6 +164,16 @@ class Venv(PEXCommand):
                 )
             ),
         )
+        parser.add_argument(
+            "--non-hermetic",
+            dest="hermetic",
+            action="store_false",
+            default=True,
+            help=(
+                "Don't rewrite console script shebangs in the venv to pass `-sE` to the interpreter; "
+                "for example, to enable running venv scripts with a custom `PYTHONPATH`."
+            ),
+        )
         cls.register_global_arguments(parser, include_verbosity=False)
 
     def run(self, pex):
@@ -196,6 +206,7 @@ class Venv(PEXCommand):
             collisions_ok=self.options.collisions_ok,
             symlink=False,
             scope=self.options.scope,
+            hermetic=self.options.hermetic,
         )
         if self.options.pip:
             try:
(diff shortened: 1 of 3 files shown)


tests/integration/tools/commands/test_venv.py
```diff
@@ -265,3 +265,75 @@ def colors_package_path(venv_dir):
...
+def test_non_hermetic_issue_2004(tmpdir):
+    # type: (Any) -> None
+
+    pex_root = os.path.join(str(tmpdir), "pex_root")
+    pylint_pex = os.path.join(str(tmpdir), "pylint.pex")
+    run_pex_command(
+        args=[
+            "-m",
+            "pylint",
+            "pylint==2.15.8",
```
danxmoran, line 279: Hrm, CI failing (at least) because this version isn't available for Python 2.7. Will see if I can make this work across versions.

pex/tools/commands/venv.py
```diff
@@ -164,6 +164,16 @@ def add_arguments(cls, parser):
                 )
             ),
         )
+        parser.add_argument(
+            "--non-hermetic",
```
jsirois, line 168: So, the name says more than it does. The venv pex is still hermetic, its only the console scripts that are not. I am personally unafraid of long names - blame Java? - and would be super happy with `--non-hermetic-scripts`.

jsirois (approved):
Look great - thanks @danxmoran.

pex/tools/commands/venv.py
danxmoran: 👍 will rename

jsirois:
@danxmoran I've got #2011 out for the CI fix here.

jsirois:
@danxmoran I'm going to rebase this to get green and do the Pex release if possible tonight. I hate getting my state mucked with so I just wanted to give you a heads-up.

PR #2011, opened 2022-12-08
Fix bitrot in test, packaging 22.0 was _just_ released.

jsirois:


jsirois:
I broke this out of #2007 since it's also affecting #2010.

danxmoran (approved):

PR #2014, opened 2022-12-14
Add PEX_DISABLE_VARIABLES to lock down a PEX run.

### LINKED ISSUE #2013: Support scrubbing all PEX_* env vars
In order to scrub PEX_* env vars to keep a PEX hermetic, someone launching a PEX must either clear the environment which takes care - many things need PATH / HOME, etc. If either a PEX could be built with a PexInfo setting to ignore PEX_* env vars or the Pex runtime provided away to turn them off, or both, then locking down a PEX would be easier.

See https://github.com/pantsbuild/scie-pants/issues/3 for motivation.

jsirois:
This is more robust than having to scrub the environment of all PEX_* vars before running a PEX.

Fixes #2013

### DIFF
--- a/pex/variables.py
+++ b/pex/variables.py
@@ -213,7 +213,7 @@ class Variables(object):
     @staticmethod
     def _maybe_get_bool_var(
         name,  # type: str
-        env,  # type: Mapping[str, str]
+        env,  # type: Mapping[str, Optional[str]]
     ):
         # type: (...) -> Optional[bool]
         value = env.get(name, None)
--- a/pex/variables.py
+++ b/pex/variables.py
@@ -236,7 +236,7 @@ class Variables(object):
             self._environ = {
                 key: value
                 for key, value in env.items()
-                if key == "PEX_DISABLE_VARIABLES" or not key.startswith("PEX")
+                if key == "PEX_DISABLE_VARIABLES" or not key.startswith("PEX_")
             }
         else:
             self._environ = dict(env)
@@ -356,12 +356,12 @@ class Variables(object):
         self._environ = self._environ.copy()
         if disable_env:
             for k in list(self._environ):
-                if k != "PEX_DISABLE_VARIABLES" and k.startswith("PEX"):
+                if k != "PEX_DISABLE_VARIABLES" and k.startswith("PEX_"):
                     self._environ.pop(k)
         for k, v in kw.items():
             if v is None:
                 self._environ.pop(k, None)
-            elif disable_env and k != "PEX_DISABLE_VARIABLES" and k.startswith("PEX"):
+            elif disable_env and k != "PEX_DISABLE_VARIABLES" and k.startswith("PEX_"):
                 self._environ.pop(k, None)
             else:
                 self._environ[k] = v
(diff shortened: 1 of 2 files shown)


jsirois:
The motivation is to keep https://github.com/pantsbuild/setup/issues/105 being fixed in the transition from `./pants` to `scie-pants`.

pex/variables.py
```diff
@@ -210,15 +210,40 @@ def _get_kv(cls, variable):
...
-            rc_values = self.from_rc(rc).copy()
-            rc_values.update(self._environ)
-            self._environ = rc_values
+        # type: (Optional[Mapping[str, str]], Optional[str]) -> None
+        env = environ if environ is not None else os.environ
+        if self._maybe_get_bool_var("PEX_DISABLE_VARIABLES", env) is True:
+            self._environ = {
+                key: value
+                for key, value in env.items()
+                if key == "PEX_DISABLE_VARIABLES" or not key.startswith("PEX")
```
benjyw, line 239:
```suggestion
                if key == "PEX_DISABLE_VARIABLES" or not key.startswith("PEX_")
```
To match the documentation below? Not that there would be a difference in practice I suppose.

pex/variables.py
```diff
@@ -332,16 +350,37 @@ def patch(self, **kw):
         present. The rest will be added to the environment or else updated if already present in
         the environment.
         """
+        disable_env = self._maybe_get_bool_var("PEX_DISABLE_VARIABLES", kw)
+
         old_environ = self._environ
         self._environ = self._environ.copy()
+        if disable_env:
+            for k in list(self._environ):
+                if k != "PEX_DISABLE_VARIABLES" and k.startswith("PEX"):
```
benjyw, line 359:
```suggestion
                if k != "PEX_DISABLE_VARIABLES" and k.startswith("PEX_"):
```

pex/variables.py
```diff
@@ -332,16 +350,37 @@ def patch(self, **kw):
...
+
         yield self._environ
         self._environ = old_environ
 
+    @defaulted_property(default=False)
+    def PEX_DISABLE_VARIABLES(self):
+        # type: () -> bool
+        """Boolean.
+
+        Disable reading of all PEX_* variables (except this one) from all sources. Both PEX_*
```
benjyw, line 377: This is the documentation I'm referring to, that specifies `PEX_` with an underscore.

benjyw (approved):

pex/variables.py
jsirois: Ah, yeah. You're right, but I'll take the opposite tack and (e)VADE.

jsirois: Hrm, you're right I think. There is precedent in that file and PEXNOTTHATPEX is not really in Pex's namespace.

pex/variables.py
jsirois: Fixed.

pex/variables.py
jsirois: Fixed.

benjyw (approved):

PR #2018, opened 2022-12-30
Support REPL command history.

benjyw:
When running a REPL via Pex (e.g., with Pants's repl goal), it can be useful to support command history.

Python normally sets up command history by default when the interpreter imports `site`, but Pex bypasses that. So this change re-introduces the suggested idiom for loading and saving history, per https://docs.python.org/3/library/readline.html#example

Also replace a few hardcoded forward-slashed paths with os.path.sep.

### DIFF
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -681,7 +681,7 @@ class PEX(object):  # noqa: T000
                 try:
                     readline.read_history_file(histfile)
                     readline.set_history_length(1000)
-                except FileNotFoundError:
+                except OSError:
                     pass
 
                 atexit.register(readline.write_history_file, histfile)
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -673,6 +673,19 @@ class PEX(object):  # noqa: T000
         else:
             self.demote_bootstrap()
 
+            if self._vars.PEX_INTERPRETER_HISTORY:
+                import atexit
+                import readline
+
+                histfile = os.path.expanduser(self._vars.PEX_INTERPRETER_HISTORY_FILE)
+                try:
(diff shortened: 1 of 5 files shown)


benjyw:
Addresses https://github.com/pantsbuild/pex/pull/2019

pex/variables.py
```diff
@@ -189,8 +189,8 @@ def from_rc(cls, rc=None):
         """
         ret_vars = {}  # type: Dict[str, str]
         rc_locations = [
-            "/etc/pexrc",
-            "~/.pexrc",
+            os.path.join(os.sep, "etc", "pexrc"),
```
kaos, line 192: Does the path `/etc/pexrc` make sense on OSes where path sep is not `/` ?

kaos (approved):

pex/variables.py
benjyw: Dunno, but no harm if not.

pex/pex.py
```diff
@@ -671,6 +671,19 @@ def execute_interpreter(self):
                 sys.argv = args
                 return self.execute_content(arg, content)
         else:
+            if self._vars.PEX_INTERPRETER_HISTORY:
```
jsirois, line 674: Unfortunately this pex.py logic also exists independently here for --venv PEXes: https://github.com/pantsbuild/pex/blob/f326835f068088fd9f75929cadf02006f10dfa32/pex/venv/pex.py#L331

That should probably get this same treatment. That main though has no access to the `pex` package.

benjyw: Done! Good tip.

jsirois (approved):

PR #2020, opened 2022-12-31
Fix tests that broke due to changes in the world.

benjyw:
One broke due to a recent pyenv bug.
Another broke when numpy published a new version.

And that's why you always pin your deps!


benjyw:
Fixes https://github.com/pantsbuild/pex/issues/2021 and another bug

jsirois (approved):

PR #2028, opened 2023-01-06
Allow safe_copy to overwrite on windows

jtbraun:
`os.rename` will throw FileExistsError on Windows, but not other operating systems. `os.replace` is the portable function that will overwrite, and should be used if available.

Obviously this will still fail on python 2.7. Alternative approaches would be testing for existence and removing the destination before trying the copy (an approach used in other portions of the codebase).

### DIFF
--- a/pex/common.py
+++ b/pex/common.py
@@ -100,7 +100,8 @@ def safe_copy(source, dest, overwrite=False):
         # type: () -> None
         temp_dest = dest + uuid4().hex
         shutil.copy(source, temp_dest)
-        os.rename(temp_dest, dest)
+        replace = getattr(os, 'replace', None) or os.rename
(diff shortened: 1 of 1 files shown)


benjyw (commented):
Thanks for the fix! Do we not have a test for this?

jsirois:
PEX absolutely does not support Windows. I appreciate the help, but do not want to encourage fake fixes like this. The full CI needs to be up and running for Windows support. I'm working on that.

jsirois:
To give you an idea, there is alot of work needed to support Windows: https://github.com/pantsbuild/pex/compare/main...jsirois:pex:Windows/wip

I should be turning my focus back to that in the next few weeks.

jtbraun:
No problem. I'll close this out then.

PR #2029, opened 2023-01-06
Check for the pip log before attempting to read it.

### LINKED ISSUE #2027: Pex fails when an expected pip log doesn't exist
This reproduces the problem on an aarch64 machine (but not on other platforms):

```
$ PEX_VERBOSE=1 python3.7  -m pex --interpreter-constraint $'CPython<3.10,>=3.7' pex==2.1.112   --resolver-version pip-2020-resolver --preserve-pip-download-log

...

  File "/home/gha/pex/pex/pip/log_analyzer.py", line 97, in _check_returncode
    with open(self._log, "r") as fp:
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/pex-pip-log.5vrr25lt/pip.log'
[Errno 2] No such file or directory: '/tmp/pex-pip-log.5vrr25lt/pip.log'
```

Some logs are written (you can see them with `--preserve-pip-download-log`) but there is always one that isn't. Unclear why, or under what conditions.

benjyw:
Fixes #2027 (verified the fix on the platform that exhibits the issue)

### DIFF
--- a/pex/pip/log_analyzer.py
+++ b/pex/pip/log_analyzer.py
@@ -92,23 +92,22 @@ class LogScrapeJob(Job):
         ]
         if activated_analyzers:
             collected = []
-            if not os.path.isfile(self._log):
-                return
-            with open(self._log, "r") as fp:
-                for line in fp:
-                    if not activated_analyzers:
-                        break
-                    for index, analyzer in enumerate(activated_analyzers):
-                        result = analyzer.analyze(line)
-                        if isinstance(result.data, ErrorMessage):
-                            collected.append(result.data)
-                        if isinstance(result, LogAnalyzer.Complete):
-                            activated_analyzers.pop(index).analysis_completed()
-                            if not activated_analyzers:
-                                break
-            for analyzer in activated_analyzers:
-                analyzer.analysis_completed()
-            if not self._preserve_log:
-                os.unlink(self._log)
-            stderr = (stderr or b"") + "".join(collected).encode("utf-8")
+            if os.path.isfile(self._log):
+                with open(self._log, "r") as fp:
+                    for line in fp:
+                        if not activated_analyzers:
+                            break
+                        for index, analyzer in enumerate(activated_analyzers):
+                            result = analyzer.analyze(line)
+                            if isinstance(result.data, ErrorMessage):
+                                collected.append(result.data)
+                            if isinstance(result, LogAnalyzer.Complete):
+                                activated_analyzers.pop(index).analysis_completed()
+                                if not activated_analyzers:
+                                    break
+                for analyzer in activated_analyzers:
+                    analyzer.analysis_completed()
+                if not self._preserve_log:
+                    os.unlink(self._log)
+                stderr = (stderr or b"") + "".join(collected).encode("utf-8")
         super(LogScrapeJob, self)._check_returncode(stderr=stderr)
--- a/pex/pip/log_analyzer.py
+++ b/pex/pip/log_analyzer.py
@@ -92,6 +92,8 @@ class LogScrapeJob(Job):
         ]
         if activated_analyzers:
             collected = []
+            if not os.path.isfile(self._log):
+                return
             with open(self._log, "r") as fp:
                 for line in fp:
                     if not activated_analyzers:
--- a/pex/pip/log_analyzer.py
+++ b/pex/pip/log_analyzer.py
@@ -92,6 +92,9 @@ class LogScrapeJob(Job):
         ]
         if activated_analyzers:
             collected = []
+            # A process may fail so early that there is no log file to analyze.
+            # We assume that if this is the case, the superclass _check_returncode will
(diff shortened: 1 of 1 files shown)


jsirois (approved):
Something bigger is going on if the Pip log doesn't exist but the analysis phase needs it.

This feels like not solving a real problem.

benjyw:
Yeah, it's very unclear to me why, still trying to debug that.

benjyw:
OK, I see the issue. It's running on a system interpreter that has no `pip` module, so that process crashes out immediately. But there is also a pypi interpreter of the same version that succeeds, so I guess that makes the overall download succeed?

benjyw:
I'm not sure what I wrote previously makes sense, since we vendor pip, but that appears to be related somehow.

jsirois:
It might help to not have the system be a mystery to Pex. Any chance Pex could get a shard so this is less invisible / less likely to regress?

benjyw:
OK, the lack of a pip module was a slightly red herring. The real error on the system python is `ModuleNotFoundError: No module named 'distutils.util'`. Because on debian you have to `apt-get install python3-distutils` to get distutils.

But that is a real-world condition that we do need to handle. I'm not yet clear on why the entire pex run doesn't fail, since one of its download jobs failed.

benjyw:
The aarch64 runner is available to the pantsbuild org, so a job that targets `["self-hosted", "Linux", "ARM64"]` will get picked up. 

I'll look into setting that up for pex.

jsirois:
Yeah, that's real world and well known: https://github.com/pantsbuild/pex/issues/1027

I have no clue if its sane to attempt hardening all entrypoints to Pex code against broken distributions.

benjyw:
OK, the entire pex run didn't fail because my early `return` in this PR didn't call the superclass `_check_returncode`. Fixed that. Now this behaves as I expect, and fails on the actual underlying error and not the missing log.

benjyw:
This is now the correct and obvious error, at the same point as the previous error, but not masked by it.

benjyw:
This was not true before because my first attempt at this PR did a silent `return` instead of delegating to the superclass result/error handling. But now we get a non-confusing and obvious error now, not a silent error later.

pex/pip/log_analyzer.py
```diff
@@ -92,21 +92,22 @@ def _check_returncode(self, stderr=None):
...
-                        if isinstance(result, LogAnalyzer.Complete):
-                            activated_analyzers.pop(index).analysis_completed()
-                            if not activated_analyzers:
-                                break
-            for analyzer in activated_analyzers:
-                analyzer.analysis_completed()
-            if not self._preserve_log:
-                os.unlink(self._log)
-            stderr = (stderr or b"") + "".join(collected).encode("utf-8")
+            if os.path.isfile(self._log):
```
jsirois, line 95: I'm claiming this is a bug. If there are activated analyzers there must be a log for them to analyze. If there is not, that's an error - or should be. That error is now masked by silently doing a shoulder shrug afaict.

jsirois: If there is some superclass contract wonkiness (or dressed-up "subtletly"), maybe that can be called out here. "We do this crazy thing because of this other crazy thing ...".

benjyw: Not sure I follow. If the interpreter is busted in some way (e.g., the distutils issue) then it will error out before pip has a chance to create the log. That seems to me to be the real error, which was previously masked by the log-reading error, but is now allowed to express itself. 

I'm not sure what the wonkiness is here? I read this code as "when the process exits, analyze its log for useful info, and then propagate its exit code, augmented with this useful info". If there is no log, there is nothing to augment it with, but we still propagate the exit code.  I guess the wonky case might be a process that succeeds but write no log? We can check for that.

jsirois: The wonkiness is that in the small there is nothing wrong. We have analyzers, they want to analyze, there should be a log for them to analyze. If there is not a log and something went wrong somewhere else, its very odd that control gets back here with no signal saying - things failed. You're taking the log file not existing as that signal. Perhaps that fine if that's always true. Maybe at least call it out in a comment? It feels like there is more than one layer involved here and the current layer is having to know something about other layers to make the bold claim that the file that is its reason for existence is not available but that's ok.

jsirois: I've read more through the surrounding code an a comment is likely the best that can be done. I'm happy following up and adding that.

benjyw: I still don't really understand your point. "there should be a log for them to analyze" was a previous assumption, but it turns out to be incorrect. Sometimes there will be no log, if the pip process fails early enough. The failure of that process is the one that is useful to show to the user, not the absence of a log file. 

To put it another way, what if instead of no log there were an empty log? This PR wouldn't have been necessary, the analyzers would operate vacuously, but the exact same thing would happen. 

I'll add a comment, but feel free to follow up with a better one.

PR #2030, opened 2023-01-06
Some tweaks to get tests passing on aarch64

benjyw:
Doesn't actually set up aarch64 CI, as this is not perceived to be necessary at this time.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,6 +97,26 @@ jobs:
         uses: pantsbuild/actions/run-tox@e63d2d0e3c339bdffbe5e51e7c39550e3bc527bb
         with:
           tox-env: py${{ matrix.python-version[0] }}${{ matrix.python-version[1] }}-pip${{ matrix.pip-version }}
+  cpython-unit-tests-aarch64:
+    name: (aarch64) Pip ${{ matrix.pip-version }} TOXENV=py${{ matrix.python-version[0] }}${{ matrix.python-version[1] }}
+    needs: org-check
+    runs-on: [ 'self-hosted', 'Linux', 'ARM64' ]
+    strategy:
+      matrix:
+        include:
+          - python-version: [ 2, 7 ]
+            pip-version: 20
+          - python-version: [ 3, 11 ]
+            pip-version: 20
+          - python-version: [ 3, 11 ]
+            pip-version: 22_3
+    steps:
+      - name: Checkout Pex
+        uses: actions/checkout@v3
+      - name: Run Unit Tests
+        uses: pantsbuild/actions/run-tox@e63d2d0e3c339bdffbe5e51e7c39550e3bc527bb
+        with:
+          tox-env: py${{ matrix.python-version[0] }}${{ matrix.python-version[1] }}-pip${{ matrix.pip-version }}
   pypy-unit-tests:
     name: (PyPy ${{ join(matrix.pypy-version, '.') }}) Pip ${{ matrix.pip-version }} TOXENV=pypy${{ join(matrix.pypy-version, '') }}
     needs: org-check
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -97,6 +97,26 @@ jobs:
         uses: pantsbuild/actions/run-tox@e63d2d0e3c339bdffbe5e51e7c39550e3bc527bb
         with:
           tox-env: py${{ matrix.python-version[0] }}${{ matrix.python-version[1] }}-pip${{ matrix.pip-version }}
+  cpython-unit-tests-aarch64:
```
benjyw, line 100: Because these run on a self-hosted runner with Pythons already set up just-so, the setup, expose and cache interpreter steps are not relevant, so these are separate jobs rather than matrix entries in the existing ones.

pex/testing.py
```diff
@@ -58,8 +58,12 @@
 NOT_CPYTHON27 = IS_PYPY or PY_VER != (2, 7)
 IS_LINUX = platform.system() == "Linux"
 IS_MAC = platform.system() == "Darwin"
-IS_NOT_LINUX = not IS_LINUX
-NOT_CPYTHON27_OR_OSX = NOT_CPYTHON27 or IS_NOT_LINUX
+machine = platform.machine()
+IS_LINUX_X86_64 = IS_LINUX and machine == "x86_64"
```
benjyw, line 62: When I had to skip a handful of tests that implicitly assume they're running on Linux x86_64, I did so by subtracting ("run unless on aarch64") instead of "only run on x86_64" so that if we add more platforms in the future (notably, Windows) these tests will fail on those platforms and at least force us to consider what to do with them.

tests/test_interpreter.py
```diff
@@ -192,7 +192,12 @@ def test_iter_interpreter_path_filter(self, test_interpreter1, test_interpreter2
 
     def test_iter_interpreter_path_filter_symlink(self, test_interpreter1, test_interpreter2):
         # type: (str, str) -> None
-        with temporary_dir() as bin_dir:
+        with temporary_dir() as tmpdir:
+            # PythonInterpreter._find_pyvenv_cfg looks in the binary's parent dir for a
+            # pyvenv.cfg file. We use a subdir to ensure that a random such file in /tmp
```
benjyw, line 197: This actually happened :)

tests/integration/test_integration.py
```diff
@@ -809,6 +811,7 @@ def test_invalid_entry_point_verification_3rdparty(tmpdir):
     ).assert_failure()
 
 
+@pytest.mark.skipif(IS_LINUX_ARM64 or IS_MAC_ARM64, reason="No p537 wheel published for ARM yet.")
```
benjyw, line 814: In the future we can publish p537 for aarch64, and enable this test for that platform. But it would require moving it to the pantsbuild org so that it has access to the self-hosted runners, and futzing with its CI config, and I didn't want to wait on all that to get this PR going.

jsirois: It's easy enough to just use Circle CI to publish for that. I'll set that up.

benjyw:
True. So far getting this working has only involved tweaking or skipping tests, not any changes to pex itself. 

I'm on the fence here. 

I'd like to at least merge the test changes, in case we ever need to run them on aarch64 in the future.

jsirois:
Totally up to you. I just did not want mystery errors on machines I couldn't access. You figured out the (non) mystery which is what I care about. I don't like lingering mysteries.

benjyw:
Will pass on the actual aarch64 CI jobs part, at least for now. There are some subtle flakiness issues that I'd rather not spend time digging at.

But we might as well merge the test changes, as they make the state of those tests more accurate.

jsirois (approved):

Issue #2026, opened 2023-01-06
Using `--complete-platform` with `--resolve-local-platforms` should build sdists when local platform provides a subset of complete-platforms


shalabhc:
# Summary

If a builder platform is capable of building sdists into wheels that are compatible with a target platform, then `pex --complete-platform=cp-generated-from-target.json --resolve-local-platforms` should build any pex on the builder, including any sdists.

# Example where this doesn't work

* builder: PyPA [official](https://github.com/pypa/manylinux#docker-images) wheel builder `quay.io/pypa/manylinux2014_x86_64` 
* target: `python:3.10-slim` [source](https://github.com/docker-library/python/blob/7a54933c66171020473b9a4f06aec4e8ffb43a45/3.10/slim-bullseye/Dockerfile)

# Steps to reproduce

1. Generate the complete-platform.json ([py310-complete-platform.json.zip](https://github.com/pantsbuild/pex/files/10356605/py310-complete-platform.json.zip))
    ```
    % docker run --platform=linux/amd64 -it python:3.10-slim /bin/bash -c 'pip3 install pex==2.1.119 --quiet --quiet --quiet; pex3 interpreter inspect --markers --tags;' > py310-complete-platform.json
    ```
2. Prepare the builder
    ```
    % docker run --platform=linux/amd64 -it -v ${PWD}:/curdir quay.io/pypa/manylinux2014_x86_64            
    [root@3a4bc9eae7f5 /]# export PATH=/opt/python/cp310-cp310/bin/:$PATH
    [root@3a4bc9eae7f5 /]# python3.10 --version
    Python 3.10.9
    [root@3a4bc9eae7f5 /]# which pip3
    /opt/python/cp310-cp310/bin/pip3
    [root@3a4bc9eae7f5 /]# pip3 install pex
    Collecting pex
      Downloading pex-2.1.119-py2.py3-none-any.whl (2.9 MB)
         ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 2.9/2.9 MB 6.0 MB/s eta 0:00:00
    Installing collected packages: pex
    Successfully installed pex-2.1.119
    
    [root@3a4bc9eae7f5 curdir]# cd /curdir
    [root@3a4bc9eae7f5 curdir]# export _PEX_FILE_LOCK_STYLE=bsd
    ```

3. Build the PEX (note [there is no wheel](https://pypi.org/project/pendulum/) for `pendulum` for Python 3.10)
[...]

PR #2031, opened 2023-01-09
Fix interpreter resolution when using --complete-platform with --resolve-local-platforms

shalabhc:
Fixes https://github.com/pantsbuild/pex/issues/2026.

Previously the interpreter resolution was more strict than necessary. The `supported_tags` were only checked if the most specific tag of the complete platform matched the platform of the local interpreter. This prevented resolution in cases when it was possible and safe (ie all `supported_tags` of the interpreter as present in the `supported_tags` of the complete-platform).

Added an additional test for the above case. Fixed existing code and also improved various existing tests by adding `expected_complete_platforms`.

### DIFF
--- a/pex/resolve/target_configuration.py
+++ b/pex/resolve/target_configuration.py
@@ -190,7 +190,7 @@ class TargetConfiguration(object):
                 map(str, requested_platforms)
             )
             with TRACER.timed(
-                "Searching for local interpreters matching ".format(", ".join(platform_strs))
+                "Searching for local interpreters matching {}".format(", ".join(platform_strs))
             ):
                 candidate_interpreters = OrderedSet(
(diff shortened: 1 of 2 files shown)


pex/resolve/target_configuration.py
```diff
@@ -199,27 +177,28 @@ def resolve_targets(self):
...
+            # interpreter.
+
+            platform_strs = list(map(str, requested_complete_platforms)) + list(
+                map(str, requested_platforms)
+            )
             with TRACER.timed(
-                "Searching for local interpreters matching {}".format(
-                    ", ".join(map(str, all_platforms))
-                )
+                "Searching for local interpreters matching ".format(", ".join(platform_strs))
```
jsirois, line 193: The fmt shard error caught this - missing `{}`.

PR #2033, opened 2023-01-09
Fix loose `--venv` PEXes to be robust to moves.

### LINKED ISSUE #2023: A loose layout, venv-with-symlink PEX creates brittle symlinks
Pex version `2.1.113`

I found this through Pants, but ultimately seems like a PEX issue/fix.

```console
$ pex --no-venv-copies --layout=loose --venv prepend awscli --script aws -o  awscli.pex
$ python awscli.pex help
<help output>
$ mv awscli.pex awscli2.pex/
$ python awscli2.pex
  File "/home/josh/.pex/venvs/ea7cb7998ed2db4f784178989a4034013cd10e01/ddab8011daaee380698ac2fb9701af18c90c03f6/bin/aws", line 19, in <module>
    import awscli.clidriver
ModuleNotFoundError: No module named 'awscli
```

See also:
```console
$ ls -l /home/josh/.pex/venvs/ea7cb7998ed2db4f784178989a4034013cd10e01/ddab8011daaee380698ac2fb9701af18c90c03f6/lib/python3.8/site-packages/awscli
lrwxrwxrwx 1 josh josh 107 Jan  3 11:49 /home/josh/.pex/venvs/ea7cb7998ed2db4f784178989a4034013cd10e01/ddab8011daaee380698ac2fb9701af18c90c03f6/lib/python3.8/site-packages/awscli -> ../../../../../../../../../<CWD>/awscli.pex/.deps/awscli-1.27.41-py3-none-any.whl/awscli
```

In Pants-land, this manifests as being a symlink to an ephemeral Pants sandbox which has since gone bye-bye.

---

What I might expect is that on first run, the wheels get copied into the PEX root and then the venv symlinks to the extracted wheels in the pex root so moving/deleting the original pex doesn't invalidate the venv.

jsirois:
Both packed and zipped PEXes undergo an unzip into the PEX_ROOT prior to
any other action, establishing a known private cache of their contents.
Loose PEXes skip this step and run directly from their loose contents.
This is fine and speedy for the "zipapp" execution mode, but leads to an
incorrect venv cache in symlink mode where venv entries symlink back out
to the loose PEX and can thus be invalidated by moves outside Pex
control. Force loose PEXes in `--venv` mode to use copies so that their
cached venv is robust to external moves.

Fixes #2023

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -876,7 +876,11 @@ def do_main(
             compress=options.compress,
         )
         if options.seed != Seed.NONE:
-            seed_info = seed_cache(options, pex, verbose=options.seed == Seed.VERBOSE)
+            seed_info = seed_cache(
+                options,
+                PEX(pex_file, interpreter=interpreter),
+                verbose=options.seed == Seed.VERBOSE,
+            )
             print(seed_info)
     else:
         if not _compatible_with_current_platform(interpreter, targets.platforms):
--- a/tests/integration/test_issue_2023.py
+++ b/tests/integration/test_issue_2023.py
@@ -1,0 +1,83 @@
+# Copyright 2023 Pants project contributors (see CONTRIBUTORS.md).
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+import os.path
+import shutil
+import subprocess
+import sys
+from textwrap import dedent
+
+import pytest
+from colors import colors
+
+from pex.layout import Layout
+from pex.testing import run_pex_command
+from pex.typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from typing import Any, List
+
+
+@pytest.mark.parametrize(
+    "layout", [pytest.param(layout, id=layout.value) for layout in Layout.values()]
+)
+@pytest.mark.parametrize(
+    "execution_mode_args",
+    [
+        pytest.param([], id="UNZIP"),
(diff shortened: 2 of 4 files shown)


thejcannon (approved):
Tests look good to me.

pex/bin/pex.py
```diff
@@ -876,7 +876,11 @@ def do_main(
             compress=options.compress,
         )
         if options.seed != Seed.NONE:
-            seed_info = seed_cache(options, pex, verbose=options.seed == Seed.VERBOSE)
+            seed_info = seed_cache(
```
jsirois, line 879: Tests exposed the fact seeding operated against the (loose) tmp PEX being build up and not the final built PEX. This fixes that - there was never a reason to seed from anything but the final PEX.

tests/integration/test_issue_2023.py
```diff
@@ -0,0 +1,83 @@
...
+
+@pytest.mark.parametrize(
+    "layout", [pytest.param(layout, id=layout.value) for layout in Layout.values()]
+)
+@pytest.mark.parametrize(
+    "execution_mode_args",
+    [
+        pytest.param([], id="UNZIP"),
+        pytest.param(["--venv", "--venv-site-packages-copies"], id="VENV (copies)"),
+        pytest.param(["--venv", "--no-venv-site-packages-copies"], id="VENV (symlinks)"),
```
jsirois, line 29: `--layout loose` + this was the 1 broken case fixed.

benjyw (approved):

PR #2035, opened 2023-01-12
Log all pip invocations

huonw:
This adds logging for `pip` invocations, to make it easier to understand how pex is calling `pip`, such as when a single invocation is taking a long time (https://github.com/pantsbuild/pex/issues/2036).

This mirrors similar logging for Python invocations: https://github.com/pantsbuild/pex/blob/1f8e25a714e52c45a50a6d6ab2ee7cb9e21a92bb/pex/interpreter.py#L713-L716

### DIFF
--- a/pex/compatibility.py
+++ b/pex/compatibility.py
@@ -259,3 +259,9 @@ else:
         if not prefix:
             return ""
         return os.path.join(*prefix)
+
+
+if PY3:
+    from shlex import quote as quote
+else:
+    from pipes import quote as quote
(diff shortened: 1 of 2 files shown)


jsirois (approved):
I don't see the harm, but the use case seems a bit exotic too. Just reading the code gives the same info and this is a pretty static arrangement. the `pip download + pip wheel + pip install` that services `pex ...` has been the same since day 1 of switching to Pip 3+ years ago as has the `pip download` used to satisfy `pex3 lock create`, etc.

PR #2041, opened 2023-01-18
Handle file:// URL deps in distributions.

### LINKED ISSUE #2038: Failed to resolve compatible distributions when building Pex from .whl with local dependencies
I have a `pyproject.toml` file as below at `~/code/mylibrary/`:

```
[tool.poetry]
name = "mylibrary"
version = "0.1.0"

[tool.poetry.dependencies]
python = ">=3.8,<4"
anotherlibrary = {path = "../anotherlibrary", develop=true}
[tool.poetry.group.dev.dependencies]

[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"
```

I build a .whl for `mylibrary` and here's what its `mylibrary-0.1.0.dist-info/METADATA` file contains:

```
Metadata-Version: 2.1
Name: mylibrary
Version: 0.1.0
Requires-Python: >=3.8,<4
Classifier: Programming Language :: Python :: 3
Classifier: Programming Language :: Python :: 3.8
Classifier: Programming Language :: Python :: 3.9
Classifier: Programming Language :: Python :: 3.10
Classifier: Programming Language :: Python :: 3.11
Requires-Dist: anotherlibrary @ file:///Users/rajaiealkorani/code/anotherlibrary
Description-Content-Type: text/markdown
```

I am trying to build a Pex file by passing that .whl as a dependency, but I am getting the error below:

```
pex --resolver-version pip-2020-resolver dist/mylibrary-0.1.0-py3-none-any.whl -o /tmp/testing.pex 
Failed to resolve compatible distributions:
1: mylibrary==0.1.0 requires anotherlibrary@ file:///Users/rajaiealkorani/code/anotherlibrary but no version was resolved
```

Is this a bug in Pex, or is this kind of dependency specification not supported?

I'm able to run `pip install mylibrary-0.1.0-py3-none-any.whl` in a clean virtual environment as long as `file:///Users/rajaiealkorani/code/anotherlibrary` exists

I am using Pex `2.1.120`

jsirois:
This has been a very long standing bug. The root issue is that
`pip download`, although it processes these deps and fails if not
present on the local filesystem, only handles resolving recursive
dependencies and it does not "download" the local distribution from its
referenced path on the file system to the `pip download` download dir.
As such, the normal Pex post-processing is foiled for these deps.

The fix provided here is in common infra used by both normal Pex builds
and `--lock` builds and uses a recursive post-processing step that
gathers all such dependencies from the downloaded distributions and
injects them into the normal Pex post-processing chain.

Fixes #2038

### DIFF
--- a/pex/resolver.py
+++ b/pex/resolver.py
@@ -16,7 +16,8 @@ from pex import targets
 from pex.atomic_directory import AtomicDirectory, atomic_directory
 from pex.auth import PasswordEntry
 from pex.common import safe_mkdir, safe_mkdtemp
-from pex.dist_metadata import Distribution, Requirement
+from pex.compatibility import unquote, urlparse
+from pex.dist_metadata import DistMetadata, Distribution, Requirement
 from pex.fingerprinted_distribution import FingerprintedDistribution
 from pex.jobs import Raise, SpawnedJob, execute_parallel
 from pex.network_configuration import NetworkConfiguration
@@ -45,7 +46,17 @@ from pex.util import CacheHelper
 from pex.variables import ENV
 
 if TYPE_CHECKING:
-    from typing import DefaultDict, Iterable, Iterator, List, Mapping, Optional, Sequence, Tuple
+    from typing import (
+        DefaultDict,
+        Iterable,
+        Iterator,
+        List,
+        Mapping,
+        Optional,
+        Sequence,
+        Set,
+        Tuple,
+    )
 
     import attr  # vendor:skip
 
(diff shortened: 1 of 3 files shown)


pex/resolver.py
```diff
@@ -657,6 +668,55 @@ def _spawn_install(
...
+                if urlinfo.scheme != "file":
+                    continue
+                dist_path = unquote(urlinfo.path).rstrip()
+                if not os.path.exists(dist_path):
+                    raise Unsatisfiable(
+                        "The {wheel} wheel has a dependency on {url} which does not exist on this "
+                        "machine.".format(wheel=install_request.wheel_file, url=requirement.url)
+                    )
+                if dist_path.endswith(".whl"):
+                    already_analyzed[requirement.project_name] = InstallRequest.create(
```
jsirois, line 700: Hrm, that's not right. This dep is not already analyzed - it still needs to hae its dist metadata loaded and its requirements processed. Fix coming...

jsirois: Ok, the issue was Pip actually. Now with recursion tested.

benjyw (approved):

PR #2047, opened 2023-02-05
Export a lockfile in an ad-hoc JSON graph format.

benjyw:
It's useful to see the lockfile in graph form, for visualization, reasoning about how some transitive dep gets pulled in, etc. 

This change emits a very simple JSON graph that other tools can process. 

For example, this will turn a lockfile into a .png of a DOT graph:

```
echo "digraph {\
  $(pex lock export --format=graph lockfile.json  | \
  jq -r '.edges | to_entries[] | .key as $src | .value[] | "\"" + ($src) + "\" -> \"" + (.) + "\""') \
}"  | dot -Tpng
```

One use case is to allow Pants to display this data in the `dependencies` goal, which users have asked for, e.g., in https://github.com/pantsbuild/pants/issues/12733

### DIFF
--- a/pex/cli/commands/lock.py
+++ b/pex/cli/commands/lock.py
@@ -3,9 +3,10 @@
 
 from __future__ import absolute_import, print_function
 
+import json
 import sys
 from argparse import Action, ArgumentError, ArgumentParser, ArgumentTypeError, _ActionsContainer
-from collections import OrderedDict
+from collections import OrderedDict, defaultdict
 from operator import attrgetter
 
 from pex import pex_warnings
@@ -65,6 +66,7 @@ class ExportFormat(Enum["ExportFormat.Value"]):
         pass
 
     PIP = Value("pip")
+    GRAPH = Value("graph")
     PEP_665 = Value("pep-665")
 
 
@@ -237,8 +239,12 @@ class Lock(OutputMixin, JsonMixin, BuildTimeCommand):
             choices=ExportFormat.values(),
             type=ExportFormat.for_value,
             help=(
-                "The format to export the lock to. Currently only the {pip!r} requirements file "
-                "format using `--hash` is supported.".format(pip=ExportFormat.PIP)
+                "The format to export the lock to. Export results may be a subset of the full "
+                "lockfile, targeting a specific interpreter and platform. Currently only "
+                "`{pip!r}` - a Pip requirements file format using `--hash`, and `{graph!r} - "
+                "an ad-hoc adjacency graph format, are supported.".format(
+                    pip=ExportFormat.PIP, graph=ExportFormat.GRAPH
+                )
             ),
         )
         export_parser.add_argument(
@@ -492,9 +498,11 @@ class Lock(OutputMixin, JsonMixin, BuildTimeCommand):
 
     def _export(self):
         # type: () -> Result
-        if self.options.format != ExportFormat.PIP:
+        if self.options.format not in [ExportFormat.PIP, ExportFormat.GRAPH]:
             return Error(
-                "Only the {pip!r} lock format is supported currently.".format(pip=ExportFormat.PIP)
+                "Only the `{pip!r}` and `{graph!r}` formats are supported currently.".format(
+                    pip=ExportFormat.PIP, graph=ExportFormat.GRAPH
+                )
             )
 
         lockfile_path, lock_file = self._load_lockfile()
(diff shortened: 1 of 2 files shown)


benjyw:
This is a speculative change. I'm not convinced this captures all the nuances of lockfiles (e.g., right now it ignores extras, at the least it should probably list them in the vertex name). I want some feedback on the approach before beefing it up and writing tests.

pex/cli/commands/lock.py
```diff
@@ -555,25 +563,41 @@ def _export(self):
...
+                                "--hash={algorithm}:{hash}".format(
+                                    algorithm=fingerprint.algorithm, hash=fingerprint.hash
+                                )
+                                for fingerprint in fingerprints
+                            ),
+                        )
                     )
-                )
+        else:
+            vertices = set()
```
jsirois, line 587: FYI: The current bit of Pex producing graphs is here: https://github.com/pantsbuild/pex/blob/97a2497e0938ece709310a03e7e41b5c26992952/pex/tools/commands/graph.py#L30-L83. That does take markers into account to good effect in the rendered graph, showing which requirements are not active in the present PEX being run by the present interpreter on the present machine. Of course the format is the dot adjacency list format and the side-band information about activation is stored in labels and colors.

If it's still the case that Pants doesn't want to define a format and Pex must my requirements are:
+ Embedding a version for the format: allow an escape hatch for backwards incompatible evolution (but with the intent to try hard to never actually use it)
+ Eventually merge the existing model / logic driving the tool example above with this: one kicks out dot, the other JSON.

pex/resolve/locked_resolve.py
```diff
@@ -345,6 +345,7 @@ def create(
         target,  # type: Target
         direct_requirements,  # type: Iterable[Requirement]
         resolved_artifacts,  # type: Iterable[_ResolvedArtifact]
+        adjacency_list,  # type: Dict[Pin, Set[Pin]]
```
jsirois, line 348: A LockedResolve already has all the information needed. There should be no need to add graph rendering-specific data or methods to it. I think this is just a sign a (stripped-down) model for exported graph data is needed.

pex/cli/commands/lock.py
```diff
@@ -555,25 +563,41 @@ def _export(self):
...
+            edges = defaultdict(list)
+            for src_pin, dst_pins in resolved.adjacency_list.items():
+                dst_pin_strs = sorted(str(dst_pin) for dst_pin in dst_pins)
+                vertices.add(str(src_pin))
+                vertices.update(dst_pin_strs)
+                edges[str(src_pin)] = dst_pin_strs
+            graph = {
```
jsirois, line 594: The single target selected above on 508-510 is controlled by command line flags like --python, --python-path, --interpreter-constraint, --platform and --complete-platform, but none of these need be present either. It seems like some node to store information about the rendered graph context would make sense; i.e.: for a universal lock, this would give the context for environment marker activation / deactivation assuming you include activation information in the graph and for a multi-lock lock file, it also indicates which lock is being talked about (say the CPython 3.7 locked resolve from a lock file containing 3 locks, 1 for 3.7, 1 for 3.8 and 1 for 3.9.

jsirois (commented):
I really think Pants should define a model if not a format. Presumably, internally, it will map all resolve graph information into that model before rendering, allowing JVM, Python, etc backends to map to a model and the core graphing goal to do the rendering in 1 spot in a consistent way that even in the future allows for mixed-language graphs. Definition of that model would help ensure the Pex format defined here need not thrash, since Pants needs have been thought through and worked out as evidenced by the model.

benjyw:
Yeah, not a bad idea to make progress on Pants ingesting the data, to validate the format.

pex/resolve/locked_resolve.py
benjyw: If this is the case, we could derive the graph entirely in Pants, no? But it seemed to me (naively?) that this logic had to be here because request_resolve() is where we know who actually requires who, in the context of a target.

jsirois: > If this is the case, we could derive the graph entirely in Pants, no?

Yup. Just be ready to break / fix when Pex changes its private, undocumented format.

> But it seemed to me (naively?) that this logic had to be here because request_resolve() is where we know who actually requires who, in the context of a target.

I think naively - maybe. So how will Pants use the graph info? If the lock is universal, which it is in Pants case, will `./pants dependencies ...` include this info now and if so, what the heck will that pick for a target? Will it solve for a local interpreter that matches the constraints of all targets requested and use that if success and fail otherwise? If so and success, you already know the target up in Pants and can annotate the graph display with that info; i.e. this is the dependency graph when Python 3.7 is used, but the results may be different for 3.8 and 3.9 (and the repo is like Pants itself and supports >=3.7,<3.10).

I think Pants really ought to have what it wants to do sorted. The sorting can probably best be done today using the lockfile json directly. Then, once sorted, a solid idea of the requirements for a stable PEX grap export format will be in hand and we can make sure it works for the not-pants cases, like non-universal locks with multiple resolves contained within and not just 1 univeral resolve, etc.

cosmicexplorer:
Are you familiar with the `pip install --report` json output? I worked on that with @sbidoul (see pypa/pip#10748 and pypa/pip#10771). I have been planning to add that to pex for a while now.

jsirois:
@benjyw since this is over 2 years old, I'm going to close. Hopefully Pants is working on a coherent dependency graph system and a spec for that graph format is forthcoming.

benjyw:
Fine with me. I'm not doing much on existing Pants backends any more, just some engine work and occasionally noodling on a new Python backend. If this comes up again in either context I'll revisit.

Issue #1969, opened 2022-10-31
pex build fails due to existing work-directory


christopherfrieler:
Beginning with version 2.1.105 building the pex file in our CI pipeline fails with the following message:

> .../python3.8/site-packages/pex/atomic_directory.py:176: PEXWarning: [pid:XX, tid:XXX, cwd:...]: After obtaining an exclusive lock on <PEX_ROOT>/isolated/.2f4fc85fa2be055a2975ce1147100c0d5c7e663a.atomic_directory.lck, failed to establish a work directory at <PEX_ROOT>/isolated/2f4fc85fa2be055a2975ce1147100c0d5c7e663a.workdir due to: [Errno 17] File exists: '<PEX_ROOT>/isolated/2f4fc85fa2be055a2975ce1147100c0d5c7e663a.workdir'
>   pex_warnings.warn(
> .../python3.8/site-packages/pex/atomic_directory.py:187: PEXWarning: [pid:XX, tid:XXX, cwd:...]: Continuing to forcibly re-create the work directory at <PEX_ROOT>/isolated/2f4fc85fa2be055a2975ce1147100c0d5c7e663a.workdir.
>   pex_warnings.warn(
> Failed to spawn a job for .../bin/python: [Errno 17] File exists: '<PEX_ROOT>/isolated/2f4fc85fa2be055a2975ce1147100c0d5c7e663a.workdir/pex/./venv'

It seems to have to do with #1905 introduced in version 2.1.105, but we have no clue, why this is happening in our CI pipeline, while building the .pex file on MacOS developer machines works. It looks like something else is creating that directory, but there is only one pex command in the pipeline job and the PEX_ROOT is not cached.

Our build environment uses:
- the Red Hat UBI 8.4 Docker image
- Python 3.8
- poetry 1.1, which manages pex as a dev-dependency

Then we build the pex with
`poetry run pex --inherit-path --python=python3.8 --requirement requirements.txt --find-links dist/ our_module --output-file dist/final.pex`

Any idea why this is happening or what else we could check would be helpful.

jsirois:
@christopherfrieler that warning message looks like the one in 2.1.112. I addded it to help debug a probable race or wrong POSIX assumption that has been hard to track down. I know this is painful for you, but I'm very happy to have a repro case from you! Can you try setting `_PEX_FILE_LOCK_STYLE=bsd` in your CI environment and see if that changes anything?

christopherfrieler:
[...]

PR #2052, opened 2023-02-10
Fix `execute_parallel` "leaking" a thread.

jsirois:
Although the thread was not leaked per-se, it could run after `execute_parallel` returned which could cause parallel atomic_directory posix locks to fail.

Fixes one case in #1969.

### DIFF
--- a/tests/integration/test_issue_1336.py
+++ b/tests/integration/test_issue_1336.py
@@ -4,6 +4,7 @@
 import os
 import subprocess
 
+from pex.compatibility import commonpath
 from pex.testing import PY310, ensure_python_interpreter, run_pex_command
 from pex.typing import TYPE_CHECKING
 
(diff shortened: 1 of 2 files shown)


tests/integration/test_issue_1336.py
```diff
@@ -16,4 +17,24 @@ def test_pip_leak(tmpdir):
     python = ensure_python_interpreter(PY310)
     pip = os.path.join(os.path.dirname(python), "pip")
     subprocess.check_call(args=[pip, "install", "setuptools_scm==6.0.1"])
-    run_pex_command(args=["--python", python, "bitstring==3.1.7"], python=python).assert_success()
+
```
jsirois, line 20: This was hanging and made me suspect the new jobs join was hanging. Clearly this test was not running before though since the LHS is designed to hang! That Pex invocation neither output a PEX (-o) nor passed CLI args to the ephemeral PEX (--).

jsirois:
The needed background here is that Pex is serial, except for ~3 spots where it uses parallel execution. 2 are thread pool spawns and joins - fan out, fan in - done, nothing happening parallel to the thread pool lifetime. These use bsd locks inside the thread pool. The other case are the execute-parallel where process parallelism is used instead of thread pools. These case too all fully execute the parallel process set and join before doing other work and its these cases that leaked a thread here which could overlap further - supposedly serial - execution and thus foil posix locks mysteriously.

benjyw (approved):

danxmoran (approved):

PR #2053, opened 2023-02-12
Lock sdists with `prepare-metadata-for-build-wheel`.

### LINKED ISSUE #2050: Creating foreign locks (`--style universal` or `--platform` / `--complete-platform`) can fail for some sdists.
Currently locks that contain sdist-artifact-only projects build a wheel from the sdist to get the METADATA needed for the lock:
https://github.com/pantsbuild/pex/blob/9491a4548247f6541e4216246a0a3f9234a0794d/pex/resolve/lockfile/create.py#L197-L205

This can fail if the sdist builds a platform-specific wheel and the local platform is either incompatible with the sdist or does not have the proper libraries installed. Pex should be able to avoid these failures for well behaved platform-specific sdists by using https://peps.python.org/pep-0517/#prepare-metadata-for-build-wheel to get wheel METADATA instead. The bulk of the machinery already exists here and all that needs be changed is to add a new public function that calls the `prepare-metadata-for-build-wheel` PEP-517 hook instead:
https://github.com/pantsbuild/pex/blob/4fd613ab4fe50ac27396f2339ad9aae2d63c398b/pex/build_system/pep_517.py#L79-L136

Unfortunately, this hook is not mandatory though; so a fallback to building a wheel will need to be maintained.

jsirois:
PEP-517 specifies an optional `prepare-metadata-for-build-wheel` hook.
This can be used to extract metadata from an sdist without actually
building a wheel from the sdist when supported by the PEP-517 build
backend. Augment the Pex lock creation code to 1st try this and only
fall back to building a wheel when the optional hook is not present.

This allows locking sdist-only platform-specific projects that are
not compilable on the current platform by avoiding compiling them.

Fixes #2050

### DIFF
--- a/pex/build_system/pep_517.py
+++ b/pex/build_system/pep_517.py
@@ -5,23 +5,25 @@ from __future__ import absolute_import
 
 import os
 import subprocess
-from subprocess import CalledProcessError
+import sys
 from textwrap import dedent
 
 from pex import third_party
 from pex.build_system.pep_518 import BuildSystem, load_build_system
-from pex.dist_metadata import Distribution
+from pex.common import safe_mkdtemp
+from pex.dist_metadata import DistMetadata, Distribution
+from pex.jobs import Job, SpawnedJob
 from pex.pip.version import PipVersion, PipVersionValue
 from pex.resolve.resolvers import Resolver
-from pex.result import Error
+from pex.result import Error, try_
 from pex.tracer import TRACER
-from pex.typing import TYPE_CHECKING, cast
+from pex.typing import TYPE_CHECKING
 from pex.util import named_temporary_file
 
 if TYPE_CHECKING:
-    from typing import Dict, Union
+    from typing import Any, Dict, Iterable, Mapping, Optional, Text, Tuple, Union
 
-_DEFAULT_BUILD_BACKEND = "setuptools.build_meta:__legacy__"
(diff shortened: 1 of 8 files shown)


benjyw (approved):
Do we have a sense of how many sdists (that don't also publish wheels) implement this hook? My guess would have been "not many". But would be great to be wrong about that.

jsirois:
It's not a function of sdists (generally), it's a function of build backends. I tested flit (Pex uses it), Poetry and setuptools. All support it. That's why I jumped through hoops in the test to create an in-tree custom build backend for evdev. I just delegate to setuptools but, importantly, I only delegate the 2 mandatory PEP-517 hooks.

benjyw:
Ah right. I'm surprised that it's possible to implement this hook without cooperation from the sdist, but cool.

jsirois:
Well, setuptools / `setup(...)` is the only bad citizen I know of, all the rest are declarative; as such, is setuptools supports this, then that means ~all sidsts support it. Now, in terms of setuptools, I think it still has to support this by eval'ing the setup.py, but that does not cause C-extensions to build unless the setup.py is hand-crafted nasty. For projects like evdev that use `setuptools.Extension` to point at C-files, the `python setup.py egg_info` distutils Command avoids building those. Looking at the output of the test in verbose mode, setuptools.build_meta is, in fact, calling egg_info.

benjyw (approved):

PR #2056, opened 2023-02-12
Support the latest Pip releases: 22.3.1 & 23.0

jsirois:


.github/workflows/ci.yml
```diff
@@ -65,10 +65,10 @@ jobs:
             pip-version: 20
           - os: ubuntu-20.04
             python-version: [ 3, 11 ]
-            pip-version: 22_2
+            pip-version: 22_3_1
```
jsirois, line 68: I just rolled off 22.2 and 22.3 here; so they are no longer covered. I figure, to be sustainable, I have to do something like this. The rolled-off Pips will have had a good long hammering during their time in CI; and assuming these 2 slots span just 1 ~major version bump (Pip actually uses the last 2 digits of the year for major; so it's not semver major - god love Pip), It seems incredibly unlikely this will lead to issues. If the runtime patches have to change in a way that does not support older versions of Pip, I'll revisit of course, but right now the patches work across all supported Pip versions.

benjyw (approved):

PR #2068, opened 2023-02-25
Introduce `--non-hermetic-venv-scripts`.

### LINKED ISSUE #2065: Pex files do not allow for OpenTelemetry auto-instrumentation
We are using pantsbuild to set up a developer platform with an emphasis on a nice, frictionless developer experience. A part of this experience is codeless auto-instrumentation such as is provided by Opentelemetry with [auto instrumentation](https://opentelemetry.io/docs/instrumentation/java/automatic/). Specifically, we are looking to auto-instrument using the [Kubernetes Operator](https://opentelemetry.io/docs/instrumentation/python/automatic/operator/) offered by OpenTelemetry.

Auto-instrumentation with OpenTelemetry works by monkey-patching libraries, creating passthrough functions that wrap existing functions with logging statements. The way this works is kind of obscure to me, and is something I have to figure out a bit more before I can give constructive input into how to support this.

Auto-instrumentation of non-pexified applications can be as easy as running (for example from a clean Python3.9 docker container):
```
pip install flask opentelemetry-distro opentelemetry-instrumentation-flask

cat << EOF > app.py
import flask
app = flask.Flask(__name__)
@app.route("/")
def helloworld():
  return "<h1>Hello world!</h1>"
app.run()
EOF

opentelemetry-instrument \
  --traces_exporter console \
  --metrics_exporter console \
 python app.py
```

This will spin up a Flask server that also emits telemetry. This separation of code and logging is super useful for maintaining a clean developer experience.

This does not work out-of-the-box with Pex, however. Ideally, one would be able to run the following to achieve the same thing:

```
pip install pex opentelemetry-distro opentelemetry-instrumentation-flask

cat << EOF > app.py
import flask
app = flask.Flask(__name__)
@app.route("/")
def helloworld():
  return "Hello from Pex"
app.run()
EOF

pex flask -o app.pex --exe ./app.py
opentelemetry-instrument \
  --metrics_exporter console \
  --traces_exporter console \
  ./app.pex
```

This does, as expected, not allow OpenTelemetry to inject any instrumentation ,and will not result in any telemetry being emitted. It would be pretty awesome if it did though! I would appreciate any thoughts and ideas for how to implement this.

jsirois:
This provides the same option when building a PEX as the
`--non-hermetic-scripts` does when creating a venv from a PEX using
`pex-tools`. In addition, both means of creating a non-hermetic venv
are extended to the venv `pex` script itself.

Closes #2065

### DIFF
--- a/pex/sh_boot.py
+++ b/pex/sh_boot.py
@@ -175,6 +175,7 @@ def create_sh_boot_script(
         set -eu
 
         VENV="{venv}"
+        VENV_PYTHON_ARGS="{venv_python_args}"
 
         # N.B.: This ensures tilde-expansion of the DEFAULT_PEX_ROOT value.
         DEFAULT_PEX_ROOT="$(echo {pex_root})"
@@ -190,7 +191,12 @@ def create_sh_boot_script(
             # interpreter to use is embedded in the shebang of our venv pex script; so just
             # execute that script directly.
             export PEX="$0"
-            exec "${{INSTALLED_PEX}}/bin/python" -sE "${{INSTALLED_PEX}}" "$@"
+            if [ -n "${{VENV_PYTHON_ARGS}}" ]; then
+                exec "${{INSTALLED_PEX}}/bin/python" "${{VENV_PYTHON_ARGS}}" "${{INSTALLED_PEX}}" \\
+                    "$@"
+            else
+                exec "${{INSTALLED_PEX}}/bin/python" "${{INSTALLED_PEX}}" "$@"
+            fi
         fi
 
         find_python() {{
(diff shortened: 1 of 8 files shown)


tests/integration/venv_ITs/test_issue_2065.py
```diff
@@ -0,0 +1,142 @@
...
+
+if TYPE_CHECKING:
+    from typing import Any
+
+
+def test_venv_pex_script_non_hermetic(tmpdir):
+    # type: (Any) -> None
+
+    # A console script that injects an element in the PYTHONPATH.
+    ot_simulator_src = os.path.join(str(tmpdir), "src")
```
jsirois, line 21: As named, this simulates OpenTelemetry auto-instrumentation - actually turning the manual test  detailed here https://github.com/pantsbuild/pex/issues/2065#issuecomment-1444451584 into an IT seemed too fragile / likely flaky.

pex/sh_boot.py
```diff
@@ -190,7 +191,12 @@ def create_sh_boot_script(
             # interpreter to use is embedded in the shebang of our venv pex script; so just
             # execute that script directly.
             export PEX="$0"
-            exec "${{INSTALLED_PEX}}/bin/python" -sE "${{INSTALLED_PEX}}" "$@"
```
jsirois, line 193: I had missed this 1 spot. That's the last.

benjyw (approved):

PR #2107, opened 2023-03-28
Support --pip overrides via PEX deps.

### LINKED ISSUE #2105: feature(pex.tools): prefer a pex-file-provided package versions when `--pip` flag is also provided
Given a `reqs.pex` file that includes a version of `pip`, `pex-tools reqs.pex --pip` should prefer the versions of packages included in the pex.

This may mean that pex has to downgrade the version of pip (and possibly setuptools, which pip depends on) to match whatever is included in the pex file.

Today, pex creates the virtualenv using the appropriate commands to include `pip`. Then it installs the requirements from the pex file, but any packages that already exist in the virtualenv, like pip, do not get downgraded to the version in the pex file. This feature request should reverse that.

pex-tools should report a warning about the collision between the pex file and the `--pip` flag (since the `--pip` flag is the only reason the virtualenv will have contents before installing things from the pex file). That warning should be silenced by `--collisions-ok`.

### Background: `pants`

When running `./pants export ...`, pants uses pex-tools `venv --pip --collisions-ok` for mutable virtualenvs.
https://github.com/pantsbuild/pants/blob/8fef19f9050d42c949edf781125cdebd6851fc57/src/python/pants/backend/python/goals/export.py#L239-L253
(that link has the following, which I'm including here since github doesn't include cross-repo snippets):
```python
                    complete_pex_env.create_argv(
                        os.path.join(tmpdir_under_digest_root, pex_pex.exe),
                        *[
                            os.path.join(tmpdir_under_digest_root, requirements_pex.name),
                            "venv",
                            "--pip",
                            "--collisions-ok",
                            output_path,
                        ],
                    ),
                    {
                        **complete_pex_env.environment_dict(python=requirements_pex.python),
                        "PEX_MODULE": "pex.tools",
                    },
                ),
```

For StackStorm, we depend on pip+setuptools at runtime, and I want to make sure that the locked version is always used in the venv.

### Slack

This request follows on a discussion in slack:
- [slack question](https://pantsbuild.slack.com/archives/C046T6T9U/p1679940252822339) ([linen mirror](https://chat.pantsbuild.org/t/10116580/u04s45aha-with-pex-tools-1-reqs-pex-venv-pip-some-path-is-th#04e53c65-b5a2-45e9-bef4-77e052c8e920))
- [slack conversation](https://pantsbuild.slack.com/archives/C046T6T9U/p1679940303328629) ([linen mirror](https://chat.pantsbuild.org/t/10116582/there-is-not-you-d-need-to-run-pip-after-creating-the-venv-t#aa48bee8-5a2b-4c32-92e9-599544c57d5e))

jsirois:
Now, when creating a venv using pex tools and the `--pip` option from a
PEX that includes pip or setuptools or both, those embedded versions
will be used in place of the conflicting versions installed via `--pip`
if `--collisions-ok` is also specified; otherwise the venv pex tool will
error.

Closes #2105

### DIFF
--- a/pex/tools/commands/venv.py
+++ b/pex/tools/commands/venv.py
@@ -10,7 +10,6 @@ import subprocess
 from argparse import ArgumentParser
 from collections import OrderedDict
 from subprocess import CalledProcessError
-from typing import Iterable, Union
 
 from pex import pex_warnings
 from pex.common import safe_delete, safe_rmtree
@@ -30,7 +29,7 @@ from pex.venv.pex import populate_venv
 from pex.venv.virtualenv import PipUnavailableError, Virtualenv
 
 if TYPE_CHECKING:
-    from typing import Optional
+    from typing import Iterable, Optional, Union
 
     import attr  # vendor:skip
 else:
@@ -154,10 +153,14 @@ def ensure_pip_installed(
         ),
     )
     if not collisions_ok:
-        return Error("{message}.\nConsider re-running without --pip.".format(message=message))
+        return Error(
+            "{message}\nConsider re-running either without --pip or with --collisions-ok.".format(
+                message=message
+            )
+        )
 
     pex_warnings.warn(
-        "{message}.\nUninstalling venv versions and using versions from the PEX.".format(
+        "{message}\nUninstalling venv versions and using versions from the PEX.".format(
             message=message
         )
     )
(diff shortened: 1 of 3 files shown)


cognifloyd (approved):
This looks great to me. Thanks!

benjyw (approved):

PR #2114, opened 2023-04-16
Add support for Pip 23.1

ShantanuKumar:
The release notes are here:
+ https://pip.pypa.io/en/stable/news/#v23-1
+ https://setuptools.pypa.io/en/stable/history.html#v67-6-1

### DIFF
--- a/pex/pip/version.py
+++ b/pex/pip/version.py
@@ -113,4 +113,11 @@ class PipVersion(Enum["PipVersionValue"]):
         requires_python=">=3.7",
     )
 
+    v23_1 = PipVersionValue(
+        version="23.1",
+        setuptools_version="67.6.1",
+        wheel_version="0.38.4",
+        requires_python=">=3.7",
+    )
+
     VENDORED = v20_3_4_patched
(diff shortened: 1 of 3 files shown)


ShantanuKumar:
@jsirois Tagging you for the review. I mostly copied the changes from https://github.com/pantsbuild/pex/pull/2072

pex/pip/version.py
```diff
@@ -113,4 +113,11 @@ def values(cls):
         requires_python=">=3.7",
     )
 
+    v23_1 = PipVersionValue(
+        version="23.1",
+        setuptools_version="67.6.1",
+        wheel_version="0.38.4",
```
jsirois, line 119: This should be 0.40.0 assuming it works: https://pypi.org/project/wheel/#history
That was released just prior to Pip 23.1 and I aim to take the latest of setuptools and wheel up to the given Pip release date barring other issues.

jsirois (approved):
Thanks @ShantanuKumar. If wheel 0.40.0 doesn't work for some reason, just let me know; otherwise I'll merge this after that upgrade.

pex/pip/version.py
ShantanuKumar: ah ok! Yeah missed that.

PR #2116, opened 2023-04-17
Add support for `--pip-version latest`.

jsirois:
This allows an opt-in to always using the latest Pip supported by Pex
when upgrading through Pex versions. This is generally desirable for
most Pex consumers.

### DIFF
--- a/pex/pip/version.py
+++ b/pex/pip/version.py
@@ -5,6 +5,7 @@ from __future__ import absolute_import
 
 from pex.dist_metadata import Requirement
 from pex.enum import Enum
+from pex.pep_440 import Version
 from pex.targets import LocalInterpreter, Target
 from pex.third_party.packaging.specifiers import SpecifierSet
 from pex.typing import TYPE_CHECKING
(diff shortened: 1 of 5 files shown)


pex/pip/version.py
```diff
@@ -68,7 +78,8 @@ def values(cls):
         return cls._values
 
     v20_3_4_patched = PipVersionValue(
-        version="20.3.4-patched",
+        name="20.3.4-patched",
+        version="20.3.4+patched",
```
benjyw, line 82: I didn't grok why the `name/version` distinction is necessary. For the currently supported versions at least ordering by the (pre-change) `version` seems like it would work? And I don't see anything in the versions of pip releases that would defeat it even if we added new versions in the future.

benjyw (approved):

pex/pip/version.py
jsirois: The `20.3.4-patched` is not a valid PEP-440 version. I did not want to break the CLI value though; so split the name from the version conceptually to handle this 1 case.

ShantanuKumar (approved):

PR #2118, opened 2023-04-17
Expose flags to override `setuptools` and `wheel` version.

### LINKED ISSUE #1895: Expose setuptools and wheel version flags for use with non-vendored Pips.
Right now these are hard-coded but it would be good to allow experimentation with these by end users:
https://github.com/pantsbuild/pex/blob/63592214c569c03b60c59766f96fdd2832e698b6/pex/pip/version.py#L77-L86

Pex does not rely on them in any direct way; so a user should be able to try out / pick up a bug fix in either or even downgrade when prodlems are discovered but not fixed.

danxmoran:
Closes #1895

### DIFF
--- a/pex/pip/version.py
+++ b/pex/pip/version.py
@@ -13,6 +13,20 @@ if TYPE_CHECKING:
     from typing import Iterable, Optional, Tuple
 
 
+def _to_requirement(
+    project_name,  # type: str
+    project_version=None,  # type: Optional[str]
+):
+    # type: (...) -> str
+    return (
+        "{project_name}=={project_version}".format(
+            project_name=project_name, project_version=project_version
+        )
+        if project_version
+        else project_name
+    )
+
+
 class PipVersionValue(Enum.Value):
     def __init__(
         self,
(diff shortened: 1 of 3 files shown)


pex/pip/version.py
```diff
@@ -25,29 +39,24 @@ def __init__(
...
+        self.setuptools_requirement = _to_requirement("setuptools", setuptools_version)
+        self.wheel_requirement = _to_requirement("wheel", wheel_version)
         self.requires_python = SpecifierSet(requires_python) if requires_python else None
 
     @property
     def requirements(self):
         # type: () -> Iterable[str]
         return self.requirement, self.setuptools_requirement, self.wheel_requirement
 
+    def override_setuptools_version(self, version):
```
danxmoran, line 52: Adding methods to allow mutating these previously-immutable (at least as immutable as Python lets you get) enum values makes my skin crawl a bit, but it was the "cleanest" way I could figure out how to expose the overrides. My other ideas involved changing the constructor for `PipVersion` or `PipVersionValue`, which would have sprawled to every usage of a `PipVersion.xyz` in the codebase...

jsirois: Agreed - this is not awesome. It seems like the right thing to do is separate setuptools and wheel versions from the PipVersion enum; so you have just the PipVersion enum, then an external default mapping of those to setuptools and wheel versions and the new flag overrides that external default mapping.

danxmoran: 👍 I'll rework

PR #2128, opened 2023-04-21
Fix CI.

jsirois:
The April 19th release of virtualenv 20.22 breaks the resolve of
poetry 1.3.2.

### DIFF
--- a/tests/integration/test_issue_2038.py
+++ b/tests/integration/test_issue_2038.py
@@ -26,8 +26,16 @@ if TYPE_CHECKING:
 def test_wheel_file_url_dep(tmpdir):
     # type: (Any) -> None
 
+    constraints = os.path.join(str(tmpdir), "constraints.txt")
+    with open(constraints, "w") as fp:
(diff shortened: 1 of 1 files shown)


kaos (approved):

Eric-Arellano (approved):

tests/integration/test_issue_2038.py
```diff
@@ -26,8 +26,16 @@
 def test_wheel_file_url_dep(tmpdir):
     # type: (Any) -> None
 
+    constraints = os.path.join(str(tmpdir), "constraints.txt")
+    with open(constraints, "w") as fp:
+        # The 20.22 release introduces a change that breaks resolution od poetry 1.3.2; so we pin
```
benjyw, line 31:
```suggestion
        # The 20.22 release introduces a change that breaks resolution of poetry 1.3.2; so we pin
```

Typo, but not worth a CI round. Could tack it onto a future change.

benjyw (approved):

tests/integration/test_issue_2038.py
jsirois: Fixed in #2129.

Issue #2110, opened 2023-04-02
Support equivalent of `pip install --target=...` to export distributions and source to flat directory


huonw:
`pip install` has a `--target`/`-t` option that installs the requested requirements directly to a specific directory, rather than into the usual structure with `site-packages` etc. It would be nice if pexes could support this too. Currently, there's two related `PEX_TOOLS=1` commands, but neither quite do exactly this:

- `PEX_TOOLS=1 ./some-pex venv ...`: constructs the full venv structure
- `PEX_TOOLS=1 ./some-pex repository extract ...`: only exports the distributions, not the source

Motivation: Direct installation like this is handy for building artefacts for certain cases, such as FaaS platforms like Lambda: their docs recommend just splatting all dependencies into the top level of a zip file, along with whatever first party code: https://docs.aws.amazon.com/lambda/latest/dg/python-package.html#python-package-create-package-with-dependency (steps 4-6).

More discussion: https://pantsbuild.slack.com/archives/C046T6T9U/p1680247328906249

huonw:
Looking at the code that handles this in `pip` suggests they do a 'normal' install to a temporary directory, and then move some of the files from that install into the target directory: https://github.com/pypa/pip/blob/51afe45cda2854d810d9714e3891e256b1509737/src/pip/_internal/commands/install.py#L514-L566

This would suggest to me that, for a lot of use cases, just doing a venv export and then copying out the `lib/.../site-packages` directory might be enough, without having to implement this dedicated feature?

jsirois:
> `PEX_TOOLS=1 ./some-pex repository extract ...`: only exports the distributions, not the source
Just noting this is not true. There is a --source option that does create and export an sdist of all the 1st party PEX code with proper `install_requires` metadata such that you can just grab that sdist alone and then copy it somewhere and `pip install the-sdist.tar.gz` and get everything back (non-reproducibly of course since `install_requires` entries may not be pinned, and even if they are they will almost never include the full transitive set; so transitive deps may be unpinned).

jsirois:
This is getting tacked on to the feature work for #1752. There will be a `--flat` (not sure of spelling yet) flag you can tack on to `pex3 venv -d right/here {--python,--platform,--complete-platform}* reqs*|--pex PEX|--lock LOCK --scope {all,deps,srcs}` to just export to a flat dir instead of a venv site packages dir.

Issue #1752, opened 2022-05-09
Pex should support creating a venv directly


zmanji:
With the lockfile and venv features it's very easy now to create reproducible venvs with pex. It's very handy for setting up development environments without Docker and you can ensure all of the dependencies are wheels and not sdists.

I have been doing this:

1. Create a lock file with `pex3 lock create`. This lockfile contains the resolves for all third party dependencies for all platforms I care about. I can use `--no-build` to ensure all of the dependencies are wheels.
2. With the above lockfile create a dependencies pex with `pex  --lock pex.lock --venv -o depencies.pex`
3. Create the venv:  `PEX_TOOLS=1 ./dependencies.pex venv  ./venv`
4. Iterate on my first party code by running `PEX_EXTRA_SYS_PATH='./srcs' ./out.pex.venv/pex -m 'mymodule`. In my case `srcs` contains multiple packages.

I install the virtualenv for two reasons:

1. Having a virtualenv allows for tools like PyCharm to pick up all of my dependencies.
2. On my machine, I have observed that executing the pex directly has about 30ms of overhead which is just enough for me to go through with step 3 above.

I think pex should be able to collapse steps 2 and 3 above into a single command invocation. I would envision something like:
```
pex3 venv create --lock pex.lock -o venv
```

jsirois:
FWIW @zmanji the new `--sh-boot` combined with `--venv` should drop the overhead of warm invocations over native venv to ~1ms. So FYI for those cases where you don't need to know the venv path to tell an IDE.

jsirois:
I ended up providing `pex3 lock {export,subset,venv} ...` instead. The `export` sub-command already existed but now supports passing `-r` requirements files to use to subset the lock. The `subset` and `venv` commands are new and both support optional sub-setting. The `subset` sub-command creates a new lockfile with a transitive closure subset of the original lock file. The `venv` sub-command` is just like the PEX_TOOLS `venv` tool, except that it is faster. I'm using the latter with nox which has both sped up session set up enough that running my full suite is now ~40% faster than using nox's `session.install(...)` and made all tooling and packaging reproducible. PR(s) coming shortly.

PR #2135, opened 2023-04-28
Introduce `pex3 venv inspect`.

jsirois:
This command allows inspecting venvs created by Pex as well as those
created by other tools.

Work towards #1752 and #2110

### DIFF
--- a/pex/venv/virtualenv.py
+++ b/pex/venv/virtualenv.py
@@ -11,6 +11,7 @@ import shutil
 import sys
 from contextlib import closing
 from fileinput import FileInput
+from textwrap import dedent
 
 from pex.atomic_directory import AtomicDirectory, atomic_directory
 from pex.common import is_exe, safe_mkdir, safe_open
@@ -18,12 +19,13 @@ from pex.compatibility import commonpath, get_stdout_bytes_buffer
 from pex.dist_metadata import Distribution, find_distributions
 from pex.executor import Executor
 from pex.fetcher import URLFetcher
-from pex.interpreter import PythonInterpreter
+from pex.interpreter import PythonInterpreter, PyVenvCfg
 from pex.orderedset import OrderedSet
 from pex.tracer import TRACER
 from pex.typing import TYPE_CHECKING, cast
 from pex.util import named_temporary_file
 from pex.variables import ENV
+from pex.version import __version__
 
 if TYPE_CHECKING:
     from typing import Iterator, Optional, Tuple, Union
@@ -116,6 +118,8 @@ def find_site_packages_dir(
 
 
 class Virtualenv(object):
+    VIRTUALENV_VERSION = "16.7.12"
+
     @classmethod
     def enclosing(cls, python):
         # type: (Union[str, PythonInterpreter]) -> Optional[Virtualenv]
@@ -174,7 +178,9 @@ class Virtualenv(object):
         ):
             # N.B.: PyPy3.6 and PyPy3.7 come equipped with a venv module but it does not seem to
             # work.
-            virtualenv_py = pkgutil.get_data(__name__, "virtualenv_16.7.12_py")
+            virtualenv_py = pkgutil.get_data(
+                __name__, "virtualenv_{version}_py".format(version=cls.VIRTUALENV_VERSION)
+            )
             with named_temporary_file(mode="wb") as fp:
                 fp.write(virtualenv_py)
                 fp.close()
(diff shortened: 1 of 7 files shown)


jsirois:
N.B.: `pex3 venv create` will follow this, but this establishes the tool sub-command namespace and is useful in its own right to, for example, determine the proper site-packages directory for a pex venv (there can be more than one) if you want to copy out its contents.

jsirois:
For example, a Pants venv created by the `./pants` script:
```
$ python -mpex.cli venv inspect ~/.cache/pants/setup/bootstrap-Linux-x86_64/2.14.1_py39/ -i2
```
```json
{
  "venv_dir": "/home/jsirois/.cache/pants/setup/bootstrap-Linux-x86_64/2.14.1_py39",
  "provenance": {
    "created_by": "virtualenv 20.4.7",
    "is_pex": false,
    "pex_version": null
  },
  "include_system_site_packages": false,
  "interpreter": {
    "binary": "/home/jsirois/.cache/pants/setup/bootstrap-Linux-x86_64/2.14.1_py39/bin/python",
    "base_binary": "/usr/bin/python3.9",
    "version": "3.9.16",
    "sys_path": [
      "/usr/lib/python39.zip",
      "/usr/lib/python3.9",
      "/usr/lib/python3.9/lib-dynload",
      "/home/jsirois/.cache/pants/setup/bootstrap-Linux-x86_64/2.14.1_py39/lib/python3.9/site-packages"
    ]
  },
  "script_dir": "/home/jsirois/.cache/pants/setup/bootstrap-Linux-x86_64/2.14.1_py39/bin",
  "scripts": [
    "normalizer",
    "pants",
    "pex",
    "pex-tools",
    "pex3",
    "pip",
    "pip3",
    "pip3.9",
    "wheel",
    "wheel-3.9",
    "wheel3",
    "wheel3.9"
  ],
  "site_packages": "/home/jsirois/.cache/pants/setup/bootstrap-Linux-x86_64/pants.2KoUMU/install/lib/python3.9/site-packages",
  "distributions": [
    "PyYAML==6.0",
    "ansicolors==1.1.8",
    "certifi==2022.12.7",
    "charset-normalizer==3.0.1",
    "fasteners==0.16.3",
    "humbug==0.2.7",
    "idna==3.4",
    "ijson==3.1.4",
    "importlib-resources==5.0.7",
    "packaging==21.3",
    "pantsbuild.pants==2.14.1",
    "pex==2.1.108",
    "pip==23.0",
    "psutil==5.9.0",
    "pyparsing==3.0.9",
    "python-lsp-jsonrpc==1.0.0",
    "requests==2.28.2",
    "setproctitle==1.2.2",
    "setuptools==63.4.3",
    "six==1.16.0",
    "toml==0.10.2",
    "types-PyYAML==6.0.3",
    "types-setuptools==62.6.1",
    "types-toml==0.10.8",
    "typing_extensions==4.3.0",
    "ujson==5.7.0",
    "urllib3==1.26.14",
    "wheel==0.36.2"
  ]
}
```

zmanji (approved):
This code lgtm.

It might be a good idea to write an integration test against a more modern virtualenv or other widely used virtualenv creating tool (pipenv?).

jsirois:
> It might be a good idea to write an integration test against a more modern virtualenv or other widely used virtualenv creating tool (pipenv?).

Good, idea - done for virtualenv. Pipenv is not even official any more if you follow PyPA discourse ... it's Hatch. I'll just stick to virtualenv if you agree since that invented all this and is used by several tools behind the scenes. I may rue the choice though, I'm sure there are PEP-405 violators out there and I'll find that out the hard way :/.

zmanji (approved):

pex/venv/virtualenv.py
```diff
@@ -250,6 +278,26 @@ def venv_dir(self):
...
 
+    @property
+    def _pyvenv_cfg(self):
+        # type: () -> Optional[PyVenvCfg]
+        if not hasattr(self, "__pyvenv_cfg"):
+            self.__pyvenv_cfg = PyVenvCfg.find(self._interpreter.binary)
+        return self.__pyvenv_cfg
+
+    @property
+    def created_by(self):
```
huonw, line 289: The code that writes a `pyvenv.cfg` file above explicitly sets `created_at = ...`, but it seems like this `created_at` method wouldn't reflect that value back out, instead showing `virtualenv ...`; is that expected?

huonw (approved):

pex/venv/virtualenv.py
jsirois: It is expected. I did choose a confusing name I guess reading the code, but the name was picked for the reader of the file.

The `created_by`, as with all fields except `home` and `include-system-site-packages`, are arbitrary an un-defined by PEP-405. I just wanted to leave a marker that this was _not_ created by virtualenv to help avoid confusing someone poking about in the venv later. Although modern virtualenv does create a pyvenv.cfg, this ancient single file version Pex vendors does not.

Clearly though I'm just causing different confusion! I did consider emitting a `#`-style comment line with the same content.

huonw: All good, sounds reasonable. If it really mattered (it doesn't to me), one could use a synonym in the `pyvenv.cfg`, e.g. `creator = ...` or something, unless there's some non-PEP-405 convention that's guiding the choice.

jsirois: No convention. There is in fact no similar field at all in any pyvenv.cfg I've seen. This appears to be unique.

jsirois: I may do something about this, at least comment, in the `pex3 venv create` PR CI burn cycle.

PR #2145, opened 2023-05-04
Add `pex3 lock export-subset`.

jsirois:
Due to lack of foresight, `pex3 lock export` has a CLI that is poisoned
for use in a consistent way with the rest of PEX while at the same time
introducing subset support; so an `export-subset` sub-command is broken
out sharing the vast majority of the `export` code.

This is immediately useful to a-scie/lift for supporting Pex lock files
on Windows where there is no Pex support yet but seems useful in general
on the same interoperability front the `export` goal was introduced for:
speaking the ~lingua-franca of Pip.

### DIFF
--- a/pex/cli/commands/lock.py
+++ b/pex/cli/commands/lock.py
@@ -33,6 +33,7 @@ from pex.resolve.lockfile.updater import (
     VersionUpdate,
 )
 from pex.resolve.path_mappings import PathMappings
+from pex.resolve.requirement_configuration import RequirementConfiguration
 from pex.resolve.resolved_requirement import Fingerprint, Pin
 from pex.resolve.resolver_options import parse_lockfile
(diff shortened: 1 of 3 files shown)


zmanji:
The code and tests lgtm. I'm not an expert as to how `lock export` works but the behaviour in the tests lgtm.

zmanji (approved):

stuhood (approved):

benjyw (approved):
Do I understand correctly that the CLI poison is the use of a positional arg for the lockfile, while also wanting to use positional args for the requirement subset? It would be possible to have the first positional arg be the lockfile and subsequent ones be the requirements, no? I agree that is janky though.

jsirois:
Exactly. Yeah, that seems like a confusing, or at least easy to get wrong, interface for users.

PR #2157, opened 2023-06-10
Fix poetry bitrot in an IT.

jsirois:


tests/integration/test_issue_2038.py
```diff
@@ -30,7 +30,10 @@ def test_wheel_file_url_dep(tmpdir):
     with open(constraints, "w") as fp:
         # The 20.22 release introduces a change that breaks resolution of poetry 1.3.2; so we pin
         # low.
-        fp.write("virtualenv<20.22")
+        print("virtualenv<20.22", file=fp)
+        # The poetry-plugin-export 1.4.0 release requires poetry>1.5 but poetry 1.3.2 floats the
```
benjyw, line 34: They have a circular dependency?

benjyw (approved):

tests/integration/test_issue_2038.py
jsirois: Yup! Pip handles those, the version constraints were just conflicting here. It was `poetry old -> ... <The plugin loose> -> poetry newer`.

Issue #2160, opened 2023-06-20
Bad `installed_wheels` dir when a wheel's `top_level.txt` dirname collides with `sources`


thejcannon:
Steps to reproduce:

Here's a `pex` (renamed to `.zip` for GitHub's sake) [local.zip](https://github.com/pantsbuild/pex/files/11805798/local.zip). It has a `top_level.txt` of `top_level`. Rename to `local.whl`

Then, next to that file:
```console
$ mkdir -p sources/top_level/
$ echo 'print("hello")' > sources/top_level/mymain.py
$ pex --output-file binary.pex --venv prepend --requirements-pex local.pex --entry-point top_level.mymain $'--sources-directory=sources' --layout packed
$ ./binary/__main__.py
hello
$ ls ~/.pex/installed_wheels/a571cccbc564d1e7da4efe2416aca4bb049256ab4d0b30553f9d4de8207fb283/top_level-0.0.0-cp38-cp38-linux_x86_64.whl/top_level/
__init__.py  mymain.py  __pycache__
```

`mymain.py` is not part of the wheel, and shouldn't be in there.

jsirois:
There is a bit of chaff in the example. The key bit is `--venv` and _not_ also `--venv-site-packages-copies` (for example, layout is irrelevant) - use `--venv-site-packages-copies`and all is well.

With just `--venv` you see:
```
$ ./binary.pex
hello
$ ls -l ~/.pex/venvs/s/fe373d35/venv/lib/python3.8/site-packages/
total 12
-rw-r--r-- 1 jsirois jsirois  230 Jun 30 18:14 PEX_EXTRA_SYS_PATH.pth
lrwxrwxrwx 1 jsirois jsirois  152 Jun 30 18:14 top_level -> ../../../../../../installed_wheels/a571cccbc564d1e7da4efe2416aca4bb049256ab4d0b30553f9d4de8207fb283/top_level-0.0.0-cp38-cp38-linux_x86_64.whl/top_level
drwxr-xr-x 2 jsirois jsirois 4096 Jun 30 18:14 top_level-0.0.0.dist-info
```

And that should be enough to see the issue. Now, for 3rdparty deps, this sort of collision (namespace package) is handled via a `pex-ns-pkgs.pth` you don't see here:
https://github.com/pantsbuild/pex/blob/c2fa8130d3dfb38f19f550a5bb0a3c69b7459f0d/pex/venv/installer.py#L459-L495
[...]

PR #2162, opened 2023-07-01
Fix CI bitrots.

jsirois:
Several issues are fixed:
+ IPython can inject ~PS1-style prefixes depending on the IPython 
  version and Python version in-play. Avoid having to worry about this
  with a dedicated test communication channel separate from stdio.
+ The PyPy 7.3.12 release includes HPy in the stdlib (including its
  dist-info) which highlights a bug in `Virtualenv` when iterating venv
  distributions; so we stick to iterating distributions found in the
  site-packages directories.
+ The setup-python action no longer supports Python 2.7; so we switch to
  pyenv for that.
+ A test involving `jaraco-collections==3.5.1` is broken by the recent
  pydantic 2.x release; so we bound low.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -186,13 +186,6 @@ jobs:
     strategy:
       matrix:
         include:
-          # N.B.: We need 20.04 for Python 2.7 tool cache support.
-          - os: macos-12
-            python-version: [ 2, 7 ]
-            pip-version: 20
-          - os: ubuntu-20.04
-            python-version: [ 2, 7 ]
-            pip-version: 20
           - os: ubuntu-22.04
             python-version: [ 3, 7 ]
             pip-version: 22_3_1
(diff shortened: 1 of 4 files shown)


jsirois:
I encountered a local failure in the test's pex_root prefix test while working #2160.

.github/workflows/ci.yml
```diff
@@ -48,16 +48,6 @@ jobs:
     strategy:
       matrix:
         include:
-          # N.B.: We need 20.04 for Python 2.7 & 3.5 tool cache support.
-          - os: macos-12
-            python-version: [ 2, 7 ]
```
jsirois, line 53: These did in fact go away on June 19th; so fixing this as well in this PR.

jsirois:
I'm going to land this to get CI green, land #2164 and continue on Python 3.12 support which #2164 affects.

benjyw (approved):

PR #2167, opened 2023-07-02
Automate release announcements.

jsirois:
This re-uses the scie-pants & a-scie infra and converts CHANGES.rst to
CHANGES.md so that infra can be used as-is.


.github/workflows/release.yml
```diff
@@ -84,19 +84,51 @@ jobs:
...
-
-            ## ${{ needs.determine-tag.outputs.release-version }}
-
-            TODO: Add CHANGES.rst entries.
+          body_path: ${{ steps.prepare-changelog.outputs.changelog-file }}
           draft: false
           prerelease: false
           files: dist/pex
+          fail_on_unmatched_files: true
+          discussion_category_name: Announcements
```
jsirois, line 105: I created the Announcements GH discussions category just now.

.github/workflows/release.yml
```diff
@@ -84,19 +84,51 @@ jobs:
...
+                  "type": "section",
+                  "text": {
+                    "type": "mrkdwn",
+                    "text": "Pex ${{ needs.determine-tag.outputs.release-version }} is released:\n* https://pypi.org/project/pex/${{ needs.determine-tag.outputs.release-version }}/\n* https://github.com/pantsbuild/pex/releases/tag/${{ needs.determine-tag.outputs.release-tag }}"
+                  }
+                }
+              ]
+            }
+        env:
+          SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
```
jsirois, line 134: I added pantsbuild/pex to the list of repos this org secret is exposed to.

benjyw: Great, you noticed I added it at the org level. This should mean we can delete it from scie-pants, but I wanted to test first.

benjyw:
BTW I have set up SES and had it moved out of the AWS sandbox, so we can also send email automatically now. I will get that set up in the main pantsbuild repo soon to demonstrate.

jsirois:
Ok. I have no clue what an SES is. Pex has never sent an email so I don't think I'll start unless the pantsbuild org has some strong reason to.

jsirois:
I imagine you know about this, but in case not, seems free no extra AWS needed?: https://github.com/marketplace/actions/send-email

Last I used Google smtp was in maybe 2009, but it worked then anyhow for a personal account like pantsbuild@gmail.com.

benjyw:
Yeah, I looked at that, but AFAICT it means CI must know pantsbuild@gmail.com's account password, which, since we use that account for a bunch of things, seemed imprudent. 

I suppose I could have created pantsbuild-announce@gmail.com or something, but SES was easy to set up, and we already use AWS for other stuff.

And that was just FYI, I see no need to introduce new emails where there weren't any before.

benjyw (approved):

PR #2170, opened 2023-07-05
Adapt to removed stdlib APIs in Python 3.12.

### LINKED ISSUE #2137: Replace deprecated imp module with importlib
This project uses the [`imp` module](https://docs.python.org/3/library/imp.html) which has been deprecated since Python 3.4 and removed in 3.12:

* Raised `PendingDeprecationWarning` since [3.4 (2014)](https://github.com/python/cpython/commit/e4f41deccf94ccc798b1eb1f44657ade66669a60)
* Raised `DeprecationWarning` since [3.5 (2015)](https://github.com/python/cpython/commit/c0d91aff9a3b91307b26e8b7c34dfbf27bbdd43a)
* Updated `DeprecationWarning` to say removal in 3.12 since [3.10 (2021)](https://github.com/python/cpython/commit/dc6d3e1e4c0c1e4b2210edab8fb4762569dc2936)
* Removal planned for [3.12](https://github.com/python/cpython/issues/98040) [(2023)](https://github.com/python/cpython/pull/98573)

Python 3.12 is [set for release on 2023-10-02](https://devguide.python.org/versions/) and this library is one of the [top 5,000 most-downloaded from PyPI](https://hugovk.github.io/top-pypi-packages/).

Please could you upgrade to use `importlib`? The [`imp` docs](https://docs.python.org/3.11/library/imp.html) have suggestions on what to use to replace each function and constant.

jsirois:
Closes #2137

### DIFF
--- a/pex/pex_builder.py
+++ b/pex/pex_builder.py
@@ -123,12 +123,12 @@ elif '__loader__' in locals():
     __entry_point__ = __loader__.archive
   else:
     try:
-      from pkgutil import ImpLoader
-      if isinstance(__loader__, ImpLoader):
-        __entry_point__ = __entry_point_from_filename__(__loader__.get_filename())
-    except ImportError:
       from importlib.abc import Loader
       if isinstance(__loader__, Loader):
+        __entry_point__ = __entry_point_from_filename__(__loader__.get_filename())
+    except ImportError:
+      from pkgutil import ImpLoader
+      if isinstance(__loader__, ImpLoader):
         __entry_point__ = __entry_point_from_filename__(__loader__.get_filename())
 
 if __entry_point__ is None:
--- a/pex/pex_builder.py
+++ b/pex/pex_builder.py
@@ -121,15 +121,11 @@ if '__file__' in locals() and __file__ is not None and os.path.exists(__file__):
 elif '__loader__' in locals():
   if hasattr(__loader__, 'archive'):
     __entry_point__ = __loader__.archive
-  else:
-    try:
-      from importlib.abc import Loader
-      if isinstance(__loader__, Loader):
-        __entry_point__ = __entry_point_from_filename__(__loader__.get_filename())
-    except ImportError:
-      from pkgutil import ImpLoader
-      if isinstance(__loader__, ImpLoader):
-        __entry_point__ = __entry_point_from_filename__(__loader__.get_filename())
+  elif hasattr(__loader__, 'get_filename'):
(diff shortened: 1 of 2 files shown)


jsirois:
This was proved out in my Python 3.12 branch here: https://github.com/pantsbuild/pex/pull/2168
Broken off as an independent chunk. This PR just proves this works with all the existing Pythons.

pex/pex_builder.py
```diff
@@ -123,13 +123,13 @@ def __entry_point_from_filename__(filename):
...
   else:
     try:
-      from pkgutil import ImpLoader
-      if isinstance(__loader__, ImpLoader):
-        __entry_point__ = __entry_point_from_filename__(__loader__.get_filename())
-    except ImportError:
       from importlib.abc import Loader
       if isinstance(__loader__, Loader):
         __entry_point__ = __entry_point_from_filename__(__loader__.get_filename())
+    except ImportError:
```
jsirois, line 129: This order is a bit better since it stresses the code now as well as getting rid of warnings under newer Pythons. It penalizes existing Python 2.7 users with a small speed bump for the cold cache case in the ~ms range though.

Eric-Arellano (approved):
Thank you for splitting this out. It may be worth a comment in the new code explaining which branch corresponds to which Python versions.

benjyw (approved):
Agree with Eric re documenting the control flow

pex/pex_builder.py
jsirois: Alright - took the feedback but went further given the prompt to think.

Eric-Arellano (approved):

pex/pex_builder.py
```diff
@@ -119,10 +119,12 @@ def __entry_point_from_filename__(filename):
 if '__file__' in locals() and __file__ is not None and os.path.exists(__file__):
   __entry_point__ = __entry_point_from_filename__(__file__)
 elif '__loader__' in locals():
-  from pkgutil import ImpLoader
   if hasattr(__loader__, 'archive'):
     __entry_point__ = __loader__.archive
-  elif isinstance(__loader__, ImpLoader):
+  elif hasattr(__loader__, 'get_filename'):
```
jsirois, line 124: Although it's still true Pex contains vendored code that uses ImpLoader (inside Pip), this adjustment as well as the support for find_spec / [PEP-451](https://peps.python.org/pep-0451) below is enough to close #2137.

PR #2171, opened 2023-07-05
Properly cache pyenv interpreters.

jsirois:
This also expands scope to cover all deprecated (or soon-to-be)
interpreters we use since GitHub has now proved they do yank support
eventually.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -118,7 +118,7 @@ jobs:
       - name: Checkout Pex
         uses: actions/checkout@v3
       - name: Setup Python ${{ join(matrix.python-version, '.') }}
-        uses: jsirois/actions/pyenv@6b4b66ba805bdd553071b95d946467a64df55029
+        uses: pantsbuild/actions/pyenv@1356087bfd1be04670c2ffde4ba94a9c6898ecd7
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -117,10 +118,9 @@ jobs:
       - name: Checkout Pex
         uses: actions/checkout@v3
       - name: Setup Python ${{ join(matrix.python-version, '.') }}
-        uses: gabrielfalcao/pyenv-action@v14
```
jsirois, line 120: As noted in https://github.com/pantsbuild/actions/pull/24 - the caching for gabrielfalcao/pyenv-action is manual and I could not get it to work; thus the switch here.

.github/workflows/ci.yml
```diff
@@ -117,10 +118,9 @@ jobs:
...
+        uses: jsirois/actions/pyenv@6b4b66ba805bdd553071b95d946467a64df55029
```
jsirois, line 121: I'll flip from jsirois to pantsbuild once https://github.com/pantsbuild/actions/pull/24 lands.

huonw (approved):

.github/workflows/ci.yml
jsirois: Done.

PR #2172, opened 2023-07-06
Add support for Python 3.12.

jsirois:
The crux here is supporting a version of Pip that works in 3.12. There
is no such released version yet; so this change adds an unreleased
Pip version but goes to some length to hide this version from users
and make it only activatable by those in the know / CI. What follows
is fixing or adjusting many tests. The result is Pex known to work with
Python 3.12 ahead of its release by several months and the spectre of
Pex 3 / a Pex branch split, etc., dispelled.

It turns out Pex can still ship supporting Python 2.7, 3.5, etc.
along side supprting 3.12. The main trick here is to use
`python3.12 -mvenv` to spirit up a bootstrap Pip that works at least
enough to install the unreleased Pip that truly works with Python 3.12.
Previously all Pip version bootstrapping was handled exclusively by the
vendored Pip.


pex/pip/installation.py
```diff
@@ -82,19 +88,61 @@ def _vendored_installation(interpreter=None):
     )
 
 
+def _bootstrap_pip(
```
jsirois, line 91: This is the new trick that allows a Pip that supports Python 3.12 to be bootstrapped without resorting to a resolve using the vendored Pip (the technique on line 142 below). Not much of a trick, pretty obvious in hindsight.

pex/pip/version.py
```diff
@@ -65,16 +85,51 @@ def requires_python_applies(self, target):
...
         if not hasattr(self, "_latest"):
-            self._latest = max(PipVersionValue._iter_values(), key=lambda pv: pv.version)
+            self._latest = max(
+                (version for version in PipVersionValue._iter_values() if not version.hidden),
+                key=lambda pv: pv.version,
+            )
         return self._latest
 
 
+class DefaultPipVersion(object):
```
jsirois, line 95: Previously VENDORED was the global default but that no longer works. This allows dynamically choosing the correct default given the current ambient interpreter: VENDORED for everything <3.12, unreleased Pip (for now) for everything newer.

.github/workflows/ci.yml
```diff
@@ -51,15 +51,27 @@ jobs:
...
             pip-version: 22_3_1
+            tox-env-python: python
           - os: ubuntu-22.04
             python-version: [ 3, 11 ]
             pip-version: 23_1_2
+            tox-env-python: python
+          - os: macos-12
+            python-version: [ 3, 12, "0-beta.3" ]
+            pip-version: 23_2
+            tox-env-python: python3.11
```
jsirois, line 70: It turned out tox doesn't work (or I couldn;t get it to work) with 3.12; thus this complication. More generally, trying to support a new Python version, especially when it yanks core stdlib like distutils, is hard if you're early. Eventually tox will support 3.12 and this complication will be removed.

pex/pip/version.py
```diff
@@ -93,57 +149,67 @@ def values(cls):
...
+        requires_python=">=3.7,<3.12",
+    )
+
+    v23_2 = PipVersionValue(
+        version="23.2.dev0+8a1eea4a",
+        requirement="pip @ git+https://github.com/pypa/pip@8a1eea4aaedb1fb1c6b4c652cd0c43502f05ff37",
+        setuptools_version="67.8.0",
+        wheel_version="0.40.0",
         requires_python=">=3.7",
+        hidden=True,
```
jsirois, line 210: Some complication was added here to be able to use a Pip version I want to yank, but still uphold the guaranty of not breaking users. This flag and the attendant complication above all work to make it so a release of Pex between now and the moment it can be upgraded to a Pip that support Python 3.12 won't expose a version of Pip on the CLI that will later get yanked.

pex/environment.py
```diff
@@ -651,6 +655,20 @@ def _declare_namespace_packages(cls, resolved_dists):
                 )
                 for index, (dist, ns_packages) in enumerate(namespace_packages_by_dist.items())
             )
+            if not pkg_resources:
```
jsirois, line 658: This change and many test fixes revolve around the full switch to `importlib.resources` in Python 3.12. The setuptools maintainers are killing `pkg_resources` as of Python 3.12 and `pkg_resources`, like Pex, relies on import hooks to enable vendored code it uses. The setuptools did not update those hooks to work in a Python 3.12 world (Pex had to in #2170); so even when `pkg_resources` is present, it fails to import under 3.12.

tests/integration/test_issue_940.py
```diff
@@ -2,14 +2,38 @@
...
+
+import pytest
 
 from pex.common import temporary_dir
 from pex.testing import built_wheel, run_pex_command, run_simple_pex
 
 
+@pytest.mark.skipif(
+    sys.version_info[:2] >= (3, 12),
+    reason="We need to use setuptools<66 ut Python 3.12+ require greater.",
```
Eric-Arellano, line 16:
```suggestion
    reason="We need to use setuptools<66 but Python 3.12+ require greater.",
```

tox.ini
```diff
@@ -17,6 +19,9 @@ commands =
     python scripts/print_env.py
 
 [testenv]
+# N.B.: We need modern setuptools downloaded OOB by virtualenv to work with the Python 3.12 case.
```
Eric-Arellano, line 22: Nit so the comment is more accessible:

```suggestion
# N.B.: We need modern setuptools downloaded out-of-bound by virtualenv to work with the Python 3.12 case.
```

tox.ini
```diff
@@ -3,6 +3,8 @@ isolated_build = true
 skip_missing_interpreters = True
 minversion = 3.25.1
 requires =
+    # Ensure tox and virtualenv compatible back through 2.7.
+    tox<4
```
Eric-Arellano, line 7: To make clear you're not talking about Tox 2:

```suggestion
    # Ensure tox and virtualenv compatible back through Python 2.7.
    tox<4
```

pex/common.py
```diff
@@ -630,6 +630,8 @@ def get_parent_dir(path):
 
             def maybe_write_parent_dirs(path):
                 # type: (str) -> None
+                if path == strip_prefix:
+                    return
```
Eric-Arellano, line 634: What's the context for this change?

pex/environment.py
```diff
@@ -651,6 +655,20 @@ def _declare_namespace_packages(cls, resolved_dists):
...
+                current_interpreter = PythonInterpreter.get()
+                pex_warnings.warn(
+                    "The legacy `pkg_resources` package cannot be imported by the "
+                    "{interpreter} {version} interpreter at {path}.\n"
+                    "These distributions will fail to work properly:\n{dists}".format(
```
Eric-Arellano, line 663: Should this be an eager error instead? The warning below makes sense—we can salvage the situation but it's not kosher. Here, this will eventually result in an error for the user, right?

Eric-Arellano: Also maybe worth mentioning this line:

> These distributions should fix their `install_requires` to include `setuptools

Otherwise the user doesn't know a possible path forward for this issue, unless they already are familiar with this niche problem.

Eric-Arellano (approved):
Excellent. Thank you, John.

I'm comfortable with the PR size. But fwiw, if you want to make it smaller so we can give stronger reviews, it could be worth breaking out the `PipVersion.DEFAULT` change in a precursor PR to remove a lot of the noise here.

tox.ini
jsirois: I guess you proved your point. It's out of band.

pex/common.py
jsirois: Good catch actually. I missed an item on my cleanup checklist. This gets rid of a `..` entry in the `--layout packed` zips currently created by Pex. In Python 3.12 zipfile, these are now illegal. I need to change the cache format so that new Pex under 3.12 does not see the old zips since it will blow up if it does.

jsirois:
> PipVersion.DEFAULT

That does not cleanly break out, it needs a new PipVersion to make any sense in isolation and bringing that in brings in the rest.

tests/integration/test_issue_940.py
jsirois: Fixed.

tox.ini
jsirois: Fixed.

tox.ini
jsirois: Fixed.

pex/environment.py
jsirois: An eager error would be bad: we tank the ship for an import that may never happen. When a tool tries to do too much, there is a very maddening side to it and its this.

Your second suggestion is reasonable though.

jsirois: Actually its not really reasonable, since that is not a path forward. The distributions simply won't work at all under 3.12 (unless non-ns-package parts are imported per the point above, in which case moot).

pex/common.py
jsirois: Ok, I needed to refresh myself. Without this (the old behavior) you get packed wheel zips like:
```
Archive:  /home/jsirois/.pex/installed_wheel_zips/858d7ad606d58d8b13b4ac842dd378ff37e94387abc4b79e281576af39c64a49/py_spy-0.3.8-py2.py3-none-manylinux_2_5_x86_64.manylinux1_x86_64.whl
Zip file size: 3126344 bytes, number of entries: 10
drwxr-xr-x  2.0 unx        0 b- defN 80-Jan-01 00:00 ../
-rw-r--r--  2.0 unx       75 b- defN 80-Jan-01 00:00 .layout.json
drwxr-xr-x  2.0 unx        0 b- defN 80-Jan-01 00:00 .prefix/
drwxr-xr-x  2.0 unx        0 b- defN 80-Jan-01 00:00 .prefix/bin/
-rwxr-xr-x  2.0 unx  9838248 b- defN 80-Jan-01 00:00 .prefix/bin/py-spy
drwxr-xr-x  2.0 unx        0 b- defN 80-Jan-01 00:00 py_spy-0.3.8.dist-info/
-rw-r--r--  2.0 unx        4 b- defN 80-Jan-01 00:00 py_spy-0.3.8.dist-info/INSTALLER
-rw-r--r--  2.0 unx    16821 b- defN 80-Jan-01 00:00 py_spy-0.3.8.dist-info/METADATA
-rw-r--r--  2.0 unx        0 b- defN 80-Jan-01 00:00 py_spy-0.3.8.dist-info/REQUESTED
-rw-r--r--  2.0 unx      123 b- defN 80-Jan-01 00:00 py_spy-0.3.8.dist-info/WHEEL
10 files, 9855271 bytes uncompressed, 3125164 bytes compressed:  68.3%
```

The issue there is the 1st `../` entry. When Python 3.12 `zipfile` sees that, it says:
```
        if not arcname:
>           raise ValueError("Empty filename.")
E           ValueError: Empty filename.

../../../.pyenv/versions/3.12.0b3/lib/python3.12/zipfile/__init__.py:1761: ValueError
```

More directly:
+ 3.11
  ```
  $ mkdir /tmp/py311-unzip && python3.11 -mzipfile -e /home/jsirois/.pex/installed_wheel_zips/858d7ad606d58d8b13b4ac842dd378ff37e94387abc4b79e281576af39c64a49/py_spy-0.3.8-py2.py3-none-manylinux_2_5_x86_64.manylinux1_x86_64.whl /tmp/py311-unzip
  ```
+ 3.12
  ```
  $ mkdir /tmp/py312-unzip && python3.12 -mzipfile -e /home/jsirois/.pex/installed_wheel_zips/858d7ad606d58d8b13b4ac842dd378ff37e94387abc4b79e281576af39c64a49/py_spy-0.3.8-py2.py3-none-manylinux_2_5_x86_64.manylinux1_x86_64.whl /tmp/py312-unzip
  <frozen runpy>:128: RuntimeWarning: 'zipfile.__main__' found in sys.modules after import of package 'zipfile', but prior to execution of 'zipfile.__main__'; this may result in unpredictable behaviour
    File "<frozen runpy>", line 198, in _run_module_as_main
    File "<frozen runpy>", line 88, in _run_code
    File "/home/jsirois/.pyenv/versions/3.12.0b3/lib/python3.12/zipfile/__main__.py", line 77, in <module>
      main()
    File "/home/jsirois/.pyenv/versions/3.12.0b3/lib/python3.12/zipfile/__main__.py", line 44, in main
      zf.extractall(curdir)
    File "/home/jsirois/.pyenv/versions/3.12.0b3/lib/python3.12/zipfile/__init__.py", line 1720, in extractall
      self._extract_member(zipinfo, path, pwd)
    File "/home/jsirois/.pyenv/versions/3.12.0b3/lib/python3.12/zipfile/__init__.py", line 1761, in _extract_member
      raise ValueError("Empty filename.")
  ValueError: Empty filename.
  ```

jsirois: I'll try to whip up a unit test for the new pex/common.py guard / behavior on top of the cache re-structure avoidance fix.

jsirois:
I experimented with the latest Pip main (a few new commits on top of the version used here for Python 3.12 support / testing) and there will need to be more changes to support Pip 23.2 - not Python 3.12 related though. I'll proceed to land this on green as a result since more work is ahead.

PR #2174, opened 2023-07-11
Prepare for Pip 23.2 release.

jsirois:
A recent Pip commit has broken
`pip --use-deprecated legacy-resolver download ...` which is our
backwards-compatible default use case. Whether or not Pip decides to
fix, the legacy resolver will be yanked at some point and it seems
reasonable to just require any Pex user on Python 3.12 and thus Pip 23.2
to only use the Pip 2020 resolver.

See: https://github.com/pypa/pip/issues/12138

### DIFF
--- a/pex/resolve/resolver_configuration.py
+++ b/pex/resolve/resolver_configuration.py
@@ -4,12 +4,12 @@
 from __future__ import absolute_import
 
 import itertools
-import os
 
 from pex.auth import PasswordEntry
 from pex.enum import Enum
 from pex.jobs import DEFAULT_MAX_JOBS
 from pex.network_configuration import NetworkConfiguration
+from pex.pep_440 import Version
 from pex.pip.version import PipVersion, PipVersionValue
 from pex.typing import TYPE_CHECKING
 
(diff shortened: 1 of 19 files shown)


kaos (approved):
lgtm.

Quite a few moving parts in the implementation I don't have a sense on, but I agree with taking the described actions outlined in the PR description.

jsirois:
Thanks for taking a look @kaos.

pex/resolve/resolver_configuration.py
```diff
@@ -31,6 +31,26 @@ class ResolverVersion(Enum["ResolverVersion.Value"]):
     class Value(Enum.Value):
         pass
 
+    @staticmethod
+    def _supports_legacy_resolver(pip_version=None):
+        # type: (Optional[PipVersionValue]) -> bool
+        pip_ver = pip_version or PipVersion.DEFAULT
+        return pip_ver.version < Version("23.2")
```
jsirois, line 38: It looks like the underlying issue will get fixed before 23.2 is released (https://github.com/pypa/pip/pull/12140). I'll probably follow up, leaving all the code in this change in place but changing this function to just return `True` for all cases so that `< Version("X.Y.Z")` can go back in when Pip does actually drop support for the legacy resolver.

PR #1705, opened 2022-04-04
Support uncompressed PEXes.

jsirois:
Closes #1686

Issue #1686, opened 2022-03-23
Expose compression control for Pex zipping operations.


jsirois:
The experiments [here](https://github.com/pantsbuild/pex/issues/1675#issuecomment-1072451741) and [here](https://github.com/pantsbuild/pex/issues/1675#issuecomment-1075737047) in #1675 prove out that the zip compression level provides a significant swing in tradeoffs. It's probably worth exposing this to let users decide where they want to land in the tradeoff spectrum.

jsirois:
For a modest sized PEX the gain is modest, ~25% speed improvement creating the PEX zip, ~no runtime perf impact and ~350% size increase:
```
$ hyperfine -w1 'python -mpex jupyter -c jupyter -o jupyter.default.pex' 'python -mpex jupyter -c jupyter -o jupyter.no-compress.pex --no-compress'
Benchmark 1: python -mpex jupyter -c jupyter -o jupyter.default.pex
  Time (mean ± σ):      8.695 s ±  0.288 s    [User: 6.429 s, System: 0.253 s]
  Range (min … max):    8.422 s …  9.351 s    10 runs
 
Benchmark 2: python -mpex jupyter -c jupyter -o jupyter.no-compress.pex --no-compress
  Time (mean ± σ):      7.065 s ±  0.109 s    [User: 4.665 s, System: 0.276 s]
  Range (min … max):    6.928 s …  7.254 s    10 runs
 
Summary
  'python -mpex jupyter -c jupyter -o jupyter.no-compress.pex --no-compress' ran
    1.23 ± 0.05 times faster than 'python -mpex jupyter -c jupyter -o jupyter.default.pex'

$ ls -lh jupyter.*.pex
-rwxr-xr-x 1 jsirois jsirois 25M Apr  4 11:05 jupyter.default.pex
-rwxr-xr-x 1 jsirois jsirois 88M Apr  4 11:06 jupyter.no-compress.pex

$ hyperfine -w1 './jupyter.default.pex --version' './jupyter.no-compress.pex --version'
Benchmark 1: ./jupyter.default.pex --version
  Time (mean ± σ):     861.9 ms ±   6.8 ms    [User: 812.1 ms, System: 49.6 ms]
  Range (min … max):   852.6 ms … 873.5 ms    10 runs
 
Benchmark 2: ./jupyter.no-compress.pex --version
  Time (mean ± σ):     871.5 ms ±   9.5 ms    [User: 826.1 ms, System: 45.1 ms]
[...]

Issue #1675, opened 2022-03-17
Large amount of time spent in "zipping" when building a pex from a `--pex-repository`


stuhood:
For the attached [lock.txt](https://github.com/pantsbuild/pex/files/8289004/lock.txt) (superset) and [subset.txt](https://github.com/pantsbuild/pex/files/8289005/subset.txt) requirements, building a PEX using `--pex-repository` (on an AWS p2x instance inside of Docker) takes ~220s, primarily inside the "zipping" phase:
```
pex: Zipping PEX file.
pex: Zipping PEX file.: 221244.8ms
```

The full PEX command to build the subset is:
```console
$HOME/.pyenv/versions/3.10.1/bin/python ./pex \
  --tmpdir.tmp \
  --jobs 8 \
  --python-path .... \
  --output-file example.pex \
  --no-emit-warnings \
  --manylinux manylinux2014 \
  --venv prepend \
  --requirements-pex local_dists.pex \
  --pex-repository default_lockfile.pex \
  --interpreter-constraint CPython==3.8.* \
  --entry-point ... \
  --sources-directory=source_files \
  albumentations<2.0.0,>=1.1.0 \
  boto3==1.20.24 \
  botocore==1.23.24 \
  docstring-parser<0.14.0,>=0.13 \
  jpeg4py<0.2.0,>=0.1.4 \
  jsonargparse[signatures]<5.0.0,>=4.3.1 \
  numpy<2.0.0,>=1.22.3 \
  nvidia-dali-cuda110>=1.11.0 \
  opencv-python<5.0.0,>=4.5.5 \
  pytorch-lightning<2.0.0,>=1.5.10 \
  requests<3.0.0,>=2.27.1 \
  s3fs==2022.2.0 \
  setuptools==59.5.0 \
  torch<2.0.0,>=1.10.2 \
  torchvision<0.12.0,>=0.11.3 \
  wandb<0.13.0,>=0.12.11 \
  --layout zipapp
```

jsirois:
Why is layout zipapp? The packed layout was built to minimize zip time / cache hits, etc.

jsirois:
Ok, assuming there is some reason for needing zipapp, 1st seeing what it takes for native tools:
```
Python 3.10 fails lock download on:
ERROR: Could not find a version that satisfies the requirement onnxruntime==1.10.0
ERROR: No matching distribution found for onnxruntime==1.10.0

Trying 3.9:
[...]

PR #1431, opened 2021-08-31
Introduce the spread app layout.

jsirois:
Beyond the top level `pex` and `__main__.py` files, the layout is an
implementation detail, but the current structure sheds some light on the
cache-friendly characteristics:

Given loose sources:
```
$ cat src/main.py
from pex.version import __version__; print(__version__)
```

A spread layout PEX looks like:
```
$ python -mpex pex -Dsrc -emain -opex.spread.venv --spread \
  --seed verbose --venv prepend | jq .
{
  "pex_root": "/home/jsirois/.pex",
  "python": "/usr/bin/python3.9",
  "pex": "/home/jsirois/.pex/venvs/d2f743b9c1ebb156f794419c01b3422653cbdb61/2d1d404c3de23b1810386195be7410700b1feb14/pex"
}
$ tree -a pex.spread.venv
pex.spread.venv
├── .bootstrap
├── .deps
│   └── pex-2.1.46-py2.py3-none-any.whl
├── __main__.py -> src/__main__.py
├── pex -> __main__.py
├── PEX-SPREAD-INFO
└── src
    ├── __main__.py
    ├── main.py
    └── PEX-INFO
```

And the runtime spreading is suggested by the new PEX-SPREAD-INFO
manifest:
```json
{
  "sources": [
    "PEX-INFO",
    "__main__.py",
    "main.py"
  ],
  "spreads": [
    {
      "strip_zip_relpath": false,
      "unpack_relpath": "bootstraps/a54b6ae5e64e5b229388fdffc8adac141f3c416b",
      "zip_relpath": ".bootstrap"
    },
    {
      "strip_zip_relpath": true,
      "unpack_relpath": "installed_wheels/f627f0368a0e29be24aa8cadba74044b9ad990d7/pex-2.1.46-py2.py3-none-any.whl",
      "zip_relpath": ".deps/pex-2.1.46-py2.py3-none-any.whl"
    }
  ]
}
```

The layout adds new PEX_ROOT caches for the `.bootstrap` zip and
installed wheel chroot zips such that neither bootstraps - which are
tied to a version of Pex, nor installed wheel chroot zips, which are
constant for a given distribution version, are created more than once.

Closes #1424

PR #2175, opened 2023-07-19
consume packed wheel cache in zipapp creation

### LINKED ISSUE #2158: research area: parallel zip creation
# Problem

I was creating a pex file to resolve dependencies while playing around with fawkes (https://github.com/shawn-shan/fawkes), and like most modern ML projects, it contains many large binary dependencies. This meant that while resolution (with the 2020 pip resolver) was relatively fast, the creation of the zip itself took about a minute after that (without any progress indicator), which led me to prefer a venv when iterating on the fawkes source code.

## Use Case Discussion
I understand pex's supported use cases revolve much more around robust reproducible deployment scenarios, where taking a single minute to zip up a massive bundle of dependencies is more than acceptable, and where making use of the battle-tested stdlib `zipfile.ZipFile` is extremely important to ensure pex files can be executed as well as inspected on all platforms and by all applications. However, for use cases like the one described above, where the pex is going to be created strictly for the local platform, I think it would be really convenient to avoid having to set up a stateful venv.

## Alternatives
I could just create a pex file for the dependencies, and use that to launch the python process that runs from source code, and indeed that is what we decided on to implement pantsbuild/pants#8793, which worked perfectly for the Twitter ML infra team's jupyter notebooks. But (assuming this is actually possible) I would still personally find a feature that zips up pex files much faster to be useful for a lot of "I really just wanna hack something together" scenarios where I don't necessarily want to have to set up a two-phase build process like that.

# Implementation Strategy
After going through the proposal of pypa/pip#8448 to hack the zip format to minimize wheel downloads, which was implemented much more thoughtfully in pypa/pip#8467 as [`LazyZipOverHTTP`](https://github.com/pypa/pip/blob/5190ef71cc77a6a6b8f40e4b1d78b4e6c35fe231/src/pip/_internal/network/lazy_wheel.py#L40-L47), and then realizing I had overpromised the potential speedup at https://github.com/pypa/pip/issues/7049#issuecomment-1268726609, I am wary of just assuming that hacking around with the zip format will necessarily improve performance over the battle-tested synchronous stdlib implementation.

However, it seems plausible that the process of compressing individual entries could be parallelized. `pigz` describes their methodology at https://github.com/madler/pigz/blob/cb8a432c91a1dbaee896cd1ad90be62e5d82d452/pigz.c#L279-L340, and there is a codebase named [`fastzip`](https://github.com/fastzip/fastzip) that does this in python using threads, with a great discussion of [performance bottlenecks](https://github.com/fastzip/fastzip#performance-bottlenecks) and a few notes on [compatibility issues](https://github.com/fastzip/fastzip#future-plans-from-drawbacks).

## End-User Interface
Due to the likely compatibility issues (both with executing the parallel method at all, as well as consuming the resulting zip file), it seems best to put this behind a flag, and probably to explicitly call it experimental (I like the way pip does `--use-feature=fast-deps`, for example), and possibly even to print out a warning to stderr when executing or processing any pex files created this way. To enable the warning message, or in case we want to do any other special-case processing of pex files created this way, we could put a key in `PEX-INFO`'s `.build_properties`, or perhaps add an empty sentinel file named `.parallel-zip` to the output.

cosmicexplorer:
*Closes #2158.*

# Problem
*#1675 has another repro.*

Generating a `--compress --layout zipapp` (the default) regularly exhibits pathologically slow performance **(46s zip / 1m1s total)** on many real-world inputs (especially tensorflow). This occurs even when all dists are already downloaded from pypi:
```bash
; time pex -vvvvvvvvv --layout zipapp --resolver-version=pip-2020-resolver 'numpy>=1.19.5' 'keras==2.4.3' 'mtcnn' 'pillow>=7.0.0' 'bleach>=2.1.0' 'tensorflow-gpu==2.5.3' -o test.pex
# ...
# Saving PEX file to test.pex
# pex: Zipping PEX file.: 45659.1ms
# 57.18s user 3.02s system 97% cpu 1:01.57 total
; ls -lAvFh test.pex
# -rwxrwxr-x 1 cosmicexplorer cosmicexplorer 617M Jul 26 09:17 test.pex*
; du -b test.pex
# 616754775       test.pex
```

## Alternatives

#1705 introduced `--no-compress`, which drastically improves the zip performance **(2.2s zip => 20.6x speedup / 19.3s total = 2.4x speedup)**:
```bash
; time pex -vvvvvvvvv --layout zipapp --no-compress --resolver-version=pip-2020-resolver 'numpy>=1.19.5' 'keras==2.4.3' 'mtcnn' 'pillow>=7.0.0' 'bleach>=2.1.0' 'tensorflow-gpu==2.5.3' -o test.pex
# ...
# Saving PEX file to test.pex
# pex: Zipping PEX file.: 2209.7ms
# pex  14.11s user 3.06s system 89% cpu 19.285 total
```

but as #1686 describes, it also introduces a significant tradeoff in file size **(1.6G => 2.5x blowup)**:
```bash
; ls -lAvFh test.pex
# -rwxrwxr-x 1 cosmicexplorer cosmicexplorer 1.6G Jul 26 09:20 test.pex*
; du -b test.pex
# 1541737605      test.pex
```

The cache entries are also many times larger:
```bash
; ls $(find ~/.pex/packed_wheels -name '*tensorflow_gpu*whl')
# -rw-rw-r-- 1 cosmicexplorer cosmicexplorer 468M Aug  1 09:10 /home/cosmicexplorer/.pex/packed_wheels/7a4763a35d0824ebb172b00f8d0241ff231404c4b7d97dd5ea870d5afca336a4/tensorflow_gpu-2.5.3-cp38-cp38-manylinux2010_x86_64.whl
# -rw-rw-r-- 1 cosmicexplorer cosmicexplorer 1.2G Aug  1 09:16 /home/cosmicexplorer/.pex/packed_wheels/7a4763a35d0824ebb172b00f8d0241ff231404c4b7d97dd5ea870d5afca336a4/un-compressed/tensorflow_gpu-2.5.3-cp38-cp38-manylinux2010_x86_64.whl
```

# Solution
1. Introduce `pex.ziputils.MergeableZipFile` to incorporate the contents of other zip files without decompressing them.
    - Introduce `pex.common.copy_file_range()` to copy over a limited range of the contents of another file handle.
2. Introduce the `--cache-dists` option to toggle the use (and/or creation) of the packed wheel caches for `.bootstrap/` and `.deps/` subdirectories.
3. Consume the existing `--layout packed` caches for `--layout zipapp` when `--cache-dists` is provided.

### DIFF
--- a/pex/common.py
+++ b/pex/common.py
@@ -39,6 +39,8 @@ if TYPE_CHECKING:
         Union,
     )
 
+    _DateTime = Tuple[int, int, int, int, int, int]
+
 
 # We use the start of MS-DOS time, which is what zipfiles use (see section 4.4.6 of
 # https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT).
@@ -138,10 +140,7 @@ def safe_copy(source, dest, overwrite=False):
         do_copy()
 
 
-_COPY_BUFSIZE = 64 * 1024
-
-
-def copy_file_range(source, destination, length, buffer_size=_COPY_BUFSIZE):
+def copy_file_range(source, destination, length, buffer_size=io.DEFAULT_BUFFER_SIZE):
     # type: (io.BufferedIOBase, io.BufferedIOBase, int, int) -> None
     """Implementation of shutil.copyfileobj() that only copies exactly `length` bytes."""
     # We require a BufferedIOBase in order to avoid handling short reads or writes.
@@ -194,7 +193,14 @@ class PermPreservingZipFile(zipfile.ZipFile, object):
         pass
 
     @classmethod
-    def zip_entry_from_file(cls, filename, arcname=None, date_time=None):
+    def zip_entry_from_file(
+        cls,
+        filename,  # type: str
+        arcname=None,  # type: Optional[str]
+        date_time=None,  # type: Optional[Tuple[int, ...]]
+        compression=zipfile.ZIP_STORED,  # type: int
+    ):
+        # type: (...) -> PermPreservingZipFile.ZipEntry
         """Construct a ZipEntry for a file on the filesystem.
 
         Usually a similar `zip_info_from_file` method is provided by `ZipInfo`, but it is not
@@ -213,16 +219,20 @@ class PermPreservingZipFile(zipfile.ZipFile, object):
             arcname += "/"
         if date_time is None:
             date_time = time.localtime(st.st_mtime)
-        zinfo = zipfile.ZipInfo(filename=arcname, date_time=date_time[:6])
+        zinfo = zipfile.ZipInfo(filename=arcname, date_time=cast(_DateTime, date_time[:6]))
         zinfo.external_attr = (st.st_mode & 0xFFFF) << 16  # Unix attributes
         if isdir:
             zinfo.file_size = 0
             zinfo.external_attr |= 0x10  # MS-DOS directory flag
+            # Always store directories decompressed, because they are empty but take up 2 bytes when
+            # compressed.
             zinfo.compress_type = zipfile.ZIP_STORED
             data = b""
         else:
             zinfo.file_size = st.st_size
-            zinfo.compress_type = zipfile.ZIP_DEFLATED
+            # File contents may be compressed or decompressed. Decompressed is significantly faster
+            # to write, but caching makes up for that.
+            zinfo.compress_type = compression
             with open(filename, "rb") as fp:
                 data = fp.read()
         return cls.ZipEntry(info=zinfo, data=data)
@@ -677,12 +687,12 @@ class Chroot(object):
                 zip_entry = zf.zip_entry_from_file(
                     filename=filename,
                     arcname=os.path.relpath(arcname, strip_prefix) if strip_prefix else arcname,
-                    date_time=DETERMINISTIC_DATETIME.timetuple()
-                    if deterministic_timestamp
-                    else None,
+                    date_time=(
+                        DETERMINISTIC_DATETIME.timetuple() if deterministic_timestamp else None
+                    ),
+                    compression=compression,
                 )
-                # FIXME: this ignores the zinfo.compress_type value from zip_entry_from_file()!
-                zf.writestr(zip_entry.info, zip_entry.data, compression)
+                zf.writestr(zip_entry.info, zip_entry.data)
 
             def get_parent_dir(path):
                 # type: (str) -> Optional[str]
--- a/pex/common.py
+++ b/pex/common.py
@@ -6,6 +6,7 @@ from __future__ import absolute_import, print_function
 import atexit
 import contextlib
 import errno
+import io
 import itertools
 import os
 import re
@@ -135,6 +136,26 @@ def safe_copy(source, dest, overwrite=False):
             do_copy()
     else:
         do_copy()
+
+
+_COPY_BUFSIZE = 64 * 1024
+
+
+def copy_file_range(source, destination, length, buffer_size=_COPY_BUFSIZE):
+    # type: (io.BufferedIOBase, io.BufferedIOBase, int, int) -> None
+    """Implementation of shutil.copyfileobj() that only copies exactly `length` bytes."""
+    # We require a BufferedIOBase in order to avoid handling short reads or writes.
+    remaining_length = length
+    if buffer_size > length:
+        buffer_size = length
+    cur_buf = bytearray(buffer_size)
+    while remaining_length > buffer_size:
+        assert source.readinto(cur_buf) == buffer_size
+        assert destination.write(cur_buf) == buffer_size
+        remaining_length -= buffer_size
+    remainder = source.read(remaining_length)
+    assert len(remainder) == remaining_length
+    assert destination.write(remainder) == remaining_length
 
 
 # See http://stackoverflow.com/questions/2572172/referencing-other-modules-in-atexit
@@ -281,16 +302,30 @@ def safe_mkdir(directory, clean=False):
         return directory
 
 
+def _ensure_parent(filename):
+    # type: (str) -> None
+    parent_dir = os.path.dirname(filename)
+    if parent_dir:
+        safe_mkdir(parent_dir)
+
+
 def safe_open(filename, *args, **kwargs):
     """Safely open a file.
 
     ``safe_open`` ensures that the directory components leading up the specified file have been
     created first.
     """
-    parent_dir = os.path.dirname(filename)
-    if parent_dir:
-        safe_mkdir(parent_dir)
+    _ensure_parent(filename)
     return open(filename, *args, **kwargs)  # noqa: T802
+
+
+def safe_io_open(filename, *args, **kwargs):
+    # type: (str, Any, Any) -> io.IOBase
+    """``safe_open()``, but using ``io.open()`` instead.
+
+    With the right arguments, this ensures the result produces a buffered file handle on py2."""
+    _ensure_parent(filename)
+    return cast("io.IOBase", io.open(filename, *args, **kwargs))
 
 
 def safe_delete(filename):
@@ -608,9 +643,13 @@ class Chroot(object):
         # type: () -> None
         shutil.rmtree(self.chroot)
 
+    # This directory traversal, file I/O, and compression can be made faster with complex
+    # parallelism and pipelining in a compiled language, but the result is much harder to package,
+    # and is still less performant than effective caching. See investigation in
+    # https://github.com/pantsbuild/pex/issues/2158 and https://github.com/pantsbuild/pex/pull/2175.
     def zip(
         self,
-        filename,  # type: str
+        output_file,  # type: Union[str, io.IOBase]
         mode="w",  # type: str
         deterministic_timestamp=False,  # type: bool
         exclude_file=lambda _: False,  # type: Callable[[str], bool]
@@ -628,7 +667,7 @@ class Chroot(object):
             selected_files = self.files()
 
         compression = zipfile.ZIP_DEFLATED if compress else zipfile.ZIP_STORED
-        with open_zip(filename, mode, compression) as zf:
+        with open_zip(output_file, mode, compression) as zf:
 
             def write_entry(
                 filename,  # type: str
@@ -642,6 +681,7 @@ class Chroot(object):
                     if deterministic_timestamp
                     else None,
                 )
+                # FIXME: this ignores the zinfo.compress_type value from zip_entry_from_file()!
                 zf.writestr(zip_entry.info, zip_entry.data, compression)
 
             def get_parent_dir(path):
--- a/pex/common.py
+++ b/pex/common.py
@@ -219,7 +219,7 @@ class PermPreservingZipFile(zipfile.ZipFile, object):
             arcname += "/"
         if date_time is None:
             date_time = time.localtime(st.st_mtime)
-        zinfo = zipfile.ZipInfo(filename=arcname, date_time=cast(_DateTime, date_time[:6]))
+        zinfo = zipfile.ZipInfo(filename=arcname, date_time=cast("_DateTime", date_time[:6]))
         zinfo.external_attr = (st.st_mode & 0xFFFF) << 16  # Unix attributes
         if isdir:
             zinfo.file_size = 0
@@ -659,7 +659,7 @@ class Chroot(object):
     # https://github.com/pantsbuild/pex/issues/2158 and https://github.com/pantsbuild/pex/pull/2175.
     def zip(
         self,
-        output_file,  # type: Union[str, io.IOBase]
+        output_file,  # type: Union[str, io.IOBase, io.BufferedRandom]
         mode="w",  # type: str
         deterministic_timestamp=False,  # type: bool
         exclude_file=lambda _: False,  # type: Callable[[str], bool]
--- a/pex/pex_builder.py
+++ b/pex/pex_builder.py
@@ -869,10 +869,10 @@ class PEXBuilder(object):
                 #     computation, because the compressed entries are simply copied over as-is.
                 #
                 #     Note that these cached zips are created in the --layout packed format, without
-                #     the .bootstrap/ or .deps/ prefixes we need to form a proper zipapp. Our zip
-                #     file merging solution edits each entry's .filename with the appropriate
-                #     prefix, but we will still need to generate intermediate directory entries
-                #     before adding the prefixed files in order to unzip correctly.
+                #     the .bootstrap/ or .deps/ prefixes we need to form a proper Pex zipapp
+                #     layout. Our zip file merging solution edits each entry's .filename with the
+                #     appropriate prefix, but we will still need to generate intermediate directory
+                #     entries before adding the prefixed files in order to unzip correctly.
 
                 # Reuse the file handle to zip into. This isn't necessary (we could close and reopen
                 # it), but it avoids unnecessarily flushing to disk. The ZipFile class will re-parse
@@ -927,6 +927,5 @@ class PEXBuilder(object):
                         compress=compress,
                         labels=self._DIRECT_SOURCE_LABELS,
                     )
-
 
         chmod_plus_x(filename)
--- a/pex/pex_builder.py
+++ b/pex/pex_builder.py
@@ -7,6 +7,7 @@ import hashlib
 import logging
 import os
 import shutil
+import zipfile
 
 from pex import pex_warnings
 from pex.atomic_directory import atomic_directory
@@ -17,9 +18,9 @@ from pex.common import (
     is_pyc_temporary_file,
     safe_copy,
     safe_delete,
+    safe_io_open,
     safe_mkdir,
     safe_mkdtemp,
-    safe_open,
 )
 from pex.compatibility import commonpath, to_bytes
 from pex.compiler import Compiler
@@ -37,9 +38,10 @@ from pex.targets import Targets
 from pex.tracer import TRACER
 from pex.typing import TYPE_CHECKING
 from pex.util import CacheHelper
+from pex.ziputils import MergeableZipFile, buffered_zip_archive
 
 if TYPE_CHECKING:
-    from typing import Dict, Optional
+    from typing import ClassVar, Dict, Iterable, Optional, Tuple
 
 
 # N.B.: __file__ will be relative when this module is loaded from a "" `sys.path` entry under
@@ -95,14 +97,14 @@ def __maybe_run_venv__(pex, pex_root, pex_path):
 
   venv_dir = venv_dir(
     pex_file=pex,
-    pex_root=pex_root, 
+    pex_root=pex_root,
     pex_hash={pex_hash!r},
     has_interpreter_constraints={has_interpreter_constraints!r},
     pex_path=pex_path,
   )
   venv_pex = os.path.join(venv_dir, 'pex')
   if not __execute__ or not is_exe(venv_pex):
-    # Code in bootstrap_pex will (re)create the venv after selecting the correct interpreter. 
+    # Code in bootstrap_pex will (re)create the venv after selecting the correct interpreter.
     return venv_dir
 
   TRACER.log('Executing venv PEX for {{}} at {{}}'.format(pex, venv_pex))
@@ -441,6 +443,7 @@ class PEXBuilder(object):
         self._header = header
 
     def _add_dist_dir(self, path, dist_name, fingerprint=None):
+        # type: (str, str, Optional[str]) -> str
         target_dir = os.path.join(self._pex_info.internal_cache, dist_name)
         if self._copy_mode == CopyMode.SYMLINK:
             self._copy_or_link(path, target_dir, label=dist_name)
@@ -557,6 +560,7 @@ class PEXBuilder(object):
         elif self._copy_mode == CopyMode.SYMLINK:
             self._chroot.symlink(src, dst, label)
         else:
+            assert self._copy_mode == CopyMode.LINK
             self._chroot.link(src, dst, label)
 
     def _prepare_bootstrap(self):
@@ -638,7 +642,7 @@ class PEXBuilder(object):
         """Package the PEX application.
 
         By default, the PEX is packaged as a zipapp for ease of shipping as a single file, but it
-        can also be packaged in spread mode for efficiency of syncing over the network
+        can also be packaged in a packed layout for efficiency of syncing over the network
         incrementally.
 
         :param path: The path where the PEX should be stored.
@@ -702,6 +706,28 @@ class PEXBuilder(object):
         )
         self.set_header(script)
 
+    def _setup_pex_info(self):
+        # type: () -> Tuple[PexInfo, str]
+        pex_info = self._pex_info.copy()
+        pex_info.update(PexInfo.from_env())
+
+        bootstrap_hash = pex_info.bootstrap_hash
+        if bootstrap_hash is None:
+            raise AssertionError(
+                "Expected bootstrap_hash to be populated for {}.".format(self._pex_info)
+            )
+
+        return (pex_info, bootstrap_hash)
+
+    _DIRECT_SOURCE_LABELS = (
+        "executable",
+        "importhook",
+        "main",
+        "manifest",
+        "resource",
+        "source",
+    )  # type: ClassVar[Iterable[str]]
+
     def _build_packedapp(
         self,
         dirname,  # type: str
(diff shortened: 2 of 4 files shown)


cosmicexplorer:
Reading @jsirois's thoughts again:
> Since this would only be used by the Pex CLI (build-time), a rust implementation is actually how I'd approach this. It's the right language for the job, the problem really has nothing to do with Pex (it has much wider applicability), and there are great tools for turning a crate into an easily consumable Python platform-specific distribution.

I'll take a look at pyoxy and friends to convert the crate into a python distribution. I'll leave this as a draft until then, and I'll reping reviewers at that time.

cosmicexplorer:
Ah, wonderful, I've just found the docs for pyo3 https://docs.rs/pyo3/latest/pyo3/.

jsirois:
> Ah, wonderful, I've just found the docs for pyo3 https://docs.rs/pyo3/latest/pyo3/.

Great. Yeah @cosmicexplorer that's exactly what I meant. I've been away climbing here and haven't taken a look yet, but if this all looks good, consuming as a native wheel was the idea.

cosmicexplorer:
Ok, just had a lot of fun converting this into a pyo3 package and the result runs even faster! This branch isn't passing because I've added some hacks to get it running, but I will be working to remove those!

cosmicexplorer:
Just refactored this immensely after @jsirois's suggestion to use @stuhood's strategy of `zip -FF`.

jsirois (commented):
A very cursory review that stopped short for another gut check. The warm cache speed of zip creation improving is nice to have - full stop. That said, it really does beg the question of why not just use `--layout packed` which has existed for quite some time and has the same speed benefits with the addition of friendlier network transfer characteristics / remote cache characteristics (which is why it was invented for Pants). You can rsync a packed PEX and win. You cannot do the same with the PEX built faster locally by this PR, but still being a totally new blob.

pex/pex_builder.py
```diff
@@ -769,22 +776,96 @@ def zip_cache_dir(path):
...
+        """Zip up each fingerprinted dist into an atomic shared cache location."""
+        merge_deps = {}  # type: Dict[str, str]
+        for dist_label, fingerprint in self._pex_info.distributions.items():
+            cached_dist = os.path.join(
+                self._pex_info.pex_root,
+                "stitched_dists",
+                fingerprint,
+                dist_label,
+            )
+            with atomic_directory(os.path.dirname(cached_dist)) as atomic_zip_dir:
```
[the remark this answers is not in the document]
jsirois: For reference #1431 which was followed up a few days later to `s/--spread/--layout packed/`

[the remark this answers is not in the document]
cosmicexplorer: Unfortunately, yes. The `--layout packed` cache zips (correctly) strip `Layout.DEPS_DIR='.deps'` from every entry:
```bash
> unzip -l test.pex/.deps/tensorflow_gpu-2.5.3-cp38-cp38-manylinux2010_x86_64.whl | head -n10 
Archive:  test.pex/.deps/tensorflow_gpu-2.5.3-cp38-cp38-manylinux2010_x86_64.whl
  Length      Date    Time    Name
---------  ---------- -----   ----
       83  1980-01-01 00:00   .layout.json
        0  1980-01-01 00:00   tensorflow_gpu-2.5.3.dist-info/
        4  1980-01-01 00:00   tensorflow_gpu-2.5.3.dist-info/INSTALLER
    13777  1980-01-01 00:00   tensorflow_gpu-2.5.3.dist-info/LICENSE
     2810  1980-01-01 00:00   tensorflow_gpu-2.5.3.dist-info/METADATA
        0  1980-01-01 00:00   tensorflow_gpu-2.5.3.dist-info/REQUESTED
      111  1980-01-01 00:00   tensorflow_gpu-2.5.3.dist-info/WHEEL
````

while the new ones for `--layout zipapp` have the prefix (which is needed for the zip concatenation to work):
```bash
> unzip -l /home/cosmicexplorer/.pex/stitched_dists/7a4763a35d0824ebb172b00f8d0241ff231404c4b7d97dd5ea870d5afca336a4/tensorflow_gpu-2.5.3-cp38-cp38-manylinux2010_x86_64.whl | head -n10
Archive:  /home/cosmicexplorer/.pex/stitched_dists/7a4763a35d0824ebb172b00f8d0241ff231404c4b7d97dd5ea870d5afca336a4/tensorflow_gpu-2.5.3-cp38-cp38-manylinux2010_x86_64.whl
  Length      Date    Time    Name
---------  ---------- -----   ----
        0  1980-01-01 00:00   .deps/
        0  1980-01-01 00:00   .deps/tensorflow_gpu-2.5.3-cp38-cp38-manylinux2010_x86_64.whl/
       83  1980-01-01 00:00   .deps/tensorflow_gpu-2.5.3-cp38-cp38-manylinux2010_x86_64.whl/.layout.json
        0  1980-01-01 00:00   .deps/tensorflow_gpu-2.5.3-cp38-cp38-manylinux2010_x86_64.whl/tensorflow_gpu-2.5.3.dist-info/
        4  1980-01-01 00:00   .deps/tensorflow_gpu-2.5.3-cp38-cp38-manylinux2010_x86_64.whl/tensorflow_gpu-2.5.3.dist-info/INSTALLER
    13777  1980-01-01 00:00   .deps/tensorflow_gpu-2.5.3-cp38-cp38-manylinux2010_x86_64.whl/tensorflow_gpu-2.5.3.dist-info/LICENSE
     2810  1980-01-01 00:00   .deps/tensorflow_gpu-2.5.3-cp38-cp38-manylinux2010_x86_64.whl/tensorflow_gpu-2.5.3.dist-info/METADATA
```

It seems maybe reasonable to generate both of these entries at once from the `packed_wheels` cache dir, but I really don't want to wait TWICE as long to generate two compressed zips for each dist we haven't cached yet.

It's possible we could avoid duplicating the compression work without using two separate cache directories, but it might get complicated.

In general, I see this `stitched_dists` cache as strictly for speeding up the default `--layout zipapp` experience (which means it might contain e.g. invalid zip data in order to feed into `zip -FF`, and may be removed later), while `packed_wheels` contains a pristine directory to conform to the packed pex interface. Does that seem like a reasonable split?

cosmicexplorer:
> That said, it really does beg the question of why not just use `--layout packed`

I understand that `--layout packed` was created specifically for users seeking to maximize cacheability, especially across rsync, and I don't plan to put any more effort into making `--layout zipapp` more cacheable.

I think this change produces a really drastic and noticeable performance improvement for an extremely common use case (especially the use of pex in prototyping), and I was really excited to figure out how to avoid introducing any new types of output formats. I hope this change can be completely invisible to users (except as a noticeable perf bump).

cosmicexplorer:
I used the term "prototyping" very loosely. I was merely emphasizing how often `--layout zipapp` is employed for the user's convenience. Our docs at https://pex.readthedocs.io/en/v2.1.140/buildingpex.html#saving-pex-files don't mention the other layouts, so a lot of people will be using these commands.

Possibly more importantly, I assume this speedup will also be felt by pants, which often reconstitutes various intermediate pexes full of requirements, using `--layout zipapp`, in order to perform some task like running pytest. I was halfway to making a repro case in the pants repo but gave up for now.

jsirois:
> I used the term "prototyping" very loosely. I was merely emphasizing how often --layout zipapp is employed for the user's convenience. Our docs at https://pex.readthedocs.io/en/v2.1.140/buildingpex.html#saving-pex-files don't mention the other layouts, so a lot of people will be using these commands.

Ok. The docs are in a sorry state to be sure, but that can be remedied. Since Pex has a rule of not breaking existing users, `--layout zipapp` is the default, so more convenient in that you need not type. But there is shell aliases to help with that.

If you don't have an actual use case in mind, that's not awesome. There is an expense here in ~duplicating cached zips and Pants / Pex are already both notorious amongst users for excessive cache sizes. If I were working on this problem, I'd want to take a hard look at the existing packed layout and the needs of the zipapp layout and seeing if there were a way to make them be able to share the same zips. Without that, this feature definitely needs to be behind a flag (`--i-opt-in-to-cache-doubling` - clearly not spelled like that!). Now you already mentioned being behind a flag, so I think you're on board there. But, then the convenience argument above goes poof. So I'd really like to either understand a unique use case that makes this better than --layout packed or else an idea about how to change things so that cached zips can be shared. 

> Possibly more importantly, I assume this speedup will also be felt by pants, which often reconstitutes various intermediate pexes full of requirements, using --layout zipapp, in order to perform some task like running pytest. I was halfway to making a repro case in the pants repo but gave up for now.

Pants in fact uses `--layout packed` exclusively for all internal Python operations. It only produces a zip when asked via `package` (AWS Lambda zips, and pex_binary targets with the default zipapp layout).

jsirois:
I wanted to note I won't be back online until ~Friday August 4th.

cosmicexplorer:
> Ok. The docs are in a sorry state to be sure, but that can be remedied.

That wasn't my intention -- please don't assume antagonism. I was attempting to justify my impression that `--layout zipapp` was a widely used standard/default for pex, which is why I had assumed this would be a more immediately attractive improvement. I would love to help improve the docs situation here!

> so more convenient in that you need not type

`--layout zipapp` is also more convenient for many deployment scenarios in that it produces a single output file, which is one of the main use cases I have for the pex utility. 

> Pants in fact uses `--layout packed` exclusively for all internal Python operations. It only produces a zip when asked via the package goal (AWS Lambda zips, and pex_binary targets with the default zipapp layout).

Ok great, I appreciate taking the time to provide the larger context here. In the case of producing a `pex_binary()` with many large 3rdparty deps, such as often occurs with machine learning code, pants users currently have to choose between waiting over a minute or accepting a massive uncompressed zip, even where the exact same command has just been run and everything possible is cached.

> an idea about how to change things so that cached zips can be shared.

I think python's `zipfile` may be able to copy over entries without decompressing and recompressing them the way the rust `zip` crate does. That would be how I would accomplish this while renaming the entries (since the names are not part of the compressed data).

cosmicexplorer:
After a bit of investigation, it seems all the necessary tools to reuse `--layout packed` cache archives for `--layout zipapp` are available in the python `zipfile` library through 2.7.

cosmicexplorer:
Ok, I didn't think it would be possible, but @jsirois (who is off climbing now) was absolutely right to suggest investigating the reuse of `--layout packed` cache entries. This solution:
- doesn't rely on any external binaries or libraries,
- doesn't introduce any new cache entries,
- doesn't rely on any private or undocumented methods,
- **does** produce standards-compliant zip files! You'll notice it looks much more similar to the previous implementation.

The magic trick is in `pex.ziputils.MergeableZipFile`, a subclass of `zipfile.ZipFile` with a method `.merge_archive()` that loops over entries from a source `ZipFile` and accesses its (public!) file pointer attribute `.fp` to directly copy over compressed file contents.

@Eric-Arellano @stuhood: absolutely zero rush on this, but since @jsirois is out right now, please feel free to leave any thoughts/comments whenever you find the time.

As a side note, I'm planning to submit a proposal to PackagingCon about this + pypa/pip#12186 today/this week ([the cfp closes August 6th](https://packaging-con.org/)).

cosmicexplorer:
One thing it would be very helpful to get input on: there are two changes in behavior in the current diff (both of which are subjective, not necessary to achieve the goals of this diff, and can be reverted):
1. **source files now come before `.bootstrap/` in the output of `unzip -l` for `--layout zipapp`**: https://github.com/pantsbuild/pex/blob/7130bd1c57bd00cd81581e0b901c4aba740b5a15/pex/pex_builder.py#L855-L858
2. **all entries in `~/.pex/{bootstrap_zips,packed_wheels}/` have `deterministic_timestamp=True`**: https://github.com/pantsbuild/pex/blob/7130bd1c57bd00cd81581e0b901c4aba740b5a15/pex/pex_builder.py#L816-L820

I suspect the right thing to do is to revert to the prior behavior in this change, then split out separate PRs to propose both of these behavioral changes, but I would like to know whether these changes in behavior are considered modifying the public API.

jsirois:
I'm reviewing presently, but as to:

> 1. source files now come before .bootstrap/ in the output of unzip -l for --layout zipapp:

This is not good for 2 reasons:
1. Zips are tail oriented; so having `__main__.py` in the tail is desirable. That file is executed on every run (unless you use `--sh-boot`) from within the zip and if it was already loaded into the page cache when the end of central directory record was read, that's a small overhead win.
2. When I run `zipinfo my.pex` today, the `.bootstrap/`, etc scroll up and away and I see my user code at the bottom in the console.

Python is so slow compared to (SSD) disk accesses that 1 may not be all that relevant any longer and 2 is eye of the beholder - presumably your zipinfo environment is in a pager and so you see the 1st page 1st. Either way, I think this change is, on the whole, not justified.

> 2. all entries in ~/.pex/{bootstrap_zips,packed_wheels}/ have deterministic_timestamp=True: 

Seems right to me.

cosmicexplorer:
> When I run zipinfo my.pex today, the .bootstrap/, etc scroll up and away and I see my user code at the bottom in the console.

Oops! I was so used to `.deps/` being quite large for the test cases I've been playing with here that I hadn't realized you're describing the more usual case! I had been repeatedly piping everything through `less`. This one is easy to revert; however:
> Zips are tail oriented; so having __main__.py in the tail is desirable.

If I'm hearing you right, it sounds like you're advocating for something a little different: the ordering of:
1. `.bootstrap/`
2. `.deps/`
3. sources, metadata and runner, incl `PEX-INFO` and `__main__.py`

Right now, it's `.bootstrap/`, then sources, then `.deps/` is at the end. To address both of your reasons, it sounds like we should change the behavior from the current? I can push a diff to show what that would look like!

pex/pex_builder.py
```diff
@@ -784,24 +847,87 @@ def _build_zipapp(
...
+
+                # (3) Engage our hacky solution to reuse the same caches as for --layout packed.
+                #
+                #     This reads from a single compressed file per cached resource instead of
+                #     traversing any directory trees. If compress=True, this will also significantly
+                #     reduce the total number of bytes to read and avoid a lot of single-threaded
+                #     computation, because the compressed entries are simply copied over as-is.
+                #
+                #     Note that these cached zips are created in the --layout packed format, without
+                #     the .bootstrap/ or .deps/ prefixes we need to form a proper zipapp. Our zip
```
jsirois, line 889: "proper zipapp" - it's a quibble perhaps, but the directory structure is a pure (internal) Pex thing and has nothing to do with zipapps, which only really care about zip files, the Python zipimporter and `__main__.py`. Maybe "proper Pex zipapp layout".

cosmicexplorer:
Oops! Didn't realize that *was* the current behavior! However, regarding the following:
> Zips are tail oriented; so having main.py in the tail is desirable.

Would it be useful to put `__main__.py` at the literal very end, then?

pex/ziputils.py
```diff
@@ -242,3 +247,93 @@ def isolate_zip(self, out_fp):
...
+            self.fp.write(zinfo.FileHeader())
+
+            # > (3.2) Seek to the start of the source entry's file contents.
+            source_zf.fp.seek(source_offset + source_header_len, io.SEEK_SET)
+            # > (3.3) Copy over the file data verbatim from the source zip's underlying
+            #       file handle.
+            copy_file_range(source_zf.fp, self.fp, zinfo.compress_size)
+
+            # (4) Hack the synthesized ZipInfo onto the destination zip's infos.
+            self.filelist.append(zinfo)
```
jsirois, line 334: So the hack tally is `self.{fp,filelist,NameToInfo}` afaict. Techincally "public", but certainly not documented as such. CI proves this is OK across CPython 2.7, 3.5, 3.7, 3.11 and 3.12 as well as PyPy 2.7 and 3.9. CI will continue to be extensive; so I'm not too worried in the medium term, but it does seem like this should all be behind a flag given the combination of the hackyness and the new cache burden (which is you thankfully eliminated for `--layout packed` users like Pants, but still exists for folks who never use(d) that option). You seemed to be on board with a feature flag ealier - does that still seem reasonable?

Ideally the original tack of a library that handled the zipping, possibly including parallelization in addition to the ability here of merging would take the place of all this to both eliminate the hack factor and speed things up in the cold cache case not to mention make these features available outside Pex and outside Python.

jsirois:
> Would it be useful to put __main__.py at the literal very end, then?

Yes, but `__pex__/__init__.py` also has the same role when the PEX is used as a sys.path entry; i.e.: `PYTHONPATH=my.pex python -c 'from __pex__ import cowsay; ...'` (this is a newish feature you may not be aware of). That said, I think some PR later should probably make the tail order exactly:
```
__pex__/
__pex__/__init__.py
__main__.py
```

The `__main__.py` entrypoint case will probably always be more commonly used than the `__pex__` magic import hook.

jsirois:
Just noticed this:

>> Ok. The docs are in a sorry state to be sure, but that can be remedied.

> That wasn't my intention -- please don't assume antagonism.

I actually mean it. The docs are in a sorry state. They are horrible. I have not spent any time on them in the ~5 years I've been working as the primary Pex maintainer; meanwhile I have spent time adding a lot of features for Pants; so the docs are severely out of date. I do plan to circle back though. I've vetted out a whole system over at https://a-scie/lift based on Sphinx that I will be applying here along with fully re-worked content.

pex/common.py
```diff
@@ -137,6 +138,26 @@ def do_copy():
         do_copy()
 
 
+_COPY_BUFSIZE = 64 * 1024
```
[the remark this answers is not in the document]
cosmicexplorer: This is just what cpython does on non-windows: https://github.com/python/cpython/blob/407d7fda94268d70b92e2e685ca9891ec3641e78/Lib/shutil.py#L48

pex/common.py
```diff
@@ -640,6 +679,7 @@ def write_entry(
                     if deterministic_timestamp
                     else None,
                 )
+                # FIXME: this ignores the zinfo.compress_type value from zip_entry_from_file()!
```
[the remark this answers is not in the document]
cosmicexplorer: You're right -- that was intentionally left as a note to self. It's easy to fix, that comment was meant as a reminder to do so only after I had finished other things.

pex/ziputils.py
jsirois: Perhaps no flag is just fine. I'm a bit conflicted here. I still do not see the real use case for this feature, I'd personally use `--layout packed` for both iteration and distribution (via rsync over ssh), but Pex has never treated new caches as opt-in. Its cache behavior is currently uniformly untunable.

cosmicexplorer: > You seemed to be on board with a feature flag ealier - does that still seem reasonable?

Yes, absolutely! Pip still has the zip file magic I proposed to them behind `--use-feature=fast-deps` for similar reasons, although that's had a lot more people contribute to it since then.

> the new cache burden (which is you thankfully eliminated for --layout packed users like Pants, but still exists for folks who never use(d) that option).

Yes: for this reason alone I think it should be behind a feature flag.

> Ideally the original tack of a library that handled the zipping, possibly including parallelization in addition to the ability here of merging would take the place of all this to both eliminate the hack factor and speed things up in the cold cache case not to mention make these features available outside Pex and outside Python.

Well yes, that definitely hasn't evaporated; in fact, it's quite useful, and indeed takes advantage of rust's structured parallelism, and it's a CLI, a rust library, and a pyo3 python library: https://github.com/cosmicexplorer/medusa-zip. I also got the hang of using github actions and cibuildwheels in the meantime and it's now published to crates.io (`{,lib,py}medusa-zip`) and pypi (`medusa-zip`) upon every tagged release. I've been continuing to speak with @jonjohnsonjr regarding the similarity of this work to Chainguard's use case efficiently merging tarballs and/or containers (see chainguard-dev/apko#781) and considering whether we can share any code. `libarchive` exists, but its goal is compatibility with the widest range of formats, not performance, and it seems plausible that a general framework for efficiently merging the contents of archives might have applicability to a wide range of build tools and package managers.

cosmicexplorer: >  outside Python.

I might propose this `ZipFile#merge_archive()` method to `python-ideas` too, although it seems possibly too specialized to be in the stdlib.

jsirois: Yeah, that seems too specialized. FWICT they have been on a somewhat lazy warpath to do less in the stdlib, not more.

cosmicexplorer:
> That said, I think some PR later should probably make the tail order exactly:

On it! #2209

cosmicexplorer:
@jsirois you already approved this, but I assumed you'd want to review again after I introduced the `--cache-dists` option to gate the use of the packed wheel cache (`--cache-dists` is default off for zipapps, but default on for packed, in order to maintain the current behavior). I have also significantly shorted the commit message, although in this case I've kept the Problem/Solution structure as I feel that describing the tradeoff vs `--no-compress` in the Alternatives section was worth specifically calling out.

pex/pex_builder.py
```diff
@@ -664,17 +672,34 @@ def build(
             else:
                 shutil.rmtree(tmp_pex, True)
 
+        if not compress:
+            if cache_dists is True:
+                pex_warnings.warn(
```
jsirois, line 677: So, say there is some legitimate use case, aka the user knows better. You've now pissed them off by nagging them. It's much cleaner if you just fail fast if that makes sense to do. I'm not commenting on whether or not there is a legitimate use case, my intent is just to point out the cost of combinatorial feature explosion and the nanny vs. document vs. trust the user tradeoffs.

jsirois:
Re the commit description, I'm still not a fan of the formula - you can say the exact same thing without the damn template. But that's just my grumpy dislike of the 5 paragraph essay. More importantly you just whiff the actual alternative, which is the packed layout. I think you should address why that doesn't work.

pex/common.py
```diff
@@ -192,16 +219,20 @@ def zip_entry_from_file(cls, filename, arcname=None, date_time=None):
...
             zinfo.file_size = 0
             zinfo.external_attr |= 0x10  # MS-DOS directory flag
+            # Always store directories decompressed, because they are empty but take up 2 bytes when
+            # compressed.
             zinfo.compress_type = zipfile.ZIP_STORED
             data = b""
         else:
             zinfo.file_size = st.st_size
-            zinfo.compress_type = zipfile.ZIP_DEFLATED
+            # File contents may be compressed or decompressed. Decompressed is significantly faster
```
jsirois, line 233: This seems like a gratuitous comment. You're selling here or else speaking to an end user. Clearly this is not the place to speak to the user and the coder doesn't need to be sold to, if a function takes a param ... it takes a param. If its a public function maybe sell in the API doc.

jsirois (changes requested):
I think adding the uncompressed option to the existing packed layout is a mistake. Seems good otherwise.

jsirois (approved):
As I said, I'll be offline ~presently. Hopefully this approve over-rides the prior request changes. I would like to see the change, but I trust you to do the right thing.

Thanks for putting up with my opinions here. You've started a contribution spike and I'm grateful for that, but I'm pretty open about not viewing all contribution as a good thing by default. As such I'm just trying to set up my expectations / the current Pex code base opinions clearly to help make further contribution more straightforward for us both. My time on this project is limited as I'm sure yours is too; so making things efficient seems to benefit us both.

jsirois:
@cosmicexplorer although this PR is approved, another thing to weigh is #2298 which provides even greater speedups with 0 hacks by skipping zipping entirely with a bonus of skipping unzipping (pre-installing) as well.

cosmicexplorer:
Hey @jsirois, thanks so much for your thoughtful review. I'll be rebasing and responding to your comments, although I understand that pex has changed a bit in the meantime. I think this is a useful change and I have the time to see it through.

cosmicexplorer:
Ah, I'll read through #2298 before doing that. I know you've been pushing back on adding this PR's complexity to pex for a while, and I've since found that some of the tricks here are probably better suited for the rust zip crate. I'll read that change and re-evaluate this now. Thanks.

cosmicexplorer:
Ok, it seems like this change will still provide a speedup in zipapp creation, without incurring the (very slight = O(100ms)) cold-start performance slowdown of `--no-pre-install-wheels`. I personally think it's still quite useful to get a build performance improvement "for free" with `--layout zipapp`, since I believe that's still the default and rather convenient. If you agree, I'll rebase and follow up with this until it's acceptable.

cosmicexplorer:
Ok, I totally buy this comment: https://github.com/pex-tool/pex/pull/2175#discussion_r1288753565. I think I'm willing to close this and move the zip file magic to my work on the zip crate. I was really surprised/impressed that we were able to do this, but `--no-pre-install-wheels` seems like the right way to do this. Thanks for talking me through this: I really appreciate your patience and I always learn from your expertise.

PR #2181, opened 2023-07-21
Add support for selecting packages and modules.

### LINKED ISSUE #2134: Add option to include a Python package/module in the Pex
The `-D DIR` option adds directory entries in `DIR` at the root of the Pex import namespace.

This is awkward given a typical workspace layout like

```
.github/workflows/...
pyproject.toml
README.md
.venv/...
myapp/__init__.py
```

as giving `-D .` includes lots of junk that is also in the workspace (up to and including whole virtualenvs) and giving `-D myapp` does not create a `myapp` package in the Pex, it puts `__init__.py` into the root instead.

A workspace layout with a `src/` directory works but it means that a plain `python` command cannot import `myapp`.

I would propose adding an option (say `-P`) to include a single Python package or module in the PEX.

With this option:

* `-P myapp` would copy the `myapp` package into the PEX and make it importable as `myapp`
* `-P myapp.py` would copy the `myapp.py` module into the PEX and make it importable as `myapp`

jsirois:
This complements the existing `-D` / `--sources-directory` support for
adding local sources and resources with finer-grained control over
what files are included in the PEX file. Notably, this allows cleanly
packaging projects with no `setup.py` / `pyproject.toml` based build
when the projects have their Python code at the top level mixed with
other files that should not be included in the PEX (e.g.: build scripts,
CI configuration, documentation, etc.).

Fixes #2134

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -16,8 +16,6 @@ import sys
 from argparse import Action, ArgumentDefaultsHelpFormatter, ArgumentError, ArgumentParser
 from textwrap import TextWrapper
 
-import attr
-
 from pex import pex_warnings
 from pex.argparse import HandleBoolAction
 from pex.commands.command import (
@@ -52,7 +50,12 @@ if TYPE_CHECKING:
     from argparse import Namespace
     from typing import Dict, Iterable, Iterator, List, Optional, Set, Tuple
 
+    import attr  # vendor:skip
+
     from pex.resolve.resolver_options import ResolverConfiguration
+else:
+    from pex.third_party import attr
+
 
 CANNOT_SETUP_INTERPRETER = 102
 INVALID_OPTIONS = 103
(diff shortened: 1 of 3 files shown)


jsirois:
N.B.: I wanted to convert this Pex PEX packaging to dogfood via `-P pex` replacing `pex_requirement` below:
https://github.com/pantsbuild/pex/blob/de7e6eccd2614d2be50be1db5e3e9b5cb753775b/scripts/package.py#L28-L56

Tests caught the fact that folks rely on the pex3 console script though (Pants for one via `PEX_SCRIPT=pex3 ./pex.pex ...) and console scripts ~require a build (`setup.py` / `pyproject.toml`).

jsirois:
@lordmauve / @khaledh, if either of you is available to take a look, I'd be grateful.

benjyw (approved):

PR #2184, opened 2023-07-22
Fix `pex.pex_bootstrapper.bootstrap_pex_env` leak.

### LINKED ISSUE #2183: The deprecated `pex.pex_bootstrapper.bootstrap_pex_env` API does not scrub `pex.third_party`.
The `pex.pex_bootstrapper.bootstrap_pex_env` is maintained for legacy consumers, including https://github.com/pantsbuild/lambdex and https://github.com/kwlzn/pyuwsgi_pex. As such, despite the `__pex__` magic import package prefix which obviates the need for the entrypoint generally and the `pex3 venv create --prefix` tool for the Lambdex use case in particular, the API should be fixed to scrub `pex.third_party` vendored imports and the Pex bootstrap from the `sys.path` like the `pex.pex.PEX.execute` standard entry point does. Without this, errors like the following are seen for user-code that depends on the latest versions of attrs - which Pex vendors at ~21.5.0 (this is from the Lambdex use case with a user code dependency on attrs 23.1.0 and using Pex 2.1.134):
```
Unable to import module 'lambdex_handler': cannot import name 'AttrsInstance' from 'pex.vendor._vendored.attrs.attr' (/var/task/.bootstrap/pex/vendor/_vendored/attrs/attr/__init__.py)
```

jsirois:
Previously the PEX bootstrap code and third party vendored code were not
demoted and uninstalled respectively like they are for normal PEX
execution.

Fixes #2183

### DIFF
--- a/tests/integration/test_issue_2183.py
+++ b/tests/integration/test_issue_2183.py
@@ -1,0 +1,83 @@
+# Copyright 2023 Pants project contributors (see CONTRIBUTORS.md).
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+import os.path
+import subprocess
+from textwrap import dedent
+
+import pytest
+
+from pex.common import safe_open
+from pex.compatibility import commonpath
+from pex.pep_440 import Version
+from pex.pep_503 import ProjectName
+from pex.pex import PEX
+from pex.testing import PY_VER, make_env, run_pex_command
+from pex.typing import TYPE_CHECKING
+from pex.venv.virtualenv import Virtualenv
+
+if TYPE_CHECKING:
+    from typing import Any
+
+
+@pytest.mark.skipif(
+    PY_VER < (3, 7) or PY_VER >= (3, 12),
+    reason=(
+        "The test requires use of attrs 23.1.0 which requires Python >= 3.7 and Lambdex further "
+        "requires a released version of Pex that supports Python 3.12."
+    ),
+)
+def test_lambdex_with_incompatible_attrs(tmpdir):
+    # type: (Any) -> None
+
+    src = os.path.join(str(tmpdir), "src")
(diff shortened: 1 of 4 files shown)


tests/integration/test_issue_2183.py
```diff
@@ -0,0 +1,83 @@
...
+            "-o",
+            lambda_zip,
+            pex,
+        ]
+    ).assert_success()
+
+    venv_dir = os.path.join(str(tmpdir), "venv_dir")
+    venv = Virtualenv.create(venv_dir=venv_dir)
+    output = (
+        subprocess.check_output(
```
jsirois, line 76: Before the fix this fails with:
```
tests/integration/test_issue_2183.py::test_lambdex_with_incompatible_attrs Traceback (most recent call last):
  File "/tmp/pytest-of-jsirois/pytest-21/test_lambdex_with_incompatible0/lambda.zip/lambdex_handler.py", line 54, in <module>
  File "/tmp/pytest-of-jsirois/pytest-21/test_lambdex_with_incompatible0/lambda.zip/.bootstrap/pex/vendor/_vendored/setuptools/pkg_resources/__init__.py", line 2481, in resolve
  File "/home/jsirois/.pex/unzipped_pexes/9632e41cb34f0318a82cb257fe31f81485c7b6c0/example.py", line 3, in <module>
    from attr import AttrsInstance
ImportError: cannot import name 'AttrsInstance' from 'pex.vendor._vendored.attrs.attr' (/tmp/pytest-of-jsirois/pytest-21/test_lambdex_with_incompatible0/lambda.zip/.bootstrap/pex/vendor/_vendored/attrs/attr/__init__.py)
```

Eric-Arellano (approved):

PR #2188, opened 2023-07-23
Eliminate warnings for default use.

### LINKED ISSUE #2186: Pex warns by default under Python 3.12
Looks like:
```
$ pex cowsay -ccowsay -- 'Moo!'
/home/jsirois/bin/pex.venv/lib/python3.12/site-packages/pex/pip/installation.py:231: PEXWarning: The Pip requested for Bootstrapping Pip 23.2 was pip @ git+https://github.com/pantsbuild/pip@386a54f097ece66775d0c7f34fd29bb596c6b0be but it does not work with any of the targets selected.

Pip 20.3.4-patched requires Python <3.12 and the following target does not apply:
1. /home/jsirois/bin/pex.venv/bin/python3.12

Using Pip 23.2 instead.
  pex_warnings.warn(
  ____
| Moo! |
  ====
    \
     \
       ^__^
       (oo)\_______
       (__)\       )\/\
           ||----w |
           ||     ||
$ pex -V
2.1.139
```

So, despite making accommodations to use Pip 23.2 by default specifically for Python 3.12, it warns anyhow.

jsirois:
Previously `pex` would warn for default usage scenarios under both
Python 3.11 and Python 3.12. For Python 3.12, the `--pip-version`
defaulted to 23.2 but a warning about vendored Pip 20.3.4 was issued
anyway; this is now fixed. For both Python 3.11 and Python 3.12, the
vendored version of packaging used the deprecated `sre_constants`
stdlib module; this is now fixed with packaging 23.1 being vendored
and selected for Python >= 3.7.

Fixes #2186


pex/vendor/__init__.py
```diff
@@ -180,12 +180,14 @@ def iter_vendor_specs(filter_requires_python=False):
...
             "packaging", "20.9", import_path="packaging_20_9", constraints=("pyparsing<3",)
         )
-    if not filter_requires_python or sys.version_info[:2] >= (3, 6):
+    if not filter_requires_python or sys.version_info[:2] == (3, 6):
         # N.B.: The pyparsing constraint is needed because our import re-writer (RedBaron) chokes on
         # newer versions.
         yield VendorSpec.pinned(
             "packaging", "21.3", import_path="packaging_21_3", constraints=("pyparsing<3",)
         )
+    if not filter_requires_python or sys.version_info[:2] >= (3, 7):
```
jsirois, line 189: This 3rd version increases Pex size ~100k from 4103623 bytes to 4205804 bytes.

pex/bin/pex.py
```diff
@@ -886,8 +886,8 @@ def main(args=None):
             except target_configuration.InterpreterConstraintsNotSatisfied as e:
                 die(str(e), exit_code=CANNOT_SETUP_INTERPRETER)
 
-            resolver_configuration = finalize_resolve_config(
-                resolver_configuration, targets, context="PEX building"
+            resolver_configuration = try_(
```
jsirois, line 889: The new test caught this old bug.

scripts/package.py
```diff
@@ -59,7 +59,7 @@ def build_pex_pex(output_file: PurePath, verbosity: int = 0) -> None:
 
 def describe_git_rev() -> str:
     git_describe = subprocess.run(
-        ["git", "describe"], check=True, capture_output=True, encoding="utf-8"
+        ["git", "describe"], check=True, stdout=subprocess.PIPE, encoding="utf-8"
```
jsirois, line 62: Testing the packaging and pytest Python 3.6 edge case (`tox -epy36-integration`) caught this bug: `capture_output` was only added in 3.7. CI only runs unit tests for 3.5, ITs start at 3.7.

Eric-Arellano (approved):

jsirois:
Hrm, so packaging 22.0+ drop support for `LegacyVersion` (an internal type really), but that means support for arbitrary equality `===` in version specifiers is also dropped. As such, this is a backwards incompatible change which Pex doesn't do. I'll back out this portion of the change for now while I think about a path forward for the packaging warnings.

tests/integration/test_issue_1179.py
```diff
@@ -21,11 +21,14 @@ def test_pip_2020_resolver_engaged():
 
     # The Pip legacy resolver cannot solve the following requirements but the 2020 resolver can.
     # Use this fact to prove we're plumbing Pip resolver version arguments correctly.
-    pex_args = ["boto3==1.15.6", "botocore>1.17<1.18.7", "--", "-c", "import boto3"]
+    pex_args = ["boto3==1.15.6", "botocore>1.17,<1.20", "--", "-c", "import boto3"]
```
jsirois, line 24: Modern packaging switched from pyparsing to a be-spoke parser and fixed the parsing bug that allowed `botocore>1.17<1.18.7` previously (missing comma / AND). The `<1.18.7` -> `<1.20` is to force a naive conflict that pip-legacy-resolver can't handle - the actual point of the test. That version, 1.19.63, is now encoded in the test assertions.

PR #2195, opened 2023-07-25
add types to Chroot

cosmicexplorer:
Split off from #2193.


pex/common.py
```diff
@@ -548,10 +565,11 @@ def write(self, data, dst, label=None, mode="wb", executable=False):
         self._ensure_parent(dst)
         with open(os.path.join(self.chroot, dst), mode) as wp:
             wp.write(data)
-        if executable:
-            chmod_plus_x(wp.name)
+            if executable:
```
jsirois, line 568: What was this for? I'm guessing you're wanting to pretend Python doesn't work the way it does with locals being visible everywhere in function scope. This used to bother me as well, but I got used to it. This is fine I guess, but generally avoid this sort of edit I think and just embrace the language you're using.

jsirois (approved):

jsirois:
Same as https://github.com/pantsbuild/pex/pull/2193#issuecomment-1650272580 - I restarted the failed shard for what appears to be an unrelated spurious error (as well as a hung mac shard going on 3h).

Issue #838, opened 2020-01-03
Be able to use user's pip.conf in pip resolution


fhoering:
We have our own internal PyPi server and therefore need to do pip resolution with this server.

Before what we were doing is to inject a custom `--index-url` into bdist_pex 
```
python setup.py bdist_pex .. --index-url=http://criteo/repository/pypi/simple
```
and also via a [custom fetcher](https://github.com/criteo/cluster-pack/blob/master/cluster_pack/packaging.py#L122) when using pex via the Python API.

Using custom configs is currently forbidden because vendored pip runs in [isolated mode](https://github.com/pantsbuild/pex/blob/master/pex/pip.py#L56)

It would be nice be able to use pip.conf now as it would remove the need to add the custom server into the code (we already have `pip.conf` file set up on all our platforms)

Could we introduce an additional flag [here](https://github.com/pantsbuild/pex/blob/master/pex/bin/pex.py#L118)? Something like "--no-pip-isolated" for example.

jsirois:
Related: #336

Issue #336, opened 2017-01-05
pex does not use .pip/pip.conf to resolve packages


salimfadhley:
I'm able to install some packages with Pip like this:

`(venv) [me@myhost glances]$ pip install -r requirements_glances.txt`

However when I attempt an equivalent syntax with Pex I get the following:

```
(venv) [me@myhost glances]$ pex -r requirements_glances.txt -o glances.pex
Could not satisfy all requirements for glances:
    glances
```

It looks like Pex doesn't honor my site config (in ~/.pip/pip.conf) - is there any way to explicitly change the index that Pex uses when resolving requirements?

kwlzn:
```
$ pex -h |grep -A12 "Resolver options"
  Resolver options:
    Tailor how to find, resolve and translate the packages that get put
    into the PEX environment.

    --pypi, --no-pypi, --no-index
                        Whether to use pypi to resolve dependencies; Default:
                        use pypi
    -f PATH/URL, --find-links=PATH/URL, --repo=PATH/URL
                        Additional repository path (directory or URL) to look
                        for requirements.
    -i URL, --index=URL, --index-url=URL
                        Additional cheeseshop indices to use to satisfy
                        requirements.
```

so, `pex -r requirements_glances.txt --no-pypi -i http://your-custom-index -o glances.pex`

salimfadhley:
One issue: The private repository is password protected, do you know if there's a way to provide some authentication information?

`pex -r requirements_glances.txt --no-pypi -i http://<user>:<password>@<server>:8081/artifactory/api/pypi/virtual-pypi/simple -o glances.pex
`
This gives the exact same "could not satisfy all requirements" error.

Also is there a way to debug what HTTP requests Pex is actually making? 

I suppose the work-around is just to pre-fetch all the requirements before running Pex, but I'd rather build everything directly from Artifactory if I can.

sixninetynine:
@salimfadhley I believe that pre-fetching and then using `--no-pypi` is _currently_ your best bet when not using pypi directly. 

@kwlzn I recall there being an issue or PR suggesting tighter integration with `pip` (or at least the underlying libraries powering `pip`). Seems like a lot of the issues in the tracker are regarding compatibility with pip. What are your thoughts on reworking some of the package resolution stuff? Could be useful for folks that work with `conda` as well.

salimfadhley:
Something still seems to be wrong:

```
(venv) [user@server glances]$ ls ./wheelhouse/
[...]

PR #2198, opened 2023-07-27
Add `--devpi` support to tox test environments.

jsirois:
The new `--devpi` support is opt-in for now and CI opts-in. The speed 
gains are modest and variable in CI where the network connection has
low latency and high bandwidth, but this also should improve stability 
there with less network flakes (anecdotally true from a few weeks of 
using this locally and retrying in CI). It's also the right thing to do
vs. "hammering" PyPI for no reason.


jsirois:
This really shines in poor network connection environments. I develop in the 1st world, but often from coffee shops with poor latency and / or bandwidth. In my current work stint this translates to a difference from:
```
$ rm -rf ~/.pex && time tox -epy311-integration
...
===== 533 passed, 15 skipped, 1 xfailed, 2 warnings in 1449.61s (0:24:09) ======
____________________________________________________________________________________________________ summary _____________________________________________________________________________________________________
  py311-integration: commands succeeded
  congratulations :)

real    24m14.599s
user    37m35.162s
sys     4m49.554s
```

To:
```
$ rm -rf ~/.pex && time tox -epy311-integration -- --devpi --shutdown-devpi -vvs
...
====== 533 passed, 15 skipped, 1 xfailed, 2 warnings in 374.44s (0:06:14) ======
INFO Killing devpi server http://127.0.0.1:39423/root/pypi/+simple/ @ 211086
____________________________________________________________________________________________________ summary _____________________________________________________________________________________________________
  py311-integration: commands succeeded
  congratulations :)

real    6m31.697s
user    78m9.170s
sys     9m41.141s
```

All at the modest cost of ~700MB of disk space:
```
$ du -sh ~/.pex_dev/devpi/
678M    /home/jsirois/.pex_dev/devpi/
```

pex/resolve/resolver_configuration.py
```diff
@@ -24,7 +25,7 @@
     from pex.third_party import attr
 
 
-PYPI = "https://pypi.org/simple"
+PYPI = os.environ.get("_PEX_TEST_DEFAULT_INDEX", "https://pypi.org/simple")
```
benjyw, line 28: There's something a little wrinkly about having an affordance for internal tests in production code. It's not terrible, just awkward. 

Could the tests set a custom index via `--no-pypi --index=...`? Or would that then not be testing some path we care about? Possibly we can test that we correctly default to real PyPI in one place and rely on a custom index elsewhere?

Or maybe in pex3 we can switch to having the pypi url be the default value of `--index` (which is its effect in the code anyway), and thus combine `--pypi` and `--index`, and remove some complexity in the options and here?

benjyw (approved):
The code looks fine, but curious why this is necessary vs. CI-caching the download cache?

pex/resolve/resolver_configuration.py
jsirois: Agreed, but passing these arguments in hundreds of tests was not attractive. This actually is intertwined with support for pip.conf in #336 & #838, but one thing at a time.

jsirois:
> The code looks fine, but curious why this is necessary vs. CI-caching the download cache?

The Pip download cache doesn't help stop resolves hitting PyPI unless all top level requirements are pinned and the corresponding downloaded dists have absent or else fully pinned Requires-Dist metadata.

benjyw:
I'm wondering about leveraging this in Pants tests as well, for the same reasons... Seems like that should be possible?

benjyw:
Tracked here: https://github.com/pantsbuild/pants/issues/19628

jsirois:
It should be, yes 

That said, an issue with my growable cache strategy here is it's already up against the GH 10GB total cache size limit. I'll be switching to a cronned docker image rebuild for the Linux shards where that image will contain a single combined devpi cache across all the tested Python + Pip version permutations instead of a GH devpi cache per permutation, which is what I have now and totals ~9.5GB across the ~30 shards.

jsirois:
Finally circling back with numbers to illustrate the difference between no cache, the pex cache, just the pip cache, and just the devpi cache. In short, the devpi cache is by far the most effective in terms of size / speedup tradeoff:
```
1. No caches (no --devpi):
$ rm -rf ~/.pex && time tox -epy311-integration
...
real    23m52.846s
user    31m56.460s
sys     3m31.465s

2. Full ~/.pex cache (no --devpi):
$ time tox -epy311-integration
...
real    16m6.267s
user    38m7.546s
sys     3m57.831s

3. Just the Pip caches (no --devpi):
$ find ~/.pex -maxdepth 1 \! -name "pip*" | xargs rm -rf && time tox -epy311-integration
...
real    24m50.162s
user    31m33.147s
sys     3m31.200s

$ du -sh ~/.pex && echo && du -csh ~/.pex/pip*
1.7G    /home/jsirois/.pex

484M    /home/jsirois/.pex/pip
37M     /home/jsirois/.pex/pip.pex
520M    total

4. No ~/.pex cache but full devpi cache:
$ rm -rf ~/.pex && time tox -epy311-integration -- --devpi -vvs
...
real    5m24.542s
user    63m27.080s
sys     6m15.611s

$ du -sh ~/.pex_dev/devpi
678M    /home/jsirois/.pex_dev/devpi
```

benjyw:
Excellent. Are these numbers representative of CI or of the coffee shop wifi case?

PR #2202, opened 2023-08-02
Ensure files are writable after copy to chroot.

phaer:
Bootstrapping vendored dependencies seems to require files copied into the chroot to be writable, that however seems only to be true if the src directory is writable as well.

So this change intends to ensure files copied to the chroot are always writable by the owner, while leaving other permissions intact. 

It seems to fix the issue described in https://github.com/NixOS/nixpkgs/issues/246448#issuecomment-1659284891

What do you think, @jsirois?

### DIFF
--- a/pex/common.py
+++ b/pex/common.py
@@ -502,6 +502,10 @@ class Chroot(object):
         # type: (str) -> None
         safe_mkdir(os.path.dirname(os.path.join(self.chroot, path)))
 
+    def _ensure_writeable(self, path):
+        path = os.path.join(self.chroot, path)
+        os.chmod(path, os.stat(path).st_mode | stat.S_IWRITE)
+
     def copy(self, src, dst, label=None):
         # type: (str, str, Optional[str]) -> None
         """Copy file ``src`` to ``chroot/dst`` with optional label.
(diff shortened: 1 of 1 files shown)


adisbladis:
I can confirm this solves running `pex3` on NixOS for me.

jsirois:
Ok, bug tracked here: https://github.com/pantsbuild/pex/issues/2203
It's very easy to repro without nix as done in that ticket example and using a similar technique in a test will form a nice basis for ensuring this buggy behavior doesn't backslide going forward. I'll take up that ticket presently and see if I can't make the underlying `.bootstrap` preparation more robust to being repetitive / over-writing itself.

jsirois:
@phaer I've put up an alternate approach here: #2205
If you and / or @adisbladis could give it a look and a spin, I'd be grateful.

phaer:
@jsirois Thanks for checking and the quick patch! Can confirm that it works with nix :)

Definitely the better fix, glad your fix uncovered a deeper issue.

Nitpick regarding the "cp" test: I think that's not entirely accurate as `shutil.copy` does copy the files mode, so it's more comparable to `cp -r` iiuc.

jsirois:
@phaer excellent.

On the nit, I'm a bit confused. Modern `cp` on Linux (my case) does preserve perms by default (no need for `-p`, although you mention `-r`, which I used, presumably as a typo). Further, my example runs `tree -p` after the `cp` to prove the new tree has the same read only perms.

phaer:
@jsirois You are right, I misread - sorry for the noise. I also overlooked that the files don't need to be writable at all as long as there's no duplicate copy, just saw that your new tests cover that - perfect :)

PR #2197, opened 2023-07-25
Removing testing modules from prod code.

jsirois:
Although these were harmless, removing them was made natural with the
introduction of a dedicated test runner script which will be expanded to
include spinning up a `devpi-server` in coming work to bring down
integration test times.

PR #2205, opened 2023-08-03
Fix the Pex CLI to work when installed read-only.

### LINKED ISSUE #2203: The Pex CLI does not work when run from a read-only venv.
Works fine in a typical venv:
```
$ python3.10 -mvenv pex.venv
$ pex.venv/bin/pip -q install -U pip
$ pex.venv/bin/pip -q install -U pex
$ rm -rf ~/.pex
$ pex.venv/bin/pex3 lock create cowsay --indent 2 -o lock.json
running egg_info
writing cowsay.egg-info/PKG-INFO
writing dependency_links to cowsay.egg-info/dependency_links.txt
writing entry points to cowsay.egg-info/entry_points.txt
writing top-level names to cowsay.egg-info/top_level.txt
reading manifest file 'cowsay.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
writing manifest file 'cowsay.egg-info/SOURCES.txt'
running dist_info
creating /tmp/tmplrzt483o/build/cowsay.egg-info
writing /tmp/tmplrzt483o/build/cowsay.egg-info/PKG-INFO
writing dependency_links to /tmp/tmplrzt483o/build/cowsay.egg-info/dependency_links.txt
writing entry points to /tmp/tmplrzt483o/build/cowsay.egg-info/entry_points.txt
writing top-level names to /tmp/tmplrzt483o/build/cowsay.egg-info/top_level.txt
writing manifest file '/tmp/tmplrzt483o/build/cowsay.egg-info/SOURCES.txt'
reading manifest file '/tmp/tmplrzt483o/build/cowsay.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
writing manifest file '/tmp/tmplrzt483o/build/cowsay.egg-info/SOURCES.txt'
creating '/tmp/tmplrzt483o/build/cowsay.dist-info'
adding license file "LICENSE.txt" (matched pattern "LICEN[CS]E*")
$ pex.venv/bin/pex --lock lock.json cowsay -c cowsay -- 'Moo!'
  ____
| Moo! |
  ====
    \
     \
       ^__^
       (oo)\_______
       (__)\       )\/\
           ||----w |
           ||     ||
```

But in a read-only venv Pex commands fail like so:
```
$ rm -rf ~/.pex
$ chmod -R -w pex.venv/lib/python3.10/site-packages
$ pex.venv/bin/pex3 lock create cowsay --indent 2 -o lock.json
Failed to spawn a job for /home/jsirois/support/pex/phaer/pex.venv/bin/python3.10: [Errno 13] Permission denied: '/home/jsirois/.pex/pip/20.3.4-patched/pip.pex/d391aba428cc037d796ca2e7c83909a1db40f322.lck.work/.bootstrap/pex/__init__.py'
```

The issue here is that Pex trips over itself when constructing the `.bootstrap` by attempting to copy `pex/__init__.py` and other root `pex.*` package `__init__.py` files multiple times. The 1st copy works, but since the source file is read only, the destination file is as well, and subsequent copies (over-writes), fail.

This issue was brought to light by https://github.com/NixOS/nixpkgs/issues/246448 where read-only venvs - which are not typically a thing, are used by Nix.

jsirois:
When Pex is run from a read-only environment such as a venv that has
been made read-only after the fact or a nix environment, it must not
attempt to over-write its own vendored `.bootstrap` code in the process
of creating PEX runtime environments. Fix `.bootstrap` creation to only
write bootstrap files once.

Fixes #2203

### DIFF
--- a/pex/pex_builder.py
+++ b/pex/pex_builder.py
@@ -562,25 +562,25 @@ class PEXBuilder(object):
     def _prepare_bootstrap(self):
         from . import vendor
 
-        vendor.vendor_runtime(
-            chroot=self._chroot,
-            dest_basedir=self._pex_info.bootstrap,
-            label="bootstrap",
-            # NB: We use pip here in the builder, but that's only at buildtime and
-            # although we don't use pyparsing directly, packaging.markers, which we
-            # do use at runtime, does.
-            root_module_names=["attr", "packaging", "pkg_resources", "pyparsing"],
-        )
+        # NB: We use pip here in the builder, but that's only at build time, and
+        # although we don't use pyparsing directly, packaging.markers, which we
+        # do use at runtime, does.
+        root_module_names = ["attr", "packaging", "pkg_resources", "pyparsing"]
+        include_dist_info = set()
         if self._pex_info.includes_tools:
             # The `repository extract` tool needs setuptools and wheel to build sdists and wheels
             # and distutils needs .dist-info to discover setuptools (and wheel).
-            vendor.vendor_runtime(
-                chroot=self._chroot,
-                dest_basedir=self._pex_info.bootstrap,
-                label="bootstrap",
-                root_module_names=["setuptools", "wheel"],
-                include_dist_info=True,
-            )
+            for project in "setuptools", "wheel":
+                root_module_names.append(project)
+                include_dist_info.add(project)
+
+        prepared_sources = vendor.vendor_runtime(
+            chroot=self._chroot,
+            dest_basedir=self._pex_info.bootstrap,
+            label="bootstrap",
+            root_module_names=root_module_names,
+            include_dist_info=include_dist_info,
+        )
 
         bootstrap_digest = hashlib.sha1()
         bootstrap_packages = ["third_party", "venv"]
@@ -591,9 +591,11 @@ class PEXBuilder(object):
                 dirs[:] = bootstrap_packages
 
             for f in filter_pyc_files(files):
-                if f.endswith("testing.py"):
-                    continue
                 abs_src = os.path.join(root, f)
+                # N.B.: Some of the `pex.*` package files (__init__.py) will already have been
+                # prepared when vendoring the runtime above; so we skip them here.
+                if abs_src in prepared_sources:
+                    continue
                 with open(abs_src, "rb") as fp:
                     data = fp.read()
                 self._chroot.write(
--- a/tests/integration/test_issue_2203.py
+++ b/tests/integration/test_issue_2203.py
@@ -1,0 +1,57 @@
+# Copyright 2023 Pants project contributors (see CONTRIBUTORS.md).
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+import os.path
+import stat
+import subprocess
+
+from pex.common import safe_rmtree
+from pex.typing import TYPE_CHECKING
+from pex.venv.virtualenv import Virtualenv
+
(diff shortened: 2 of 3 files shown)


tests/integration/test_issue_2203.py
```diff
@@ -0,0 +1,57 @@
...
+
+    assert_pex_works()
+
+    write_mask = stat.S_IWUSR | stat.S_IWGRP | stat.S_IWOTH
+    for root, dirs, files in os.walk(venv.site_packages_dir, topdown=False):
+        for path in files + dirs:
+            abs_path = os.path.join(root, path)
+            os.chmod(abs_path, os.stat(abs_path).st_mode & ~write_mask)
+
+    assert_pex_works()
```
jsirois, line 57: This failed in the same manner as described in https://github.com/pantsbuild/pex/issues/2203 and https://github.com/NixOS/nixpkgs/issues/246448 prior to the fix.

pex/pex_builder.py
```diff
@@ -591,9 +591,11 @@ def _prepare_bootstrap(self):
                 dirs[:] = bootstrap_packages
 
             for f in filter_pyc_files(files):
-                if f.endswith("testing.py"):
```
benjyw, line 594: Why this change?

benjyw (approved):

pex/pex_builder.py
jsirois: Ah, right. You reviewed that one actually in #2197. I noticed this miss in that PR while I was in this code.

pex/pex_builder.py
```diff
@@ -591,9 +591,11 @@ def _prepare_bootstrap(self):
...
-                    continue
                 abs_src = os.path.join(root, f)
+                # N.B.: Some of the `pex.*` package files (__init__.py) will already have been
+                # prepared when vendoring the runtime above; so we skip them here.
+                if abs_src in prepared_sources:
+                    continue
```
cosmicexplorer, line 598: This particular part regarding files that "will already have been prepared" and how it relates to writing files into the read-only directory is confusing to me and I would appreciate a quick note on how we avoid writing to a read-only dir with this deduplication process if you have a moment. 

This exact line/`if` check is the part that fixes #2203, right? I'm confused as to why not writing duplicate files (which clearly seems good, and probably improves performance too) is fixing the issue with writing to a read-only directory.
- I only see mutating a single `Chroot` instance here, which I assume is a directory we're allowed to write to.
- I see that we reference `_ABS_PEX_PACKAGE_DIR` in this method, which I assume is the bootstrap directory (which may be read-only) that we were previously writing files to, but I don't see how we were writing any files to it.
- I also see that `vendored_sources` stops us from writing the same files to the chroot, which is another deduplication; but that just looks like an optimization (and when I remove it, the new test still passes)?

Thanks for your time, but no pressure to respond, as this is mostly out of curiosity.

cosmicexplorer (commented):
non-blocking comment, thanks for including me!

pex/pex_builder.py
cosmicexplorer: nvm; found the reasoning at #2203! sorry for ping
> The issue here is that Pex trips over itself when constructing the .bootstrap by attempting to copy pex/__init__.py and other root pex.* package __init__.py files multiple times. The 1st copy works, but since the source file is read only, the destination file is as well, and subsequent copies (over-writes), fail.

PR #2209, opened 2023-08-04
colocate __main__.py with the central directory record of the zipapp

cosmicexplorer:
When the Python interpreter executes a zipapp, it will import `__main__.py`, while `__pex__/__init__.py` will be loaded if the pex file is on the `PYTHONPATH`. This file is loaded from the zip file handle with `zipimport`, and the PEX bootstrap script at that location will then import some of the code in `.bootstrap/` before the zipapp is unpacked and compiled. Because a zip file's central directory records are located at the end of the file, placing this code at the end of the zip file list will reduce the amount of seeking the Python interpreter needs to perform in order to execute the bootstrap script. This was previously discussed in https://github.com/pantsbuild/pex/pull/2175#issuecomment-1665710187 and https://github.com/pantsbuild/pex/pull/2175#issuecomment-1665800458.

Changes involved:
1. Make `Chroot#zip()` respect the relative ordering of labels when collecting from `self.filesets`.
2. Put `"main"` and `"importhook"` labels at the end of the zipapp.

### DIFF
--- a/pex/pex_builder.py
+++ b/pex/pex_builder.py
@@ -4,6 +4,7 @@
 from __future__ import absolute_import
 
 import hashlib
+import itertools
 import logging
 import os
 import shutil
@@ -39,7 +40,7 @@ from pex.typing import TYPE_CHECKING
 from pex.util import CacheHelper
 
 if TYPE_CHECKING:
-    from typing import Dict, Optional
+    from typing import ClassVar, Dict, Iterable, Iterator, Optional, Tuple
 
 
 # N.B.: __file__ will be relative when this module is loaded from a "" `sys.path` entry under
@@ -95,14 +96,14 @@ def __maybe_run_venv__(pex, pex_root, pex_path):
 
   venv_dir = venv_dir(
     pex_file=pex,
-    pex_root=pex_root, 
+    pex_root=pex_root,
     pex_hash={pex_hash!r},
     has_interpreter_constraints={has_interpreter_constraints!r},
     pex_path=pex_path,
   )
   venv_pex = os.path.join(venv_dir, 'pex')
   if not __execute__ or not is_exe(venv_pex):
-    # Code in bootstrap_pex will (re)create the venv after selecting the correct interpreter. 
+    # Code in bootstrap_pex will (re)create the venv after selecting the correct interpreter.
     return venv_dir
 
   TRACER.log('Executing venv PEX for {{}} at {{}}'.format(pex, venv_pex))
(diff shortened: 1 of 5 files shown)


pex/pex_builder.py
```diff
@@ -710,15 +764,20 @@ def _build_packedapp(
...
-        pex_info.update(PexInfo.from_env())
+        pex_info = self._setup_pex_info()
 
         # Include user sources, PEX-INFO and __main__ as loose files in src/.
-        for fileset in ("executable", "importhook", "main", "manifest", "resource", "source"):
-            for f in self._chroot.filesets.get(fileset, ()):
-                dest = os.path.join(dirname, f)
-                safe_mkdir(os.path.dirname(dest))
-                safe_copy(os.path.realpath(os.path.join(self._chroot.chroot, f)), dest)
+        with TRACER.timed("copying over uncached sources", V=9):
```
jsirois, line 770: Ok, this seems fine - but boy oh boy does it seem _incredibly_ complicated vs just ordering the labels correctly in the existing tuple on the LHS and writing a good block comment just above.

jsirois (approved):

jsirois:
@cosmicexplorer I'm a huge non-fan of Problem / Solution formulas for writing commit messages. We're adults and I think you can just write a good message. Quoting people - not helpful. Just explaining the change and its reason should stand on its own.

pex/pex_builder.py
cosmicexplorer: You're right, and this change could be reduced quite a bit by providing the labels literally (for example, we could avoid changing this method at all). I thought it was useful to document the meanings of each label, since they're not really documented elsewhere unless you grep for the literal strings `"executable"`, `"importhook"`, etc. I was also thinking that naming each section with a classmethod would make it easier for someone to subclass `PEXBuilder` and insert their own files into the appropriate section while retaining the performance benefit for zipapps. If that's not a supported use case, we could very well reduce the size of this change by a lot.

cosmicexplorer:
> I'm a huge non-fan of Problem / Solution formulas for writing commit messages. 
Ok, I will keep that in mind for the future.

> Quoting people - not helpful. Just explaining the change and its reason should stand on its own.
In this case it's not actually documented anywhere else that the python interpreter accesses `__main__.py` or `__pex__/__init__.py` from the same zip file handle (which I confirmed by looking through the CPython source), so I thought the quote was precisely the rationale needed for this change. However, given this feedback, I have rewritten the OP to be much more concise while simply linking to instead of directly quoting the prior discussions. Is that closer to what you're looking for?

cosmicexplorer:
You're right; I have fixed the wording to clarify that `__pex__/__init__.py` is loaded if the pex file is on the `PYTHONPATH`.

jsirois:
@cosmicexplorer since this is just over 1.5 years old, I'm going to close. You've lost interest and the speed gains are likely in the noise / I've lost interest in pursuing this nugget.

Issue #2155, opened 2023-05-29
Non-deterministic output due to non-deterministic os.walk and unix file permissions of original files


ento:
### Context
I'm using Pants to package up a GitHub Actions action I started writing in Python. The packaged pex file is checked in to the repo, so that when the action gets run, all dependencies get pulled down as part of the action itself, similar to [how actions/typescript-action uses ncc](https://github.com/actions/typescript-action#publish-to-a-distribution-branch
).

### Problem and current workaround
It appeared to work fine until I added a CI check that builds a pex file in the CI environment and verifies the checked in pex file is identical to the one that just got built: even though none of the source files nor dependencies changed, the check didn't pass.

I tracked down the causes and was able to use a forked pex repo with a few changes to make the check pass: https://github.com/pantsbuild/pex/compare/main...ento:pex:deterministic-bootstrap

1. `os.walk` was yielding directories in a non-deterministic order, apparently. Fixed by applying `.sort()` in a few places.
   * I wrote a few tests that reproduces this issue: https://github.com/pantsbuild/pex/compare/main...ento:pex:deterministic-os-walk
2. My local environment and the CI environment had different umask values, and because pex preserves filesystem permissions, the resulting zip archive ended up being different too. Fixed by applying something like a umask when adding entries to the zip archive.
   * Although this worked in my case, I don't think the way I did it really guarantees reproducibility: umask at the OS level works fine when it comes to reproducibility because the default permission is something constant when a new filesystem entry is created (although I don't know if that's the same across all (POSIX-compliant?) OS'es), but in the case of pex, there's no default permission per se and it uses the permission of the entry on disk as the starting point.

### Questions
I can work on opening a PR or two to address these sources of non-determinism if that's something good to incorporate into pex.

1. For `os.walk`, I'm thinking of adding a wrapper around `os.walk` like `deterministic_os_walk` and using it at least in the two places I needed to patch. Would it be a good idea to use it everywhere, or should it be limited to just these two places? Or is this something out of scope of guarantees that pex is willing to provide?
2. Is providing deterministic builds when it comes to unix file permissions something within scope of pex?  If so, what would be a good approach?
   * I did think of specifying the umask in the CI environment and not handling it within pex when I was trying to come up with a fix, but enforcing the umask value in local dev environments is going to be hard, unless all contributors' machines are set up the same way, as the value would need to be set before you clone the GitHub actions repo.
[...]

PR #2220, opened 2023-08-14
Traverse directories in stable order when building a PEX

ento:
Addresses one part of #2155.

Adds a wrapper for `os.walk` that sorts directory entries as it traverses directories. As discussed in the issue, `os.walk` may yield directory entries in a different order depending on (possibly) the file system.

I don't know if the two `os.walk` calls that this PR replaces are enough for addressing this class of non-determinism in pex, but these were sufficient for solving the issue in my case. Happy to expand the use of the wrapper if there are other places that it should be used.

### DIFF
--- a/pex/common.py
+++ b/pex/common.py
@@ -694,8 +694,8 @@ class Chroot(object):
                         yield full_path, path
                         continue
 
-                    for root, _, files in os.walk(full_path):
-                        for f in sorted(files):
+                    for root, _, files in deterministic_walk(full_path):
+                        for f in files:
                             if exclude_file(f):
                                 continue
                             abs_path = os.path.join(root, f)
--- a/pex/common.py
+++ b/pex/common.py
@@ -30,6 +30,7 @@ if TYPE_CHECKING:
         Dict,
         Iterable,
         Iterator,
+        List,
         NoReturn,
         Optional,
         Set,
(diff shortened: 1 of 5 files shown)


testing/__init__.py
```diff
@@ -791,7 +791,7 @@ def __call__(self, *args, **kwargs):
     def _increment_counter(self, counter_key):
         # type: (str) -> int
         self._counters[counter_key] += 1
-        return self._counters[counter_key]
+        return cast(int, self._counters[counter_key])
```
jsirois, line 794: It would be better to type `self._counters = defaultdict(int)  # type: DefaultDict[str, int]` where you declare / initialize, but it would be arguably even better to use `collections.Counter` (with similar declaration site typing) which is exactly built for this.

tests/test_common.py
```diff
@@ -140,6 +142,39 @@ def test_chroot_zip():
             assert b"data" == zip.read("directory/subdirectory/file")
 
 
+def test_chroot_zip_is_deterministic():
+    # type: () -> None
+    with temporary_dir() as tmp:
+        root_dir = os.path.join(tmp, "root")
+        os.mkdir(root_dir)
+        dir_a = os.path.join(root_dir, "a")
+        os.mkdir(dir_a)
```
jsirois, line 151: The `os.mkdir`s are fine, but touch takes care of this.

pex/common.py
```diff
@@ -236,6 +237,24 @@ def open_zip(path, *args, **kwargs):
         yield zip
 
 
+def deterministic_walk(*args, **kwargs):
+    # type: (*Any, **Any) -> Iterator[Tuple[str, List[str], List[str]]]
+    """Walk the specified directory tree in deterministic order.
+
+    Takes the same parameters as os.walk and yields tuples of the same shape.
```
jsirois, line 244: It seems like calling out the `topdown` restriction, would be good.

jsirois (approved):
Thanks @ento. This LGTM.

tests/test_common.py
ento: Ah, forgot to follow up here. Looking into it now

ento: Done. Fingers crossed the new commit doesn't end up needing multiple CI runs.

Also thanks for the other review comments - I learned a few things.

jsirois: > Fingers crossed the new commit doesn't end up needing multiple CI runs.

The change in #2198, which just landed aheaded of this PR, should help with things going forward. If you continue to do development on Pex, you too can take advantage with `tox -e<test env> -- --devpi`.

jsirois:
Thanks again @ento. This should go out in 2.1.144 tomorrow.

jsirois:
Well, tomorrow came and went, but this is now released @ento:
+ https://pypi.org/project/pex/2.1.144/
+ https://github.com/pantsbuild/pex/releases/tag/v2.1.144

PR #2227, opened 2023-08-22
Use a data-only image for cache in CI.

jsirois:
The previous cache strategy is left in place for Mac, but it was too
expensive for the 10GB limit GitHub imposes on total cache size and was
leading to cache thrash.

The new strategy sidesteps the cache size limit by creating a cache
offline in a data-only image and switching the Linux CI jobs - the
lion's share - to run via docker / `./dtox.sh`. It takes ~3 minutes to
download the images now whereas loading the cache took ~30 seconds
previously, but there is no longer any worry about cache size limits.
Obviously running the same thing you run locally in CI is a big benefit
and was one of the goals when I introduced `./dtox.sh` in ~2018. The
downside of the image load times could possibly be overcome by using the
GH cache with docker save / load, but this is good enough for now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ env:
 concurrency:
   group: CI-${{ github.ref }}
   # Queue on all branches and tags, but only cancel overlapping PR burns.
-  cancel-in-progress: ${{ github.ref != 'refs/heads/main' || !startsWith(github.ref, 'refs/tags/') }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' && !startsWith(github.ref, 'refs/tags/') }}
 jobs:
   org-check:
     name: Check GitHub Organization
@@ -23,7 +23,7 @@ jobs:
       - name: Noop
         run: "true"
   checks:
-    name:  TOXENV=format-check,typecheck,vendor-check,package -- --additional-format sdist --additional-format wheel
+    name:  tox -e format-check,typecheck,vendor-check,package
     needs: org-check
     runs-on: ubuntu-22.04
     steps:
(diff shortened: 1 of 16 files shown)


tox.ini
```diff
@@ -217,3 +218,13 @@ commands =
     devpi-server \
     --indent 2 \
     -o testing/devpi-server.lock
+
+[testenv:build-cache-image]
```
jsirois, line 222: Automating this (and re-creation / push of the base image as needed well) is tracked in #2233.

testing/bin/run_tests.py
```diff
@@ -20,27 +20,50 @@
...
+        default="INFO",
+        choices=["DEBUG", "INFO", "WARN", "WARNING", "ERROR", "CRITICAL"],
+        help="Set the logging level (case insensitive).",
+    )
     parser.add_argument("--color", default=None, action="store_true", help="Force colored logging.")
     parser.add_argument(
         "--devpi", action="store_true", help="Proxy PyPI through a local devpi server."
     )
+    parser.add_argument(
+        "--require-devpi",
```
jsirois, line 54: This is used in a subsequent commit in this PR to ensure the job that builds up the cache is actually populating a devpi-server data dir - the ~whole point.

testing/bin/run_tests.py
```diff
@@ -76,27 +99,49 @@ def main():
...
             timeout=options.devpi_timeout,
             max_connection_retries=options.devpi_max_connection_retries,
             request_timeout=options.devpi_request_timeout,
         )
-        if not launch_result:
-            logger.warning("Failed to launch devpi server. Continuing without it...")
-        else:
+        if isinstance(launch_result, devpi.LaunchResult):
             os.environ["_PEX_TEST_DEFAULT_INDEX"] = launch_result.url
+            os.environ["PIP_INDEX_URL"] = launch_result.url
```
jsirois, line 122: This was a leak I noticed when fixing one of the broken tests in a later commit in this PR. Some test code uses `pip install`, etc. directly and that was still going out to PyPI.

dtox.sh
```diff
@@ -4,41 +4,69 @@ set -euo pipefail
...
     --build-arg GROUP="$(id -gn)" \
     --build-arg GID="$(id -g)" \
-    --tag pantsbuild/pex:user \
-    --label user_hash="${user_hash}" \
+    --tag pantsbuild/pex/user:latest \
+    --tag "pantsbuild/pex/user:${user_hash}" \
     "${ROOT}/docker/user"
 fi
 
+if [[ "${CACHE_MODE}" == "pull" ]]; then
```
jsirois, line 52: This is actually the only bit that has to do with the new cache data-only image, the other changes in this script support interop between local build and seeding with `*_MODE=pull` as well as test fixes (altering a few volume mount paths).

.github/workflows/ci.yml
```diff
@@ -42,370 +42,70 @@ jobs:
...
-    env:
-      _TOX_ENV: py${{ matrix.python-version[0] }}${{ matrix.python-version[1] }}-pip${{ matrix.pip-version }}-integration
+        tox-env:
+          - py27-pip20
+          - py35-pip20
+          - py311-pip20
+          - py311-pip22_3_1
+          - py311-pip23_1_2
+          - py312-pip23_2
+          - pypy310-pip20
```
jsirois, line 60: I dropped pypy27 testing as a pragmatic / cowardly decision. The base image pypy runs into an OpenSSL linking issue in 1 integration test and after much effort I gave up. Unlike CPython 2.7, I'm pretty sure folks have moved on from PyPy 2.7. I'm guessing there are no support contracts for the latter like there are for the former.

.github/workflows/ci.yml
```diff
@@ -42,370 +42,70 @@ jobs:
...
-            pip-version: 20
-          - os: ubuntu-22.04
-            pypy-version: [ 3, 9 ]
-            pip-version: 22_3_1
-          - os: ubuntu-22.04
-            pypy-version: [ 3, 9 ]
-            pip-version: 23_1_2
-    env:
-      _TOX_ENV: pypy${{ join(matrix.pypy-version, '') }}-pip${{ matrix.pip-version }}-integration
+          - python-version: [ 3, 11 ]
```
jsirois, line 97: I dropped CPython 2.7 tests for Mac to keep things simple and because its almost a certainty Mac (~I need the latest hot thing, and they do not support their old things) + Ancient Python is just not a thing.

dtox.sh
jsirois: And this is a pretty particular trick I had to fight a bit for. I'll add a comment about the necessity of an empty volume + a run to cause the local volume to be filled from the data image.

docker/cache/populate_cache.sh
```diff
@@ -0,0 +1,32 @@
...
+fi
+
+function run_tox() {
+  local env="$1"
+  tox -e "${env}" -- --color --devpi --require-devpi -vvs
+  if (( $? == 42 )); then
+    echo >&2 "tox -e ${env} failed to start or connect to the devpi-server, exiting..."
+    exit 1
+  elif (( $? != 0 )); then
+    echo >&2 "tox -e ${env} failed, continuing..."
```
jsirois, line 18: This is very deliberate. If 2 tests fail out of ~500 (common enough when running without a devpi-server cache!), the devpi-server cache is likely ~fully populated for the run and we got what we needed. Compound this with ~19 permutations of IT being run through this - a ~4 hour process on a fast machine - and `set -e` would lead to no cache image ever successfully building.

benjyw: So the idea is to manually eyeball that the failures were only a small % when building the data-only image?

benjyw (approved):

docker/cache/populate_cache.sh
jsirois: Not really (that's hard to do if you have a docker that uses BuildKit anyway with its default fancy tui logging). I just compare sizes of the dev cache:
```
$ ./dtox.sh inspect
jsirois@db935e314f12:/development/pex$ du -shc /development/pex_dev/*
844M    /development/pex_dev/devpi
1.6G    /development/pex_dev/pyenv
2.4G    total
```
844M is a little bigger than my local dev cache, which makes sense since I don't always run all tox environments locally and which tests are run or what exactly is downloaded does vary a little bit by interpreter version.

When automating the cache image build, I'm guessing no action will be needed in practice, the failure rates seen in the wild in ITs are O(1/500) when an IT network flakes and not all IT runs in CI do (most don't). If I do take action, I think a cache size comparison to the prior cache image - maybe no more than 1% shrinkage, would be the easiest check that is both somewhat robust and tunable in an easy way.

PR #2229, opened 2023-08-27
Raise the flit-core limit for Python 3

tjni:
flit-core is on version 3.9 now so I hope we can unpin this.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["flit_core >=2,<3"]
+requires = ["flit_core>=2,<4"]
 build-backend = "flit_core.buildapi"
 
 [tool.flit.metadata]


jsirois:
@tjni why do you hope this? IIRC the upper bound was mandated by Pex's continued support for Python 2.7, which is non-negotiable in the Pex 2.x release series.

tjni:
I'm in an environment that is only using newer versions of flit-core. Your rationale for pinning makes sense, I updated the PR to keep the pin in Python 2 environments.

jsirois:
I'm sorry to keep pressing, but in what environment do you care about build requirements for a project that distributed a universal wheel? Pex vendors what would otherwise be its install requirements in order to be usable in any runtime environment. I have never thought of doing the same for its build requirement. The more detail you can provide that explains how this current setup affects you would be great. Although this seems harmless I'd like to understand the why before proceeding.

tjni:
Sure, I understand and don't mind the questions. I work on packaging in [nixpkgs](https://github.com/NixOS/nixpkgs) and we build everything from source. We also don't keep around more than one version of a package (with rare exceptions), so we spend some energy on getting packages to work well with newer versions of their dependencies.

jsirois:
Ok, that makes sense then. We'll see what CI says, but I think your 1st change may have been sufficient although I prefer at least an upper bound on the major version on the grounds major versions are "allowed" to break folks and I don't want my build to break. I'm not sure if you've been `tox -e<env>` to test this out locally or not.

tjni:
I'm happy to add the upper limit as well as try out just the first one if this CI run is happy.

jsirois:
Ok, I think just a single `flit_core>=2,<4` should do it. Pip should pick the right flit for Python 2 & 3 respectively and the upper bound keeps the semver major version of the current 3.9.0 stable release so we don't pick up any breaks allowed in 4.x releases by mistake. This means nix may suffer pain in the future, but I imagine caring about build requirements already puts nix in this position more generally.

tjni:
Let me update it and we can see what CI says.

> This means nix may suffer pain in the future, but I imagine caring about build requirements already puts nix in this position more generally.

I appreciate you allowing the bound relaxation. I didn't realize you were still supporting Python 2, so doing this certainly adds more maintenance surface area, although I hope not a lot.

jsirois (approved):
Thanks @tjni. This will be released shortly in Pex 2.1.145 via #2231.

Issue #2235, opened 2023-09-10
`secrets` not in `sys.path` when using `--venv` 


yjabri:
# Overview
This arose from a discussion on [slack](https://pantsbuild.slack.com/archives/C046T6T9U/p1694200846761159). The repo https://github.com/yjabri/debug-pex-import describes how to reproduce the bug in the `README.md`.

The culprit appears to be [pex_bootstrapper.py#L672](https://github.com/pantsbuild/pex/blob/main/pex/pex_bootstrapper.py#L672). Just prior to the line, in a debugger I can successfully `import secrets`. (A cause of `flask` failing to import). However just after the line, in a new debugger session `import secrets` raises a `ModuleNotFoundError` exception.

## Some details

The line [https://github.com/pantsbuild/pex/blob/main/pex/pex_bootstrapper.py#L672](pex_bootstrapper.py#L672 ) converts my `sys.path` from

```python
['/home/yjabri/Projects/debug-pex-import',
'/home/yjabri/.pyenv/versions/3.10.13/lib/python310.zip',
'/home/yjabri/.pyenv/versions/3.10.13/lib/python3.10',
'/home/yjabri/.pyenv/versions/3.10.13/lib/python3.10/lib-dynload',
'',
'/home/yjabri/.local/lib/python3.10/site-packages',
'/home/yjabri/.pyenv/versions/3.10.13/lib/python3.10/site-packages',
'dist/.pex' ]
```

to

```python
['/home/yjabri/.pex/venvs/6c8eaec5b07b6fe5d8c257cd72d1297dfafdfc8e/779eb2cc0ca9e2fdd204774cbc41848e4e7c5055/lib/python3.10/site-packages',
'dist/.pex']
 ```
 
 My understanding is that this doesn't include standard library which would be in  `'/home/yjabri/.pyenv/versions/3.10.13/lib/python3.10'`
 
For reference, `python dist/.pex -c "import sys; print(sys.path)"` outputs

```python
['',
'/home/yjabri/.pex/venvs/6c8eaec5b07b6fe5d8c257cd72d1297dfafdfc8e/779eb2cc0ca9e2fdd204774cbc41848e4e7c5055',
'/home/yjabri/.pyenv/versions/3.10.13/lib/python310.zip',
'/home/yjabri/.pyenv/versions/3.10.13/lib/python3.10',
'/home/yjabri/.pyenv/versions/3.10.13/lib/python3.10/lib-dynload',
[...]

PR #2236, opened 2023-09-15
Fix non executable venv sys path bug

yjabri:
This PR fixes the bug  identified in https://github.com/pantsbuild/pex/issues/2235 when importing dependencies from a non executable pex  created with the `--venv ` flag. 

For example, the snippet

```
import sys
sys.path.append('flask.pex')  # via `pex Flask -o flask.pex --venv`
import __pex__
import flask
```

would mangle `sys.path` removing necessary paths such as `/usr/lib/python3.9` when executing against `/usr/bin/python3.9`.

### DIFF
--- a/tests/integration/test_issue_2235.py
+++ b/tests/integration/test_issue_2235.py
@@ -13,7 +13,7 @@ def test_standard_library_is_included(
 ):
     # type: (...) -> None
 
-    built_pex_path = os.path.join(tmpdir, "flask.pex")
+    built_pex_path = os.path.join(str(tmpdir), "flask.pex")
     run_pex_command(
         args=[
             "Flask==2.3.3",
@@ -23,7 +23,7 @@ def test_standard_library_is_included(
         ]
     ).assert_success()
 
-    output_path = os.path.join(tmpdir, "output.txt")
+    output_path = os.path.join(str(tmpdir), "output.txt")
     script = """
 import sys
 error = ""
@@ -41,7 +41,7 @@ with open("{}", 'w') as f:
         built_pex_path, output_path
     )
 
-    script_path = os.path.join(tmpdir, "script.py")
+    script_path = os.path.join(str(tmpdir), "script.py")
     with open(script_path, "w") as f:
         f.write(script)
 
--- a/tests/integration/test_issue_2235.py
+++ b/tests/integration/test_issue_2235.py
@@ -29,6 +29,7 @@ def test_standard_library_is_included(
             "Flask==2.3.3",
             "-o",
             "flask.pex",
+            "--venv",
         ],
         check=True,
         env=os.environ,
--- a/tests/integration/test_issue_2235.py
+++ b/tests/integration/test_issue_2235.py
@@ -49,7 +49,7 @@ with open("{}", 'w') as f:
     with open(script_path, "w") as f:
         f.write(script)
 
-    subprocess.run([sys.executable, script_path], check=True)  # type: ignore
+    subprocess.check_call([sys.executable, script_path])
 
     with open(output_path, "r") as f:
         output = f.read()
--- a/tests/integration/test_issue_2235.py
+++ b/tests/integration/test_issue_2235.py
@@ -4,45 +4,32 @@ import sys
 
 import pytest
 
-from pex.compatibility import PY3
-from testing import PY38, ensure_python_interpreter
+from testing import run_pex_command
 
 
 @pytest.mark.skipif(sys.version_info < (3, 8), reason="Flask 2.3.3 requires Python3.8+.")
 def test_standard_library_is_included(
     tmpdir,  # type: Any
-    pex_project_dir,  # type: str
 ):
     # type: (...) -> None
 
-    # N.B.: The package script requires Python 3.
-    python = sys.executable if PY3 else ensure_python_interpreter(PY38)
-
-    package_script = os.path.join(pex_project_dir, "scripts", "package.py")
-    pex_pex = os.path.join(str(tmpdir), "pex")
-    subprocess.check_call(args=[python, package_script, "--pex-output-file", pex_pex])
-
-    subprocess.run(
-        [
-            sys.executable,
-            pex_pex,
+    built_pex_path = os.path.join(tmpdir, "flask.pex")
+    run_pex_command(
+        args=[
             "Flask==2.3.3",
             "-o",
-            "flask.pex",
+            built_pex_path,
             "--venv",
-        ],
-        check=True,
-        env=os.environ,
-    )
+        ]
+    ).assert_success()
 
-    output_path = "{}/output.txt".format(tmpdir)
-
+    output_path = os.path.join(tmpdir, "output.txt")
     script = """
 import sys
 error = ""
 
 try:
-    sys.path.append("flask.pex")
+    sys.path.append("{}")
     import __pex__
     import flask
 except Exception as e:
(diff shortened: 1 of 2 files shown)


tests/integration/test_issue_2235.py
```diff
@@ -0,0 +1,67 @@
...
+    tmpdir,  # type: Any
+    pex_project_dir,  # type: str
+):
+    # type: (...) -> None
+
+    # N.B.: The package script requires Python 3.
+    python = sys.executable if PY3 else ensure_python_interpreter(PY38)
+
+    package_script = os.path.join(pex_project_dir, "scripts", "package.py")
+    pex_pex = os.path.join(str(tmpdir), "pex")
```
jsirois, line 22: Instead of building the Pex PEX and running that; just use testing.run_pex_command as is done here: https://github.com/pantsbuild/pex/blob/5c0c29c192bd99335b5a1ab5b68a0f8556cd97aa/tests/integration/test_issue_1936.py#L17-L18

tests/integration/test_issue_2235.py
```diff
@@ -0,0 +1,67 @@
...
+    subprocess.check_call(args=[python, package_script, "--pex-output-file", pex_pex])
+
+    subprocess.run(
+        [
+            sys.executable,
+            pex_pex,
+            "Flask==2.3.3",
+            "-o",
+            "flask.pex",
```
jsirois, line 31: You just leaked flask.pex to CWD (the root of the repo). Use a path formed from `tmpdir` instead.

tests/integration/test_issue_2235.py
```diff
@@ -0,0 +1,67 @@
...
+    f.write(error)
+""".format(
+        output_path
+    )
+
+    script_path = "{}/script.py".format(tmpdir)
+    with open(script_path, "w") as f:
+        f.write(script)
+
+    subprocess.run([sys.executable, script_path], check=True, env=os.environ)
```
jsirois, line 61: This runs the script directly (~`python /tmp/pytest-of.../script.py`), you want to run the script via the PEX (~`python the.pex /tmp/pytest-of.../script.py`).

tests/integration/test_issue_2235.py
```diff
@@ -0,0 +1,67 @@
...
+except Exception as e:
+    error = str(e)
+
+with open("{}", 'w') as f:
+    f.write(error)
+""".format(
+        output_path
+    )
+
+    script_path = "{}/script.py".format(tmpdir)
```
jsirois, line 57: Please use `os.path.join` - these are papercuts that will make finishing the Windows support branch harder.

tests/integration/test_issue_2235.py
```diff
@@ -0,0 +1,67 @@
...
+
+@pytest.mark.skipif(sys.version_info < (3, 8), reason="Flask 2.3.3 requires Python3.8+.")
+def test_standard_library_is_included(
+    tmpdir,  # type: Any
+    pex_project_dir,  # type: str
+):
+    # type: (...) -> None
+
+    # N.B.: The package script requires Python 3.
+    python = sys.executable if PY3 else ensure_python_interpreter(PY38)
```
jsirois, line 19: You already guaranteed Python>=3.8 via the skipif above; so this conditional is not needed.

jsirois (commented):
It looks like you're not done @yjabri, but some observations so far.

tests/integration/test_issue_2235.py
yjabri: I want to run the script directly. Is it correct to assume that `import __pex__` from a `pex` built with `--venv` in a `python`, `ipython`, or `jupyer` should be supported as long as interpreter constraints are satisfied?

tests/integration/test_issue_2235.py
yjabri: Thanks, this is much simpler.

yjabri:
Pushed a new commit that

* Incorporated all but one suggestion - left a comment explaining my rationale.
* Checked the test still fails with `Failed: Test failed with exception: No module named 'contextvars'`
* Nothing leaks to CWD

tests/integration/test_issue_2235.py
jsirois: Aha - right. I was being dumb. No, you're right here.

So ... `env=os.environ` is a noop - that's the default when you pass `env=None`, in other words, just inherit the ambient environment. Did you mean to set PYTHONPATH in the env? If so `env=testing.make_env(PYTHONPATH=...)` will do the right thing.

tests/integration/test_issue_2235.py
```diff
@@ -51,14 +38,14 @@ def test_standard_library_is_included(
...
+        built_pex_path, output_path
     )
 
-    script_path = "{}/script.py".format(tmpdir)
+    script_path = os.path.join(tmpdir, "script.py")
     with open(script_path, "w") as f:
         f.write(script)
 
-    subprocess.run([sys.executable, script_path], check=True, env=os.environ)
+    subprocess.run([sys.executable, script_path], check=True)
```
jsirois, line 48: Ok, gotcha. You fixed the extraneous env since you directly manipulate sys.path in the script.

jsirois:
Ok, @yjabri this test now LGTM. I won't let CI run though until you add the fix. Thanks for taking a go at this.

tests/integration/test_issue_2235.py
jsirois: As noted in your 2nd commit - I see you realized this env usage was extraneous already.

yjabri:
Thanks for the quick feedback! To make sure I submit a solid fix, I want to ensure I give myself enough time to get acquainted with  the codebase. If I don't get to it today, I'm going to aim for Sunday.

tests/integration/test_issue_2235.py
```diff
@@ -0,0 +1,54 @@
...
+    run_pex_command(
+        args=[
+            "Flask==2.3.3",
+            "-o",
+            built_pex_path,
+            "--venv",
+        ]
+    ).assert_success()
+
+    output_path = os.path.join(tmpdir, "output.txt")
```
jsirois, line 26: Does this work?! Perhaps it does in Python 3. As a general rule I do `os.path.join(str(tmpdir), "output.txt")` for tests, but perhaps that coercion is only needed for Python 2.7, which you don't test against here; so are safe?

yjabri:
I think the test name could be better. Maybe something along the lines of `test_non_executable_venv`? If that sounds good I'll update the name of the PR to `Fix non executable venv bug`

jsirois:
I'm happy with the file name and test name as they are. Only change if you'd like to.

jsirois:
@yjabri while fixing `tox -eformat-run,typecheck` it would also be good to update the PR title and description.

yjabri:
Looks like `subprocess.run` was only introduced in python 3.5.

I think adding `# type: ignore` to `subprocess.run([sys.executable, script_path], check=True)` is okay because of the python version constraint.

jsirois:
How about just use `subprocess.check_call`.

jsirois:
@yjabri the CI issue is of my making. I'm actively working on CI improvements and it looks like you got a disk full due to those. You'll need to merge main to pick up fixes when I let you know to do so in a few hours.

yjabri:
Sounds good.

jsirois:
Alright @yjabri, if you merge main now, CI should be good to go.

jsirois (approved):
Thanks @yjabri. I'll get a release spun up assuming this goes green.

yjabri:
Thanks for all of the assistance @jsirois! I appreciated the quick feedback and how smooth it was to contribute.

PR #2237, opened 2023-09-16
Improve CI caching.

jsirois:
Only write caches on main and include caching of docker images. This
should keep the cache size under 10GB at ~8GB and take the Linux Docker
overhead lower.


benjyw (approved):

jsirois:
The PR description was way off, I forgot about cache zstd compression. Things came in at ~3.8GB; so plenty of headroom to the 10GB limit.

PR #2239, opened 2023-09-16
Remove docker-cache step.

jsirois:
The docker-cache step actually increased wall time by more than a
minute.


jsirois:
Not only was the caching slower, but it was actively causing disk full issues on PRs (see 1st annotation here: https://github.com/pantsbuild/pex/actions/runs/6208806535?pr=2236).

benjyw (approved):
Ah well, it was worth a try.

PR #2243, opened 2023-09-18
Add support for `--use-pip-config`.

### LINKED ISSUE #336: pex does not use .pip/pip.conf to resolve packages

### LINKED ISSUE #838: Be able to use user's pip.conf in pip resolution

jsirois:
When `--use-pip-config` is specified, both `PIP_*` env vars and Pip
configuration files will be read by the Pip Pex delegates to.

Fixes #336
Fixes #838

### DIFF
--- a/pex/resolve/resolver_options.py
+++ b/pex/resolve/resolver_options.py
@@ -3,6 +3,7 @@
 
 from __future__ import absolute_import
 
+import os
 from argparse import Action, ArgumentTypeError, Namespace, _ActionsContainer
 
 from pex import pex_warnings
@@ -108,6 +109,7 @@ def register(
             "fast indicating the problematic selected interpreters."
         ),
     )
+    register_use_pip_config(parser)
 
     register_repos_options(parser)
     register_network_options(parser)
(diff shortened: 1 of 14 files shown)


huonw (approved):

pex/resolve/resolver_options.py
```diff
@@ -251,6 +253,40 @@ def register(
...
+    # type: (_ActionsContainer) -> None
+    """Register an option to control Pip config hermeticity.
+
+    :param parser: The parser to register the Pip config hermeticity option with.
+    """
+    parser.add_argument(
+        "--use-pip-config",
+        "--no-use-pip-config",
+        dest="use_pip_config",
+        default=None,
```
benjyw, line 266: Why not `False` ?

benjyw (approved):

Issue #2249, opened 2023-09-19
Unable to run a script in interactive mode `-i`


gauthamnair:
## Issue description and reproduction steps

A pex with no pre-defined entry point or console script will run a script or module (`-m`) that it is passed as a CLI arg similarly to how the [python command line does](https://docs.python.org/3/using/cmdline.html#interface-options)

```
$ cat << EOF > foo.py
> bar = 42
> print(f'hello {bar}')
> EOF
$ python foo.py
hello 42
```

```
~/code/pantsbuild/pex (main) $ python -m pex -o testpex.pex
~/code/pantsbuild/pex (main) $ ./testpex.pex foo.py
hello 42
```

the python command line allows you to run the target in interactive mode by [passing the `-i` flag](https://docs.python.org/3/using/cmdline.html#cmdoption-i). It executes the target but then hands you interactive control.
```
$ python -i foo.py
hello 42
>>> bar
42
>>> quit()
```
Trying something similar with a `.pex` does not work. The target executes successfully but you get an error stack trace after it. Then you end up in a REPL but it doesn't have the expected names loaded.
```
~/code/pantsbuild/pex (main) $ ./testpex.pex -i foo.py
hello 42
Traceback (most recent call last):
  File "/Users/gauthamnair/.pex/unzipped_pexes/53e3acded6c4b7940375dd07ace6edae8800941b/__main__.py", line 108, in <module>
    bootstrap_pex(__entry_point__, execute=__execute__, venv_dir=__venv_dir__)
  File "/Users/gauthamnair/.pex/unzipped_pexes/53e3acded6c4b7940375dd07ace6edae8800941b/.bootstrap/pex/pex_bootstrapper.py", line 627, in bootstrap_pex
    pex.PEX(entry_point).execute()
  File "/Users/gauthamnair/.pex/unzipped_pexes/53e3acded6c4b7940375dd07ace6edae8800941b/.bootstrap/pex/pex.py", line 560, in execute
    sys.exit(self._wrap_coverage(self._wrap_profiling, self._execute))
SystemExit
>>> bar
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
NameError: name 'bar' is not defined
```

## Motivation

Similar to https://github.com/pantsbuild/pex/issues/2248 this was discovered while trying out support in pants for supplying args to a `pants repl` command. Running `-i some_script.py` is one of the more common modifiers to boot an interactive session.

## Diagnosis

[...]

PR #2250, opened 2023-09-19
Fix hang when running repl with python options

### LINKED ISSUE #2248: Interactive interpreter hangs if pex called with a python cli option
## Issue description and reproduction steps

A pex that has no pre-defined entry point or console script is expected to open up an interactive interpreter if it isn't passed a script, module, or command as an argument:
```
~/code/pantsbuild/pex (main) $ python -m pex -o testpex.pex
~/code/pantsbuild/pex (main) $ ./testpex.pex
Python 3.9.17 (main, Aug  8 2023, 12:13:14) 
[Clang 14.0.3 (clang-1403.0.22.14.1)] on darwin
(InteractiveConsole)
>>> print('hello')
hello
>>> quit()
~/code/pantsbuild/pex (main) $ 
```

If we pass a [python command line option](https://docs.python.org/3/using/cmdline.html) like `-q`, which normally quiets the repl preamble, the pex command hangs indefinitely:
```
~/code/pantsbuild/pex (main) $ python -q
>>> quit()
~/code/pantsbuild/pex (main) $ ./testpex.pex -q
# hangs ... and requires Ctrl-C
```

## Motivation

This came up while testing a fix to address 
 https://github.com/pantsbuild/pants/issues/3423 so we can pass through args to `pants repl` for the python shell. For a python shell, pants runs a `requirements.pex` as an interactive process to give a repl to the user.

Usually if you want to pass args to a python repl it is going to be either interpreter options like these or it will be `-i` to run a script/module in interactive mode.

## Diagnosis

Under the hood it hangs because pex gets into an infinite loop of re-executing itself with the same argument. We deal with python options inside `Pex.execute_interpreter`:

```python
args = sys.argv[1:]
python_options = []
for index, arg in enumerate(args):
    # Check if the arg is an expected startup arg.
    if arg.startswith("-") and arg not in ("-", "-c", "-m"):
        python_options.append(arg)
    else:
        args = args[index:]
        break

# The pex was called with Python interpreter options
if python_options:
    return self.execute_with_options(python_options, args)
```
The first block of code is attempting to partition the input args into python options like `-q` or `-B` and a remainder of command line args that should be passed through when we `execute_with_options`. But when the args supplied are _only python options_, we do not set `args = []` because the `args = args[index:]` does not get triggered. Therefore we end up execing something like this:
```python
# cmdline
['<path_to_python>/python3.9', '-q', '<path_to_unzipped_pex>/__main__.py', '-q']
```
later in `os.execv(python, cmdline)` within `execute_with_options` which puts us in an infinite loop because the pex will encounter the `-q` again as a python option.

This is simple to fix, a PR will be forthcoming.

gauthamnair:
Fixes #2248 .

Hang is averted, but not all options work as they would be intended. For example `-q` is not quiet. But that can be separated to another issue.

### DIFF
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -620,6 +620,9 @@ class PEX(object):  # noqa: T000
             else:
                 args = args[index:]
                 break
+        else:
+            # All the args were python options
+            args = []
 
         # The pex was called with Python interpreter options
         if python_options:
--- a/pex/venv/installer.py
+++ b/pex/venv/installer.py
@@ -849,6 +849,9 @@ def _populate_first_party(
                     else:
                         args = args[index:]
                         break
+                else:
+                    # All the args were python options
+                    args = []
 
                 # The pex was called with Python interpreter options, so we need to re-exec to
                 # respect those:
--- a/tests/integration/venv_ITs/test_issue_2248.py
+++ b/tests/integration/venv_ITs/test_issue_2248.py
@@ -1,0 +1,81 @@
+# Copyright 2022 Pants project contributors (see CONTRIBUTORS.md).
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+import os
+import subprocess
+from textwrap import dedent
+
+import pytest
+from colors import colors
+
+from pex.typing import TYPE_CHECKING
+from testing import IntegResults, run_pex_command
+
+if TYPE_CHECKING:
+    from typing import Any, List
+
+
+# N.B.: To test that we are running a REPL we'll show that it computes result even after
+# assertion errors and exceptions that would normally halt a script.
+# To check that it forwards python options we use -O to deactivate asserts
+# See: https://docs.python.org/3/using/cmdline.html#cmdoption-O
+@pytest.mark.parametrize(
+    "execution_mode_args",
+    [
+        pytest.param([], id="ZIPAPP"),
+    ],  # pytest.param(["--venv"], id="VENV")]
+)
+def test_repl_python_options(
+    execution_mode_args,  # type: List[str]
+    tmpdir,  # type: Any
+):
+    # type: (...) -> None
+
+    pex = os.path.join(str(tmpdir), "pex")
+    run_pex_command(args=["ansicolors==1.1.8", "-o", pex] + execution_mode_args).assert_success()
+
+    repl_commands = dedent(
+        """
+        import colors
+        assert False
+        raise Exception("customexc")
+        result = 20 + 103
+        print(colors.green(f"Worked: {result}"))
+        quit()
+        """
+    )
+
+    def execute_pex(disable_assertions):
+        # type: (bool) -> IntegResults
+        args = [pex]
+        if disable_assertions:
+            args.append("-O")
+        process = subprocess.Popen(
+            args=args,
+            stdin=subprocess.PIPE,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        )
+        stdout, stderr = process.communicate(input=repl_commands.encode("utf-8"))
+        return IntegResults(
+            output=stdout.decode("utf-8"),
+            error=stderr.decode("utf-8"),
+            return_code=process.returncode,
+        )
+
+    # The assertion will fail and print but since it is a REPL it will keep going
+    # and compute the result
+    result = execute_pex(disable_assertions=False)
+    result.assert_success()
+    assert "InteractiveConsole" in result.error
+    assert "AssertionError" in result.error
+    assert "customexc" in result.error
+    assert colors.green("Worked: 123") in result.output
+
+    # The -O will disable the assertion, but the regular exception will still get raised.
+    result = execute_pex(disable_assertions=True)
+    result.assert_success()
+    assert "InteractiveConsole" in result.error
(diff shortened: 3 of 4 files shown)


jsirois:
@gauthamnair thanks for the fix. Proof / preventing backslide would be good though. Can you add to the tests introduced in #1746 and #1748? In particular, #1748 will point out a whole parallel launcher for `--venv` mode PEXes this fix misses.

gauthamnair:
@jsirois been working on it quite a bit despite the apparent silence. I'm having a surprising amount of trouble getting the options-related tests to work for me (even off of main). Hopefully will get through it soon.

gauthamnair:
Thanks for the tip @jsirois . I was trying to run pytest directly instead of going through tox (initial attempts at tox were not working for me). So I sat down and finally learned tox properly last night and I went through the tox.ini in detail for having been silly and of course at the end of it I find that `$ tox -e py39 -- t
ests/test_pex.py -k dashc` does actually work completely fine for me.

I've added a test for the existing fix. Also includes a test for the original no-options repl emulation because I couldn't quite find one. The assertions are a bit awkward and I worry that asserting on the presence of `>>> ` and/or on `InteractiveConsole` output could be brittle. The latter especially is just an impl detail of what `code.interact()` chooses to say when it starts up. Short of adding a test dependency on [pexpect](https://pexpect.readthedocs.io/en/stable/) I can't figure out a more satisfying demo.

Moving on to fixing the case of venv pexes.

jsirois:
@gauthamnair thanks. I agree console tests are a bit awkward, but I think what you have does the job.
Just let me know when you're ready - until then I'll save some trees by not clicking "Approve and run" got the CI workflow.

gauthamnair:
@jsirois I added support for the `--venv` type and an integration test that is a copy-and-pare-down of the one introduced in https://github.com/pantsbuild/pex/pull/1748 . I ran `format-run` successfully but I have not been able to get _all_ the tests to pass yet on local (some even seem to hang). Can you trigger CI?

jsirois:
> I have not been able to get all the tests to pass yet on local (some even seem to hang).

Ok. I find that disturbing and would like to get the tests to work everywhere. They work on my machine and in Mac and Linux CI; but I'd like them to work for everyone. It may even be that them not all working on your machine uncovers a real bug in the production code and not just a badly written test.

gauthamnair:
Ok, I think I see what the ci error is. syntax error on py2 because the integration test uses the py3 print-function syntax instead of py2. Will fix it.

jsirois:
Ok, great. You should be able to repro locally with `./dtox -epy27-integration -- -vvsk test_repl_python_options -n2`. And if you haven't found the `--devpi` option to the tox / dtox.sh test envs / https://github.com/pantsbuild/pex/blob/main/testing/bin/run_tests.py, it's useful in porportion to the narrowness / flakiness of you internet connection.

gauthamnair:
Yes, was able to reproduce local with regular `tox` and pushed a fix. It was the f-string actually that was the issue.

Now `tox -e py27-integration,py39-integration -- -v -k test_repl_python_options` is working fine for me.

But hold off on running CI. I should make sure that all the tests can pass on my local and if not help to figure out why. I'll try `tox -epy39` and `tox -epy39-integration -- --devpi` and check back in when they are good.

gauthamnair:
Ok, accounted for all regular tests `tox -epy39`. Two files were failing because my repo clone workspace was not clean. The third test is failing because of inability to find openssl, usual stuff. 

But the fourth test was hanging for real due to #2250 

I'll turn my attention to the integration tests later, after the 'sprint' day. @jsirois would you mind re-triggering CI?

gauthamnair:
@jsirois all checks seem to be passing (no doubt because you have been rerunning them). There is also the test-hanging PR you have approved https://github.com/pantsbuild/pex/pull/2251 . Lmk what usually happens after this stage.

Meanwhile, would you suggest I turn my attention to #2249 ? I am not sure to what extent we need the repl emulation to work super faithfully. It will require a bit more moving around code. Lmk if it is worth the review cycles or we can set that one aside/close.

jsirois:
I'll get this PR reviewed and merged (assuming I don't spot a bug). The other PR is merged.

As to what to work on, that's up to you. If it fixes a bug / use case you personally actually care about (i.e.: you'd personally use / benefit from the fix), then I'm happy to review. I only really get grumpy for half thought out fixes that don't bother to understand what's going on in detail. Those should be able to be avoided by reading and then asking clarifying questions about what's going on in some part of the codebase.

pex/pex.py
```diff
@@ -620,6 +620,9 @@ def execute_interpreter(self):
             else:
                 args = args[index:]
                 break
+        else:
+            # All the args were python options
```
jsirois, line 624: The comment helps, and this is just me talking to the void, but I have to do this to re-prove to myself how for / else works everytime - such odd syntax:
```python
>>> def loop(brk):
...     for x in range(5):
...         if brk and x == 3:
...             break
...         print(x)
...     else:
...         print("else")
...
>>> loop(False)
0
1
2
3
4
else
>>> loop(True)
0
1
2
>>>
```

jsirois (approved):
Thanks @gauthamnair - this looks good.

PR #2252, opened 2023-09-29
Upgrade Python 3.12 to rc3.

jsirois:
The release notes are here:
  https://docs.python.org/3.12/whatsnew/changelog.html#python-3-12-0-release-candidate-3


dtox.sh
```diff
@@ -28,6 +28,7 @@ elif [[ "${BASE_MODE}" == "pull" ]]; then
 fi
 
 USER_INPUT=(
+  "${BASE_INPUT[@]}"
```
jsirois, line 31: Using `./dtox.sh inspect` then `python3.12 -V` in the inspect shell was not getting me 3.12.0rc3! Old bug.

benjyw (approved):

PR #2253, opened 2023-09-30
Add `--check` support for zipapps.

jsirois:
By default Pex now warns for un-useable PEX zipapps it creates, but it
can be made to error or do nothing at all.

Fixes #2247

PR #2255, opened 2023-09-30
Avoid spurious disk full job errors on Ubuntu.

jsirois:


jsirois:
These were spurious - depended on what exact host configuration a job landed on it seems - but now with the new test in #2253 are reliable.

jsirois:
I'm going to submit TBR to get main green.

benjyw (commented):
LGTM

PR #2256, opened 2023-09-30
Allow PyPI fallback on Mac CI runners.

jsirois:
Also guard against premature devpi server death by just warning on
failure to shut down the devpi server.


jsirois:
Hrm, yeah - not so easy, Adding an extra index adds extra artifacts / artifact URLs to locks and that breaks several test's expectations.

Issue #2097, opened 2023-03-15
Option to exclude dependencies from pex file


JoostvDoorn:
Pex currently relies on Pip to resolve dependencies and build the PEX file. However, there are cases where users may want to exclude certain packages, either because they are already available on the target environment or because they are too large to include in the PEX file.

The proposed solution is to add an exclude flag to Pex, which would allow users to specify a list of packages to exclude from the PEX file. During the assembly phase, Pex would exclude these packages and their transitive dependencies, resulting in a smaller PEX file.

I believe this would be a useful feature for many users. We would want to use it with Pants to exclude dependencies such as pyspark for an environment that includes pyspark already.

This issue follows up on the comment here: https://github.com/pantsbuild/pex/issues/2082#issuecomment-1456942556

Other related issues have been filed previously in pants and pex as well. https://github.com/pantsbuild/pex/issues/1146 https://github.com/pantsbuild/pants/issues/11324

kaos:
You'd have to find a way to ensure the version already present in the target env is compatible with the version janked from the pex or you might find yourself with hard to debug issues.

jsirois:
Yeah, as @kaos points out this is definitely buyer beware. It at least require the `--ignore-errors` PEX build time option or the `PEX_IGNORE_ERRORS=1` runtime env var option in combination with `--inherit-path` / `PEX_INHERIT_PATH` or `PEX_EXTRA_SYS_PATH` or `PEX_PATH`. An advanced implementation could expand the current PEX boot resolve process:
https://github.com/pantsbuild/pex/blob/ecf9b95544c53970628e982b7132a345bbc931cd/pex/environment.py#L506-L509

Such that `iter_distributions` looks at more than the PEX internal distribution cache:
https://github.com/pantsbuild/pex/blob/ecf9b95544c53970628e982b7132a345bbc931cd/pex/environment.py#L266-L275

This would allow not using `--ignore-errors` / `PEX_IGNORE_ERRORS=1` and still validating the boot resolve, but now via a mixed `sys.path`. This would only work though if the mixed-in `sys.path` elements had the distributions installed with their dist-info metadata. For some systems, system packages are not installed this way. A quick check of Ubuntu 22.04 shows the system Python packages use `egg-info` metadata.

jsirois:
Ok, I've spent some time prototyping this - the advanced implementation described above. This works great but its also very sensitive. It turns out two installs of the same wheel can lead to a different hash due to ~inconsequential differences. For example:
```
[...]

PR #2262, opened 2023-10-09
Fix site-packages identification.

jsirois:
Fix site-packages identification.

In the process of working a new feature, it was discovered that
`sys.path` manipulations leaked into `PythonInterpreter` when they were
performed before the `__pex__` import hook. This is fixed by always
going through the interpreter disk cache path, which has properly
isolated `sys.path` for quite some time, eliminating the one exception
for the current interpreter. Fixing this case caused several tests to
fail that exposed several eager uses of `PythonInterpreter.get()` that
would execute in import of various modules. Those cases are fixed to
either be lazy or use `sys.executable`.

Along the way, fix up the ~TODO for migrating from `distutils.sysconfig`
to `sysconfig` for determining the site-packages directories. This
should preempt issues with vendor (e.g.: Debian) supplied Python 3.12+.

### DIFF
--- a/pex/interpreter.py
+++ b/pex/interpreter.py
@@ -114,18 +114,32 @@ class PythonIdentity(object):
             # include a getsitepackages function.
             TRACER.log("The site module does not define getsitepackages: {err}".format(err=e))
 
+        # The distutils package was deprecated in 3.10 and removed in 3.12. The sysconfig module was
+        # introduced in 3.2 but is not usable for our purposes until 3.11. We need
+        # `get_default_scheme` to get the current interpreter's installation scheme, which was made
+        # public in 3.10, but not made correct for venv interpreters until 3.11.
+        try:
+            import sysconfig
+
+            get_default_scheme = getattr(sysconfig, "get_default_scheme", None)
+            if get_default_scheme and sys.version_info[:2] >= (3, 11):
+                scheme = get_default_scheme()
+                yield sysconfig.get_path("purelib", scheme)
+                yield sysconfig.get_path("platlib", scheme)
+                return
+        except ImportError:
+            pass
+
+        # The distutils.sysconfig module is deprecated in Python 3.10 but still around. It goes away
+        # in 3.12 with viable replacements in sysconfig starting in Python 3.11. See above where we
+        # use those replacements preferentially, when available.
         try:
             from distutils.sysconfig import get_python_lib
 
             yield get_python_lib(plat_specific=False)
             yield get_python_lib(plat_specific=True)
-        except ImportError as e:
-            # The distutils.sysconfig module is deprecated in Python 3.10 but still around.
-            # Eventually it will go away with replacements in sysconfig that we'll add at that time
-            # as another site packages source.
-            TRACER.log(
-                "The distutils.sysconfig module does not define get_python_lib: {err}".format(err=e)
-            )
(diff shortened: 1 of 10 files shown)


jsirois:
This all came up in work towards #2097 and #1361.

pex/interpreter.py
```diff
@@ -979,9 +998,6 @@ def _spawn_from_binary(cls, binary):
         cached_interpreter = cls._PYTHON_INTERPRETER_BY_NORMALIZED_PATH.get(canonicalized_binary)
         if cached_interpreter is not None:
             return SpawnedJob.completed(cached_interpreter)
-        if canonicalized_binary == cls.canonicalize_path(sys.executable):
-            current_interpreter = cls(PythonIdentity.get())
```
jsirois, line 983: This was the bug. Here `PythonIdentity.get()` slurps up the ambient `sys.path` which may have been mucked with.

jsirois:
Ok, and the CI failure was real. The vendoring process is very titchy, very bootstrappy. The `PythonInterpreter.get` semantics going from in-mem to through the file cache system needs the vendored code to pre-exist, which triggered this new bootstrap tangle.

pex/pex.py
```diff
@@ -51,41 +51,48 @@ class IsolatedSysPath(object):
     @staticmethod
     def _expand_paths(*paths):
         # type: (*str) -> Iterator[str]
+        seen = set()
         for path in paths:
+            if path in seen:
+                continue
+            seen.add(path)
```
huonw, line 58: Is this set just an optimisation, or is trying to be a guarantee that all output paths are unique?

If it's meant to be a guarantee, as written, it seems possible that this could be violated, e.g.:

- the `abspath` codepath: `paths == ["a", "/path/to/cwd/a"]` while running with `/path/to/cwd` and thus get a sequence like `["a", "/path/to/cwd/a", "/path/to/cwd/a"]` (where the first two correspond to the first input element, and the last corresponds to the second input element)
- the `realpath` codepath: `paths == ["/a", "/link-to-a"]` where `os.path.realpath("/link-to-a") == "/a"` 
- the combination of the two: AFAIK, `os.path.realpath` will absolutise a relative path, and thus `not isabs(path)` implies `os.path.realpath (realpath) != path` and both extra `yield`s will happen

Maybe this private function is not ever actually called with problematic args, though?

huonw (approved):
I'm not sure I understand the moving parts in detail, but the code certainly seems to match the PR description, AFAIK. 👍

pex/resolve/target_options.py
```diff
@@ -61,16 +61,12 @@ def register(
         ),
     )
 
-    current_interpreter = PythonInterpreter.get()
-    program = sys.argv[0]
     singe_interpreter_info_cmd = (
```
benjyw, line 64: Fix the typo to `single_` while you're here?

benjyw (approved):

pex/pex.py
jsirois: The questions the code raised are enough here to justify simplifying at the expense of slightly more work. Fixed.

pex/resolve/target_options.py
jsirois: Fixed.

benjyw (approved):

PR #2263, opened 2023-10-09
Migrate many types from str to Text.

jsirois:
This supports an up coming refactor of `pex.{dist_metadata,pep_376}`
which also requires re-formulating the pyc filters from `Iterable` ->
`Iterable` form to boolean predicates to make typing easier at the
loss of ~nothing.

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -23,7 +23,7 @@ from pex.commands.command import (
     global_environment,
     register_global_arguments,
 )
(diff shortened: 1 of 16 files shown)


jsirois:
Like #2262, more work in support of #2097 and #1361.

jsirois:
I believe the CI errors are fixed by #2262. I'll wait on that and then merge it here to see if that's true.

huonw (approved):

benjyw (approved):

PR #2266, opened 2023-10-12
Fix ReadTheDocs generation.

benjyw:
Sphinx is incompatible with Jinja2==3.1.0. 

RTD used to pin Jinja2 below that version, but they stopped
doing so when they changed the set of packages they install
by default:
https://blog.readthedocs.com/defaulting-latest-build-tools/

So now we must pin explicitly ourselves.


jsirois (approved):
Thanks Benjy.

Issue #2267, opened 2023-10-13
Once we release 2.1.150, Restore RTD "default version" to it if needed.


benjyw:
A change in RTD's own settings caused the 2.1.148 doc build to break. 

The breakage has been fixed in https://github.com/pantsbuild/pex/pull/2266, so the next release's docs will build. But until then, I've manually configured the "default version" on RTD's Admin -> Advanced Settings, which pointed to "v2.1.148" to point to "latest", so that https://pex.readthedocs.io/ isn't broken. ("latest" points to the latest SHA, not the latest release, AFAICT).

Once 2.1.149 is released we may need to manually repoint it (or this may happen automatically, I haven't dug into how RTD normally updates this).

jsirois:
It looks like this automation rule handles activating any new incoming tag as the default version; so the ship should right itself automatically: https://readthedocs.org/dashboard/pex/rules/regex/865/

jsirois:
It looks like this automation rule handles activating any new incoming tag as the default version; so the ship should right itself automatically: https://readthedocs.org/dashboard/pex/rules/regex/865/

jsirois:
Alrighty. They are changing all the things - new failure: https://readthedocs.org/projects/pex/builds/22435029/:

![image](https://github.com/pantsbuild/pex/assets/157586/ab6edb1a-f7d8-48a0-8a85-41f9cffe8f06)

The fix for that is outlined here: https://blog.readthedocs.com/migrate-configuration-v2/

benjyw:
> It looks like this automation rule handles activating any new incoming tag as the default version; so the ship should right itself automatically: https://readthedocs.org/dashboard/pex/rules/regex/865/

I'd hoped that was the case, but I wasn't sure if my manually resetting it would somehow prevent it automatically righting itself (I didn't dig into how those get set, but I concur that this automation rule should do the trick).

jsirois:
Yeah - it did do the trick. The now-failing-in-a-new-way RTD build has undone your manual pointing fix. You should find https://pex.readthedocs.io -> https://pex.readthedocs.io/en/v2.1.149/ which is broken now.

jsirois:
Ok, #2275 has fixed the RTD build and I've pointed the default version at latest for now. Hopefully the 2.1.150 Pex release triggers a self-righting of the default version.

jsirois:
[...]

PR #2275, opened 2023-11-03
Fix `.readthedocs.{yml,yaml}` duplication.

jsirois:
The fix in #2274 that introduced `.readthedocs.yaml` (the name
documented by RTD currently to use for the config file) was blind to the
fact we already had a `.readthedocs.yml` file. It appears to be the case
that `.readthedocs.yml` is both read by RTD and preferred; so RTD build
errors persisted. Fix this by merging the config files and sticking with
the current reccomended name of `.readthedocs.yaml`.

More work towards #2267.


benjyw (approved):

cognifloyd (approved):

jsirois:
Alright - that righted the RTD build. I've pointed the default version at latest; so the 2.1.150 release should self-right, but I'll leave #2267 open until then to confirm.

PR #2280, opened 2023-11-05
Link my sponsorship info for large projects.

jsirois:


jsirois:
I added you two as reviewers as Pantsbuild leads. I set up a personal sponsorship for larger projects a few months ago and just linked it to the scie jump and lift repos. The other intended audience is pex. This gets a bit weird though since Pantsbuild has its own sponsorship; thus you should have the say here. I think the underlying weirdness is that Pex is housed under Pantsbuild as an org, but I'm effectively the only person maintaining it / grokking what's needed to implement larger projects currently and for a while now.

benjyw (approved):
I think this is great, and there is sufficient distinction between pex and pants in the minds of potential sponsors. If and when we gather substantial sponsorship for Pants we can look into diverting some of those funds to Pex, but I see no harm in having this as a separate thing.

stuhood (approved):

PR #2281, opened 2023-11-06
Implement support for `--exclude <req>`.

jsirois:
When excluding a requirement from a PEX, any resolved distribution
matching that requirement, as well as any of its transitive dependencies
not also needed by non-excluded requirements, are elided from the PEX.

At runtime these missing dependencies will not trigger boot resolve
errors, but they will cause errors if the modules they would have
provided are attempted to be imported. If the intention is to load the
modules from the runtime environment, then `--pex-inherit-path` /
`PEX_INHERIT_PATH` or `PEX_EXTRA_SYS_PATH` knobs must be used to allow
the PEX to see distributions installed in the runtime environment.
Clearly, you must know what you're doing to use this option and not
encounter runtime errors due to import errors. Be ware!

A forthcoming `--provided` option, with similar effects on the PEX
contents, will both automatically inherit any needed missing
distributions from the runtime environment and require all missing
distributions are found; failing fast if they are not.

Work towards #2097.


jsirois:
N.B.: The diffstat is large, but a little more than half is a lock file I check in for an integration test; so this is just under 500 real delta lines.

pex/bin/pex.py
```diff
@@ -290,6 +291,22 @@ def configure_clp_pex_options(parser):
...
     )
 
+    group.add_argument(
+        "--exclude",
+        dest="excluded",
+        default=[],
+        type=str,
+        action="append",
+        help=(
+            "Adds a requirement to exclude from the built PEX. Any distribution included in the "
```
kaos, line 301: The wording I find could be a little misleading, as we're only adding to the list of things to exclude, and it is the excluding bit that is interesting. (I think the note about this option being able to be used multiple times at the end of the help text to be sufficient regarding the additiveness of the option.)

How about something like `"Names a requirement to exclude ..."` instead?

kaos (approved):
lgtm. I've only done a cursory pass.

pex/bin/pex.py
jsirois: Ok. I guess the other "adds" you find misleading is thinking this adds the requirement to the PEX?

I think `s/Names/Specifies/` sits better with me since a requirement in a name + a version sepcifier and although most people proably will just use the name part, they really can specify a full requirement, including marker - for example to exclude only for mac in a multiplatform PEX for example. Does that work for you?

kaos: Specifies works just as well, yes 👍🏽 

Yes, it's a bit of a double negative with "add to exclude". I did understand it, but can also see it being an easy mistake to make. (I may have paused and read the sentence a second time to make sure I parsed it correctly..)

jsirois: Fixed.

pex/pex_info.py
```diff
@@ -527,12 +540,14 @@ def update(self, other):
                 other.interpreter_constraints
             )
         self._requirements.update(other.requirements)
+        self._excluded.update(other.excluded)
```
huonw, line 543: Is the behaviour when `self` includes something excluded by `other` (or vice versa) something that's been explicitly considered?

It looks like this code will consider those requirements as "excluded". Is that intentional or accidental? If the latter, is it desirable?

pex/environment.py
```diff
@@ -348,12 +349,23 @@ def _resolve_requirement(
...
+        exclude_configuration=ExcludeConfiguration(),  # type: ExcludeConfiguration
     ):
         # type: (...) -> Iterator[_DistributionNotFound]
         requirement_key = _RequirementKey.create(requirement)
         if requirement_key in resolved_dists_by_key:
             return
 
+        excluded_by = exclude_configuration.excluded_by(requirement)
+        if excluded_by:
+            TRACER.log(
```
huonw, line 361: Just confirming that the `V=1` verbosity level for these logs was intentional, vs. something slightly quieter like `V=2` or similar?

pex/dependency_manager.py
```diff
@@ -0,0 +1,108 @@
...
+                    excluded_by = exclude_configuration.excluded_by(candidate_dist.distribution)
+                    if excluded_by:
+                        excludes = " and ".join(map(str, excluded_by))
+                        TRACER.log(
+                            "Skipping adding {candidate}: excluded by {excludes}".format(
+                                candidate=candidate_dist.distribution, excludes=excludes
+                            )
+                        )
+                        for root_req in self._requirements:
+                            if candidate_dist.distribution in root_req:
```
huonw, line 84: With a significant number of excludes, these nested loops are potentially O(n**2) (or worse), is that a concern?

tests/test_dependency_manager.py
```diff
@@ -0,0 +1,155 @@
...
+    )
+
+    pex_info = PexInfo.default()
+    pex_builder = PEXBuilder(pex_info=pex_info)
+    dependency_manager.configure(pex_builder, excluded=["c"])
+    pex_builder.freeze()
+
+    assert ["a", "b"] == list(pex_info.requirements)
+    assert ["c"] == list(pex_info.excluded)
+    expected_dists = [dist_graph.dist(name) for name in ("A", "B", "D", "F")]
```
huonw, line 152: Double checking my understanding of the behaviour, if `"c"` _wasn't_ excluded, would the expected result be all of the letters, including `C` and `E`?

tests/integration/test_excludes.py
```diff
@@ -0,0 +1,77 @@
...
+
+    assert expected_import_error_msg in stderr.decode("utf-8"), stderr.decode("utf-8")
+
+    venv_dir = os.path.join(str(tmpdir), "venv")
+    venv = Virtualenv.create(venv_dir)
+    pip = venv.install_pip()
+
+    # N.B.: The constraining lock requirement is the one expressed by requests: certifi>=2017.4.17
+    # The actual locked version is 2023.7.22; so we stress this crease and use a different, but
+    # allowed, version.
```
huonw, line 64: Just confirming behaviour again: for this `--exclude` option, the version isn't actually checked, AIUI, right? For instance, this could install `certifi==1.0.0` and the PEX would start fine and an `import certifi` statement would execute (although with no guarantee of any sort of correct behaviour on the part of `requests` with the out-of-range dependency).

That is, potentially the crease to stress would be installing a non-allowed version, to compare/contrast to `--provided`'s future behaviour?

pex/environment.py
```diff
@@ -502,14 +515,21 @@ def resolve(self):
...
+                )
             )
         return self._resolved_dists
 
-    def resolve_dists(self, reqs):
-        # type: (Iterable[Requirement]) -> Iterable[FingerprintedDistribution]
+    def resolve_dists(
+        self,
+        reqs,  # type: Iterable[Requirement]
+        exclude_configuration=ExcludeConfiguration(),  # type: ExcludeConfiguration
```
huonw, line 530: `pex/resolve/pex_repository_resolver.py` also calls this function; should it propagate that pex's exclusions?

pex/environment.py
```diff
@@ -348,12 +349,23 @@ def _resolve_requirement(
         resolved_dists_by_key,  # type: MutableMapping[_RequirementKey, FingerprintedDistribution]
         required,  # type: bool
         required_by=None,  # type: Optional[Distribution]
+        exclude_configuration=ExcludeConfiguration(),  # type: ExcludeConfiguration
```
huonw, line 352: Minor thing, but I notice that both callers of this private function currently provide this argument. Could it potentially be a required argument, so that future refactoring doesn't accidentally leave it off when it should be provided?

pex/dependency_manager.py
```diff
@@ -0,0 +1,108 @@
...
+else:
+    from pex.third_party import attr
+
+
+@attr.s
+class DependencyManager(object):
+    _requirements = attr.ib(factory=OrderedSet)  # type: OrderedSet[Requirement]
+    _distributions = attr.ib(factory=OrderedSet)  # type: OrderedSet[FingerprintedDistribution]
+
+    def add_from_pex(self, pex):
```
huonw, line 34: I notice that this function loses any exclusions in `pex`. Is that correct?

It seems like this might lead to attempting to find some dependency that isn't available, because its required by one of this PEX's requirements, but has been excluded?

pex/bin/pex.py
```diff
@@ -808,11 +825,15 @@ def build_pex(
     pex_info.strip_pex_env = options.strip_pex_env
     pex_info.interpreter_constraints = interpreter_constraints
 
-    for requirements_pex in options.requirements_pexes:
-        pex_builder.add_from_requirements_pex(requirements_pex)
```
huonw, line 812: I'm guessing you're aware that this removed call seems to be the only caller of this `add_from_requirements_pex` method?

huonw (approved):
Nice! I like it.

Lots of questions because, as usual, I'm not sure I fully understand all the moving parts (but I'm getting closer!).

pex/dependency_manager.py
jsirois: I was concerned about the outer double loop and so constructed `dists_by_project_name` to limit the outer inner; so that's the best it can be afaict using standard data structures.

Here there is an additional single loop (that scales with root input requirements as opposed to excludes) that could only be sped up by either killing the warning, or keeping it and similarly using a pre-built multimap keyed on project_name. The biggest input requirement sets I've seen are on the order of 100. At full exclusion, that's 10k checks. I'd be surprised if that amounted to more than ~500ms total, but I've gone ahead and added the multimap.

tests/test_dependency_manager.py
jsirois: Correct. That's implicitly checked by the 100s of tests that already build and exercise PEXes elsewhere.

tests/integration/test_excludes.py
jsirois: You've got that all correct.

pex/pex_info.py
jsirois: It has, but not in this context. There is actually a test for the case when the same PexInfo includes contradictory root requirements and excludes. That case warns at build time and is silent at run time. This fits my general philosophy of trying not to piss off an advanced user (the warning is even questionable in my mind). As the semantics stand, a user with a batch system that creates these command lines can have excludes trump requirements and they just get a warning. If they want this to trigger a fail-fast error, they can add that behavior in their code that generates the command line. If I went with the nanny bias, however, they would be no such recovery for the aforementioned batch user. They would be forced to curse and submit a patch adding some new option to relax the check.

pex/environment.py
jsirois: No. That use case flow is part of the 1-2 (it's one of the 3 ways step 2 is implemented):
1. add repository pex dists to dep manager
2. add resolved dists to dep manager
3. now that the full set of dists is known, can the proper excludes be calculated and no earlier.

pex/environment.py
jsirois: Sure. I did not like the re-ordering this forced, which mixes topic importance to my reading, but done.

pex/dependency_manager.py
jsirois: Yes, I think you've found a bug here.

pex/bin/pex.py
jsirois: I had missed that in the shuffles. Thanks - killed.

pex/dependency_manager.py
jsirois: Ok, added a new integration test that fails without the fix. Thanks for catching that.

pex/pex_info.py
jsirois: After more thought, I'm even less convinced the warning is a good idea. The obvious use case is using Pex to package but some other means to lock. In other words, something like `pex -r requirements.lock --exclude build --exclude my --exclude subset ...`. Some other tool provides the lock in Pip requirements style (hopefully with hashes), but you can still use Pex to lock a subset - here by passing `locked-reqs - desired-reqs` as the `--exclude`s. The reachability behavior documented for `--exclude` is what builds the proper subset.

Since the warning can be reverted without breaking people though, I'll proceed as-is and see if the the use case I realized even comes up / comes up enough to get complaints about the warnings / having to `--no-emit-warnings`.

PR #2286, opened 2023-11-09
Exclude pyc dirs, not include, when hashing code

### LINKED ISSUE #2285: `venv create` no longer includes `--sources-directory` contents when all files are nested
It seems like there was a regression from 2.1.148 -> 2.1.149 with the behaviour of `venv create` with a `--pex-repository` that was created with `--sources-directory`: those sources aren't included in the final venv.

Reproducer:

```shell
cd $(mktemp -d)

# create our dummy file
mkdir -p source_files/foo
touch source_files/foo/bar.py # NB.1
# touch source_files/qux.py # NB.2

for version in v2.1.148 v2.1.149; do
    curl -s -L https://github.com/pantsbuild/pex/releases/download/$version/pex > pex-$version
    chmod +x pex-$version

    # NB.3
    ./pex-$version --output-file=repository-$version.pex --sources-directory=source_files

    # NB.4
    PEX_SCRIPT=pex3 ./pex-$version venv create --dest-dir=dest-$version --pex-repository=repository-$version.pex --layout=flat

    # what was included?
    tree dest-$version
done
```

Running that shows that the contents of the `dest-...` directory depends on the version, without the `bar.py` file when using v2.1.149, but should be the same:

```
dest-v2.1.148
└── foo
    └── bar.py

1 directory, 1 file
dest-v2.1.149

0 directories, 0 files
```

Ablative studies:

- uncommenting `NB.2` line (to have two files) passes ✅ (both versions have both `foo/bar.py` and `qux.py`)
- _replacing_ the `NB.1` with `NB.2` (to just `qux.py` at the top level) passes ✅ 
- _always_ using v2.1.148 on line `NB.3` (create the pex) and v2.1.149 on line `NB.4` (create the venv) passes ✅ 
- v2.1.149 for `NB.3` and v2.1.148 for `NB.4` fails ❌ 
- I think third-party dependencies work okay, but haven't confirmed in this reduced setting
- This reproduces without `--layout`, but the output is simpler with `--layout=flat`

(First observed in https://github.com/pantsbuild/pants/pull/20149.)

huonw:
This fixes #2285 by fixing a minor typo in f9a9d949d95efd0153452dc255625656d628ee17 / #2263 that caused the computation of a PEXes code hash to not recur into the correct directories.

I think this explains the symptoms in #2285 due to caching: if a PEX only contains files in subdirectories, the code hash will be `da39a3ee5e6b4b0d3255bfef95601890afd80709` (the sha256 hash of no data) and thus potentially the `venv create` might be referring to an incorrectly cached installation.

### DIFF
--- a/pex/util.py
+++ b/pex/util.py
@@ -87,7 +87,7 @@ class CacheHelper(object):
         hashing.dir_hash(
             directory=directory,
             digest=digest,
-            dir_filter=is_pyc_dir,
+            dir_filter=lambda d: not is_pyc_dir(d),
             file_filter=lambda file_path: not is_pyc_file(file_path)
             and not file_path.startswith("."),
         )
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -91,6 +91,19 @@ def test_dir_hash():
         ), "All content under __pycache__ directories should be ignored."
 
 
+def test_dir_hash_recur_issue_2285():
+    # type: () -> None
+    with temporary_dir() as tmp_dir:
+        # explicitly confirm that we recur into directories
+        hash_empty = CacheHelper.dir_hash(tmp_dir)
+        with safe_open(os.path.join(tmp_dir, "a", "b.py"), "w") as fp:
+            fp.write("contents1")
+
+        hash_full = CacheHelper.dir_hash(tmp_dir)
+
+        assert hash_full != hash_empty
+
+
 try:
     import __builtin__ as python_builtins  # type: ignore[import]
 except ImportError:
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -6,6 +6,8 @@ import subprocess
 from hashlib import sha1
 from textwrap import dedent
 
+import pytest
+
 from pex.common import safe_mkdir, safe_open, temporary_dir, touch
 from pex.pex import PEX
 from pex.pex_builder import PEXBuilder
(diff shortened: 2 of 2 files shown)


huonw:
I haven't got my dev environment to work with the Python 3.12 changes and don't have time to today, so this is as yet untested (i.e. I don't know if the test fails before/passes after). The bisection script from #2285 does work, though.

jsirois (approved):
My time today is limited as well, but this is certainly a fix of something.

huonw:
I've got a dev environment working again 🎉 

That test was wrong (testing `dir_hash`, bug was in `pex_code_hash`). I've rewritten the existing tests for `dir_hash` to also test `pex_code_hash`, which catches the bug, plus added an assertion for the hidden file handling.

jsirois (approved):

jsirois:
> ... haven't yet got my dev environment to work for running test with the Python 3.12 changes ...
@huonw fwiw, if you have docker, then you need no dev env and can do things just like CI:
https://github.com/pantsbuild/pex/blob/db5b5100d0d739964bfc46943bc9c8f9e1bed480/.github/workflows/ci.yml#L99-L101

Basically:
1. (one time only) `BASE_MODE=pull CACHE_MODE=pull ./dtox.sh --version`
2. `./dtox.sh ...regular tox args...`

You don't need to do step 1, in which case these images will be built locally, but that is slow. If you have a decent network connection, pulling them is probably more convenient. What this nets you is the full suite of Python interpreters Pex supports - both CPython and PyPy - ready to go in the image. It also gets you a pre-seeded PyPI cache when you use `./dtox.sh -epy... -- --devpi ...`. You can omit the `CACHE_MODE=pull` if you don't use the `--devpi` feature of the build or if you'd rather have it build that cache up locally.

PR #2287, opened 2023-11-10
Prepare the 2.1.152 release

huonw:


huonw:
I have noticed there's some issue juggling that is typically connected to a release too, but I'm a bit unclear on it, so I have not attempted it. Please let me know if/how I should be doing that.

jsirois (approved):

jsirois:
> I have noticed there's some issue juggling that is typically connected to a release too, but I'm a bit unclear on it, so I have not attempted it. Please let me know if/how I should be doing that.

As with everything else Pex, I just continued precedent as much as possible from prior maintainers. I can't remember if that started with Wickman or Kris, but it is, at this point juggling as you say and of little use since I have changed release flow to early and often and there is no real buildup of changes before a release any longer. I think there is no need here and I should just potentially stop that practice and stick to what is documented in RELEASE.rst.

Thanks for identifying the issue in the release and fixing. I'll merge this and do the release since my time is limited again.

PR #2298, opened 2023-12-13
Add support for `--no-pre-install-wheels` and `--max-install-jobs`.

### LINKED ISSUE #2292: Pex zip-creation takes a very long time for `torch>=2`
Hey!

Not sure if actionable, but maybe there's something here that can be done. I was investigating another issue today and ended up seeing a very slow Pants package step ~5 minutes. The issue reproduces with the simple command line `pex -vvv torch>=2 -o t2.2.pex`. This takes ~280 seconds on my machine, of which ~210-220 is spent purely in the zip step:

```
<snip>
pex: Building pex: 70298.8ms
pex:   Adding distributions from pexes: : 0.1ms
pex:   Resolving distributions for requirements: torch: 70294.7ms
pex:     Resolving requirements.: 70294.6ms
pex:       Resolving for:
  /usr/bin/python3.10: 55574.5ms
pex:       Calculating project names for direct requirements:
  PyPIRequirement(line=LogicalLine(raw_text='torch', processed_text='torch', source='<string>', start_line=1, end_line=1), requirement=Requirement(name='torch', url=None, extras=frozenset(), specifier=<SpecifierSet('')>, marker=None), editable=False): 0.1ms
pex:       Installing 22 distributions: 9352.3ms
pex:       Checking install: 2.7ms
pex:   Configuring PEX dependencies: 3.4ms
pex: Zipping PEX file.: 213135.5ms
```
This turns out to a 2.5 GB pex, which admittedly is on the fat side. Unzipping this beast takes ~30 seconds, and zipping it with regular `zip` takes ~230 seconds. `zip -1` takes ~100 seconds and adds ~10% to the size. `zip -0` takes 12 seconds but doubles the size. Seeing as compression seems to add the majority of the runtime, I did a very quick hack (outside of pex) where I move the compress step to a process pool (since it's CPU-heavy). With that, I get ~30 seconds at level 1, or about ~60 seconds on level 6. So 3-4x speed increase. It may be able to push this a bit higher by playing with ordering.

I also played around with the store-only-by-suffix capabilities, but it seems like the .so's make up the bulk of both the compression potential and time: only compressing text-like files gives a ~4.3 GB zip in 20 seconds.

With all that said, I'm mostly curious if this is something that has been discussed elsewhere (found nothing while searching), and what kind of solution might be palatable relative to the gains that can be made. I'm willing to contribute something based on the work I've done so far, or investigate other suggested approaches.

jsirois:
The `--no-pre-install-wheels` option causes built PEXes to use raw
`.whl` files. For `--layout zipapp` this means a single `.whl` file is
`STORED` per dep, and for `--layout {packed,loose}` this means the loose
`.deps/` dir contains raw `.whl` files. This speeds up all PEX builds by
avoiding pre-installing wheel deps (~unzipping into the `PEX_ROOT`) and
then, in the case of zipapp and packed layout, re-zipping. For large
dependencies the time savings can be dramatic.

Not pre-installing wheels comes with a PEX boot cold-start performance
tradeoff since installation now needs to be done at runtime. This is
generally a penalty of O(100ms), but that penalty can be erased for some
deployment scenarios with the new `--max-install-jobs` build option / 
`PEX_MAX_INSTALL_JOBS` runtime env var. By default, runtime installs are
performed serially, but this new option can be set to use multiple
parallel install processes, which can speed up cold boots for large
dependencies.

Fixes #2292

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -848,7 +848,7 @@ def build_pex(
         )
     ):
         try:
-            dependency_manager.add_from_installed(
+            dependency_manager.add_from_resolved(
                 resolve(
                     targets=targets,
                     requirement_configuration=requirement_configuration,
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -30,6 +30,7 @@ from pex.inherit_path import InheritPath
 from pex.interpreter_constraints import InterpreterConstraints
 from pex.layout import Layout, ensure_installed
 from pex.orderedset import OrderedSet
+from pex.pep_427 import InstallableType
 from pex.pex import PEX
 from pex.pex_bootstrapper import ensure_venv
 from pex.pex_builder import Check, CopyMode, PEXBuilder
@@ -892,6 +893,11 @@ def build_pex(
                     resolver_configuration=resolver_configuration,
                     compile_pyc=options.compile,
                     ignore_errors=options.ignore_errors,
+                    result_type=(
+                        InstallableType.INSTALLED_WHEEL_CHROOT
+                        if options.pre_install_wheels
+                        else InstallableType.WHEEL_FILE
+                    ),
                 )
             )
         except Unsatisfiable as e:
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -176,7 +176,7 @@ def configure_clp_pex_options(parser):
             "always yield slightly faster PEX cold boot times; so they are used by default, but "
             "they also slow down PEX build time. As the size of dependencies grows you may find a "
             "tipping point where it makes sense to not pre-install wheels; either because the "
-            "increased cold boot time is is irrelevant to your use case or marginal compared to "
+            "increased cold boot time is irrelevant to your use case or marginal compared to "
             "other costs. Note that you may be able to use --max-install-jobs to decrease cold "
             "boot times for some PEX deployment scenarios."
         ),
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -165,6 +165,41 @@ def configure_clp_pex_options(parser):
         ),
     )
     group.add_argument(
+        "--pre-install-wheels",
+        "--no-pre-install-wheels",
+        dest="pre_install_wheels",
+        default=True,
+        action=HandleBoolAction,
+        help=(
+            "Whether to pre-install third party dependency wheels. Pre-installed wheels will "
+            "always yield slightly faster PEX cold boot times; so they are used by default, but "
+            "they also slow down PEX build time. As the size of dependencies grows you may find a "
+            "tipping point where it makes sense to not pre-install wheels; either because the "
+            "increased cold boot time is is irrelevant to your use case or marginal compared to "
+            "other costs. Note that you may be able to use --max-install-jobs to decrease cold "
+            "boot times for some PEX deployment scenarios."
+        ),
+    )
+    group.add_argument(
+        "--max-install-jobs",
+        dest="max_install_jobs",
+        default=1,
+        type=int,
+        help=(
+            "The maximum number of parallel jobs to use when installing third party dependencies "
+            "contained in a PEX during its first boot. By default, this is set to 1 which "
+            "indicates dependencies should be installed in serial. A value of 2 or more indicates "
+            "dependencies should be installed in parallel using exactly this maximum number of "
+            "jobs. A value of 0 indicates the maximum number of parallel jobs should be "
+            "auto-selected taking the number of cores into account. Finally, a value of -1 "
+            "indicates the maximum number of parallel jobs should be auto-selected taking both the "
+            "characteristics of the third party dependencies contained in the PEX and the number "
+            "of cores into account. The third party dependency heuristics are intended to yield "
+            "good install performance, but are opaque and may change across PEX releases if better "
+            "heuristics are discovered. Any other value is illegal."
+        ),
+    )
+    group.add_argument(
         "--check",
         dest="check",
         default=Check.WARN,
@@ -824,6 +859,8 @@ def build_pex(
     pex_info.pex_root = options.runtime_pex_root
     pex_info.strip_pex_env = options.strip_pex_env
     pex_info.interpreter_constraints = interpreter_constraints
+    pex_info.deps_are_wheel_files = not options.pre_install_wheels
+    pex_info.max_install_jobs = options.max_install_jobs
 
     dependency_manager = DependencyManager()
     excluded = list(options.excluded)  # type: List[str]
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -28,7 +28,7 @@ from pex.dependency_manager import DependencyManager
 from pex.enum import Enum
 from pex.inherit_path import InheritPath
 from pex.interpreter_constraints import InterpreterConstraints
-from pex.layout import Layout, maybe_install
+from pex.layout import Layout, ensure_installed
 from pex.orderedset import OrderedSet
 from pex.pex import PEX
 from pex.pex_bootstrapper import ensure_venv
@@ -1100,9 +1100,7 @@ def seed_cache(
 
         with TRACER.timed("Seeding caches for {}".format(pex_path)):
             final_pex_path = os.path.join(
-                maybe_install(pex=pex_path, pex_root=pex_root, pex_hash=pex_hash)
-                or os.path.abspath(pex_path),
-                "__main__.py",
+                ensure_installed(pex=pex_path, pex_root=pex_root, pex_hash=pex_hash), "__main__.py"
             )
             if verbose:
                 return json.dumps(create_verbose_info(final_pex_path=final_pex_path))
--- a/pex/environment.py
+++ b/pex/environment.py
@@ -658,7 +658,9 @@ class PEXEnvironment(object):
                     "Failed to resolve requirements from PEX environment @ {pex}.\n"
                     "Needed {platform} compatible dependencies for:\n"
                     "{items}".format(
-                        pex=self._pex, platform=self._target.platform.tag, items="\n".join(items)
+                        pex=self.source_pex,
+                        platform=self._target.platform.tag,
+                        items="\n".join(items),
                     )
                 )
 
--- a/pex/environment.py
+++ b/pex/environment.py
@@ -16,7 +16,7 @@ from pex.exclude_configuration import ExcludeConfiguration
 from pex.fingerprinted_distribution import FingerprintedDistribution
 from pex.inherit_path import InheritPath
 from pex.interpreter import PythonInterpreter
-from pex.layout import ensure_installed
+from pex.layout import ensure_installed, identify_layout
 from pex.orderedset import OrderedSet
 from pex.pep_425 import CompatibilityTags, TagRank
 from pex.pep_503 import ProjectName
@@ -42,6 +42,8 @@ if TYPE_CHECKING:
     )
 
     import attr  # vendor:skip
+
+    from pex.pep_427 import InstallableType
 else:
     from pex.third_party import attr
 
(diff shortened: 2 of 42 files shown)


jsirois:
I'll attach some example perf numbers for the torch 2GB case here as soon as CI goes green. Suffice to say, this is both a build time win and run time win for that case.

jsirois:
Reviewers: Thanks in advance - this is large. I did break it up into independent commits though, none of which is larger than 550 delta-lines, with most being ~100 delta lines. So instead of ~12 PRs you get 1 with 12 commits. I needed the 1 personally to work out all the titchy details comprehensively and ensure this all worked / vet it was complicated and actually required the 2 new knobs instead of just being the new default.

jsirois:
Alright, #2292 now has perf analysis for the OP torch case. I'm still working on the 3 mac-specific IT failures, but this is ready for review.

pex/environment.py
```diff
@@ -269,16 +273,45 @@ def path(self):
...
+        return self._source_pex or self._pex
+
+    def iter_distributions(self, result_type_wheel_file=False):
+        # type: (bool) -> Iterator[FingerprintedDistribution]
+        if result_type_wheel_file:
+            if not self._pex_info.deps_are_wheel_files:
+                raise ResolveError(
+                    "Cannot resolve .whl files from PEX at {pex}; its dependencies are in the "
+                    "form of pre-installed wheel chroots.".format(pex=self.source_pex)
                 )
```
jsirois, line 288: In general I only like to add orthogonal features that can be combined freely. There are 3 Pex resolver sources: Pip, lock files and PEX repositories, and 2 PEX dependency formats: installed wheel chroots and (now) raw `.whl` files. In this 6 element matrix, only this case of resolving raw `.whl` files from a PEX repository containing dependencies in installed wheel chroot form cannot be satisfied currently (this is tested explicitly in the changes to `test_pex_repository_resolve.py`).

The `pep_427.py` work that now powers `.whl` file installs instead of `pip install` does make it possible to plug this gap in a later release though with some modification of `pep_376.py`. This will also allow the existing `PEX_TOOLS=1 ./a.pex repository extract ...` tool to produce fully correct wheels for the installed wheel chroot dependencies in PEXes of that style.

pex/bin/pex.py
```diff
@@ -164,6 +165,41 @@ def configure_clp_pex_options(parser):
...
+            "tipping point where it makes sense to not pre-install wheels; either because the "
+            "increased cold boot time is is irrelevant to your use case or marginal compared to "
+            "other costs. Note that you may be able to use --max-install-jobs to decrease cold "
+            "boot times for some PEX deployment scenarios."
+        ),
+    )
+    group.add_argument(
+        "--max-install-jobs",
+        dest="max_install_jobs",
+        default=1,
```
huonw, line 187: I'm assuming this default is to preserve behaviour; if a breaking change was possible (e.g. in 3.0.0), would you consider defaulting to `-1`?

pex/environment.py
```diff
@@ -528,10 +561,19 @@ def resolve_dists(
...
         reqs,  # type: Iterable[Requirement]
         exclude_configuration=ExcludeConfiguration(),  # type: ExcludeConfiguration
+        result_type=None,  # type: Optional[InstallableType.Value]
     ):
         # type: (...) -> Iterable[FingerprintedDistribution]
 
-        self._update_candidate_distributions(self.iter_distributions())
+        result_type_wheel_file = False
+        if result_type is not None:
+            from pex.pep_427 import InstallableType
```
huonw, line 570: I'm curious, why is this import conditional?

pex/bin/pex.py
```diff
@@ -164,6 +164,41 @@ def configure_clp_pex_options(parser):
...
+        "--no-pre-install-wheels",
+        dest="pre_install_wheels",
+        default=True,
+        action=HandleBoolAction,
+        help=(
+            "Whether to pre-install third party dependency wheels. Pre-installed wheels will "
+            "always yield slightly faster PEX cold boot times; so they are used by default, but "
+            "they also slow down PEX build time. As the size of dependencies grows you may find a "
+            "tipping point where it makes sense to not pre-install wheels; either because the "
+            "increased cold boot time is is irrelevant to your use case or marginal compared to "
```
kaos, line 179:
```suggestion
            "increased cold boot time is irrelevant to your use case or marginal compared to "
```

pex/layout.py
```diff
@@ -126,18 +149,243 @@ def record(self, dest_dir):
...
+                                with TRACER.timed({extracting_msg!r}):
+                                    layout.extract_dist(
+                                        dest_dir=spread_chroot.work_dir,
+                                        dist_relpath={dist_relpath!r},
+                                        is_wheel_file={is_wheel_file!r}
+                                    )
+
+                        symlink_dest = {symlink_dest!r}
+                        safe_mkdir(os.path.dirname(symlink_dest))
+                        os.symlink({symlink_src!r}, symlink_dest)
```
huonw, line 257: Is it worth exposing the core of this logic as a function that can be used by both both `_ensure_distributions_installed_serial` directly and by this script?

pex/layout.py
```diff
@@ -126,18 +149,243 @@ def record(self, dest_dir):
...
+            max_jobs="<auto>" if max_jobs == 0 else max_jobs, count=len(pex_info.distributions)
+        )
+    ):
+        for _ in execute_parallel(
+            inputs=inputs, spawn_func=install_distribution, max_jobs=max_jobs
+        ):
+            pass
+
+
+AVERAGE_DISTRIBUTION_SIZE_PARALLEL_JOB_THRESHOLD = 1 * 1024 * 1024  # ~1MB
```
huonw, line 309: Does this number come from experiments?

pex/layout.py
```diff
@@ -224,14 +460,42 @@ def dist_strip_prefix(self, dist_name):
...
+        is_wheel_file,  # type: bool
+    ):
+        # type: (...) -> int
+        if is_wheel_file:
+            return self._zfp.getinfo(dist_relpath).file_size
+        else:
+            return sum(
+                self._zfp.getinfo(name).file_size
+                for name in self._names
+                if name.startswith(dist_relpath)
```
huonw, line 475: I know this pattern is used elsewhere, but just asking to confirm that one can't have two unrelated dists where the `dist_relpath` of one will be a string prefix of the other? (e.g. `a/b` vs. `a/bc`)

For instance, does `dist_relpath` always end in `/`?

pex/resolver.py
```diff
@@ -610,6 +611,90 @@ def build_wheels(
         return build_results
 
 
+@attr.s(frozen=True)
+class DirectRequirements(object):
+    @classmethod
+    def calculate(
```
huonw, line 617: Is the contents of this method just moved, with no changes?

pex/resolver.py
```diff
@@ -610,6 +611,90 @@ def build_wheels(
...
+                direct_requirements_by_project_name[direct_requirement.project_name].add(
+                    direct_requirement
+                )
+            return cls(direct_requirements_by_project_name)
+
+    _direct_requirements_by_project_name = (
+        attr.ib()
+    )  # Mapping[ProjectName, OrderedSet[Requirement]]
+
+    def adjust(self, distributions):
```
huonw, line 680: What's this adjusting? (I acknowledge it's just moved code.)

tests/integration/test_integration.py
```diff
@@ -795,19 +801,67 @@ def test_pex_resource_bundling():
             assert stdout == b"hello\n"
 
 
-def test_entry_point_verification_3rdparty(tmpdir):
-    # type: (Any) -> None
+@pytest.mark.parametrize(
+    "layout", [pytest.param(layout, id=layout.value) for layout in Layout.values()]
```
huonw, line 805: Totally minor, but are you aware that `pytest.mark.parametrize` supports an `ids` parameter that can take a function, so these could potentially be something like:

```python
@pytest.mark.parametrize("layout", Layout.values(), ids=lambda layout: layout.value)
...
```

kaos (commented):
Not gone through all commits yet. Only been reading the code, trying to grok the change with limited knowledge of the surroundings.

tests/integration/test_lock_resolver.py
```diff
@@ -523,3 +524,40 @@ def assert_dists(
...
+    # N.B.: We choose ansicolors 1.1.8 since it works with all Pythons and has a universal wheel
+    # published on PyPI and cowsay 5.0 since it also works with all Pythons and only has an sdist
+    # published on PyPI. This combination ensures the resolve process can handle both building
+    # wheels (cowsay stresses this) and using pre-existing ones (ansicolors stresses this).
+    run_pex3("lock", "create", "ansicolors==1.1.8", "cowsay==5.0", "-o", lock).assert_success()
+
+    pex = os.path.join(str(tmpdir), "pex")
+    exe = os.path.join(str(tmpdir), "exe")
+    with open(exe, "w") as fp:
+        fp.write("import colors; print(colors.blue('Moo?'))")
```
huonw, line 542: I assume the `import colours` is meaning to validate that `ansicolors` is actually installed in a useful way. Should there be an  `import cowsay` too?

huonw (approved):
The numbers from #2292 are awesome, nice!

pex/bin/pex.py
jsirois: Probably not. Parallel installs slow down most cases and although -1 tries to guess the divide in the cases, I'm sure it does so poorly. My tests were on a 16 core SSD machine and I'm not confident on the heuristic for my machine let alone other disk / CPU combos.

pex/environment.py
jsirois: Because Python is hella slow and this is in the boot hot path. Normal imports in this file add ~50ms to boot overhead.

pex/layout.py
jsirois: Yes. As noted elsewhere, on 1 machine and with limited cases. I'm very much not confident in the heuristic.

pex/layout.py
jsirois: I'm not sure how you get much better. This code is pretty small and the sticking point is 7 parameters.

pex/layout.py
jsirois: A dist relative path is `.deps/*.whl`. Using the wheel name here is critical to PEX functioning (runtime dist resolution perf in multi-platform PEXes). Since the suffix is fixed and conspires to be illegal in any of the 5 possible wheel name component slots, this is safe.

pex/resolver.py
jsirois: Yup.

pex/resolver.py
jsirois: The name of the extracted class is indicative here, it adjusts a distribution to note the direct top level requirement it satisfies, if any. Most dists usually don't get adjusted since they satisfy only transitive requirements. The need for associating dists to the direct requirements they satisfy goes back to handling local project directory dependencies. These have no project name known except after the resolve process which builds them into a wheel with a name and version.

tests/integration/test_integration.py
jsirois: I was not aware. I'm not sure I like that though! It's not obvious the two align when you split things up like that. 10 values, 6 ids. Clearly they align in your example, but it takes a second to make sure of that.

pex/bin/pex.py
jsirois: Fixed.

pex/layout.py
jsirois: Commented.

tests/integration/test_lock_resolver.py
jsirois: Fixed.

jsirois:
Alright - thanks @huonw and @kaos. I've added the perf analysis for small PEXes to the ticket which backs up what is expected: Builds of `.whl` file-based PEXes is always at least a little bit faster, Cold 1st runs is always a little bit slower and install parallelization is generally a loss. Although I'd love to not have knobs for these two new features, it seems necessary - there are unavoidable tradeoffs that are hard to justify removing from the user's purview.

jsirois:
@kaos, I'm going to submit - I feel pretty confident I've covered the bases here - but I'll follow up with any feedback you have. There is refactoring and simplification to do as at least partly tracked by #2299 that further feedback will fall naturally into.

tests/integration/test_integration.py
jsirois: Oh, I read poorly. It's 1 function mapped over all the values. Gotcha.

huonw: Yeah, in addition to the function, it also supports passing a list with exactly the right number of IDs, but I don't like that possibility either.

PR #2300, opened 2023-12-13
Prepare the 2.1.154 release.

jsirois:


jsirois:
The CI break is unrelated and was already fixed in #2298. I just picked that (645d49f) here to get green CI, but it will go in 1st with #2298.

kaos (approved):

PR #2302, opened 2023-12-14
Since we no longer use wheel code, remove it.

jsirois:
Lots of changes over a long time have led to this point. The last small
one is in this change, installing wheel in the vendored Pip venv just in
time when wheel building is required by using vendored Pip itself and
the user configured indexes, etc.


pex/pex_builder.py
```diff
@@ -641,20 +641,11 @@ def _prepare_bootstrap(self):
         # although we don't use pyparsing directly, packaging.markers, which we
         # do use at runtime, does.
         root_module_names = ["attr", "packaging", "pkg_resources", "pyparsing"]
-        include_dist_info = set()
-        if self._pex_info.includes_tools:
-            # The `repository extract` tool needs setuptools and wheel to build sdists and wheels
```
jsirois, line 646: This is covered by pre-existing tests and was made untrue by the commits that made way for Python 3.12 support this summer. The savings in no longer vendoring these both for the Pex PEX (which `--includes-tools`) can be seen in this comparison of end products:
```
$ du -sh released/* dist/*
4.1M    released/pex
2.9M    released/pex-2.1.154-py2.py3-none-any.whl
3.9M    released/pex-2.1.154.tar.gz
3.6M    dist/pex
2.8M    dist/pex-2.1.154-py2.py3-none-any.whl
3.9M    dist/pex-2.1.154.tar.gz
```

kaos (approved):

PR #2306, opened 2023-12-16
Switch vendoring from `pip install ...` to Pex install.

jsirois:
Pex now only uses Pip for resolving distributions (`pip download ...`)
and building wheels (`pip wheel ...`), installation of wheels in chroots
and venvs is now completely handled by Pex everywhere in the code base.


pex/pep_427.py
```diff
@@ -394,13 +394,16 @@ def record_files(
         print("pex", file=fp)
     installed_files.append(InstalledWheel.create_installed_file(path=fp.name, dest_dir=dest))
 
-    if requested:
-        requested_path = os.path.join(dest, wheel.metadata_path("REQUESTED"))
-        touch(requested_path)
-        installed_files.append(
-            InstalledWheel.create_installed_file(path=requested_path, dest_dir=dest)
-        )
+    if interpreter:
```
huonw, line 397: Why the change from unconditional to conditional for these files?

huonw (approved):
Nice clean up; just one question.

pex/pep_427.py
jsirois: REQUESTED is not required at all; see: https://peps.python.org/pep-0376/#one-dist-info-directory-per-installed-distribution.

RECORD is required to support uninstalling as noted; so we still include it when an interpreter is supplied (installing in a venv) for interoperability with external tools like `pip uninstall`, but not for installed wheel chroots since those are private and never uninstalled by anyone but Pex, and, since they are chroots, this can be done with a simple recursive chroot dir remove.

So, basically, this just saves a small amount of work for the most common Pex wheel install case.

PR #2318, opened 2024-01-07
Add hidden support for Python 3.13.

jsirois:
This uses a patched version of Pip as well as a new build system to
affect spinning up CI for the Python 3.13 release in October ahead of
time without leaking details of this testing-only setup to the
production distribution.


pex/bin/pex.py
```diff
@@ -958,15 +958,21 @@ def _compatible_with_current_platform(interpreter, platforms):
 def main(args=None):
     args = args[:] if args else sys.argv[1:]
     args = [transform_legacy_arg(arg) for arg in args]
-    parser = configure_clp()
+
+    parser = catch(configure_clp)
```
jsirois, line 962: Testing out the hiddenness of pep24 from the production CLI `--pip-version` support revealed these two spots where raise exceptions led to noise backtraces instead of clean bad value messages.

pex/build_system/pep_517.py
```diff
@@ -183,8 +180,8 @@ def _invoke_build_hook(
             args=args,
             env=build_system.env,
             cwd=project_directory,
-            stdout=stdout if stdout is not None else sys.stderr.fileno(),
-            stderr=stderr if stderr is not None else sys.stderr.fileno(),
+            stdout=subprocess.PIPE,
```
jsirois, line 183: Working through an issue with non-reproducible local project locks when fiddling with the proper `setup.{cfg,py}` setup (see various comments in those files) I became keenly aware of all the sdist build output leaked to `stdout` / `stderr` here. No callers passed `stdout` / `stderr` and always piping does the right thing since `Job` knows to slurp the streams up and only report output when the `Job` fails, which is exactly what you want for a non-noisy CLI that is only noisy when it needs to be to aid debug of a build failure.

pex/pip/version.py
```diff
@@ -114,14 +115,29 @@ def __get__(self, obj, objtype=None):
...
-                        if not version.hidden and version.requires_python_applies(current_version)
-                    ),
-                    key=lambda pv: pv.version,
+                applicable_versions = tuple(
+                    version
+                    for version in PipVersionValue._iter_values()
+                    if not version.hidden and version.requires_python_applies(current_version)
                 )
+                if not applicable_versions:
+                    raise ValueError(
```
jsirois, line 124: When working through `--pip-version` pip24 hiddenness, I was getting a semi-cryptic error message from `max` getting an empty iterable. Erroring was correct, but now the message is more clear in this case.

tests/bin/test_sh_boot.py
```diff
@@ -43,28 +62,16 @@ def expected(*names):
...
-            "pypy3.10",
-            "pypy3.9",
-            "pypy3.8",
-            "pypy3.7",
-            "pypy3.6",
-            "pypy3.5",
-            "pypy2.7",
-        ]
+
+    supported_versions = sorted(
```
jsirois, line 66: The switch to the dynamic list is necessitated by the new support for mutating `Requires-Python` metadata for the Pex dist under CI for Python 3.13. In that case the list needs to include python3.13 and pypy3.13 whereas in the nominal production cases it does not.

setup.py
```diff
@@ -0,0 +1,44 @@
...
+        # sub-command options we pass below work around the otherwise default `<CWD>/build/`
+        # directory for all three which defeats concurrency in tests.
+        options={
+            "egg_info": {"egg_base": unique_build_dir("egg_base")},
+            "build": {"build_base": unique_build_dir("build_base")},
+            "bdist_wheel": {"bdist_dir": unique_build_dir("bdist_dir")},
+        },
+        # This supports expanding the supported Python range via the _PEX_REQUIRES_PYTHON env var
+        # for testing unreleased Pythons.
+        python_requires=os.environ.get("_PEX_REQUIRES_PYTHON"),
```
jsirois, line 43: This is what necessitated the build system change from `flit` to `setuptools`. AFAICT setuptools provides the only pep-517 build backend where Requires-Dist metadata can be easily altered which is what I needed to be able to do to keep Python 3.13 support officially turned off while testing against Python 3.13 alpha, beta and rc releases from now through the end of September as the 3.13 release works towards final.

jsirois: And, although this was thoroughly vetted by tests - which caught issues with the new packaging relentlessly, the eyeball looks like so:
```
$ tox -epackage -- --additional-format sdist --additional-format wheel
ROOT: will run in automatically provisioned tox, host /home/jsirois/bin/tox.venv/bin/python3.11 is missing [requires (has)]: tox<4 (4.11.4), virtualenv<20.16 (20.25.0)
ROOT: install_deps> python -I -m pip install 'tox<4' 'tox>=3.25.1' 'virtualenv<20.16'
ROOT: provision> .tox/.tox/bin/python -m tox -epackage -- --additional-format sdist --additional-format wheel
package create: /home/jsirois/dev/pantsbuild/jsirois-pex/.tox/package
package installdeps: build
package installed: build==1.0.3,packaging==23.2,pyproject_hooks==1.0.0
package run-test-pre: PYTHONHASHSEED='2295051996'
package run-test: commands[0] | python scripts/package.py --additional-format sdist --additional-format wheel
Building Pex PEX to `dist/pex` ...
Built Pex PEX @ v2.1.156-1-g6a6e403a:
sha256: 8eeaef638aa9db10766f7805c231687b36a27dff84031a0721c336cdb451b872
  size: 3668094
Building additional distribution formats to `dist`: 1.) sdist, 2.) wheel ...
Built:
  dist/pex-2.1.156-py2.py3-none-any.whl
  dist/pex-2.1.156.tar.gz
____________________________________________________________________________________________________ summary _____________________________________________________________________________________________________
  package: commands succeeded
  congratulations :)
```

+ sdist:
    ```diff
    $ diff -u1 <(tar -tzf ~/downloads/pex-2.1.156.tar.gz | sort) <(tar -tzf dist/pex-2.1.156.tar.gz | grep -v -E "/$" | sort)
    --- /dev/fd/63  2024-01-08 01:41:03.688560680 -0800
    +++ /dev/fd/62  2024-01-08 01:41:03.688560680 -0800
    @@ -1,11 +1,6 @@
    -pex-2.1.156/.github/FUNDING.yml
    -pex-2.1.156/.github/workflows/ci.yml
    -pex-2.1.156/.github/workflows/release.yml
    -pex-2.1.156/.gitignore
    -pex-2.1.156/.readthedocs.yaml
     pex-2.1.156/CHANGES.md
     pex-2.1.156/LICENSE
    +pex-2.1.156/MANIFEST.in
     pex-2.1.156/PKG-INFO
     pex-2.1.156/README.rst
    -pex-2.1.156/RELEASE.rst
     pex-2.1.156/docker/base/Dockerfile
    @@ -278,2 +273,3 @@
     pex-2.1.156/pex/vendor/_vendored/pip/.prefix/bin/pip3.8
    +pex-2.1.156/pex/vendor/_vendored/pip/__init__.py
     pex-2.1.156/pex/vendor/_vendored/pip/pip-20.3.4.dist-info/INSTALLER
    @@ -801,2 +797,4 @@
     pex-2.1.156/scripts/typecheck.py
    +pex-2.1.156/setup.cfg
    +pex-2.1.156/setup.py
     pex-2.1.156/testing/__init__.py
    ```
+ wheel:
    ```diff
    $ diff -u1 <(zipinfo -1 ~/downloads/pex-2.1.156-py2.py3-none-any.whl | sort) <(zipinfo -1 dist/pex-2.1.156-py2.py3-none-any.whl | grep -v -E "/$" |
    sort)
    --- /dev/fd/63  2024-01-08 01:42:12.688549984 -0800
    +++ /dev/fd/62  2024-01-08 01:42:12.688549984 -0800
    @@ -5,2 +5,3 @@
     pex-2.1.156.dist-info/entry_points.txt
    +pex-2.1.156.dist-info/top_level.txt
     pex/__init__.py
    @@ -256,2 +257,3 @@
     pex/vendor/_vendored/pip/.prefix/bin/pip3.8
    +pex/vendor/_vendored/pip/__init__.py
     pex/vendor/_vendored/pip/pip-20.3.4.dist-info/INSTALLER
    ```

benjyw (approved):
Sounds like this was a little painful, with the flit->setuptools switch and the pypy symlinks issue... All LGTM though!

jsirois:
> Sounds like this was a little painful, with the flit->setuptools switch and the pypy symlinks issue... All LGTM though!

A bit, but flit->setuptools was a 1 time thing to allow this style of anticipating Python releases going forward and that seems worth it. I'd much rather not be caught back on my heels any longer.

Thanks for taking a look.

PR #2321, opened 2024-01-11
Add tab-completion support to PEX repls.

jsirois:
This works for all supported Python versions even when those versions
own REPL does not support tab-completion out of the box.

### DIFF
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -696,7 +696,7 @@ class PEX(object):  # noqa: T000
                     try:
                         readline.read_history_file(histfile)
                         readline.set_history_length(1000)
-                    except (IOError, OSError):
+                    except (IOError, OSError) as e:
                         sys.stderr.write(
                             "Failed to read history file at {path} due to: {err}\n".format(
                                 path=histfile, err=e
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -685,7 +685,7 @@ class PEX(object):  # noqa: T000
                     with warnings.catch_warnings():
                         warnings.simplefilter("ignore")
                         readline.read_init_file()
-                except OSError:
+                except (IOError, OSError):
                     # No init file (~/.inputrc for readline or ~/.editrc for libedit).
                     pass
 
@@ -696,7 +696,7 @@ class PEX(object):  # noqa: T000
                     try:
                         readline.read_history_file(histfile)
                         readline.set_history_length(1000)
-                    except OSError as e:
+                    except (IOError, OSError):
                         sys.stderr.write(
                             "Failed to read history file at {path} due to: {err}\n".format(
                                 path=histfile, err=e
--- a/pex/pex.py
+++ b/pex/pex.py
@@ -10,7 +10,7 @@ import sys
 from site import USER_SITE
 from types import ModuleType
 
-from pex import bootstrap
+from pex import bootstrap, pex_warnings
 from pex.bootstrap import Bootstrap
 from pex.common import die
 from pex.dist_metadata import CallableEntryPoint, Distribution, EntryPoint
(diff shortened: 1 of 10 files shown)


tests/integration/test_integration.py
```diff
@@ -234,37 +235,95 @@ def test_pex_repl_built():
         assert b">>>" in stdout
 
 
-@pytest.mark.skipif(
-    IS_PYPY or IS_MAC,
-    reason="REPL history is only supported on CPython. It works on macOS in an interactive "
```
jsirois, line 239: This appears to never have been true. In fact, using `PEX_INTERPRETER_HISTORY` / `PEX_INTERPRETER_HISTORY_FILE` on any Python interpreter without readline support, even CPython, previously led to:
```
$ PEX_INTERPRETER_HISTORY=1 PEX_INTERPRETER_HISTORY_FILE=/tmp/check $PWD/install/bin/python3.12 empty.pex
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/jsirois/.pex/unzipped_pexes/e2c4c91895f250a4f203b1d575cb3c39208ea5a7/__main__.py", line 106, in <module>
    bootstrap_pex(__entry_point__, execute=__execute__, venv_dir=__venv_dir__)
  File "/home/jsirois/.pex/unzipped_pexes/e2c4c91895f250a4f203b1d575cb3c39208ea5a7/.bootstrap/pex/pex_bootstrapper.py", line 627, in bootstrap_pex
    pex.PEX(entry_point).execute()
  File "/home/jsirois/.pex/unzipped_pexes/e2c4c91895f250a4f203b1d575cb3c39208ea5a7/.bootstrap/pex/pex.py", line 561, in execute
    sys.exit(self._wrap_coverage(self._wrap_profiling, self._execute))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jsirois/.pex/unzipped_pexes/e2c4c91895f250a4f203b1d575cb3c39208ea5a7/.bootstrap/pex/pex.py", line 468, in _wrap_coverage
    return runner(*args)
           ^^^^^^^^^^^^^
  File "/home/jsirois/.pex/unzipped_pexes/e2c4c91895f250a4f203b1d575cb3c39208ea5a7/.bootstrap/pex/pex.py", line 499, in _wrap_profiling
    return runner(*args)
           ^^^^^^^^^^^^^
  File "/home/jsirois/.pex/unzipped_pexes/e2c4c91895f250a4f203b1d575cb3c39208ea5a7/.bootstrap/pex/pex.py", line 585, in _execute
    return self.execute_interpreter()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jsirois/.pex/unzipped_pexes/e2c4c91895f250a4f203b1d575cb3c39208ea5a7/.bootstrap/pex/pex.py", line 661, in execute_interpreter
    import readline
ModuleNotFoundError: No module named 'readline'
```
(Here I hand-built CPython3.12 with `./configure --prefix $PWD/install --without-readline`)

pex/pex.py
```diff
@@ -656,20 +657,53 @@ def execute_interpreter(self):
...
-                    sys.stderr.write(
-                        "Failed to read history file at {} due to: {}".format(histfile, e)
+            except ImportError:
+                if self._vars.PEX_INTERPRETER_HISTORY:
+                    pex_warnings.warn(
+                        "PEX_INTERPRETER_HISTORY was requested which requires the `readline` "
+                        "module, but the current interpreter at {python} does not have readline "
+                        "support.".format(python=sys.executable)
                     )
+            else:
```
huonw, line 669: `python` has `-S` that (among other things) disables completion. It looks like this is unconditional, if the modules exist.

Is there potential risks/downsides to doing this unconditionally without a knob to turn it off, beyond the (seemingly small) performance cost of this set-up?

I assume that it's harmless enough, just double checking.

huonw (approved):
Nifty!

pex/pex.py
jsirois: Yeah, I got familiar with the CPython site.py code that -S disables in order to get this all working. I can't see any harm.

tests/integration/test_integration.py
```diff
@@ -234,37 +235,97 @@ def test_pex_repl_built():
...
-    "terminal, but this test fails in CI on macOS with `Inappropriate ioctl for device`, "
-    "because readline.read_history_file expects a tty on stdout. The linux tests will have "
-    "to suffice for now.",
+try:
+    READLINE_AVAILABLE = True
```
jsirois, line 239: Hrm, autoflake.

jsirois: Fixed.

tests/integration/test_integration.py
benjyw: Hmm, archaeology on #2018 (whose CI logs have unfortunately expired) finds a commit message stating that the issue on pypy (in CI) was apparently related to terminfo, so may have been due to interaction between pypy and the CI environment rather than pypy per se, but AFAICT wasn't due to unavailability of readline. Similarly, the macos ENOTTY issue only manifested in CI, and was not related to the presence of readline.

Anyway, if this is now deemed false and the actual underlying issue is just readline then the mentions of CPython in pex/variables.py#531 and pex/variables.py#542 should be removed.

jsirois: Yeah, definitely false. CI passes for `pypy3.10` and mac and I've personally tested for `pypy{2.7,3.{5,6,7,8,9,10}}`. This PR had to plumb `TERM` to work around 1/2 of the tty issues (using pexpect / ptys was the other half).

Variables docs fixed.

benjyw: I should have dived deeper in #2018 then, sounds like TERM would have made things work.

PR #2328, opened 2024-01-15
Fix vendored attrs `sys.path` leak.

jsirois:
Whenever a given Python interpreter on a machine was 1st identified by
Pex during the PEX boot process, Pex's own vendored attrs would be
leaked onto the hermetic `sys.path` of that interpreter forevermore.
This would lead to Pex's vendored attrs not being scrubbed from the
`sys.path` during PEX boot handoff to user code.

### DIFF
--- a/pex/interpreter.py
+++ b/pex/interpreter.py
@@ -19,6 +19,7 @@ from textwrap import dedent
 
 from pex import third_party
 from pex.common import is_exe, safe_mkdtemp, safe_rmtree
+from pex.compatibility import commonpath
 from pex.executor import Executor
 from pex.jobs import Job, Retain, SpawnedJob, execute_parallel
 from pex.orderedset import OrderedSet
@@ -183,10 +184,23 @@ class PythonIdentity(object):
         # vendored attrs distribution so that its `cache_hash=True` feature can work (see the
         # bottom of pex/third_party/__init__.py where the vendor importer is installed). We ignore
         # such adjoined `sys.path` entries to discover the true base interpreter `sys.path`.
-        pythonpath = frozenset(
-            os.environ.get("PYTHONPATH", "").split(os.pathsep) + list(third_party.exposed())
+        pythonpath = os.environ.get("PYTHONPATH")
+        internal_entries = frozenset(
+            (pythonpath.split(os.pathsep) if pythonpath else []) + list(third_party.exposed())
         )
-        sys_path = OrderedSet(item for item in sys.path if item and item not in pythonpath)
+
+        def is_internal_entry(entry):
+            # type: (str) -> bool
+            if entry in internal_entries:
+                return True
+            if not os.path.isabs(entry):
+                return False
+            for internal_entry in internal_entries:
+                if internal_entry == commonpath((internal_entry, entry)):
+                    return True
+            return False
+
+        sys_path = OrderedSet(entry for entry in sys.path if entry and not is_internal_entry(entry))
 
         site_packages = OrderedSet(
             path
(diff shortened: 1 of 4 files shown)


pex/interpreter.py
```diff
@@ -183,10 +184,23 @@ def get(cls, binary=None):
...
-        sys_path = OrderedSet(item for item in sys.path if item and item not in pythonpath)
+
+        def is_internal_entry(entry):
+            # type: (str) -> bool
+            if entry in internal_entries:
+                return True
+            if not os.path.isabs(entry):
+                return False
+            for internal_entry in internal_entries:
+                if internal_entry == commonpath((internal_entry, entry)):
```
jsirois, line 199: The lack of this check was the bug. Something like `~/.pex/isolated/975c556eea71292a09d930db2ca41875066d8be6/pex/vendor/_vendored/attrs` would be (and still is) on the `sys.path` during PEX boot, but only the `~/.pex/isolated/975c556eea71292a09d930db2ca41875066d8be6` entry was in PYTHONPATH since `attrs` was imported via the `pex.third_party` custom importer, which imports from `~/.pex/isolated/975c556eea71292a09d930db2ca41875066d8be6` using the `pex.vendor._vendored.attrs` package prefix.

jsirois: The fix is good, but the comment above has some BS in it I just detected. I'll amend the comment to spell out the situation a bit better in a few hours.

pex/interpreter.py
```diff
@@ -183,10 +184,23 @@ def get(cls, binary=None):
...
+                    return True
+            return False
+
+        sys_path = OrderedSet(entry for entry in sys.path if entry and not is_internal_entry(entry))
```
huonw, line 203: Just confirming that both this `sys.path` and the `internal_entries` lists will be relatively small, so the nested loops are unlikely to be a performance problem?

pex/interpreter.py
huonw, line 199: This looks like this is conceptually equivalent to to something like:

```suggestion
                if entry.startswith(internal_entry):
```

Is that correct? 

Is `commonpath` version is required to ensure that `entry = "/foobar"`, `internal_entry = "/foo"` doesn't return an incorrect value?

pex/interpreter.py
```diff
@@ -183,10 +184,23 @@ def get(cls, binary=None):
...
+        internal_entries = frozenset(
+            (pythonpath.split(os.pathsep) if pythonpath else []) + list(third_party.exposed())
         )
-        sys_path = OrderedSet(item for item in sys.path if item and item not in pythonpath)
+
+        def is_internal_entry(entry):
+            # type: (str) -> bool
+            if entry in internal_entries:
+                return True
+            if not os.path.isabs(entry):
```
huonw, line 196: Might `internal_entries` also have relative paths in it (e.g. `PYTHONPATH="./foo/bar"`)? Does that need handling?

pex/interpreter.py
```diff
@@ -252,7 +288,7 @@ def iter_tags():
         env_markers = MarkerEnvironment(**values.pop("env_markers"))
         return cls(
             version=cast("Tuple[int, int, int]", version),
-            pypy_version=pypy_version,
+            pypy_version=cast("Optional[Tuple[int, int, int]]", pypy_version),
```
huonw, line 291: Is this change and the `__format_version__` handling related to the attrs fix? If so, can you explain how it connects? Thanks.

huonw (approved):

zmanji:
I can confirm this bug exists on my local machine by running:
```
fd -t f .  /Users/zmanji/.pex/interpreters/ -X 'jq' '.sys_path'
```

And seeing some entries have `/Users/zmanji/.pex/isolated/194b5174b81632b65d29e39489cf61230eecfa7a/pex/vendor/_vendored/attrs` which is unexpected.

pex/interpreter.py
```diff
@@ -227,12 +241,34 @@ def get(cls, binary=None):
             configured_macosx_deployment_target=configured_macosx_deployment_target,
         )
 
+    # Increment this integer version number when changing the encode / decode format or content.
+    _FORMAT_VERSION = 1
```
zmanji, line 245: Wouldn't it make sense to just tie this to the current pex version?

jsirois: I could, but at the expense of thrashing caches for no reason. The INTERP-INFO format changes rarely relative to Pex releases.

pex/interpreter.py
jsirois: The `commonpath` function is not `commonprefix` and is not subject to that sort of bug.

jsirois: Oh, perhaps you sorta got that and we're asking for confirmation. If so, yes, exactly.

pex/interpreter.py
jsirois: Correct. But even so, the cross product must be checked anyhow. There is no clever shortcut I can see.

pex/interpreter.py
jsirois: It cannot, no.

pex/interpreter.py
jsirois: There was already logic to bust the cache when a new INTERP-INFO field was added (see the comparison of # entries to 16 on LHS of diff), but in this case, the number of entries did not change (well, it did because I added this new version format field, but ignore that) but the contents of an existing entry, `sys_path`, did. The format version field allows busting the cache when no new entries are added, but contents has changed / a big fixed and it's desired to have the old cache entries auto-evicted.

pex/interpreter.py
huonw: If it _was_ a problem, I have a suspicion could build a trie from the components of `internal_entries`, and traverse it using the components of `entry`, but much better if we can avoid that complexity.

pex/interpreter.py
jsirois: Ok, there is a deeper fix that can be made such that the old check would work. The real bug is that `third_party.exposed()` is empty when it should include attrs. This is a very old bug in the `pex.third_party` custom Importer code. I'll proceed with releasing the current fix and circle back to make the fix cleaner (it'll probably be more ugly / more code, but it will be in the right place).

jsirois: Ok, actually a clean fix. This just required removing an always un-needed check present in the original introduction of the vendoring / pex.third_party mechanism in #624 way back in 2018. See #2330.

PR #2330, opened 2024-01-16
Fix `third_party.exposed()`.

jsirois:
This cleans up #2328 by fixing the root cause of the interpreter
identification leak of exposed `attrs`. The API to report exposed
`sys.path` elements, `third_party.exposed()`, worked at build time,
with a full Pex distribution available, but returned an empty iterator
at run time, with the subset of Pex available in the runtime
`.bootstrap/`.

### DIFF
--- a/pex/interpreter.py
+++ b/pex/interpreter.py
@@ -19,7 +19,6 @@ from textwrap import dedent
 
 from pex import third_party
 from pex.common import is_exe, safe_mkdtemp, safe_rmtree
-from pex.compatibility import commonpath
 from pex.executor import Executor
 from pex.jobs import Job, Retain, SpawnedJob, execute_parallel
 from pex.orderedset import OrderedSet
@@ -188,19 +187,9 @@ class PythonIdentity(object):
         internal_entries = frozenset(
             (pythonpath.split(os.pathsep) if pythonpath else []) + list(third_party.exposed())
         )
-
-        def is_internal_entry(entry):
-            # type: (str) -> bool
-            if entry in internal_entries:
-                return True
-            if not os.path.isabs(entry):
-                return False
-            for internal_entry in internal_entries:
-                if internal_entry == commonpath((internal_entry, entry)):
-                    return True
-            return False
-
-        sys_path = OrderedSet(entry for entry in sys.path if entry and not is_internal_entry(entry))
+        sys_path = OrderedSet(
+            entry for entry in sys.path if entry and entry not in internal_entries
+        )
 
         site_packages = OrderedSet(
             path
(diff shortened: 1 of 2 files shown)


pex/interpreter.py
```diff
@@ -188,19 +187,9 @@ def get(cls, binary=None):
...
-                return True
-            if not os.path.isabs(entry):
-                return False
-            for internal_entry in internal_entries:
-                if internal_entry == commonpath((internal_entry, entry)):
-                    return True
-            return False
-
-        sys_path = OrderedSet(entry for entry in sys.path if entry and not is_internal_entry(entry))
+        sys_path = OrderedSet(
```
jsirois, line 190: This is the simpler check that existed pre-#2328. It was always the right check; it's just that `third_party.exposed()` was broken at PEX runtime.

pex/third_party/__init__.py
```diff
@@ -213,13 +213,11 @@ def iter_installed_vendor_importers(
...
         # type: (...) -> Iterator[VendorImporter]
         root = cls._abs_root(root)
-        vendored_paths = set(cls._vendored_path_items())
         for importer in cls._iter_all_installed_vendor_importers():
             # All Importables for a given VendorImporter will have the same prefix.
             if importer._importables and importer._importables[0].prefix == prefix:
                 if importer._root == root:
-                    if {importable.path for importable in importer._importables} == vendored_paths:
-                        yield importer
+                    yield importer
```
jsirois, line 220: This comparison of the vendored paths at runtime (from the PEX `.bootstrap/`) would always fail since Pex only vendors a subset of all available vendored distributions in the runtime bootstrap of a PEX. The vendored Pip, for one, is never vendored in the runtime; it is only used at build time.

I'm not sure why I added this check in the 1st place back in #624 when I introduced vendoring / the `pex.third_party` import hook. Checking the root and prefix is enough to ensure the discovered `VendorImporter` is the one in question.

pex/interpreter.py
jsirois: And all is well here. Performing this change 1st led to a failing `test_boot_identification_leak` test, and then fixing below got that passing again.

jsirois:
The setuptools build backend introduced to handle pre-testing Python 3.13 in CI is quite flaky. Yet another instance: https://github.com/pantsbuild/pex/actions/runs/7545643289/job/20541741101?pr=2330#step:5:1847

I'm working on switching the build system in a separate PR to handle dynamic `Requires-Python` metadata without using setuptools to solve the files it leaks to the source root during the build process problems.

benjyw (approved):
Nice!

jsirois:
> The setuptools build backend introduced to handle pre-testing Python 3.13 in CI is quite flaky...

Fix in #2331

PR #2331, opened 2024-01-17
Switch to the `hatchling` build backend.

jsirois:
This should elminate flaky CI due to ephemeral dirs in the source root
created by the setuptools build backend which were not multi-process
safe.

The `hatchling` build backend supports metadata plugins and it performs
isolated builds making it a good replacement for the prior setuptools
hack for effecting dynamic `Requires-Python` to enable Python 3.13
shadow-support. Unfortunately, there is no nice way to specify an
in-tree plugin; so we actually use a very thin in-tree build backend
that wraps `hatchling.build`, to allow discovery of our in-tree plugin.


pex/pip/tool.py
```diff
@@ -451,7 +452,7 @@ def spawn_download_distributions(
 
         if not build_isolation:
             download_cmd.append("--no-build-isolation")
-            extra_env.update(PEP517_BACKEND_PATH=os.pathsep.join(sys.path))
+            pex_extra_sys_path.extend(sys.path)
```
jsirois, line 455: Now that Pex uses an in-tree build backend, it unearthed this latent bug where `PEP517_BACKEND_PATH` would get stomped with the backend-path. As such, both here in spawn download and lower in spawn install wheel, we now pass non-isolated-build sys.path via PEX_EXTRA_SYS_PATH instead.

scripts/package.py
```diff
@@ -46,7 +47,12 @@ def build_pex_pex(output_file: PurePath, verbosity: int = 0) -> None:
     subprocess.run(args, check=True)
 
 
-def describe_git_rev() -> str:
+def describe_rev() -> str:
+    if not os.path.isdir(".git") and os.path.isfile("PKG-INFO"):
```
jsirois, line 51: This was always a bug, I just never tried unpacking our sdist and doing a build from there before. That now works.

tests/build_system/test_pep_518.py
```diff
@@ -68,14 +68,15 @@ def test_load_build_system_pyproject(
...
     dists = {
         dist.metadata.project_name
         for dist in Virtualenv(build_system.venv_pex.venv_dir).iter_distributions()
     }
-    assert ProjectName("setuptools") in dists
+    assert ProjectName("hatchling") in dists
     subprocess.check_call(
-        args=[build_system.venv_pex.pex, "-c", "import {}".format(build_system.build_backend)]
+        args=[build_system.venv_pex.pex, "-c", "import {}".format(build_system.build_backend)],
+        env=build_system.env,
```
jsirois, line 79: This env carries the backend-path -> PEX_EXTRA_SYS_PATH which is needed to have the build see our now in-tree build backend.

tests/integration/build_system/test_pep_518.py
```diff
@@ -55,5 +55,6 @@ def test_load_build_system_pyproject_custom_repos(
     build_system = load_build_system(current_target, custom_resolver, pex_project_dir)
     assert isinstance(build_system, BuildSystem)
     subprocess.check_call(
-        args=[build_system.venv_pex.pex, "-c", "import {}".format(build_system.build_backend)]
+        args=[build_system.venv_pex.pex, "-c", "import {}".format(build_system.build_backend)],
+        env=build_system.env,
```
jsirois, line 59: Ditto https://github.com/pantsbuild/pex/pull/2331/files#r1455083605

kaos (approved):

benjyw (commented):
Late to the party, but LGTM

PR #2338, opened 2024-01-19
Upgrade readthedocs requirements.

### LINKED ISSUE #2337: Documentation website is down/missing
404 Not Found error when attempting to access documentation at: [pex.readthedocs.io](https://pex.readthedocs.io/)

image

benjyw:
Our older pins were no longer compatible with production RTD.

Fixes #2337

### DIFF
--- a/docs/rtd/requirements.txt
+++ b/docs/rtd/requirements.txt
@@ -1,10 +1,4 @@
-# N.B.: Pex RTD started failing upon release of docutil 0.18; work around by pinning low.
-# See: https://github.com/pantsbuild/pex/issues/1508
-docutils<0.18
-
-sphinx==3.2.1
-sphinx_rtd_theme==0.5.0
-
-# N.B.: Sphinx doesn't work with jinja2 3.1.0.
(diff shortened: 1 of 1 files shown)


benjyw:
I tested this by replicating the RTD build commands locally, and it seems to work. 

This stops the bleeding, but I will circle back to figure out if we can pare down/eliminate this requirements file. Since RTD is opinionated about what it supports, presumably it should be able to just work with those opinions.

benjyw:
Based on https://blog.readthedocs.com/defaulting-latest-build-tools/ RTD should now install the latest sphinx and sphinx_rtd_theme by default, and looking at the git history of our requirements.txt, I think those are all we care about. Looks like the other pins were to work around temporary compatibility issues, and then they started piling up on each other. So a followup PR will test that assumption (I'd like to merge this to get things working again before going down that path).

jsirois (approved):
Yeah, if `tox -edocs` works (it does for me), then we don't need any of this I think and just need to add a version of `tox -edocs` that does not attempt to open the docs to the GH action checks shard.

Issue #2344, opened 2024-01-23
Support syncing input requirements to a lock file more easily.


jsirois:
With #2335, the correct set of arguments to `pex3 lock update` can be crafted to minimally update to a lock to an evolved set of input requirements. As noted here https://github.com/pantsbuild/pex/issues/2334#issuecomment-1900963246 though, a nicer interface would be:
```
pex3 lock sync "req1<2" -r reqs.txt --constraints constraints.txt --lock lock.json -- pytest ...
```

The idea is to perform a minimal lock update given the input requirements and constraints (or noop if reqs have not changed) and, if `--` is present, execute the command line following it after syncing that command's venv with the lock.

This new notion of an ~auto minimal lock update would parse all input requirements and use the new features from #2335 to form the appropriate set of `-R` replacements and `-d` deletes to bring the lock in sync.

PR #2346, opened 2024-01-26
Plumb Pip's `--{no,only}-binary`.

### LINKED ISSUE #2343: Support Pip's `--only-binary` /  `--no-binary`.
This would require being able to parse the flag in requirements files as well as providing Pex top-level flags for the option as well.

jsirois:
Previously, Pex just allowed specifying `--no-wheel` or `--no-build`
globally. Now, individual projects can be slated for `--only-wheel` or
`--only-build` just as they can with Pip. In the lock case, this
information is saved so that lock updates carry the configuration
forward.

Fixes #2343


pex/pip/tool.py
```diff
@@ -401,6 +405,36 @@ def _spawn_pip_isolated_job(
...
+                yield str(project)
+            for project in build_configuration.only_builds:
+                yield "--no-binary"
+                yield str(project)
+
+        if build_configuration.prefer_older_binary:
+            yield "--prefer-binary"
+
+        if build_configuration.use_pep517 is not None:
+            yield "--use-pep517" if build_configuration.use_pep517 else "--no-use-pep517"
```
kaos, line 433: Q: what makes the pep517 option different from the other bool options?

_(realized I could probably reply to myself here by reading `pip install --help`)_
A: the difference aiui is that pep517 is used by default, so the option really is about disabling it.

pex/resolve/resolver_configuration.py
```diff
@@ -82,15 +84,105 @@ def create(
...
+            raise self.Error(
+                "Builds were disallowed, but the following project names are configured to only "
+                "allow building: {only_builds}".format(
+                    only_builds=", ".join(sorted(map(str, self.only_builds)))
+                )
+            )
+        if not self.allow_wheels and self.only_wheels:
+            raise self.Error(
+                "Resolving wheels was disallowed, but the following project names are configured "
+                "to only allow resolving pre-built wheels for: {only_wheels}".format(
```
kaos, line 138:
```suggestion
                "Resolving wheels was disallowed, but the following project names are configured "
                "to only allow resolving pre-built wheels: {only_wheels}".format(
```

kaos (approved):

pex/pip/tool.py
jsirois: Thanks for digging a bit there, but I'm still crying a bit. Did you check out Pex's own docs?:
https://github.com/pantsbuild/pex/blob/015a077d3496205824491bf97136afd719277b12/pex/resolve/resolver_options.py#L206-L224

Assuming those are read, if they still don't make sense, it would be good to know so I can fix.

jsirois: I fixed that bad URL.

jsirois:
Ok, @cburroughs I knocked this out and will release 2.1.161 as soon as it lands green here in the hour. My time got cut short this stint and I only have through Saturday evening to bang out #2344. I think I'll get that done, but in case not, this should unblock your Pants work on the sharper knife path if you don't want to potentially wait until ~February 9th when I next pick back up work after going AFK this Sunday.

cburroughs:
Thank for the update, context, and enjoy your trip!  I'll try to give it a look right now.


It lookslike the pex has an extra validation that stops the pip special `:all:` and `:none` values from flowing through:
```
$ python3.10 -m pex.cli lock create '--indent=2'  --manylinux manylinux2014 --interpreter-constraint $'CPython==3.10.*' --style=universal --pip-version=23.3.2 --resolver-version pip-2020-resolver --target-system=linux --target-system=mac --only-binary :all:  ansicolors==1.1.5 kaleido==0.1.0.post1   -o tmp/k2.json 
usage: /usr/bin/python3.10 -m pex.cli lock create [-h] [--style {strict,sources,universal}] [--target-system {linux,mac,windows}]
                                                  [--path-mapping PATH_MAPPINGS] [-o PATH] [--indent INDENT] [-r FILE or URL]
                                                  [--constraints FILE or URL] [--python PYTHON] [--python-path PYTHON_PATH]
                                                  [--interpreter-constraint INTERPRETER_CONSTRAINT] [--platform PLATFORMS]
                                                  [--complete-platform COMPLETE_PLATFORMS] [--manylinux [ASSUME_MANYLINUX]]
                                                  [--resolve-local-platforms]
                                                  [--resolver-version {pip-legacy-resolver,pip-2020-resolver}]
                                                  [--pip-version {latest,vendored,20.3.4-patched,22.2.2,22.3,22.3.1,23.0,23.0.1,23.1,23.1.1,23.1.2,23.2,23.3.1,23.3.2}]
                                                  [--allow-pip-version-fallback] [--use-pip-config] [--pypi] [-f PATH/URL]
                                                  [-i URL] [--retries RETRIES] [--timeout SECS] [--proxy PROXY] [--cert PATH]
                                                  [--client-cert PATH] [--cache-ttl DEPRECATED] [-H DEPRECATED] [--pre] [--wheel]
                                                  [--only-binary ONLY_WHEELS] [--build] [--only-build ONLY_BUILDS]
                                                  [--prefer-wheel] [--force-pep517] [--build-isolation] [--transitive] [-j JOBS]
                                                  [--preserve-pip-download-log] [-v] [--emit-warnings] [--pex-root PEX_ROOT]
                                                  [--disable-cache] [--cache-dir CACHE_DIR] [--tmpdir TMPDIR] [--rcfile RC_FILE]
                                                  [requirements ...]
/usr/bin/python3.10 -m pex.cli lock create: error: argument --only-binary/--only-wheel: The given project name ':all:' is not a valid. It must conform to the regex '^([A-Z0-9]|[A-Z0-9][A-Z0-9._-]*[A-Z0-9])$' as specified in https://peps.python.org/pep-0508/#names

```

jsirois:
Sorry about that! It looks like I edited your comment instead of replying. Very bad.

> It looks like the pex has an extra validation that stops the pip special :all: and :none values from flowing through:

Correct! That's what the pre-existing `--no-{wheel,build}` options are for. In other words these are not direct pass throughs. They only logically pass through. See here for the exact mapping step setting up Pip's args: https://github.com/pantsbuild/pex/pull/2346/files#diff-443e54ea41451726af072f7976469d8bdbb37d285fd49fc1a4bc6447588591c3R409

For context, the `--no-{wheel,build}` options were introduced back in Pex "OLD" (2014) when it had a bespoke resolver and did not use Pip. Since Pex doesn't break users, the option names and meanings stayed constant after adoption of Pip as the underlying resolver in the 2.x series.

pex/pip/tool.py
kaos: Ah, my bad for not realizing the option was of course documented for pex as well (got tunnel visioned onto the `pip` arg..).

Yea, the pex option has a lot of information to digest. Reading it through a few times makes it almost clear. I wonder if there is a typo here:
```diff
---  If PEP-517 is forced (--use-pep517 is passed)
+++  If PEP-517 is forced (--force-pep517 is passed)
```

then again, I would expect that `--force-pep517` is the same as `--use-pep517` (based on my understanding of how `ArgumentParser` works), but the help has so many layers that I get the impression that they are not.. (also, that it reads `default: None` rather than `default: True` doesn't help which would align with the help text stating that it uses pep517 by default)
So, I feel there's a bit of language ... (failed to find the word I was looking for) here.

I think this option would be clearer using an enum value, but I guess the current state is due to upholding backwards compatibility..? In which case I think making the relationship between `--force-pep517` and `--use-pep517` clearer would help, and perhaps also mentioning that the default behavior is based on not providing a value for the option (e.g. explain that this flag is a tri-bool)

I can see why you cry a bit over this.

jsirois: Ok, so to sum up - I think you figured this out, but in case not:
+ The option is a None|True|False tri-state
+ The default is None
+ The option names are all just synonyms (with --no- prefixes negating)

So the default, None, means do the right thing: if there is a project.toml build system, use it; if not concoct one based on legacy setuptools build backend.

True means use (force) pep 517 always.

False means do not use pep 517.

I have no clue what use cases benefit from the True or False cases. I just exposed this Pip knob to not get between users and Pip. Perhaps my sin was in trying to make the Pip help different / more clear in up in Pex. I clearly failed there!

kaos: Thanks. Yea, that summation aligns with my current understanding.

pex/resolve/lock_resolver.py
```diff
@@ -98,8 +101,9 @@ def __init__(
         self._network_configuration = network_configuration
         self._password_entries = password_entries
         self._cache = cache
-        self._use_pep517 = use_pep517
-        self._build_isolation = build_isolation
+        self._build_configuration = attr.evolve(
+            build_configuration, allow_wheels=False, prefer_older_binary=False
```
jsirois, line 105: This is a bug. The incoming `BuildConfiguration` from the user may say `allow_builds=False, allow_wheels=True` (the known useful example being a user constraining a `--style universal` lock to just wheels), in which case just flipping `allow_wheels=False` here leads immediately to `BuildConfiguration` failing a pre-condition check (`allow_wheels=False, allow_builds=False` makes no sense and is illegal). The `VCSArtifactDownloadManager` should complete its thought and say it uses the user `BuildConfiguration` but trumps `allow_wheels=False, allow_builds=True` since the `VCSArtifactDownloadManager` must be able to build a wheel from the VCS checkout to do anything useful at all.

PR #2356, opened 2024-02-07
Create SSLContexts in the main thread.

jsirois:
This solves #2355 without yet understanding why that issue exists.

Fixes #2355

Issue #2355, opened 2024-02-07
Generating lockfiles fails with: unknown error (_ssl.c:3161)


jsirois:
As initially reported here: https://github.com/pantsbuild/pants/issues/20467

A streamlined repro:
```Dockerfile

FROM fedora:37

RUN curl --fail -sSL -O \
    https://github.com/indygreg/python-build-standalone/releases/download/20240107/cpython-3.9.18+20240107-x86_64-unknown-linux-gnu-install_only.tar.gz && \
    curl --fail -sSL -O \
    https://github.com/indygreg/python-build-standalone/releases/download/20240107/cpython-3.9.18+20240107-x86_64-unknown-linux-gnu-install_only.tar.gz.sha256 && \
    [[ \
        "$(cat cpython-3.9.18+20240107-x86_64-unknown-linux-gnu-install_only.tar.gz.sha256)" \
        == \
        "$( \
            sha256sum cpython-3.9.18+20240107-x86_64-unknown-linux-gnu-install_only.tar.gz | \
            cut -d' ' -f1 \
        )" \
    ]] && \
    tar -xzf cpython-3.9.18+20240107-x86_64-unknown-linux-gnu-install_only.tar.gz

RUN python/bin/python3.9 -mvenv pex.venv && \
    pex.venv/bin/pip install -U pip && \
    pex.venv/bin/pip install pex

ENV PATH=$PATH:pex.venv/bin
```

```console
$ docker build . -t repro
[+] Building 0.0s (8/8) FINISHED                                                                         docker:default
 => [internal] load .dockerignore                                                                                  0.0s
 => => transferring context: 2B                                                                                    0.0s
 => [internal] load build definition from Dockerfile                                                               0.0s
 => => transferring dockerfile: 965B                                                                               0.0s
 => [internal] load metadata for docker.io/library/fedora:37                                                       0.0s
 => [1/4] FROM docker.io/library/fedora:37                                                                         0.0s
 => CACHED [2/4] RUN dnf install -y curl                                                                           0.0s
 => CACHED [3/4] RUN curl --fail -sSL -O     https://github.com/indygreg/python-build-standalone/releases/downloa  0.0s
 => CACHED [4/4] RUN python/bin/python3.9 -mvenv pex.venv &&     pex.venv/bin/pip install -U pip &&     pex.venv/  0.0s
 => exporting to image                                                                                             0.0s
 => => exporting layers                                                                                            0.0s
[...]

PR #2358, opened 2024-02-09
Explain why #2356 "works".

jsirois:
The explanation was hard won and a bit embarrasing in outcome. The
original behavior was correct afaict and PBS use on a RedHat system with
custom RedHat OpenSSL config keys should fail, obviously with a much
better error message, and prompt OpenSSL configuration evaluation on the
machine. That said, the behavior is released now and Pex stands behind
it.

It may make sense to add a `--strict-ssl` option or something similar to
restore the old behavior and let the (confusing) error bubble, perhaps
with a pointer to what may be wrong.

This should also serve to close
https://github.com/indygreg/python-build-standalone/issues/207 or at
least give Gregory enough information to decide what to do over in PBS.

Closes the loose ends in #2355.


jsirois:
@mjimlittle and @xlevus, I'd appreciate your eyes on the big comment added in this PR. In short, it explains that your RedHat systems use a RedHat proprietary OpenSSL config option that only works with the patched OpenSSL that ships with the OSes you use. There is not much I can see that PBS Python's can do about this! That said, you should know that PBS Pythons are choking on your SSL Config on those machines and may only be partially configured / ignoring some of your settings.

kaos (approved):

benjyw (approved):
Thanks, that's painful.

jsirois:
> Thanks, that's painful.

Yeah, a bit. Always fun to learn more gdb though.

I'm not sure what Pants / scie-pants want to do about this, but I'll be adding a warning over in the science PBS provider docs about this potential security hole in PBS ssl. If you'd like to provide security-conscious Pants users the option to see these errors I'm happy to field a feature request / bug report about this silent bad SSL config dropping (i.e. add the `--strict-ssl` option I mentioned in the OP). I'm even happier to review a PR if any of you want to contribute that.

PR #2359, opened 2024-02-10
Update pantsbuild(.org) -> pex-tool(.org).

jsirois:
Also add a bespoke CONTRIBUTING.md instead of pointing off to the
pantsbuild.org contributing docs.


zmanji:
🚀 I guess if the CI goes green then this migration was successful?

zmanji (approved):

jsirois:
> I guess if the CI goes green then this migration was successful?

Yup. There will be a few details after to make sure the next deploy works, but PRs will be open for business.

PR #2364, opened 2024-02-14
Fix PyPI deploy.

jsirois:
The v2.1.164 release went smoothly except for the PyPI deploy. I had to
build and deploy manually from the v2.1.164 tag with:
```
$ rm -rf dist/
$ tox -epackage -- \
--additional-format sdist \
--additional-format wheel \
--embed-docs
$ pex twine -c twine -- upload -s dist/*
```

The action wants the dist/ dir to just contain the artifacts to deploy;
no more and no less.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Check Packaging
         uses: pantsbuild/actions/run-tox@b16b9cf47cd566acfe217b1dafc5b452e27e6fd7
         with:
-          tox-env: package -- --additional-format sdist --additional-format wheel --embed-docs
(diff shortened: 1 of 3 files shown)


kaos (approved):

benjyw (approved):
Out of curiosity, why was this not an issue in the past? Is this related to the org switch, and how?

jsirois:
This is new due to `--embed-docs`, which causes html docs to be generated to `dist/docs/html` prior to building the wheel as part of the embedding machinery.

So you now got:
```
dist/
    docs/
    pex-2.1.164.tar.gz
    pex-2.1.164-py2.py3-none-any.whl
```
And https://github.com/marketplace/actions/pypi-publish tries to publish everything in `dist/` with no knobs to control inclusion or exclusion.

PR #2365, opened 2024-02-15
Add `pex --docs` and several `pex3 docs` options.

jsirois:
The `pex --docs` shortcut should provide an ergonomic way to open the
local docs for most Pex CLI users (undoubtedly few use `pex3`). The
`pex3 docs` additional options allow for most of the control needed
though, and the help for the `pex --docs` option points there.

### DIFF
--- a/pex/docs/command.py
+++ b/pex/docs/command.py
@@ -1,0 +1,98 @@
+# Copyright 2024 Pex project contributors.
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+from __future__ import absolute_import
+
+import logging
+from argparse import Namespace, _ActionsContainer
+from textwrap import dedent
+
+from pex import docs
+from pex.commands.command import try_open_file
+from pex.docs.server import SERVER_NAME, LaunchError, LaunchResult
+from pex.docs.server import launch as launch_docs_server
+from pex.result import Error, try_
+from pex.typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from typing import Optional, Union
+
+    import attr  # vendor:skip
+else:
+    from pex.third_party import attr
+
+
+logger = logging.getLogger(__name__)
+
+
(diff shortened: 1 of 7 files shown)


zmanji (approved):
I am not sure on the whole server/pidfile logic but everything lgtm.

jsirois:
Thanks for taking a look @zmanji.

> I am not sure on the whole server/pidfile logic but everything lgtm.

That is just moved code in this PR (so `pex --docs` could share the `pex3 docs` infra), and its still ~copied code from `testing.devpi` which powers the `--devpi` option for the tests and integration tests. Both CI and me have been hammering that logic for a while now; so it may be buggy, but it appears to be solid; so the bugs likely lie in rare corner cases.

pex/docs/command.py
```diff
@@ -0,0 +1,98 @@
...
+# E    X    P
+# 0x45 0x58 0x50
+#
+# This is a currently unassigned port in the 1024-49151 user port registration range.
+# We have no intention of registering! But this value plays by the loose rules and satisfies
+# the criteria of being unlikely to be in use.
+STANDARD_PORT = 45585
+
+
+def serve_html_docs(
```
jsirois, line 63: This function was pretty much a pure extraction from the run method of `pex.cli.commands.docs.Docs` so that `pex --docs` could share the infra.

pex/docs/server.py
```diff
@@ -0,0 +1,200 @@
...
+    url = attr.ib()  # type: str
+    pid = attr.ib()  # type: int
+
+    def __str__(self):
+        # type: () -> str
+        return "{url} @ {pid}".format(url=self.url, pid=self.pid)
+
+
+@attr.s(frozen=True)
+class Pidfile(object):
```
jsirois, line 46: This file was pretty much a pure extraction from `pex.cli.commands.docs` so that `pex --docs` could share the infra.

PR #2373, opened 2024-02-18
Implement pex3 lock sync.

### LINKED ISSUE #2344: Support syncing input requirements to a lock file more easily.

jsirois:
Introduce the high level `pex3 lock sync` command. This should generally
suffice for typical use cases and works as follows:

+ On first use (where the specified `--lock` does not yet exist), it
  acts just like `pex3 lock create`.
+ On subsequent uses it does a minimal synchronization of the lock based
  on the diff of the given current requirements against the requirements
  used to generate the specified `--lock`. This amounts to formulating a
  `pex3 lock update` command with the appropriate `-p`, `-R` and `-d`
  arguments.

In addition to creating and syncing a lock, it can also create and sync
a venv (--venv) based on the lock. Further, a command can be specified
to run in the synchronized venv with arguments following the `--` option
terminator.

This latter set of features allow Pex to act as a concise tool in `tox` /
`nox` / `invoke` / `make` setups to implement a simple build system.

Fixes #2344


jsirois:
Another pretty huge one. Thanks in advance for any scrutiny you can provide.

I did not convert Pex's `tox` setup to use itself as the venv provisioner in this PR to keep the already large diff size down. There will be a follow up for that though as well as similar use in a-scie/lift (science) where Pex is already heavily used, but in a cumbersome way, to drive the underlying `nox` venvs.

jsirois:
@cburroughs I sent you an invite. I'd like to have you review especially with an eye towards making sure you have everything you need for your Pants work with this change.

cburroughs:
Thank you!  I'll take a long look this week.

jsirois:
N.B.: I've noticed a few things and will be pushing a few individually reviewable commits here at the tail. None of these change the nature of the large 1st commit; so any review effort there is not wasted.

jsirois:
Ok, with 120fe0d I think this is complete and I'll stay hands off until there is time to review.

benjyw:
Nice! I'll take a close look over the next couple of days.

pex/cli/commands/lock.py
```diff
@@ -119,6 +131,237 @@ def __call__(self, parser, namespace, value, option_str=None):
...
+                    exe_path = os.path.abspath(os.path.join(entry, argv0))
+                    if is_exe(exe_path):
+                        argv0_path = exe_path
+                        break
+
+        if argv0_path:
+            venv = Virtualenv.enclosing(python=argv0_path)
+            if not venv:
+                try:
+                    venv = Virtualenv(os.path.dirname(os.path.dirname(argv0_path)))
```
huonw, line 172: I see that this double parent is used a few places in the code base so probably isn't "new"... what it's unclear what it means? What `--venv` argument value might a user pass that falls into this path?

pex/cli/commands/lock.py
```diff
@@ -119,6 +131,237 @@ def __call__(self, parser, namespace, value, option_str=None):
...
+
+        abs_venv_dir = os.path.realpath(self.venv.venv_dir)
+
+        existing_distributions_by_project_name = {
+            dist.metadata.project_name: dist for dist in self.venv.iter_distributions()
+        }  # type: Dict[ProjectName, Distribution]
+
+        to_remove = []  # type: List[Distribution]
+        to_install = []  # type: List[Distribution]
+        for distribution in distributions:
```
huonw, line 219: Confirming my understanding of transitive dependencies of requirements (e.g. `explicit` requires `hidden`, and the command is something like `pex3 lock sync explicit --venv ...`): 

1. does `distributions` include all of them, not just the top level requirements?
2. I don't see explicit tests for these sort of transitive dependencies, either for updating lockfiles (including something transitioning from transitive-only to top-level requirement and vice versa) or for updating a venv. Have I missed some? If not, is that something worth testing or is it well covered elsewhere?

pex/cli/commands/lock.py
```diff
@@ -119,6 +131,237 @@ def __call__(self, parser, namespace, value, option_str=None):
...
+                    )
+                    # Pip generates "installed-files.txt" upon installing and this file does not
+                    # include itself; so we tack that on here ourselves.
+                    to_unlink.append(
+                        os.path.realpath(os.path.join(distribution.location, installed_files))
+                    )
+            if to_unlink:
+                to_unlink_by_pin[
+                    (distribution.metadata.project_name, distribution.metadata.version)
+                ] = [file for file in to_unlink if abs_venv_dir == commonpath((abs_venv_dir, file))]
```
huonw, line 262: I've noticed this idiom is used in a quite a few places (not just this PR), is it worth a dedicated `is_prefix_of` (or something) function for it?

pex/cli/commands/lock.py
```diff
@@ -119,6 +131,237 @@ def __call__(self, parser, namespace, value, option_str=None):
...
+                "{venv}? [yN]: ".format(venv=self.venv.venv_dir)
+            ).strip().lower() not in ("y", "yes"):
+                return Error("Sync cancelled.")
+
+        if to_unlink_by_pin:
+            parent_dirs = set()  # type: Set[Text]
+            for file in itertools.chain.from_iterable(to_unlink_by_pin.values()):
+                safe_delete(file)
+                parent_dirs.add(os.path.dirname(file))
+            for parent_dir in sorted(parent_dirs, reverse=True):
```
huonw, line 279: Cool trick with the `reverse=True` 👍

pex/cli/commands/lock.py
```diff
@@ -397,15 +656,48 @@ def _add_update_arguments(cls, update_parser):
...
+            "--no-pip",
+            "--no-retain-pip",
+            action=HandleBoolAction,
+            default=False,
+            help=(
+                "When creating a venv that doesn't exist, install pip in it. When syncing an "
+                "existing venv, retain its pip, if any, even if pip is not present in the lock "
+                "file being synced. N.B.: When in `--no-retain-pip` mode (the default) only remove "
+                "pip if it is both present in the venv and not present in the lock file being "
+                "synced."
```
huonw, line 686: If a lockfile has pip in it:

- does the `--no-retain-pip` (a) install pip to a new venv (or existing one without it), and (b) update it if an existing venv has a different version?
- does `--pip` use the version from the lockfile?

If so, the docs might be slightly clearer trying to convey something like:

- In the default `--no-retain-pip` mode, `pip` isn't special, and behaves like any other package in the lock (i.e. installed/updated/removed as required)
- With `--retain-pip`, `pip` is explicitly installed (aka, not removed) in venv, even if it is not in the lockfile (and if it is in the lockfile, the lockfile's version is used?)

pex/cli/commands/lock.py
```diff
@@ -1028,3 +1272,235 @@ def process_req_edits(
...
+                    venv.install_pip(upgrade=True)
+            sync_target = SyncTarget.resolve_command(venv=venv, command=self.passthrough_args)
+        elif self.passthrough_args:
+            sync_target = try_(
+                SyncTarget.resolve_venv(self.passthrough_args[0], *self.passthrough_args[1:])
+            )
+        if not sync_target:
+            return Ok()
+
+        if self.options.dry_run:
```
huonw, line 1459: From this, it looks like `--dry-run` only applies to the sync to the venv, and the `--lock` lockfile is still created/mutated, which I find a bit surprising: I'd expect a `--dry-run` flag to mean not mutate anything "real" on disk, only caches and temporary files. It looks like the docs for `--dry-run` say "Don't update the lock file".

Am I missing something?

huonw (commented):
This is very cool. I've focused on the core change in `lock.py` and skimmed the tests, but haven't had time to look beyond that.

pex/cli/commands/lock.py
jsirois: The venv bin dir. So /a/venv/bin/script. The parent is /a/venv/bin/, and its parent is /a/venv/ - the venv dir.

pex/cli/commands/lock.py
jsirois: I guess I haven't thought so. The `commonpath` function is stdlib and this is about the only way to use it; so why obscure a stdlib idiom behind a bespoke to the Pex codebase helper.

pex/cli/commands/lock.py
jsirois: You are, partly. The `--dry-run` option is accessed by helpers that perform the update. The create path, though, does not handle dry-run and always creates the lock file. I'm not sure what to do about that. A dry-run venv mutation makes sense (the create path can sync an existing venv instead of creating a new one). A dry run new lock file creation makes a lot less sense.

pex/cli/commands/lock.py
jsirois: 1. Yes, distributions come from a full transitive lock resolve.
2. There are extensive tests of lock create and lock update in other files. This change just re-uses that 2 year old infra.

That said re 2, the transition you mention is probably interesting enough to test here.

jsirois:
I've piece-meal responded. I'll have full time to respond to all comments on 2/26 / 2/27.

Some of the questions, though are answered by tests in this PR if you get impatient.

pex/cli/commands/lock.py
jsirois: This is the grid that makes sense as I see it (keep in mind there is only a single bool here and the `--no-*` mean `False` / no Pip and the other two mean `True`, the venv should have Pip:
||Existing venv + resolve has Pip|Existing venv + resolve does not have Pip|No venv + resolve has Pip|No venv + resolve does not have Pip|
|---|---|---|---|---|
|`--pip`   |install resolve's Pip|`-m ensurepip`|install resolve's Pip|`-m ensurepip`|
|`--no-pip`|install resolve's Pip|uninstall Pip if installed|install resolve's Pip||

In other words the idea is `--no-pip` is ignored if the the resolve does in fact contain Pip. One tactic there would be to error instead: you told me no Pip but your resolve contains Pip. My thinking here is the main point is "sync"; so clearly the lock trumps. In a scenario where there are many venvs being synced, it makes sense to say `pex3 lock sync --no-pip ...` to ensure final venvs do not have pip unless in the lock (i.e.: a use case where you generally want the venv <-> lock mapping to be 1-1 / controlled by sync with no cowboying the venv by using Pip later ... unless the lock explicitly asks for Pip).

jsirois: I believe the 1 case in the grid not handled is `--pip` + "Existing venv + resolve does not have Pip". According to the grid, Pip should be installed via `-m ensurepip` but it is not. I'll fix that and add a test as well as take a second go at the option help docs.

pex/cli/commands/lock.py
jsirois: N.B.: For Windows the structure is `C:\a\venv\Scripts\script.exe` and so the `..\..` treatment holds.

pex/cli/commands/lock.py
jsirois: Ok, tested all of 2 in `test_sync_venv_transitive_to_direct_and_vice_versa`.

jsirois:
> I've piece-meal responded. I'll have full time to respond to all comments on 2/26 / 2/27.

@huonw I've fully responded as far as I can tell saving for the case of what to do with `--dry-run` when there is no lock file yet (the create case). I still have no good idea there besides the current inconsistency. Certainly printing out the contents of the lock that would be written, in JSON form or the sentency form that update `--dry-run` uses, seems not useful. Perhaps just outputting something like a `pip freeze`?:
```
Would lock 37 projects:
  ansicolors 1.1.8
  cowsay 5.0
  foo 1
  bar 1
  ...
  spam 1
```

pex/cli/commands/lock.py
jsirois: Ok, there is now a trimmed down dry run report for lock creation such that no lock file hits disk.

jsirois:
Ok, all of @huonw's feedback is addressed AFAICT; so ready for round 2 of feedback.

jsirois:
Tests are bloody, but it appears to be a GitHub outage with action repos timing out during clone and ghcr.io also timing out when downloading Docker images.

pex/cli/commands/lock.py
```diff
@@ -440,6 +746,19 @@ def add_extra_arguments(
...
+                "Create or update a Pex lock file from requirements and optionally synchronize a "
+                "venv to it."
+            ),
+            passthrough_args=(
+                "The path to a venv directory to sync with this lock or else a command line whose "
+                "1st argument resolves to a binary in a venv."
+            ),
+            func=cls._sync,
+        ) as update_parser:
+            cls._add_sync_arguments(update_parser)
```
kaos, line 761:
```suggestion
        ) as sync_parser:
            cls._add_sync_arguments(sync_parser)
```

pex/cli/commands/lock.py
```diff
@@ -1414,8 +1428,28 @@ def _sync(self):
                     pip_configuration=pip_configuration,
                 )
             )
-            with safe_open(lock_file_path, "w") as fp:
-                self._dump_lockfile(lockfile, output=fp)
+            if self.options.dry_run:
```
huonw, line 1431: This seems like an appropriate way to handle creating with `--dry-run`, thanks for getting it in.

huonw (approved):
The code looks good; thanks for the updates. I haven't (and won't, sorry) had time to execute it/play with it, yet, though; sounds like @cburroughs may be able to cover that more practical validation.

benjyw:
Sorry for the delay, got back from a ski trip yesterday, so doing a deep dive on this now.

pex/cli/commands/lock.py
```diff
@@ -397,15 +667,51 @@ def _add_update_arguments(cls, update_parser):
...
+                "Synchronize this venv to the contents of the lock after synchronizing the "
+                "contents of the lock to the passed requirements."
+            ),
+        )
+        sync_parser.add_argument(
+            "--venv-python",
+            help="Use this interpreter to create the `--venv` if the venv doesn't exist.",
+        )
+        sync_parser.add_argument(
+            "--pip",
```
benjyw, line 686: I don't have good intuition for when this feature will be useful. I suppose people might want some recent pip in their venv so they can mess around with it, but don't want to specify it explicitly in their requirements files because it's not a code dep in their repo? 

Either way, the behavior grid you posted in an earlier comment seems reasonable to me.

benjyw (approved):

pex/cli/commands/lock.py
```diff
@@ -119,6 +131,248 @@ def __call__(self, parser, namespace, value, option_str=None):
...
+                to_unlink_by_pin[
+                    (distribution.metadata.project_name, distribution.metadata.version)
+                ] = [file for file in to_unlink if abs_venv_dir == commonpath((abs_venv_dir, file))]
+        if confirm and to_unlink_by_pin:
+            for (project_name, version), files in to_unlink_by_pin.items():
+                print(project_name, version, ":", file=sys.stderr)
+                for f in files:
+                    print("    ", f, file=sys.stderr)
+            if input(
+                "Remove the outdated distributions listed above from the venv at "
```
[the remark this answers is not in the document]
jsirois: Yes. In the vacuum of a suggestion though I'll leave the wording as is.

pex/cli/commands/lock.py
```diff
@@ -119,6 +131,248 @@ def __call__(self, parser, namespace, value, option_str=None):
...
+            lock_config_updated=lock_config_updated,
+        )
+
+    def _no_updates(self):
+        # type: () -> Ok
+        return Ok(
+            "No lock update was performed.\n"
+            "The following platforms present in {lock_file} were not found on the local "
+            "machine:\n"
+            "{missing_platforms}\n"
```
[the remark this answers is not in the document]
jsirois: Only insofar as the data structure can be; i.e.: I did not concoct an at-least-1-element-container for the Lockfile. In practice though, you can't create a Lockfile without at least 1 locked resolve.

pex/cli/commands/lock.py
```diff
@@ -119,6 +131,248 @@ def __call__(self, parser, namespace, value, option_str=None):
...
+            "You might still be able to update the lock by adjusting target options like "
+            "--python-path.".format(
+                lock_file=self._lock_file_path,
+                missing_platforms="\n".join(
+                    sorted(
+                        "+ {platform}".format(platform=locked_resolve.platform_tag)
+                        for locked_resolve in self._lock_updater.lock_file.locked_resolves
```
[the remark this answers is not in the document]
jsirois: They may either not be locally present or, as the advice hopefully conveys, they may be present, but not visible (i.e.: there may be applicable Python interpreters not on the PATH or on the PEX_PYTHON_PATH, etc.). If you have a re-word suggestion, let me know.

pex/cli/commands/lock.py
jsirois: Or, the venv may be pre-existing and they don't want to blow away the Pip already there just because it isn't in the lock.

pex/cli/commands/lock.py
```diff
@@ -119,6 +131,248 @@ def __call__(self, parser, namespace, value, option_str=None):
...
+            ).strip().lower() not in ("y", "yes"):
+                return Error("Sync cancelled.")
+
+        if to_unlink_by_pin:
+            parent_dirs = set()  # type: Set[Text]
+            for file in itertools.chain.from_iterable(to_unlink_by_pin.values()):
+                safe_delete(file)
+                parent_dirs.add(os.path.dirname(file))
+            for parent_dir in sorted(parent_dirs, reverse=True):
+                if not os.listdir(parent_dir) and abs_venv_dir == commonpath(
```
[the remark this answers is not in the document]
jsirois: Yes, if the distribution has a top-level package that is >1 level deep; i.e.: Project foo-bar 1.0 distributes code starting inside the `foo.bar` package (so `foo/bar/...` with no `foo/*.py`). I have never seen this in practice, but deleting empty dirs up to the site-packages dir would be fine. That said, that is just 1 case! There is the bin dir, and the free-form territory of the data dirs. In the bin dir case, it should never be empty - there should at least by a python executable symlink to the base python. In the data dir case I'm really unsure and a bit hesitant to go further. In general, this is terrifying territory since it strays outside of the PEX_ROOT into potentially pre-existing venvs Pex did not create.

pex/cli/commands/lock.py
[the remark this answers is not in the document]
benjyw: No need to re-word, just verifying my understanding.

pex/cli/commands/lock.py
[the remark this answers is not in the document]
benjyw: There are commonly-used distributions whose top-level package is >1 level deep. A noteworthy example of this is the google-cloud-* dists. They install under `site-packages/google/cloud/` where `site-packages/google/` has no files, just subdirs. 

I figure we should do whatever is consistent with the purpose of deleting the empty dirs to begin with. As far as I can tell an empty dir should have no effect (an implicit namespace package must contain some `.py` file, even if not an `__init__.py`), so I assume removing them is for hygiene/to avoid confusing people who poke around in the site-packages? In which case deleting up to the site-packages might make sense? 

But having said all that, I guess the real thing to do is whatever `pip uninstall` does in this case.  And from my experiments, it does what your code does, so no need to change anything...

pex/cli/commands/lock.py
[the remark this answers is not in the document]
benjyw: No, this is fine, I took a more intense look at the creation of `LockUpdateRequest` below, and this makes sense in my mind now.

benjyw (approved):

cburroughs:
Thank you!  This is great and I've been kicking the tires so to speak and try it out with our internal lockfile.  A few in progress comments so far.


## Conflicting output

While syncing an internal lock file I've gotten this apparently conflicting message:
```
Updates for lock generated by universal:
  There were no updates for mypy
```

But I've been unable to figure out a smaller reproduction case so far.

## Extras/Constraints

A bug where a requirement with extras resulted in an invalid tmp constraints file for the sync.

```
$ cat tmp/dhub.txt 
ansicolors==1.1.5
acryl-datahub[datahub-rest,feast,postgres,snowflake]==0.12.1.4
altair>=4.2.0,<5.2
$ cat tmp/dhub-c.txt 
awscli >= 1.30.0
boto3 >= 1.33.0
botocore >= 1.33.0
cachetools >= 5.3.0
click >= 8.1.0
coverage >= 7.3.0
dash >= 2.10.0
dash_bootstrap_components >= 1.4.0
flake8-pep585 >= 0.1.0
freezegun >= 1.3.0
graphene >= 3.0
holidays >= 0.30
ipython >= 8.0.0
kubernetes >= 28.0.0
loguru >= 0.6.0
matplotlib >= 3.7.0
pytest-mock >= 3.9.0
pytest-xdist >= 3.0.0
pytz >= 2023.3
requests >= 2.20.0
ruamel.yaml >= 0.17.0
s3fs >= 2023.1.0
setuptools >= 69.0.0
shap >= 0.40.0
xarray >= 2023.10.1
$ python3.10 -m pex.cli lock create '--indent=2'  --manylinux manylinux2014 --interpreter-constraint $'CPython==3.10.*' --style=universal --pip-version=23.3.2 --resolver-version pip-2020-resolver --target-system=linux --target-system=mac --prefer-binary  -r tmp/dhub.txt --constraints tmp/dhub-c.txt   -o tmp/dub.json
$ python3.10 -m pex.cli lock sync '--indent=2'  --manylinux manylinux2014 --interpreter-constraint $'CPython==3.10.*' --style=universal --pip-version=23.3.2 --resolver-version pip-2020-resolver --target-system=linux --target-system=mac --prefer-binary  -r tmp/dhub.txt --constraints tmp/dhub-c.txt  --lock tmp/dub.json
Encountered 1 error updating tmp/dub.json:
1.) cp310-cp310-manylinux_2_37_x86_64: /tmp/lock_update.32vgd7eh.constraints.txt line 1:
acryl-datahub[datahub-rest,feast,postgres,snowflake]==0.12.1.4

Constraint files do not support VCS, URL or local project requirementsand they do not support requirements with extras. Search for 'We are also changing our support for Constraints Files' here: https://pip.pypa.io/en/stable/user_guide/#changes-to-the-pip-dependency-resolver-in-20-3-2020.
```


## Interpreter Constraints and friends

I see 120fe0d23552863b34750866db7cc79376d0fffb but I'm not sure how these lock options are intended to be "sync-able" changes.  That is which can switch to "new" values while trying to hold the versions constant, or only only values that are subsets of the original range.

Naive example trying to "upgrade" Python versions with `--interpreter-constraints`.  (I'm not sure this *should* be a supported operation by `lock sync`, just that I don't know how someone looking at `--help` would know.)   In contrast, `--pip-version` can be changed on a new sync.

```
$ cat tmp/a-req.txt 
ansicolors==1.1.5
requests
watchfiles
$ python3.10 -m pex.cli lock create '--indent=2'  --manylinux manylinux2014 --interpreter-constraint $'CPython==3.10.*' --style=universal --pip-version=23.3.2 --resolver-version pip-2020-resolver --target-system=linux --target-system=mac --no-build --wheel -r  tmp/a-req.txt -o tmp/a.json
$ cat tmp/a-req.txt 
ansicolors==1.1.8
requests
watchfiles
$ python3.10 -m pex.cli lock sync '--indent=2'  --manylinux manylinux2014 --interpreter-constraint $'CPython==3.11.*' --style=universal --pip-version=23.3.2 --resolver-version pip-2020-resolver --target-system=linux --target-system=mac -r   tmp/a-req.txt --lock tmp/a.json 
Failed to resolve compatible artifacts from lock tmp/a.json for 1 target:
1. /usr/bin/python3.11:
    Failed to resolve all requirements for cp311-cp311-manylinux_2_37_x86_64 interpreter at /usr/bin/python3.11 from tmp/a.json:

Configured with:
    build: False
    use_wheel: True

Dependency on watchfiles not satisfied, 1 incompatible candidate found:
1.) watchfiles 0.21 (via: watchfiles) does not have any compatible artifacts:
    https://files.pythonhosted.org/packages/70/76/8d124e14cf51af4d6bba926c7473f253c6efd1539ba62577f079a2d71537/watchfiles-0.21.0-pp310-pypy310_pp73-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
    https://files.pythonhosted.org/packages/14/d0/662800e778ca20e7664dd5df57751aa79ef18b6abb92224b03c8c2e852a6/watchfiles-0.21.0-cp310-cp310-manylinux_2_17_s390x.manylinux2014_s390x.whl
    https://files.pythonhosted.org/packages/18/c4/ad5ad16cad900a29aaa792e0ed121ff70d76f74062b051661090d88c6dfd/watchfiles-0.21.0-cp310-cp310-manylinux_2_17_armv7l.manylinux2014_armv7l.whl
    https://files.pythonhosted.org/packages/37/17/4825999346f15d650f4c69093efa64fb040fbff4f706a20e8c4745f64070/watchfiles-0.21.0-pp310-pypy310_pp73-manylinux_2_17_aarch64.manylinux2014_aarch64.whl
    https://files.pythonhosted.org/packages/41/0e/3333b986b1889bb71f0e44b3fac0591824a679619b8b8ddd70ff8858edc4/watchfiles-0.21.0-cp310-cp310-manylinux_2_17_aarch64.manylinux2014_aarch64.whl
    https://files.pythonhosted.org/packages/4e/d2/769254ff04ba88ceb179a6e892606ac4da17338eb010e85ca7a9c3339234/watchfiles-0.21.0-cp310-cp310-manylinux_2_17_ppc64le.manylinux2014_ppc64le.whl
    https://files.pythonhosted.org/packages/5b/79/ecd0dfb04443a1900cd3952d7ea6493bf655c2db9a0d3736a5d98a15da39/watchfiles-0.21.0-cp310-cp310-manylinux_2_12_i686.manylinux2010_i686.whl
    https://files.pythonhosted.org/packages/62/66/7463ceb43eabc6deaa795c7969ff4d4fd938de54e655035483dfd1e97c84/watchfiles-0.21.0-pp310-pypy310_pp73-macosx_10_7_x86_64.whl
    https://files.pythonhosted.org/packages/6e/85/ea2a035b7d86bf0a29ee1c32bc2df8ad4da77e6602806e679d9735ff28cb/watchfiles-0.21.0-cp310-cp310-macosx_10_7_x86_64.whl
    https://files.pythonhosted.org/packages/92/ff/75cc1b30c5abcad13a2a72e75625ec619c7a393028a111d7d24dba578d5e/watchfiles-0.21.0-cp310-cp310-musllinux_1_1_aarch64.whl
    https://files.pythonhosted.org/packages/9a/65/12cbeb363bf220482a559c48107edfd87f09248f55e1ac315a36c2098a0f/watchfiles-0.21.0-cp310-cp310-musllinux_1_1_x86_64.whl
    https://files.pythonhosted.org/packages/b5/e5/240e5eb3ff0ee3da3b028ac5be2019c407bdd0f3fdb02bd75fdf3bd10aff/watchfiles-0.21.0-cp310-cp310-macosx_11_0_arm64.whl
    https://files.pythonhosted.org/packages/f7/4b/b90dcdc3bbaf3bb2db733e1beea2d01566b601c15fcf8e71dfcc8686c097/watchfiles-0.21.0-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
    https://files.pythonhosted.org/packages/fe/a3/42686af3a089f34aba35c39abac852869661938dae7025c1a0580dfe0fbf/watchfiles-0.21.0-pp310-pypy310_pp73-macosx_11_0_arm64.whl

```

jsirois:
@cburroughs for Extras/Constraints you did not provide the contents of `--constraints tmp/dhub-c.txt`. Can you please provide everything needed to repro?

jsirois:
Ok, I am going to proceed without support for lockfiles with multiple locked resolves, that case did prove too much to get working in a way I'm satisfied with yesterday.

pex/cli/commands/lock.py
```diff
@@ -1002,23 +1000,42 @@ def _create_lock_update_request(
...
+                    [LocalInterpreter.create(targets.interpreter)]
+                    if lock_file.style is LockStyle.UNIVERSAL
+                    else targets.unique_targets()
+                )
+                update_requests = [
+                    ResolveUpdateRequest(target=target, locked_resolve=locked_resolve)
+                    for target in update_targets
+                ]
+            else:
+                subset_result = try_(
```
jsirois, line 1022: N.B.: This is the old logic and it still applies for lock files with more than one locked resolve within; so these locks will still be subject to the more strict locked resolve selection rules and potentially fail to sync. I'll add a comment here pointing to an issue that tracks possibly handling those style locks better with a less strict, but still correct, scheme for selecting locked resolves to sync.

jsirois:
@cburroughs this is good to go afaict.
+ [ ] "Conflicting output": It would be good to improve this, but it seems pretty harmless. Without a repro case I'd like to push forward.
+ [ ] "Extras/Constraints": I did no fix here, but I also don't repro and the code that generates the tmp constraints file in the Pex lock update code derives the constraints from Pin objects and not any original requirements; so it seems impossible for that file to contain markers or extras.
+ [x] "Interpreter Constraints and friends"

pex/cli/commands/lock.py
jsirois: Comment added and filed https://github.com/pex-tool/pex/issues/2386

jsirois:
Ok, I looked more deeply at "Extras/Constraints" and although I cannot repro using that case, I do see a bug in the code by inspection. I'll add a failing test and fix.

jsirois:
Alright @cburroughs this should be truly good to hammer now. The only outstanding known issue is the cosmetic "updates: no updates" message.

tests/integration/cli/commands/test_lock_sync.py
```diff
@@ -1421,3 +1431,33 @@ def test_sync_universal_to_universal(tmpdir):
...
+        return {
+            locked_requirement.pin.project_name: locked_requirement
+            for locked_requirement in locked_resolves[0].locked_requirements
+        }
+
+    run_pex3("lock", "sync", "shiv==1.0.5", "--indent", "2", "--lock", lock).assert_success()
+    original_locked_requirements = collect_locked_requirements()
+    assert ProjectName("sphinx-click") not in original_locked_requirements
+
+    run_pex3("lock", "sync", "shiv[rtd]==1.0.5", "--indent", "2", "--lock", lock).assert_success()
```
jsirois, line 1457: Before the fix this failed with:
```
E       AssertionError: integration test failed: return_code=1, output=, error=Encountered 1 error updating /tmp/pytest-of-jsirois/pytest-9/test_sync_extras0/lock.json:
E       1.) cp312-cp312-manylinux_2_35_x86_64: /tmp/lock_update.zcz8gryb.constraints.txt line 4:
E       shiv[rtd]==1.0.5
E       Constraint files do not support VCS, URL or local project requirementsand they do not support requirements with extras. Search for 'We are also changing our support for Constraints Files' here: https://pip.pypa.io/en/stable/user_guide/#changes-to-the-pip-dependency-resolver-in-20-3-2020.
```

cburroughs:
> Alright @cburroughs this should be truly good to hammer now. The only outstanding known issue is the cosmetic "updates: no updates" message.

Thanks!  Have been swamped but will try to get my hammer out tomorrow.

cburroughs:
> Conflicting output

Still haven't figured out what triggers this

> Extras/Constraints

:+1: 


Sorry for my slowness.  My pattern here has been to hammer on a large internal lockfile, find something that errors out, and then tease out if that's a real thing that happens in public.   Usually that has involved a bunch of stumbling over internal packages, local PyPi registries, PEBCAK, or other things that don't generalize.  I think I've found a case where if the transitive dependencies are strict (`==`)  it interferes with the generated constraints.  That is in this example I think the user intent is reasonably clear ("change the version of this single package") but the `sync` has a conflict:

```
$ python3.10 -m pex.cli lock create '--indent=2'  --manylinux manylinux2014 --interpreter-constraint $'CPython==3.10.*' --style=universal --pip-version=23.3.2 --resolver-version pip-2020-resolver --target-system=linux --target-system=mac  'acryl-datahub[postgres,snowflake]==0.12.1.5' -o tmp6/lock.json
```

```
$ python3.10 -m pex.cli lock sync '--indent=2'  --manylinux manylinux2014 --interpreter-constraint $'CPython==3.10.*' --style=universal --pip-version=23.3.2 --resolver-version pip-2020-resolver --target-system=linux --target-system=mac  'acryl-datahub[postgres,snowflake]==0.12.1.4' --lock tmp6/lock.json
ERROR: Given the lock requirements:
acryl-datahub[postgres,snowflake]==0.12.1.4

The following lock update constraints could not all be satisfied:
acryl-datahub==0.12.1.4
acryl-datahub-classify==0.0.9
acryl-sqlglot==20.4.1.dev14
aiohttp==3.9.3
aiosignal==1.3.1
altair==4.2
anyio==3.7.1
appnope==0.1.4
argon2-cffi==23.1
argon2-cffi-bindings==21.2
asn1crypto==1.5.1
asttokens==2.4.1
async-timeout==4.0.3
attrs==23.2
avro==1.11.3
avro-gen3==0.7.11
beautifulsoup4==4.12.3
bleach==6.1
blis==0.7.11
cached-property==1.5.2
catalogue==2.0.10
certifi==2024.2.2
cffi==1.16
charset-normalizer==3.3.2
click==8.1.7
click-default-group==1.2.4
click-spinner==0.1.10
colorama==0.4.6
comm==0.2.2
confection==0.1.4
cryptography==42.0.5
cymem==2.0.8
debugpy==1.8.1
decorator==5.1.1
defusedxml==0.7.1
deprecated==1.2.14
docker==7
entrypoints==0.4
exceptiongroup==1.2
executing==2.0.1
expandvars==0.12
fastjsonschema==2.19.1
filelock==3.13.3
frozenlist==1.4.1
geoalchemy2==0.14.6
great-expectations==0.15.50
greenlet==3.0.3
humanfriendly==10
idna==3.6
ijson==3.2.3
importlib-metadata==7.1
ipaddress==1.0.23
ipykernel==6.17.1
ipython==8.21
ipython-genutils==0.2
ipywidgets==8.1.2
iso3166==2.1.1
jedi==0.19.1
jinja2==3.1.3
jsonpatch==1.33
jsonpointer==2.4
jsonref==1.1
jsonschema==4.21.1
jsonschema-specifications==2023.12.1
jupyter-client==7.4.9
jupyter-core==4.12
jupyter-server==1.24
jupyterlab-pygments==0.3
jupyterlab-widgets==3.0.10
langcodes==3.3
makefun==1.15.2
markupsafe==2.1.5
marshmallow==3.21.1
matplotlib-inline==0.1.6
mistune==3.0.2
mixpanel==4.10.1
msal==1.28
multidict==6.0.5
murmurhash==1.0.10
mypy-extensions==1
nbclassic==1
nbclient==0.6.3
nbconvert==7.16.3
nbformat==5.10.2
nest-asyncio==1.6
notebook==6.5.6
notebook-shim==0.2.4
numpy==1.26.4
packaging==24
pandas==2.2.1
pandocfilters==1.5.1
parso==0.8.3
pathlib-abc==0.1.1
pathy==0.11
pexpect==4.9
phonenumbers==8.13
platformdirs==3.11
preshed==3.0.9
progressbar2==4.4.2
prometheus-client==0.20
prompt-toolkit==3.0.43
psutil==5.9.8
psycopg2-binary==2.9.9
ptyprocess==0.7
pure-eval==0.2.2
py==1.11
pycountry==23.12.11
pycparser==2.21
pydantic==1.10.14
pygments==2.17.2
pyjwt==2.8
pyopenssl==24.1
pyparsing==3.1.2
python-dateutil==2.9.post0
python-stdnum==1.20
python-utils==3.8.2
pytz==2024.1
pyyaml==6.0.1
pyzmq==24.0.1
referencing==0.34
requests==2.31
requests-file==2
rpds-py==0.18
ruamel-yaml==0.17.17
schwifty==2024.1.1.post0
scipy==1.12
send2trash==1.8.2
sentry-sdk==1.43
setuptools==69.2
six==1.16
smart-open==6.4
sniffio==1.3.1
snowflake-connector-python==3.7.1
snowflake-sqlalchemy==1.5.1
sortedcontainers==2.4
soupsieve==2.5
spacy==3.5
spacy-legacy==3.0.12
spacy-loggers==1.0.5
sqlalchemy==1.4.52
sqlparse==0.4.4
srsly==2.4.8
stack-data==0.6.3
tabulate==0.9
termcolor==2.4
terminado==0.18.1
thinc==8.1.12
tinycss2==1.2.1
toml==0.10.2
tomlkit==0.12.4
toolz==0.12.1
tornado==6.4
tqdm==4.66.2
traitlets==5.2.1.post0
typer==0.7
typing-extensions==4.10
typing-inspect==0.9
tzdata==2024.1
tzlocal==5.2
urllib3==1.26.18
vininfo==1.8
wasabi==1.1.2
wcwidth==0.2.13
webencodings==0.5.1
websocket-client==1.7
widgetsnbextension==4.0.10
wrapt==1.16
yarl==1.9.4
zipp==3.18.1

Encountered 1 error updating tmp6/lock.json:
1.) cp310-cp310-manylinux_2_37_x86_64: pid 17028 -> /home/ecsb/.pex/venvs/5c1db138e499727676fa1d0ca449202de8f6f34d/5985ed09b49a653d6596b0e14d134c5456cf1a9f/bin/python -sE /home/ecsb/.pex/venvs/5c1db138e499727676fa1d0ca449202de8f6f34d/5985ed09b49a653d6596b0e14d134c5456cf1a9f/pex --disable-pip-version-check --no-python-version-warning --exists-action a --no-input --isolated -q --cache-dir /home/ecsb/.pex/pip/23.3.2/pip_cache --log /tmp/pex-pip-log.6yqis34i/pip.log download --dest /tmp/tmp_a25uxoz/usr.bin.python3.10 --constraint /tmp/lock_update.53w8bkah.constraints.txt acryl-datahub[postgres,snowflake]==0.12.1.4 --retries 5 --timeout 15 exited with 1 and STDERR:
ERROR: Cannot install acryl-datahub[postgres,snowflake]==0.12.1.4 because these package versions have conflicting dependencies.
ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
 
 The conflict is caused by:
     acryl-datahub[postgres,snowflake] 0.12.1.4 depends on spacy==3.4.3; extra == "snowflake"
     The user requested (constraint) spacy==3.5
 
 To fix this you could try to:
 1. loosen the range of package versions you've specified
 2. remove package versions to allow pip attempt to solve the dependency conflict
```

jsirois:
@cburroughs my next work stint runs the afternoon of 3/27 through 4/1. I'll take a look then.

jsirois:
Okay, @cburroughs that last behavior you mentioned is what you get. You can fix by following the reccomendation (and then once more when you get a new but similar conflict) to arrive at a successful:
```
:; python3.10 -m pex.cli lock sync --indent 2 --interpreter-constraint "CPython==3.10.*" --style universal --pip-version 23.3.2 --resolver-version pip-2020-resolver --target-system linux --target-system mac "acryl-datahub[postgres,snowflake]==0.12.1.4" spacy==3.4.3 "wasabi>=0.9.1,<1.1.0" --lock lock.json
Updates for lock generated by universal:
  Updated acryl-datahub from 0.12.1.5 to 0.12.1.4
  Updated spacy from 3.5 to 3.4.3
  Updated wasabi from 1.1.2 to 0.10.1
Updates to lock input requirements:
  Updated 'acryl-datahub[postgres,snowflake]==0.12.1.5' to 'acryl-datahub[postgres,snowflake]==0.12.1.4'
  Added 'spacy==3.4.3'
  Added 'wasabi<1.1.0,>=0.9.1'
```

The issue here is that acryl-datahub pins a whole heck of alot of deps (instead of, say, constraining to a non-breaking semver range):
```
:; unzip -qc ~/downloads/acryl_datahub-0.12.1.4-py3-none-any.whl acryl_datahub-0.12.1.4.dist-info/METADATA | grep Requires | grep -E "==[0-9]" | cut -d';' -f1 | sort -u
Requires-Dist: acryl-datahub-airflow-plugin ==0.12.1.4
Requires-Dist: acryl-datahub-classify ==0.0.9
Requires-Dist: acryl-pyhive[hive-pure-sasl] ==0.6.16
Requires-Dist: acryl-sqlglot ==20.4.1.dev14
Requires-Dist: avro-gen3 ==0.7.11
Requires-Dist: avro-gen3 ==0.7.11
Requires-Dist: black ==22.12.0
Requires-Dist: boto3-stubs[glue,s3,sagemaker,sts] ==1.28.15
Requires-Dist: elasticsearch ==7.13.4
Requires-Dist: flake8-bugbear ==23.3.12
Requires-Dist: google-cloud-datacatalog-lineage ==0.2.2
Requires-Dist: lark[regex] ==1.1.4
Requires-Dist: looker-sdk ==23.0.0
Requires-Dist: msal ==1.22.0
Requires-Dist: mypy ==1.0.0
Requires-Dist: mypy-boto3-sagemaker ==1.28.15
Requires-Dist: spacy ==3.4.3
Requires-Dist: sql-metadata ==2.2.2
Requires-Dist: sqllineage ==1.3.8
Requires-Dist: sqlparse ==0.4.4
Requires-Dist: types-click ==0.1.12
Requires-Dist: vertica-sqlalchemy-dialect[vertica-python] ==0.0.8.1
```

That sort of mass pinning will lead to issues when the pins change in another version given that Pex updates have these rules:
0. Pex is not a resolver, it uses Pip to do resolving; as such Pex cannot backtrack, etc. Pip does all that opaquely.
1. The `pex3 lock {update,sync}` commands attempt to do a minimal update by masking off unchanged portions of the resolve (as can be divined from the input requirements alone anyhow) with constraints pinning.

Afaict, to have these sorts of situations resolve automatically would either require Pex tries, fails, retries switching constraints pins to semver range pins? But that would lead to more disruptive lock updates growing akin to re-running `pex3 lock create` - which I think is pointedly the thing you're trying to avoid / fix.

So, how do you wish to proceed? Afaict Pex cannot do better than this without becoming a resolver itself which is not in the cards in Pex 2.x. The user has to be able to read the Pip conflict output and do what I did above or else Pants needs to have an escape hatch to re-run `pex3 lock create` for locks that are overly constrained preventing update. Maybe though there is some clever strategy that uses constraints ranging instead of pinning to cut down on lock disturbance but still allow updates.

jsirois:
@cburroughs I may move forward with this iteration this afternoon. I'm getting uncomfortable with the number of changes queued behind a release and it seems like Pants is at liberty to skip this initial release of `pex3 lock sync` if you find improvements or bugs that we can follow up with.

cburroughs:
(Yeah some of these large packages with narrow pins are challenging to deal with in a few different ways.)

The mental model of "Pex is not a resolver and thus can't 'backtrack'" makes sense to me.  I'll have to think about how to present that through Pants in a way that's helpful without getting in the users way, but that was the essential design problem there anyway.

Thanks for putting this all together.

cburroughs (approved):

jsirois:
> I'll have to think about how to present that through Pants in a way that's helpful without getting in the users way

One thing to point out is `rm lock.json && pex3 lock sync ...` reverts to the old Pants behavior of re-locking every time; so, at worst, if a user can't figure out Pip conflict messages and Pants gives some guidance, you can always work your way out of an overly constrained lock by re-creating it this way even if `pex3 lock sync ...` is the only command Pants uses behind the scenes.

PR #2383, opened 2024-03-15
Fix Pex to handle venvs with multiple site-packages dirs.

jsirois:
Previously the Pex venv code would blow up when it encountered a venv
with multiple distinct site-packages directories. Although this was
safe, it was unhelpful for virtualenvs created by modern Virtualenv on
Red Hat based distributions where there is a separation between
purelib (`lib/pythonX.Y/site-packages`) and
platlib (`lib64/pythonX.Y/site-packages`). Interestingly this was not a
problem for Python stdlib `-mvenv` virtualenvs since those create the
`lib64/` platlib directory as a symlink to the `lib/` purelib
directory. It's unclear who is constructing venvs correctly here, but
given the Virtualenv layout, its definitely correct to honor the
purelib / platlib distinction when installing wheels in venvs. Pex now
does this and tests are added to confirm.

Also fix up various existing tests that failed on ARM machines where
this problem was discovered.

### DIFF
--- a/pex/pep_376.py
+++ b/pex/pep_376.py
@@ -19,6 +19,7 @@ from pex.interpreter import PythonInterpreter
 from pex.typing import TYPE_CHECKING, cast
 from pex.util import CacheHelper
 from pex.venv.virtualenv import Virtualenv
+from pex.wheel import WHEEL
 
 if TYPE_CHECKING:
     from typing import Callable, Iterable, Iterator, Optional, Protocol, Text, Tuple, Union
@@ -168,6 +169,7 @@ class InstalledWheel(object):
         prefix_dir,  # type: str
         stash_dir,  # type: str
         record_relpath,  # type: Text
+        root_is_purelib,  # type: bool
     ):
         # type: (...) -> InstalledWheel
 
@@ -179,6 +181,7 @@ class InstalledWheel(object):
             "stash_dir": stash_dir,
             "record_relpath": record_relpath,
             "fingerprint": fingerprint,
+            "root_is_purelib": root_is_purelib,
         }
         with open(cls.layout_file(prefix_dir), "w") as fp:
             json.dump(layout, fp, sort_keys=True)
@@ -187,6 +190,7 @@ class InstalledWheel(object):
             stash_dir=stash_dir,
             record_relpath=record_relpath,
             fingerprint=fingerprint,
+            root_is_purelib=root_is_purelib,
         )
 
     @classmethod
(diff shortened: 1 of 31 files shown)


tests/integration/test_excludes.py
```diff
@@ -1,144 +1,147 @@
...
+    ).assert_success()
+
+    assert ProjectName("certifi") not in frozenset(
+        dist.metadata.project_name for dist in PEX(pex).resolve()
+    )
+
+    return pex
+
+
+REQUESTS_CMD = [
```
jsirois, line 51: This REQUESTS_CMD is the only real change here. I also belatedly ran dos2unix to fix line endings which makes the GitHub diff look worse than it is.

pex/wheel.py
```diff
@@ -0,0 +1,61 @@
...
+
+@attr.s(frozen=True)
+class WHEEL(object):
+    """The .whl WHEEL metadata file.
+
+    See item 6 here for the WHEEL file contents: https://peps.python.org/pep-0427/#file-contents
+    """
+
+    class LoadError(Exception):
+        """Indicates an error loading WHEEL metadata."""
```
benjyw, line 27: It took some focus to distinguish `pex.pep_376.LoadError`, `pex.pep_427.WheelLoadError` and this new `WHEEL.LoadError`, especially when using two of them in the same file (as in pep_376.py, pep_427.py and installer.py).

AFAICT `pex.pep_376.LoadError` relates to loading the layout file for an InstalledWheel, which admittedly is slightly different, but the other two seem to mean the same thing (error loading the WHEEL metadata)?

So WDYT about consolidating at least the last two, and possibly distinguishing the other via naming, for example: 

`pex.pep_376.LoadError` -> `InstalledWheelLoadError`

testing/docker.py
```diff
@@ -0,0 +1,103 @@
...
+    is_exe(os.path.join(entry, "docker"))
+    for entry in os.environ.get("PATH", os.path.defpath).split(os.pathsep)
+)
+
+skip_unless_docker = pytest.mark.skipif(not HAVE_DOCKER, reason="This test needs docker to run.")
+
+
+@attr.s(frozen=True)
+class DockerVirtualenvRunner(object):
+    """Run's code in a venv created by Virtualenv at /virtualenv.venv in a docker container."""
```
benjyw, line 34:
```suggestion
    """Runs code in a venv created by Virtualenv at /virtualenv.venv in a docker container."""
```

benjyw (approved):

testing/docker.py
jsirois: Fixed.

pex/pep_376.py
```diff
@@ -219,17 +223,32 @@ def load(cls, prefix_dir):
 
         fingerprint = layout.get("fingerprint")
 
+        # N.B.: Caching root_is_purelib was not part of the original InstalledWheel layout data; so
+        # we materialize the property if needed to support older installed wheel chroots.
+        root_is_purelib = layout.get("root_is_purelib")
+        if type(root_is_purelib) is not bool:
```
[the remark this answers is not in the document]
jsirois: Yeah, not sure how I git there. Fixed.

pex/wheel.py
jsirois: I went with `InstalledWheel.LoadError` for the one and moved `pex.pep_427.Wheel` to `pex.wheel.Wheel` and consolidated with `pex.wheel.WheelMetadataLoadError` now that `Wheel` and `WHEEL` are in the same module.

benjyw (approved):
Thanks! That structure is much more clear.

Issue #2391, opened 2024-03-24
 `--no-pre-install-wheels --requirements-pex .. --layout packed` gives `[Errno 21] Is a directory: '.../.deps/....whl'` errors


huonw:
It seems pex emits "Is a directory" errors referring to wheels for invocations that use all three of `--requirements-pex ... --no-pre-install-wheels --layout packed`.

For instance:

```shell
curl -OLs https://github.com/pex-tool/pex/releases/download/v2.2.2/pex
chmod +x ./pex

./pex cowsay==6.1 -o reqs.pex
./pex --layout packed --no-pre-install-wheels --requirements-pex reqs.pex -o output.pex
```

Output:

```
[Errno 21] Is a directory: '/private/var/folders/sv/vd266m4d4lvctgs2wpnhjs9w0000gn/T/tmp4032w10a/.deps/cowsay-6.1-py3-none-any.whl'
```

<details><summary>Output with PEX_VERBOSE=1 on second invocation</summary>

```
pex: Laying out PEX zipfile /Users/huon/tmp/pex-loose-no-pre-install/pex: 0.1ms
pex: Executing installed PEX for /Users/huon/tmp/pex-loose-no-pre-install/./pex at /Users/huon/.pex/unzipped_pexes/ad5d02c39f095829e9578f565f825d0835bfcd73
pex:   Testing /Users/huon/.pyenv/versions/3.10.4/bin/python3.10 can resolve PEX at /Users/huon/.pex/unzipped_pexes/ad5d02c39f095829e9578f565f825d0835bfcd73: 2.8ms
pex: Using the current interpreter /Users/huon/.pyenv/versions/3.10.4/bin/python since it matches constraints and PYTHONPATH is not set.
pex: Discarding site packages path: /Users/huon/.pyenv/versions/3.10.4/lib/python3.10/site-packages
pex: Tainted path element: /Users/huon/.pyenv/versions/3.10.4/lib/python3.10/site-packages
pex: Scrubbing from user site: /Users/huon/.local/lib/python3.10/site-packages
pex: Scrubbing from site-packages: /Users/huon/.pyenv/versions/3.10.4/lib/python3.10/site-packages
[...]

Issue #2299, opened 2023-12-13
Support turning installed wheel chroots into `.whl` files with full fidelity.


jsirois:
If we can round trip from a downloaded `.whl` to an installed wheel chroot and back with full fidelity, then we can fix a long existing latent bug and improve composability of Pex options:
+ Unreported bug: `PEX_TOOLS=1 ./a.pex repository extract ...` does not produce correct wheels when the original wheel contains any data dirs (`<project>-<version>.data/*` `.whl` zip file entries).
+ `--pex-repository`: via #2298 a PEX with dependency `.whl` files can satisfy `.whl` file resolves and installed wheel chroot resolve but a PEX dependency with installed wheel chroot dependencies can only satisy resolves for installed wheel chroot dependencies and not for `.whl` files.

PR #2392, opened 2024-03-24
Guard against mismatched `--requirements-pex`.

jsirois:
Pex cannot yet handle turning installed wheel chroots back into wheel
files (see #2299). As such, until that time, it must reject attempts
to combine PEXes with different wheel packaging.

Clarifies #2391

### DIFF
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -892,28 +892,9 @@ def build_pex(
         "Adding distributions from pexes: {}".format(" ".join(options.requirements_pexes))
     ):
         for requirements_pex in options.requirements_pexes:
-            if (
-                pex_info.deps_are_wheel_files
-                != PexInfo.from_pex(requirements_pex).deps_are_wheel_files
-            ):
-                die(
-                    "The {option} option was selected but the --requirements-pex "
-                    "{requirements_pex} is built with {opposite_option}. Any --requirements-pex "
-                    "you want to merge into the main PEX must be built with {option}.".format(
-                        option=(
-                            "--no-pre-install-wheels"
-                            if pex_info.deps_are_wheel_files
-                            else "--pre-install-wheels"
-                        ),
-                        requirements_pex=requirements_pex,
-                        opposite_option=(
-                            "--pre-install-wheels"
-                            if pex_info.deps_are_wheel_files
-                            else "--no-pre-install-wheels"
-                        ),
-                    )
-                )
-            requirements_pex_info = dependency_manager.add_from_pex(requirements_pex)
+            requirements_pex_info = dependency_manager.add_from_pex(
+                requirements_pex, result_type_wheel_file=pex_info.deps_are_wheel_files
+            )
             excluded.extend(requirements_pex_info.excluded)
 
     with TRACER.timed(
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -892,6 +892,27 @@ def build_pex(
         "Adding distributions from pexes: {}".format(" ".join(options.requirements_pexes))
     ):
         for requirements_pex in options.requirements_pexes:
+            if (
+                pex_info.deps_are_wheel_files
+                != PexInfo.from_pex(requirements_pex).deps_are_wheel_files
+            ):
+                die(
+                    "The {option} option was selected but the --requirements-pex "
+                    "{requirements_pex} is built with {opposite_option}. Any --requirements-pex "
+                    "you want to merge into the main PEX must be built with {option}.".format(
+                        option=(
+                            "--no-pre-install-wheels"
+                            if pex_info.deps_are_wheel_files
+                            else "--pre-install-wheels"
+                        ),
+                        requirements_pex=requirements_pex,
+                        opposite_option=(
+                            "--pre-install-wheels"
+                            if pex_info.deps_are_wheel_files
+                            else "--no-pre-install-wheels"
+                        ),
+                    )
+                )
             requirements_pex_info = dependency_manager.add_from_pex(requirements_pex)
             excluded.extend(requirements_pex_info.excluded)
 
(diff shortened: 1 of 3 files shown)


pex/bin/pex.py
```diff
@@ -892,6 +892,27 @@ def build_pex(
         "Adding distributions from pexes: {}".format(" ".join(options.requirements_pexes))
     ):
         for requirements_pex in options.requirements_pexes:
```
jsirois, line 894: Can the Pants-aware reviewers comment on why this option is used by Pants at all? Its an odd duck option in the PEX cli and smells like a hack for Pants sake, but I can't recall exactly why this was deemed necessary. Since PEX_PATH nets you the same thing, but not as a single file, I suspect this must have to do with creating a single file PEX more quickly than it might otherwise be created? I'd love to know a bit more to help evaluate if this can be axed for Pex 3.

benjyw (approved):

pex/bin/pex.py
benjyw: Pants uses this only to support "local dists" - which is how it supports building (unversioned, unpublished) local native code into wheels (via setup.py) and then packaging those into the final .pex file along with local Python code. 

To [quote you](https://github.com/pantsbuild/pants/issues/13587#issuecomment-968160031): "The local PEX is a hack and can be "broken" since its never exposed to or used by end users. It's just there to emulate building a wheel for each local dist and then adding that wheel to another PEX."

So some alternative way to add unpublished wheels directly to a pex would be fine. There is more context in [that issue](https://github.com/pantsbuild/pants/issues/13587#issuecomment-969088777) about the perf reasons that led to the current solution, and you have suggested an alternative there that Pants could adopt and is more robust, it just takes some work.

jsirois: Ok, thanks for the pointers. I'll have to re-read. Clearly though PEX already supports this, you `-f dir/` for the dir containing the unpublished wheel and just do a normal resolve, lock, what have you.

pex/bin/pex.py
```diff
@@ -892,6 +892,27 @@ def build_pex(
...
+            if (
+                pex_info.deps_are_wheel_files
+                != PexInfo.from_pex(requirements_pex).deps_are_wheel_files
+            ):
+                die(
+                    "The {option} option was selected but the --requirements-pex "
+                    "{requirements_pex} is built with {opposite_option}. Any --requirements-pex "
+                    "you want to merge into the main PEX must be built with {option}.".format(
```
huonw, line 902: Hm, I'm a bit confused, as a test like the following seems to have matching options but still fails with the same "Is a directory" error:

```shell
pex --no-pre-install-wheels -o /tmp/first.pex cowsay
pex --no-pre-install-wheels -o /tmp/second.pex --requirements-pex /tmp/first.pex --layout packed
```

This error message suggests this should work. Am I misunderstanding it?

jsirois: It works in the test in this PR (minus packed, your grid made layout appear irrelevant).

jsirois: Aha, I read poorly. Back to drawing board.

jsirois: @huonw I don't repro:
```
:; ./dtox.sh inspect
jsirois@617430d8f0ac:/development/pex$ python -mpex --no-pre-install-wheels -o /tmp/first.pex cowsay
jsirois@617430d8f0ac:/development/pex$ python -mpex --no-pre-install-wheels -o /tmp/second.pex --requirements-pex /tmp/first.pex --layout packed
jsirois@617430d8f0ac:/development/pex$ PEX_SCRIPT=cowsay /tmp/second.pex/__main__.py -t Moo!
  ____
| Moo! |
  ====
    \
     \
       ^__^
       (oo)\_______
       (__)\       )\/\
           ||----w |
           ||     ||
jsirois@617430d8f0ac:/development/pex$
exit
```

Let me dig a bit more to see if I can gedanken whatever it is you're hitting.

pex/bin/pex.py
jsirois: Yeah @benjyw, so after re-reading, Pants gets itself into trouble by not performing a normal resolve. If it's at all possible to:
1. Build the wheel
2. Do the resolve with either `pex -f find-links/ ...` or else `pex {normal reqs...} {wheel1} {wheel2} ... {wheeln) ...`.

That is the sane thing to do, IOW treat the wheels Pants builds like any other with no PEX hacks that attempt to stitch together resolves post-facto. In general, and this includes PEX_PATH, any attempt to cobble together a sys.path outside of running a single unified resolve is asking for all sorts of trouble as a user and all sorts of complication for Pex to support since Pex is not a resolver, it just uses one.

pex/bin/pex.py
jsirois: Yeah @huonw I've got nothing. The new matrixed test passes locally and in CI as did the by-hand replication above; so I'm not sure what is different about our environments that I'm missing. I think this PR is forward progress though and, afaict, introduces no new bugs, just whittling down the existing buggy cases at worst; so I'll proceed. If you can find maybe a dockerable repro case (I used `./dtox.sh inspect` above for this reason - so we can share ~the same env) then I can address in a follow up.

jsirois: Aha, ok - Mac specific issue: https://github.com/pex-tool/pex/actions/runs/8457687342/job/23170224480?pr=2392#step:9:2348

Pain in the butt, but at least I have a repro.

huonw: Ah, sorry, I didn't cross my mind that it might be platform-specific. I'll have some time over the next few days so can help fix if doing loops through CI is too unpleasant.

jsirois: Ok, the issue is systemic to `--requirements-pex X --no-pre-install-wheels`; it's just masked on Linux. I'll add a fix patch to this PR tomorrow.

PR #2398, opened 2024-04-19
Upgrade Python 3.13 to alpha 6.

jsirois:
Also upgrade all other pythons to latest point releases.


huonw (approved):

jsirois:
Ok, the unit test failure is due to https://github.com/python/cpython/issues/118107. I'll be fixing that for the 1st 3.13 CPython beta cut, but for now I'll add a Pex patch to handle 3.13.0a6 broken-ness here.

PR #2403, opened 2024-05-08
Update Hatchling build backend

ofek:
Sorry for the trouble back then, 1.22 was a big release 😅

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,14 +5,7 @@
 # entry point metadata so the hook is discovered.
 backend-path = ["build-backend"]
 build-backend = "pex_build.hatchling.build"
-
-# The hatchling 1.22.0 release breaks our build in several ways:
-# + Our console scripts are no longer generated.
-# + The `project.readme` pyproject.toml key now has its value hydrated before our metadata plugin
-#   runs, which foils our metadata expansion since our README.rst contains inadvertant expansion
-#   tokens.
-# This pin low buys time to work through the issues.
-requires = ["hatchling<1.22.0"]
+requires = ["hatchling>=1.22.5"]
 
 [tool.hatch.metadata.hooks.pex-adjust-metadata]
 expand = {"pex_version" = "version"}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@
 # entry point metadata so the hook is discovered.
 backend-path = ["build-backend"]
 build-backend = "pex_build.hatchling.build"
-requires = ["hatchling>=1.22.5"]
+requires = ["hatchling"]
 
 [tool.hatch.metadata.hooks.pex-adjust-metadata]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -5,14 +5,7 @@
...
 build-backend = "pex_build.hatchling.build"
-
-# The hatchling 1.22.0 release breaks our build in several ways:
-# + Our console scripts are no longer generated.
-# + The `project.readme` pyproject.toml key now has its value hydrated before our metadata plugin
-#   runs, which foils our metadata expansion since our README.rst contains inadvertant expansion
-#   tokens.
-# This pin low buys time to work through the issues.
-requires = ["hatchling<1.22.0"]
+requires = ["hatchling>=1.22.5"]
```
ofek, line 8:
```suggestion
requires = ["hatchling"]
```

pyproject.toml
```diff
@@ -15,8 +15,7 @@ expand = {"pex_version" = "version"}
 
 [project]
 name = "pex"
-dynamic = ["version"]
-requires-python = ">=2.7,<3.13,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*"
+dynamic = ["version", "requires-python"]
```
jsirois, line 18: If this is now required I'm not a fan. The intent is to record the real requires-python here in the static metadata and only override in CI tests. Moving the real metadata to the metadata_hook obscures a bit too much for my liking.

ofek: The issue is that the spec says that tools must adhere to what is defined statically (e.g. they cannot be changed) and Hatchling had a bug that it was not enforcing that.

If you would prefer I can add an option to enable that unsafe behavior but it would be bad for the ecosystem. What do you think?

FYI here is a comment that explains a little bit https://discuss.python.org/t/respecting-core-metadata-2-2-when-building-from-source-distributions/48886/60

It's not exactly related but the bug was revealed during that discussion/thread.

jsirois: Ok, I think I grok the thread you pointed me to. IIUC the criux lies in PEP-621 under https://peps.python.org/pep-0621/#dynamic:

> * Build back-ends MUST raise an error if the metadata specifies a field statically as well as being listed in dynamic.

If that were, instead:
```
* When a field is specified statically as well as marked dynamic, its value must be obtained from the backend dynamic generation process.
```

All would be well.

@ofek I'll send up a PR and add you as a reviewer that sortof acheives the end I want here using non-canonical fields in pyproject.toml to carry the static metadata / template metadata. That PR may contain comments about the PyPA that you want no association with; so feel free to not comment if you're uncomfortable!

Thanks for hunting down users affected by your rocky release period though, I really appreciate it.

ofek:
Superseded by https://github.com/pex-tool/pex/pull/2408

PR #2407, opened 2024-05-10
Upgrade Python 3.13 to beta 1.

jsirois:


docker/base/install_pythons.sh
```diff
@@ -17,7 +17,7 @@ PYENV_VERSIONS=(
   3.9.19
   3.10.14
   3.12.3
-  3.13.0a6
+  3.13.0b1
```
jsirois, line 20: An item of note: this is the 1st Python release to support ZIP64 in zipimporter: aka Python 3.13 PEXes can now contain huge ML apps without `--layout {loose,packed}` shenanigans being necessary.

zmanji (approved):

tests/integration/test_issue_2355.py
```diff
@@ -37,6 +40,16 @@ def test_ssl_context(
                 ]] && \
                 tar -xzf $PBS_ARCHIVE
                 """
+            ).format(
+                extra_instructions="\n".join(
+                    # Git is needed for the git VCS URL the patched Pip needed for Python 3.13
+                    # pre-releases is resolved by.
+                    # TODO(John sirois): Remove once a Pip with Python 3.13 support is released:
+                    #   https://github.com/pex-tool/pex/issues/2406
```
huonw, line 48: I happened to notice that pip has done a 24.1b1 release which seems to include https://github.com/pypa/pip/pull/12462. Doesn't have to be this PR, but could PEX switch over to using that one for validating 3.13 rather than the git version?

huonw (approved):

tests/integration/test_issue_2355.py
jsirois: Perhaps, but it would, like the git version, be temporary and never public. I don't ever intend to add anything but production Pip versions to the `--pip-version` enum.

PR #2409, opened 2024-05-15
Fix `--exclude`.

### LINKED ISSUE #455: Building a PEX for multiple platforms doesn't affect `sys.platform` in `extras_require`
I'm trying to build a cross-platform (Mac and Linux) PEX that contains the [`doit` library](https://github.com/pydoit/doit), but it seems as if `pex` is not doing anything to make the target platform be respected by [the usages of `sys.platform` in the conditions of the `extras_require` dictionary in its `setup.py` file](https://github.com/pydoit/doit/blob/088d2ba528d65d4cf265035928b64a33eaaa1e63/setup.py#L77-L78).

The following command results in the failure I'm running into on my Mac (running 10.12.6):

```sh
➜  pex -v --disable-cache --python=python3 doit --platform macosx_10.12-x86_64 --platform linux-x86_64 -o doit.pex
  doit 0.31.1 pex :: Resolving distributions :: Packaging MacFSEvents
  cloudpickle 0.5.2
  MacFSEvents 0.8.1
pex: Target package WheelPackage('file:///private/var/folders/b7/qgjhz8kx5fvc6kz_fnx55b600000gq/T/tmpt_08qmf1/MacFSEvents-0.8.1-cp36-cp36m-macosx_10_12_x86_64.whl') is not compatible with CPython-3.6.4 / linux-x86_64
  File "/usr/local/bin/pex", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.6/site-packages/pex/bin/pex.py", line 663, in main
  File "/usr/local/lib/python3.6/site-packages/pex/bin/pex.py", line 601, in build_pex
    for dist in resolveds:
  File "/usr/local/lib/python3.6/site-packages/pex/resolver.py", line 438, in resolve_multi
    allow_prereleases):
  File "/usr/local/lib/python3.6/site-packages/pex/resolver.py", line 376, in resolve
    return resolver.resolve(resolvables_from_iterable(requirements, builder))
  File "/usr/local/lib/python3.6/site-packages/pex/resolver.py", line 209, in resolve
  File "/usr/local/lib/python3.6/site-packages/pex/resolver.py", line 177, in build
pex.resolver.Untranslateable: Package SourcePackage('https://pypi.python.org/packages/28/2e/1ff399cfd2a6a8ebb65152203c920643c2169aa507b2e96559d19baeb7c3/MacFSEvents-0.8.1.tar.gz#md5=45759553d58bab6f7c8d6187fb0fe12f') is not translateable by ChainedTranslator(WheelTranslator, EggTranslator, SourceTranslator)
```

It appears as if `pex` is succeeding in building the part of the PEX for `macosx_10.12-x86_64` (makes sense since that's my host OS), but failing on the part for `linux-x86_64`.

Shouldn't `pex` not try to use `MacFSEvents` for the `linux-x86_64` platform since the choice between `MacFSEvents` and `pyinotify` is guarded by checking `sys.platform` in the `extras_require` dictionary?

I also don't actually need either `MacFSEvents` nor `pyinotify` for my purposes, so alternatively is there a way to tell `pex` to ignore the `extras_require` dictionary?

If instead this is something that should be fixed/adjusted in the `doit` library, what do you recommend I change in a potential pull request for that library?

### LINKED ISSUE #2097: Option to exclude dependencies from pex file

jsirois:
Previously, `--exclude`s were processed after all resolution was 
complete. Not only was this sub-optimal, incurring more resolution work
(downloads, backtracks, etc.), but it also could lead to failed
interactions (metadata extraction & wheel builds) with excluded
distributions.

Fix this by plumbing `--exclude`s all the way through the Pip resolve
process. Additionally, fix the existing plumbing through the two other
resolve processes: PEX repository resolution and lock file resolution.

Fixes #455
Fixes #2097


jsirois:
This is on the big side, but it is broken up into functioning independent commits. Thanks in advance for your attention.

I will note this opens up `--override` since the deep `--exclude` plumbing already tricks Pip into following doctored requirements lists; it's just that the only doctoring right now is removal of requirements from the list and not editing of requirements in the list. No one has asked for this yet, so I'll save that for a follow up.

jsirois:
This should be all green now.

cc:@cognifloyd who inspired going a bit further and implementing deep excludes.

pex/pip/tool.py
```diff
@@ -358,7 +358,7 @@ def _spawn_pip_isolated(
         with ENV.strip().patch(
             PEX_ROOT=ENV.PEX_ROOT,
             PEX_VERBOSE=str(ENV.PEX_VERBOSE),
-            __PEX_UNVENDORED__="1",
+            __PEX_UNVENDORED__="setuptools",
```
jsirois, line 361: The issue before this change was the un-vendoring was global and just above in `pex/pip/download_observer.py`, `pex` was newly exposed to Pip to support patches that use Pex code (the `--exclude` patches leverage this). That `pex` code, including everything under `pex.vendor._vendored` was being exposed which broke `pex.third_party` vendored imports for Python2.7 where the meta-path importer has different precedence than under newer Pythons. Since Pex itself never imports setuptools, exposing just that serves both the Pex patches of Pip (which don;t need pex.third_party.setuptools) and vendored Pip itself (which needs setuptools to build sdists). Ideally Pex would not re-write setuptools imports at all (since it never does a pex.third_party import of the code), but Lambdex _does_ and this change would break Lambdex. A test caught this and serves as a backstop to not break Lambdex users.

tests/integration/test_excludes.py
```diff
@@ -145,3 +160,204 @@ def test_requirements_pex_exclude(
...
+            dedent(
+                """\
+                import os
+                import sys
+
+                from setuptools import setup
+
+
+                if "BEHAVE" not in os.environ:
+                    sys.exit("I'm an evil package.")
```
cognifloyd, line 268: 🤣

This is so true. Packages that put `sys.exit` in `setup.py` are EVIL!

tests/integration/test_issue_455.py
```diff
@@ -0,0 +1,76 @@
...
+            "latest",
+            "--platform",
+            "macosx_10.12-x86_64-cp-310-cp310",
+            "--platform",
+            "linux-x86_64-cp-310-cp310",
+            "doit==0.34.2",
+            "--exclude",
+            "MacFSEvents",
+            "--exclude",
+            "pyinotify",
```
cognifloyd, line 39: Ooh. Nice! This looks great! Blasted evil pyinotify with its `sys.exit` ...

cognifloyd (commented):
LGTM

I read through each of the commits. Impressive! 

This should reduce the pain caused by evil packages by providing an elegant escape hatch. Thank you for figuring out how to plumb excludes into pip.

pex/specifier_sets.py
```diff
@@ -0,0 +1,331 @@
...
+        parts.append("".join(str(x) for x in parsed_version.pre))
+    if parsed_version.post is not None:
+        parts.append(".post")
+        parts.append(str(parsed_version.post))
+    if parsed_version.dev is not None:
+        parts.append(".dev")
+        parts.append(str(parsed_version.dev))
+    if parsed_version.local is not None:
+        parts.append("+")
+        parts.append(parsed_version.local)
```
kaos, line 192: At least for the compatibility operator `~=`, the suffix is ignored so I imagine it would be more correct to drop these for the ceiling version. I wonder if the same is not true also for the other operators?
e.g. from the packaging version specifiers docs example, the ceiling equiv. version drops these:
```
~= 2.2.post3
>= 2.2.post3, == 2.*

~= 1.4.5a4
>= 1.4.5a4, == 1.4.*
```

pex/specifier_sets.py
```diff
@@ -0,0 +1,331 @@
...
+                    upper = new_upper
+                    excludes.remove(exclude)
+
+    # Only retain excludes whose range resides inside the final fully narrowed bounds.
+    for exclude in tuple(excludes):
+        if upper:
+            if exclude.lower.version > upper.version or (
+                exclude.lower.version == upper.version
+                and exclude.lower.inclusive
+                and upper.inclusive
```
kaos, line 302: shouldn't this inclusive check be inversed? seems we're dropping an exclude range that effectively excludes the upper bound here..

pex/specifier_sets.py
```diff
@@ -0,0 +1,331 @@
...
+            ):
+                excludes.remove(exclude)
+                continue
+        if lower:
+            if exclude.upper.version < lower.version or (
+                exclude.upper.version == lower.version and lower.inclusive
```
kaos, line 308: why is not `exclude.upper.inclusive` consulted? (or should it be `and not lower.inclusive`?)

kaos (commented):
Reviewed first commit on specifier sets.

tests/integration/test_excludes.py
jsirois: Well, I felt a bit bad about this. It's really Python that is evil here. The pyinotify project is ~17 years old and, IIRC there was no way at that time (just like today!) for an sdist to do anything but fail when you try to compile it on a platform it doesn't support. The sys.exit in the position pyinotify puts it is aggressive, but with out it the build will just fail later anyhow.

The real problem then, is the distutils / setuptools build system here. Since those require evaluating Python code (in greater degrees further back in time) to learn just project metadata, aggressive tends towards evil - the project sdist acts as a hand grenade for the only curious.

Setuptools grew setup.cfg eventually to become up to ~100% declarative and then later Python people (but crucially not Python!) woke up and forked PyPA and eventually produced PEP 517/518 and now setuptools even supports pyproject.toml declarative metadata.

I can't describe the brokenness of a language not having its story straight better than Indy Greg or the related discussion, or, the icing on the cake, the ensuing inaction:
+ https://gregoryszorc.com/blog/2023/10/30/my-user-experience-porting-off-setup.py/
+ https://discuss.python.org/t/user-experience-with-porting-off-setup-py/37502/1

It's true it can be hard for an OSS project to muster the resources to tackle a big project, but Python has some serious financial backers and it could decide to make this a priority and get the money and people-time to clean this up in, say, a 5 year period. Instead though its floundering and leaving a confusing landscape in its wake for users.

pex/specifier_sets.py
```diff
@@ -0,0 +1,331 @@
...
+    excludes = attr.ib(default=())  # type: Tuple[ExcludedRange, ...]
+
+    def __contains__(self, other):
+        # type: (Any) -> bool
+        if Range is not type(other):
+            return NotImplemented
+        return self.contains(other)
+
+    def contains(self, other):
+        # type: (Range) -> bool
```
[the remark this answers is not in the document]
jsirois: If anything I'd bias opposite. In the bespoke functions (`contains` and `intersect`), I leverage type annotations and MyPy. As a matter of keeping my own sanity, I never double-check type-checking with runtime assertions. It may be the case that I could type `__contains__` to only accept other `Range`s, but that's against the spirit of the `__contains__` interface, which is supposed to work on `Any`.

pex/specifier_sets.py
jsirois: Thanks - that's definitely a bug.

pex/specifier_sets.py
```diff
@@ -0,0 +1,331 @@
...
+    # type: (SpecifierSet) -> Iterator[Union[ArbitraryEquality, ExcludedRange, LowerBound, UpperBound]]
+
+    for spec in specifier_set:
+        if "===" == spec.operator:
+            yield ArbitraryEquality(spec.version)
+            continue
+
+        if spec.version.endswith(".*"):
+            if "~=" == spec.operator:
+                raise ValueError()
```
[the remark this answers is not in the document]
jsirois: Yup, thanks. Will fix.

pex/specifier_sets.py
jsirois: This and the following comment are hard to parse (as is the code!), without a proposed defeating example. The tests are - apparently - extensive, but I may have missed some cases. I'll see if I can cook up any defeating examples or otherwise understand what cases you're imagining here.

pex/specifier_sets.py
```diff
@@ -0,0 +1,331 @@
...
+                return False
+            elif (
+                self.upper.version == other.lower.version
+                and self.upper.inclusive
+                and other.lower.inclusive
+            ):
+                return True
+
+        # Test full containment.
+        excluded_range = Range(lower=self.lower, upper=self.upper)
```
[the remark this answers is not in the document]
jsirois: It is a bit off. I went all over the place with code arrangement here over the course of working this branch. I'll see what I can do.

pex/specifier_sets.py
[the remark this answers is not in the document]
jsirois: Actually the else comment below covers `~=` - its impossible to get a SpecifierSet object with this content, since it fails to parse / construct. I'll just kill this checking code.

pex/specifier_sets.py
jsirois: Ok, after re-evaluating what this code was doing, I was able to simplify and eliminate with a small tweak to the exclude merging code above (it now handles the discards of excludes outside the bounds as well).

pex/specifier_sets.py
jsirois: Fixed by eliminating this code altogether.

pex/specifier_sets.py
jsirois: One thing to note is that the exclude culling and merging is a simplification step that does not affect real-world results, it just takes "!=1.*,!=2.*,>=2,<4" and discards "!=1.*" and merges "!=2.*" which simplifies to ">=3,<4" which evaluates the same, but with fewer checks. May or may not be faster overall, but simpler to read in any log or user-facing output; so worth it in my mind on that basis alone.

pex/specifier_sets.py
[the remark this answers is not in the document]
jsirois: The crucial invariant that an `ExcludeRange` always has both an upper and lower bound but a `Range` does not makes this tricky from a typing standpoint. I've tried twice to unwind this and both results are net losses. Unless you object further I'll let this lay as it is.

kaos (approved):
Nice. 
Didn't spend as much time as I'd like on the remaining commits, but nothing that jumped out to me.

pex/specifier_sets.py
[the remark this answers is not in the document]
kaos: No further objections. I imagine it's doable, but I guess it'll be with trade-offs and come down to a matter of preference over style so wouldn't be more than an exercise for the reader.

tests/test_dependency_manager.py
```diff
@@ -142,7 +149,22 @@ def test_exclude_complex(dist_graph):
 
     pex_info = PexInfo.default()
     pex_builder = PEXBuilder(pex_info=pex_info)
-    dependency_manager.configure(pex_builder, excluded=["c"])
+    with pytest.raises(AssertionError) as exec_info:
+        dependency_manager.configure(
```
huonw, line 153: Sorry, but I'm a bit slow on the uptake from this code. Why does this configure call fail but the later one does not?

huonw (commented):
Late review, that would've been approval ✅ just some minor questions.

pex/pip/excludes/requires.py
```diff
@@ -0,0 +1,67 @@
...
+logger = logging.getLogger(__name__)
+
+
+def patch():
+    from pex.dist_metadata import Requirement
+    from pex.exclude_configuration import ExcludeConfiguration
+
+    # N.B.: The following environment variable is used by the Pex runtime to control Pip and must
+    # be kept in-sync with `__init__.py`.
+    excludes_file = os.environ.pop("_PEX_EXCLUDES_FILE")
```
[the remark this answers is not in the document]
jsirois: I didn't actually. This is the 1st time a patch can use Pex code at all; so the pattern was copy-pasta-ed from the pre-existing foreign platform and locker patches. I'll circle back and fix that though now that it's possible.

pex/pip/excludes/requires.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def create_requires(orig_requires):
+        def requires(self, *args, **kwargs):
+            unexcluded_requires = []
+            for req in orig_requires(self, *args, **kwargs):
+                excluded_by = exclude_configuration.excluded_by(Requirement.parse(str(req)))
+                if excluded_by:
+                    logger.debug(
+                        "[{type}: patched {orig_requires}] Excluded {dep} from {dist} due to "
+                        "configured excludes: {excludes}".format(
```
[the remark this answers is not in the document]
jsirois: I guess I thought the `[{type}: patched {orig_requires}] ...` covered that - particularly the "patched", not to mention the lack of a `pip ...` feature to do this! If you're in this deep and don't know Pex patches to implement `--exclude` _and_ you get tripped up by this - you probably have no hope of debugging things. Not everything is paint by numbers!

But people really don't like my take on things like this at all and I don't see harm. I'll add that in the follow-up that fixes the shared env var name you pointed out above.

tests/test_dependency_manager.py
jsirois: Because the input requirements are `dist_graph.root_reqs` (see: [147](https://github.com/pex-tool/pex/pull/2409/files/2967b7d9e133dfa7dfcaf4d65d89d69df2eb0994..5591f758ccd3def5e729d9c3ced965ed17f05b1e#diff-1734962c259079fe44ab215607c037d799acd31b8f29f8c0d0c41e880d4f4145R147)) which is [`["a", "b"]`](https://github.com/pex-tool/pex/pull/2409/files/2967b7d9e133dfa7dfcaf4d65d89d69df2eb0994..5591f758ccd3def5e729d9c3ced965ed17f05b1e#diff-1734962c259079fe44ab215607c037d799acd31b8f29f8c0d0c41e880d4f4145R91). The `--exclude`, however is for "c", which is a transitive dependency, but "c" has not been excluded from The `DependencyManager.distributions`. The `DependencyManager` only handles excluding root reqs in the case you both ask for a thing _and_ `--exclude` the thing. The tricky case being `pex /this/local/project --exclude "foo<1" ...` where Pex doesn't know the project name and version of the `/this/local/project` requirement until _after_ the Pip resolve and wheel build, which is when the `DependencyManager` finishes things up building the PEX.

The latter call doesn't fail because the `DependencyManager.distributions` has a hole - "C" is excluded already (by Pip).

huonw: > The DependencyManager only handles excluding root reqs in the case you both ask for a thing and --exclude the thing

Ah, okay 👍 makes sense thanks.

pex/pip/excludes/requires.py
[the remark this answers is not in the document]
jsirois: Cleaned up in #2430.

pex/pip/excludes/requires.py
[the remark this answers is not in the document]
jsirois: Added in #2430.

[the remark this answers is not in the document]
kaos: I agree with making this clearer. Usually, I'm oblivious to a lot of details like this, and don't learn more until I'm facing an actual issue prompting me to get my feet wet and my act together and learn more. And in those settings, clear messages helps as things may be confusing as-is any way (and overlooking obvious things that I know/should know in the heat of the moment).

PR #2423, opened 2024-06-08
Fix `pex3 lock export` handling of exotic reqs.

### LINKED ISSUE #2416: `pex3 lock export` can turn VCS requirements into PyPI ones, resulting in hash failures
It seems like exporting a lockfile with a VCS requirement can sometimes use a `<name>==<version>` specifier in the exported `requirements.txt`, not `<name> @ <url>` as specified in the lockfile. This results in attempting to install different artifacts, with different hashes, to the ones PEX originally selected & hashed.

```shell
pex3 lock create 'pex @ git+https://github.com/pex-tool/pex@v2.3.1' -o test.lock --indent=2
pex3 lock export test.lock -o requirements.txt
pip install --no-cache -r requirements.txt
```

`test.lock`:

```json
{
  "allow_builds": true,
  "allow_prereleases": false,
  "allow_wheels": true,
  "build_isolation": true,
  "constraints": [],
  "locked_resolves": [
    {
      "locked_requirements": [
        {
          "artifacts": [
            {
              "algorithm": "sha256",
              "hash": "e6fb2d3eb7893fc5ff61c9e615d7ea8ad2cb9a93ac2b7927b30fe5f1018b2395",
              "url": "git+https://github.com/pex-tool/pex@v2.3.1"
            }
          ],
          "project_name": "pex",
          "requires_dists": [
            "subprocess32>=3.2.7; python_version < \"3\" and extra == \"subprocess\""
          ],
          "requires_python": "!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*,<3.13,>=2.7",
          "version": "2.3.1"
        }
      ],
      "platform_tag": [
        "cp39",
        "cp39",
        "macosx_14_0_arm64"
      ]
    }
  ],
  "only_builds": [],
  "only_wheels": [],
  "path_mappings": {},
  "pex_version": "2.3.1",
  "pip_version": "20.3.4-patched",
  "prefer_older_binary": false,
  "requirements": [
    "pex"
  ],
  "requires_python": [],
  "style": "strict",
  "target_systems": [],
  "transitive": true,
  "use_pep517": null
}
```

`requirements.txt`:

```
pex==2.3.1 \
  --hash=sha256:e6fb2d3eb7893fc5ff61c9e615d7ea8ad2cb9a93ac2b7927b30fe5f1018b2395
```

Output of `pip install`:
```
...
ERROR: THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
    pex==2.3.1 from https://files.pythonhosted.org/packages/e7/d0/fbda2a4d41d62d86ce53f5ae4fbaaee8c34070f75bb7ca009090510ae874/pex-2.3.1-py2.py3-none-any.whl (from -r requirements.txt (line 1)):
        Expected sha256 e6fb2d3eb7893fc5ff61c9e615d7ea8ad2cb9a93ac2b7927b30fe5f1018b2395
             Got        64692a5bf6f298403aab930d22f0d836ae4736c5bc820e262e9092fe8c56f830

```

This example uses the tagged commit of a version, but the same thing applies to using a non-tagged commit, e.g. `pex @ git+https://github.com/pex-tool/pex@36dd2374569b5f3fbef54918e6ba3bd1ff852b61` (36dd2374569b5f3fbef54918e6ba3bd1ff852b61)

The same thing applies to `pex3 lock export-subset` too.

jsirois:
Although Pex supports locking VCS requirements and local project
requirements, it does so with a be-spoke system for fingerprinting
each; as such, the `--hash`es emitted when exporting lock files
containing these types of requirements are not actually useable in
practice. Continue to support exporting this class of lock file, but
warn of the potential problems and offer a new `--format pip-no-hash`
mode for the daring. In addition, change the requirements output for
this class of lock file to match the input requirement for best
fidelity when actually attempting to use the resulting exported
requirement file without `--hash`es.

Fixes #2416

### DIFF
--- a/pex/cli/commands/lock.py
+++ b/pex/cli/commands/lock.py
@@ -34,7 +34,14 @@ from pex.resolve import requirement_options, resolver_options, target_options
 from pex.resolve.config import finalize as finalize_resolve_config
 from pex.resolve.configured_resolver import ConfiguredResolver
 from pex.resolve.lock_resolver import resolve_from_lock
-from pex.resolve.locked_resolve import LockConfiguration, LockStyle, Resolved, TargetSystem
+from pex.resolve.locked_resolve import (
+    LocalProjectArtifact,
+    LockConfiguration,
+    LockStyle,
+    Resolved,
+    TargetSystem,
+    VCSArtifact,
+)
 from pex.resolve.lockfile import json_codec
 from pex.resolve.lockfile.create import create
 from pex.resolve.lockfile.model import Lockfile
@@ -85,6 +92,7 @@ class ExportFormat(Enum["ExportFormat.Value"]):
         pass
 
     PIP = Value("pip")
+    PIP_NO_HASHES = Value("pip-no-hashes")
     PEP_665 = Value("pep-665")
 
 
@@ -512,8 +520,8 @@ class Lock(OutputMixin, JsonMixin, BuildTimeCommand):
             choices=ExportFormat.values(),
             type=ExportFormat.for_value,
             help=(
-                "The format to export the lock to. Currently only the {pip!r} requirements file "
-                "format using `--hash` is supported.".format(pip=ExportFormat.PIP)
+                "The format to export the lock to. Currently only the Pip requirements file "
+                "formats (using `--hash` or bare) are supported."
             ),
         )
         export_parser.add_argument(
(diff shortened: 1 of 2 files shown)


jsirois:
I appreciate any and all reviews as always. That said, in this case, I'll be blocking on your opinion @huonw. There are some eye-of-the-beholder decisions here that I'd like your second eye on.

huonw:
I've only had a moment for a quick skim but looks reasonable: nice approach with the warning message, and the introduction of `--format pip-no-hash` seems sensible (and the name looks good: obviously preserving `--format pip` for backwards compatibility, but even if starting from scratch, seems better to have the "more secure" version as the easier default).

Could you indicate which parts you're thinking are worth particular attention?

jsirois:
That's it, you covered it all in that paragraph.

benjyw (approved):
Implementation looks good to me, so if @huonw is happy with the functionality...

huonw (approved):

PR #2427, opened 2024-06-11
Implement support for preserving and injecting Python args.

### LINKED ISSUE #2422: A way to configure sys.executable flags at PEX runtime
I noticed that Python flags are supported in interpreter mode using the `inject_args` flag and wonder if something similar can be supported running entry points.

jsirois:
Support preserving and injecting Python args.

Pex now supports `--inject-python-args` similar to `--inject-args` but
for specifying arguments to pass to the Python interpreter as opposed to
arguments to pass to the application code. This is supported for both
zipapp and venv execution modes as well as standard shebang launching,
launching via a Python explicitly or via the `--sh-boot` mechanism.

In addition, PEX files now support detecting and passing through Python
args embedded in shebangs or passed explicitly on the command line for
all Pythons Pex supports save for `PyPy<3.10` where there appears to be
no facility to retrieve the original argv PyPy was executed with.

Closes #2422

### DIFF
--- a/pex/pex_boot.py
+++ b/pex/pex_boot.py
@@ -148,6 +148,7 @@ def boot(
     hermetic_venv_scripts,  # type: bool
     pex_path,  # type: Tuple[str, ...]
     is_venv,  # type: bool
+    inject_python_args,  # type: Tuple[str, ...]
 ):
     # type: (...) -> int
 
@@ -170,7 +171,7 @@ def boot(
         sys.stderr.write("Could not launch python executable!\\n")
         return 2
 
-    python_args = []  # type: List[str]
+    python_args = list(inject_python_args)  # type: List[str]
     orig_args = orig_argv()
     if orig_args:
         for index, arg in enumerate(orig_args[1:], start=1):
--- a/pex/pex_boot.py
+++ b/pex/pex_boot.py
@@ -1,0 +1,164 @@
+# Copyright 2014 Pex project contributors.
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+import os
+import sys
+
+TYPE_CHECKING = False
+
+if TYPE_CHECKING:
+    from typing import NoReturn, Optional, Tuple
+
+
+def __re_exec__(
+    argv0,  # type: str
+    *extra_launch_args  # type: str
+):
+    # type: (...) -> NoReturn
+
+    os.execv(argv0, [argv0] + list(extra_launch_args) + sys.argv[1:])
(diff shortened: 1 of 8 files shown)


jsirois:
This was a good idea from @bkad and it completes the work started by @BrandonTheBuilder in #1746 at PyCon US 2022. I think Pex now supports all Python interpreter args that make sense in all modes of operation.

tests/integration/test_inject_python_args.py
```diff
@@ -0,0 +1,262 @@
...
+    args.extend(execution_mode_args)
+    args.extend(boot_mode_args)
+
+    run_pex_command(args=args).assert_success()
+    shutil.rmtree(pex_root)
+    assert b"BufferedWriter\n" == subprocess.check_output(args=[sys.executable, exe])
+    assert b"BufferedWriter\n" == subprocess.check_output(
+        args=[pex]
+    ), "Expected cold run to use buffered io."
+    # assert b"BufferedWriter\n" == subprocess.check_output(
```
zmanji, line 236: This seems like some stray code that can be deleted.

zmanji:
The tests make sense to me and capture the original request of passing `-u` to the interpreter.

zmanji (approved):

tests/integration/test_inject_python_args.py
jsirois: Yup! Actually, un-commented. I want to be running all these 2x in a row to hit the cold cache, then the warm cache cases, which re-direct through different `execv` code paths.

huonw (approved):
Nifty

pex/pex_boot.py
```diff
@@ -0,0 +1,219 @@
...
+        sys.stderr.write("Could not launch python executable!\\n")
+        return 2
+
+    python_args = list(inject_python_args)  # type: List[str]
+    orig_args = orig_argv()
+    if orig_args:
+        for index, arg in enumerate(orig_args[1:], start=1):
+            if os.path.exists(arg) and os.path.samefile(entry_point, arg):
+                python_args.extend(orig_args[1:index])
+                break
```
jsirois, line 180: I added some manual timing around this during development - I was a bit worried about the ctypes impl since there is no opt-out even if you don't use Python args and this is in the hot path. The timings were consistently sub-ms for both sys.orig_argv and the ctypes implementation. Almost all the ~O(100us) time for the ctypes version was spent in the imports above where the function is defined. The call here was much faster.

PR #2431, opened 2024-06-12
Add support for `--override`.

### LINKED ISSUE #2425: Add support for `--override`ing transitive dependencies.
The deep `--exclude` work in #2409 opens the door to easily supporting re-writing the 3rdparty requirements Pip sees during the resolve process. This would allow loosening a 3rdparty requirement range or even completely swapping one dependency for another.

I am aware of cases where folks have wanted to loosen / expand a 3rdparty requirement to allow an impossible resolve to succeed where they know the altered requirement is valid for their use case.

I am not aware of cases where completely swapping a dependency out was needed, but you could imagine using this to publish a private wheel with a project name that bears no relation to the packaged code, which is code of some 3rdparty project with fixes. Although you could usually use a VCS requirement for this sort of 3rdparty patching, VCS requirements are sdists and require wheel builds on the fly, which can be sub-optimal.

jsirois:
This allows overriding transitive requirements in a resolve when the
user knows a thrid party project's requirements are too restrictive.

Also plumb the existing `--exclude` through to locks (along with
`--override`) to allow sealing dependency manipulation configuration
into the lock.

Closes #2425


tests/integration/test_overrides.py
```diff
@@ -0,0 +1,216 @@
...
+
+    assert_overrides(
+        pex,
+        expected_overrides=[
+            'urllib3==1.21; python_version >= "3.10"',
+            'urllib3==1.20; python_version < "3.10"',
+            "idna<2.5",
+        ],
+        expected_overridden_dists={"urllib3": ["1.20", "1.21"], "idna": ["2.4"]},
+    )
```
jsirois, line 117: @cognifloyd you can do pretty crazy stuff now between `--exclude` and `--override`. Use the power wisely.

cognifloyd: Thank you! That turned out to be very elegant.

kaos (approved):

pex/dependency_configuration.py
```diff
@@ -0,0 +1,166 @@
...
+        dest="overridden",
+        default=[],
+        type=str,
+        action="append",
+        help=(
+            "Specifies a transitive requirement to override when resolving. Any distribution "
+            "requirement in the PEX's resolve that matches the override project name is replaced "
+            "with the given override requirement. This option can be used multiple times."
+        ),
+    )
```
jsirois, line 159: @benjyw I wanted to draw your attention to this change retroactively. The `--exclude` and `--override` options are very different from `--constraints`. With `--constraint` you can only narrow the requirements of some transitive dependency. With `--exclude` and `--override` you can actually _edit_ the requirements of a transitive dependency.

PR #2433, opened 2024-06-21
Fix `pex --only-binary X --lock ...`.

### LINKED ISSUE #2432: "Resolving wheels disallowed" error when using `--lock` and `--only-binary`
I'm using pex 2.3.1 on Ubuntu 22.04.

If I create a lock file using:

```
pex3 lock create --output lock.json --only-binary ansicolors ansicolors
```

and run

```
pex --lock lock.json
```

it works w/o problem.

But running:

```
pex --lock lock.json --only-binary ansicolors
```

yields the following error:

```
Resolving wheels was disallowed, but the following project names are configured to only allow resolving pre-built wheels: ansicolors
```

In Pants 2.21.0, it seems like a change that uses pex to handle binary wheels was introduced.
https://github.com/pantsbuild/pants/pull/20598

Due to this change, pants always passes both `--lock` and `--only-binary` flags, causing https://github.com/pantsbuild/pants/issues/21077

jsirois:
Since #2346, which was released in Pex 2.1.161, using `--only-binary X`
with a `--lock` would fail fast even if the lock itself was created with
`--only-binary X`, which should be compatible.

Fixes #2432

### DIFF
--- a/pex/resolve/lock_resolver.py
+++ b/pex/resolve/lock_resolver.py
@@ -103,9 +103,13 @@ class VCSArtifactDownloadManager(DownloadManager[VCSArtifact]):
         self._network_configuration = network_configuration
         self._password_entries = password_entries
         self._cache = cache
+
+        # Since a VCSArtifactDownloadManager is only used for VCS requirements, a build is both
+        # required and preferred by the user.
         self._build_configuration = attr.evolve(
-            build_configuration, allow_wheels=False, allow_builds=True, prefer_older_binary=False
+            build_configuration, allow_builds=True, prefer_older_binary=False
         )
+
         self._pip_version = pip_version
         self._resolver = resolver
         self._use_pip_config = use_pip_config
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,14 @@
 # Release Notes
 
+## 2.4.1
+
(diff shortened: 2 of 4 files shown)


pex/resolve/lock_resolver.py
```diff
@@ -103,9 +103,13 @@ def __init__(
         self._network_configuration = network_configuration
         self._password_entries = password_entries
         self._cache = cache
+
+        # Since a VCSArtifactDownloadManager is only used for VCS requirements, a build is both
+        # required and preferred by the user.
         self._build_configuration = attr.evolve(
-            build_configuration, allow_wheels=False, allow_builds=True, prefer_older_binary=False
+            build_configuration, allow_builds=True, prefer_older_binary=False
```
huonw, line 110: I was initially confused why changes to `VCSArtifactDownloadManager` would improve the behaviour for what was tested. That test seems to only use PyPI requirements, not VCS ones.

My understanding is that this error happens eagerly, on construction of this `BuildConfiguration` object, and so when a `VCSArtifactDownloadManager` is constructed, it'll eagerly do the checking and throw the error here.

Is that correct?

huonw (approved):

pex/resolve/lock_resolver.py
jsirois: Yes.

PR #2440, opened 2024-06-25
give details of package which failed parsing requirements

jasonwbarnett:
When I was trying to resolve a lockfile in pants I kept getting an error like this:

```text
pex.dist_metadata.RequirementParseError: Expected matching RIGHT_PARENTHESIS for LEFT_PARENTHESIS, after version specifier
    scikit-learn (>=1.0.*) ; extra == 'pipelines'
```

It was very difficult to understand what to do or how to proceed. I ended up hacking up the cached source to get the relevant details. This PR aims to reduce the cost of getting the information necessary to work through an error like this.

This could probably be expanded to any exception, but I know that's a code smell so I'm not sure, but I can imagine there are more errors like this that could be challenging to debug for end users.


pex/dist_metadata.py
```diff
@@ -789,14 +789,20 @@ def as_constraint(self):
...
-            files=metadata_files,
-            project_name=metadata_files.metadata.project_name,
-            version=metadata_files.metadata.version,
-            requires_dists=tuple(requires_dists(metadata_files)),
-            requires_python=requires_python(metadata_files),
-        )
+        project_name = metadata_files.metadata.project_name
+        version = metadata_files.metadata.version
+        try:
+            # type: (MetadataFiles) -> DistMetadata
```
jsirois, line 795: Pex supports `>=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*,<3.14` (See: https://pypi.org/project/pex/ Requires: sidebar); so this serves as a type comment for MyPy that works for Python 2.7, but it only works if its where it was, just under the function `def` line.

pex/dist_metadata.py
```diff
@@ -789,14 +789,20 @@ def as_constraint(self):
...
+            return cls(
+                files=metadata_files,
+                project_name=project_name,
+                version=version,
+                requires_dists=tuple(requires_dists(metadata_files)),
+                requires_python=requires_python(metadata_files),
+            )
+        except RequirementParseError as e:
+            error_message = f"Error parsing requirements for project {project_name} version {version}: {str(e)}"
```
jsirois, line 804: Similar to above, f-strings are a no-go, you'll need to switch to `"...".format(...)`.

pex/dist_metadata.py
```diff
@@ -789,14 +789,20 @@ def as_constraint(self):
 class DistMetadata(object):
     @classmethod
     def from_metadata_files(cls, metadata_files):
```
jsirois, line 791: You might get an even more informative message y adding a new `source=""` argument that `DistMetadata.load(...)` below could pass. If that were either the full location or even just the basename, that would reveal the exact artifact with the issue, the sdist or one of the wheels. In the insane world of Python packaging the metadata is not required to be uniform amongst these various artifacts all in the same release version.

jsirois:
Thanks @jasonwbarnett. A test would be great, but I'm happy to circle back and add it as well. I'll wait to kick off CI until you have fixed the 2 Python 2.7 breaks I pointed out above.

You might want to `./dtox.sh -efmt,lint,check` to at least vet your change doesn't break things in a basic way (the 1st `./dtox` run will be slow - it effectively builds a dev container). Alternatively, if you have a fast internet connection, you might try `BASE_MODE=pull ./dtox.sh -efmt,lint,check` instead, which will pull the image from https://github.com/orgs/pex-tool/packages/container/package/pex%2Fbase.

I will note if you were using `--pip-version 24.1` I think you'd get Pip failing fast and a better error message as a result today. That Pip version became supported just a few days ago in https://github.com/pex-tool/pex/releases/tag/v2.5.0.

jasonwbarnett:
@jsirois Thank you for the quick feedback. I'll take a turn on this and get back to you later today.

>I will note if you were using --pip-version 24.1 I think you'd get Pip failing fast and a better error message as a result today. That Pip version became supported just a few days ago in https://github.com/pex-tool/pex/releases/tag/v2.5.0.

I'm going to update pex in pants manually ([docs][1]) and see if the error message is better. I'll use that as additional input into this PR.

[1]: https://www.pantsbuild.org/2.21/docs/python/overview/pex#setting-pex-and-pip-versions

jasonwbarnett:
@jsirois you're totally right. After adding this to pants.toml, the error message is *so* much better.

```toml
[python]
pip_version = "latest"

[pex-cli]
version = "v2.6.0"
known_versions = [
    "v2.6.0|macos_arm64|bc8fc449fc8165c395b26b407cc6033cf25cf02a76b7ede54828d2e899c1ee97|4157988",
    "v2.6.0|macos_x86_64|bc8fc449fc8165c395b26b407cc6033cf25cf02a76b7ede54828d2e899c1ee97|4157988",
    "v2.6.0|linux_x86_64|bc8fc449fc8165c395b26b407cc6033cf25cf02a76b7ede54828d2e899c1ee97|4157988",
    "v2.6.0|linux_arm64|bc8fc449fc8165c395b26b407cc6033cf25cf02a76b7ede54828d2e899c1ee97|4157988",
]
```

Error message:

```text
Requested mlflow==1.27.0 from https://.../pypi/simple/mlflow/1.27.0/mlflow-1.27.0-py3-none-any.whl#sha256=d759f3eefad2ff509a0fbc10507224204c6f6bb8d7f437bbf0bb9961cf74ff95 has invalid metadata: .* suffix can only be used with `==` or `!=` operators
    scikit-learn (>=1.0.*) ; extra == 'pipelines'
```

jasonwbarnett:
Based on that, I think I should just abandon this unless I'm missing something. You thoughts?

jsirois:
Thanks for bringing this to my attention @jasonwbarnett. I filed #2441 to track my work on this. Should have a release out by tomorrow with improvements for old `--pip-version`s.

PR #2442, opened 2024-06-26
Add artifact attestations for release.

jsirois:
See: https://github.blog/changelog/2024-06-25-artifact-attestations-is-generally-available/

PR #2444, opened 2024-06-26
Fix GitHub Releases deployment.

jsirois:
The addition of digital attestations in #2442 broke the GitHub Releases
release by moving from default permissions to more restrictive explicit
permission. Those permissions lacked the content write permission needed
to create the release and post its artifacts.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -72,6 +72,7 @@ jobs:
     permissions:
       id-token: write
       attestations: write
+      contents: write
     steps:
       - name: Checkout Pex ${{ needs.determine-tag.outputs.release-tag }}
         uses: actions/checkout@v4
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,19 @@
 # Release Notes
 
+## 2.6.2
+
+There are no changes to Pex code or released artifacts over 2.6.1, just
+a fix to the GitHub Releases release process which #2442 broke.
+
+* Fix GitHub Releases deployment. (#2444)
+
 ## 2.6.1
+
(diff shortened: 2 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -72,6 +72,7 @@ jobs:
     permissions:
       id-token: write
       attestations: write
+      contents: write
```
jsirois, line 75: The failure caused by the lack of content write permission can be seen here: https://github.com/pex-tool/pex/actions/runs/9684301905/job/26722997959

.github/workflows/release.yml
```diff
@@ -72,6 +72,7 @@ jobs:
     permissions:
       id-token: write
       attestations: write
```
jsirois, line 74: The attestations did work: https://github.com/pex-tool/pex/attestations

benjyw (approved):

jsirois:
Ok, @benjyw that didn't quite do it. The fix in the 2.6.1 release is a nice-to-have and only really useful if you can't use `--pip-version 24.1` (You run Python<=7). I'm away until July 2nd at which point I'll dig in to righting the release ship. If an emergency release is needed, I think just reverting this PR and the attestation PR makes sense as the stop gap.

benjyw:
Thanks @jsirois, I'll keep an eye out for any urgent release needs.

benjyw:
FWIW, looks like [this issue](https://github.com/softprops/action-gh-release/issues/142) was triggered, presumably by the upgrade of softprops/action-gh-release in #2374, and nothing to do with the attestations.

PR #2455, opened 2024-07-05
Support `--project` locking and PEX building.

### LINKED ISSUE #2412: Support for pip-compile generated requirements with hashes
When attempting to build a pex for package with dependencies pinned with `pip-compile --generate-hashes`, it will fail with the following error:
```
ERROR: Can't verify hashes for these file:// requirements because they point to directories:
    file:///home/anisse/dev/minimal-pip-compile-pex
```
I made a (not so) minimal repro to exhibit this issue, the relevant code can be found here: https://github.com/anisse/pex-minimal-repros/commit/4648ec016a644f91362a5e5984957f03e7053a60

I know there might be a few similar issues, like #429, and that pex has put a lot of effort in its own lock file format, so feel free to close this issue if it's out of scope for the project.

jsirois:
Although Pex already supported local project requirements, they worked
like any other requirement and were included in locks. In the typical
case, a project would rather maintain a lock of its third party
requirements separate from itself. For projects that used requirement
files to record their unlocked dependencies, `-r` could be used instead
of locking the local project. For projects that recorded their
dependencies in a `setup.py`, `setup.cfg` or `pyproject.toml`, the
dependency information needed to be extracted from those files to create
a lock that did not lock the project code itself. Just as bad, when
constructing a PEX the source code layout of the project recorded in
those same project files needed to be replicated in some combination of
`-D`, `-P` and `-M` since resolving from a lock required all projects
needed by the PEX were represented in the lock.

With the addition of the new `--project` option to PEX builds and
`pex3 lock {create,sync}`, a project's dependencies and source code
layout can be referenced from their canonical home in the project
configuration files.

Fixes #2412.

### DIFF
--- a/pex/pex_builder.py
+++ b/pex/pex_builder.py
@@ -397,8 +397,10 @@ class PEXBuilder(object):
                 copy_mode=CopyMode.LINK if is_wheel_file else None,
             )
         else:
-            for root, _, files in os.walk(path):
+            for root, _, files in deterministic_walk(path):
                 for f in files:
+                    if is_pyc_file(f):
+                        continue
                     filename = os.path.join(root, f)
                     relpath = os.path.relpath(filename, path)
                     target = os.path.join(target_dir, relpath)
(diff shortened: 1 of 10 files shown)


pex/tracer.py
jsirois, line 1: I noticed dropped timings for the new `TRACER.timed("Collecting requirements from {count} local {projects}")` added to [`pex/cli/commands/lock.py`](https://github.com/pex-tool/pex/pull/2455/files#diff-e8d5b02d5703b5ab96946abe35a9b20be92e4d3a96d1bd5142c6ae58c4d60ca0R841-R854).

pex/pex_builder.py
jsirois, line 1: I had actually seen this `.pyc` leak before in [scie-pants](https://github.com/pantsbuild/scie-pants/issues/2), but also encountered it in the integration test [here](https://github.com/pex-tool/pex/pull/2455/files#diff-4fd169dd839506218280b873b586be1a9f712d3c10e50b07c86411fa6f6efe49R184-R186) comparing PEX files.

kaos: Great!

kaos (approved):

PR #2457, opened 2024-07-07
Fix `bdist_pex` to use `--project`.

### LINKED ISSUE #2412: Support for pip-compile generated requirements with hashes

jsirois:
This change produces the same results for existing invocations of
`python setup.py bdist_pex` but allows new uses passing locked project
requirements (either hashed requirement files or Pex lock files) via
`--pex-args`.

Fixes #2412

### DIFF
--- a/tests/integration/test_issue_2412.py
+++ b/tests/integration/test_issue_2412.py
@@ -1,0 +1,257 @@
+# Copyright 2024 Pex project contributors.
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+from __future__ import absolute_import, print_function
+
+import os.path
+import subprocess
+from textwrap import dedent
+
+import pytest
+
+import testing
+from pex.common import safe_open
+from pex.interpreter_constraints import InterpreterConstraint
+from pex.pep_503 import ProjectName
+from pex.requirements import PyPIRequirement, parse_requirement_file
+from pex.resolve.lockfile import json_codec
+from pex.typing import TYPE_CHECKING
+from pex.venv.virtualenv import InstallationChoice, Virtualenv
+from testing.cli import run_pex3
+from testing.lock import index_lock_artifacts
+
+if TYPE_CHECKING:
+    from typing import Any
+
+
+def create_entry_module(project_dir):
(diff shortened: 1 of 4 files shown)


tests/integration/test_issue_2412.py
```diff
@@ -0,0 +1,257 @@
...
+            venv.interpreter.binary,
+            "setup.py",
+            "bdist_pex",
+            "--pex-args",
+            "--disable-cache -vvvv -r requirements.txt --pip-version 24.0",
+            "--bdist-all",
+        ],
+        cwd=issue_2412_repro_project_dir,
+    )
+    assert_bdist_pex(issue_2412_repro_project_dir, expected_numpy_version=locked_numpy_version)
```
jsirois, line 169: Alright @anisse, now your exact use case works. I added alternatives below that use Pex locking instead of `pip-compile`. I will note that `pex3 lock create --style universal --interpreter-constraint <matching your project> ...` does not have the limitations described here for `pip-compile`: https://pip-tools.readthedocs.io/en/stable/#cross-environment-usage-of-requirements-in-requirements-txt-and-pip-compile The Pex `--style universal` lock is truly universal, including all possibly needed depepdency subgraphs and resolving the right one when the lock is consumed later on a target machine to create a PEX for that machine. Or even when identifying the (foreign) target machine with `--platform` or `--complete-platform` to build, for example, a Linux PEX on a Mac or vice-versa or even a Linux / Mac multi-platform PEX.

kaos (approved):

PR #2461, opened 2024-07-09
Add support for `--extra-pip-requirement`.

jsirois:
You can now inject extra requirements into the isolated Pip PEXes Pex
uses to resolve distributions. The motivating use case for this is to
use the feature Pip 23.1 introduced for forcing `--keyring-provider
import`.

Pex already supported using a combination of the following to force
non-interactive use of the keyring:
1. A `keyring` script installation that was on the `PATH`
2. A `--pip-version` 23.1 or newer.
3. Specifying `--use-pip-config` to pass `--keyring-provider subprocess`
   to Pip.

You could not force `--keyring-provider import` though since the Pips
Pex uses are themselves hermetic PEXes without access to extra installed
keyring requirements elsewhere on the system. With
`--extra-pip-requirement` you can do this with the primary benefit over
`--keyring-provider subprocess` being that you do not need to add the
username to index URLs. This is ultimately because the keyring CLI
requires username whereas the API does not; but see
https://pip.pypa.io/en/stable/topics/authentication/#keyring-support for
more information.


tests/integration/conftest.py
```diff
@@ -119,61 +117,11 @@ def mitmdump_venv(shared_integration_test_tmpdir):
 
 
 @pytest.fixture
-def run_proxy(
```
jsirois, line 122: I factored this out to `testing/mitmproxy.py` to re-use mitmproxy in the new test in its reverse proxy mode to wrap PyPI / devpi with required auth.

pex/pip/installation.py
```diff
@@ -77,15 +79,63 @@ def _pip_installation(
...
 
     return _pip_installation(
         version=PipVersion.VENDORED,
-        iter_distribution_locations=lambda: third_party.expose(
-            ("pip", "setuptools"), interpreter=interpreter
-        ),
+        iter_distribution_locations=iter_distribution_locations,
         interpreter=interpreter,
+        fingerprint=_fingerprint(extra_requirements),
     )
```
jsirois, line 139: N.B.: The PR was motivated by Pip keyring but is more generic, allowing extra pip requirements for any reason under the sun that comes up. As such, even though keyring support does not work with vendored Pip, `--extra-pip-requirement` should work with vendored Pip.

benjyw (approved):
Nice!

PR #2466, opened 2024-07-15
Add `--scie` option to produce native PEX exes.

### LINKED ISSUE #636: Subpar alternative/comparison documentation
_copied over from issue on website repo pantsbuild/pantsbuild.github.io/issues/10 for exposure_

The [PEX Design page](https://www.pantsbuild.org/pex_design.html#alternatives) doesn't list subpar as an alternative, although it seems to be.

I don't know how much is configurable between the two, or how different situations are handled (such as C extensions). It looks like pex isn't very portable, but handles dependencies at least a bit differently.

Does anyone know the two well enough to be able to give a fairly detailed comparison on the practical differences in the website repo?

### LINKED ISSUE #1007: Can PEX bundle a Python interpreter too?
Is it possible for PEX to bundle a Python interpreter too so that the target machine does not even need a Python to be installed?

### LINKED ISSUE #2096: Introduce PAR support / add an option to ship a PEX with Python included.
With the advent of [scies](https://github.com/a-scie/jump), Pex's can now be shipped as standalone native executables with the Python interpreter included without much effort. If Pex were to incorporate this feature, it might be spelled like:
```
--par {eager,lazy}
```

This would cause Pex to generate one or more binaries using scie-jump that either directly include the appropriate PBS Python archive or else lazily fetch it using ptex. The fiddly bit here will be selecting an appropriately matched PBS per Pex target Python / abbreviated `--platform` / `--complete-platform`.

jsirois:
You can now specify `--scie {eager,lazy}` when building a PEX file and
one or more additional native executable PEX scies will be produced
along side the PEX file. These PEX scies will contain a portable CPython
interpreter from [Python Standalone Builds][PBS] in the `--scie eager`
case and will instead fetch a portable CPython interpreter just in time
on first boot on a given machine if needed in the `--scie lazy` case.

Although Pex will pick the target platforms and target portable CPython
interpreter version automatically, if more control is desired over which
platforms are targeted and which Python version is used, then
`--scie-platform`, `--scie-pbs-release`, and `--scie-python-version` can
be specified.

Closes #636
Closes #1007
Closes #2096

[PBS]: https://github.com/indygreg/python-build-standalone

### DIFF
--- a/pex/scie/__init__.py
+++ b/pex/scie/__init__.py
@@ -77,11 +77,19 @@ def register_options(parser):
         dest="scie_platforms",
         default=[],
         action="append",
-        type=SciePlatform.for_value,
-        choices=SciePlatform.values(),
+        type=SciePlatform.parse,
+        choices=[
+            platform
+            for platform in SciePlatform.values()
+            if platform not in (SciePlatform.WINDOWS_AARCH64, SciePlatform.WINDOWS_X86_64)
+        ],
         help=(
             "The platform to produce the native PEX scie executable for. Can be specified multiple "
-            "times."
+            "times. You can use a value of 'current' to select the current platform. If left "
+            "unspecified, the platforms implied by the targets selected to build the PEX with are "
+            "used. Those targets are influenced by the current interpreter running Pex as well as "
+            "use of `--python`, `--interpreter-constraint`, `--platform` or `--complete-platform` "
+            "options."
         ),
     )
     parser.add_argument(
--- a/pex/scie/__init__.py
+++ b/pex/scie/__init__.py
@@ -34,6 +34,9 @@ __all__ = (
     "ScieStyle",
     "ScieTarget",
     "build",
+    "extract_options",
+    "register_options",
+    "render_options",
 )
 
 
--- a/pex/scie/__init__.py
+++ b/pex/scie/__init__.py
@@ -1,0 +1,175 @@
+# Copyright 2024 Pex project contributors.
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+from __future__ import absolute_import
+
+from argparse import Namespace, _ActionsContainer
+
+from pex.fetcher import URLFetcher
+from pex.orderedset import OrderedSet
+from pex.pep_440 import Version
+from pex.scie import science
+from pex.scie.model import (
+    ScieConfiguration,
+    ScieInfo,
+    ScieOptions,
+    SciePlatform,
+    ScieStyle,
+    ScieTarget,
+)
+from pex.typing import TYPE_CHECKING, cast
+from pex.variables import ENV, Variables
+
+if TYPE_CHECKING:
+    from typing import Iterator, Optional, Tuple, Union
+
+
+__all__ = (
+    "ScieConfiguration",
+    "ScieInfo",
+    "SciePlatform",
+    "ScieStyle",
+    "ScieTarget",
+    "build",
+)
+
+
+def register_options(parser):
+    # type: (_ActionsContainer) -> None
+
+    parser.add_argument(
+        "--scie",
+        dest="scie_style",
+        default=None,
+        type=ScieStyle.for_value,
+        choices=ScieStyle.values(),
+        help=(
+            "Create one or more native executable scies from your PEX that include a portable "
+            "CPython interpreter along with your PEX making for a truly hermetic PEX that can run "
+            "on machines with no Python installed at all. If your PEX has multiple targets, "
+            "whether `--platform`s, `--complete-platform`s or local interpreters in any "
+            "combination, then one PEX scie will be made for each platform, selecting the latest "
+            "compatible portable CPython interpreter. Note that only CPython>=3.8 is supported. If "
+            "you'd like to explicitly control the target platforms or the exact portable CPython "
+            "selected, see `--scie-platform`, `--scie-pbs-release` and `--scie-python-version`. "
+            "Specifying `--scie {lazy}` will fetch the portable CPython interpreter just in time "
+            "on first boot of the PEX scie on a given machine if needed. The URL(s) to fetch the "
+            "portable CPython interpreter from can be customized by exporting the "
+            "PEX_BOOTSTRAP_URLS environment variable pointing to a json file with the format: "
+            '`{{"ptex": {{<file name 1>: <url>, ...}}}}` where the file names should match those '
+            "found via `SCIE=inspect <the PEX scie> | jq .ptex` with appropriate replacement URLs. "
+            "Specifying `--scie {eager}` will embed the portable CPython interpreter in your PEX "
+            "scie making for a larger file, but requiring no internet access to boot. If you have "
+            "customization needs not addressed by the Pex `--scie*` options, consider using "
+            "`science` to build your scies (which is what Pex uses behind the scenes); see: "
+            "https://science.scie.app.".format(lazy=ScieStyle.LAZY, eager=ScieStyle.EAGER)
+        ),
+    )
+    parser.add_argument(
+        "--scie-platform",
+        dest="scie_platforms",
+        default=[],
+        action="append",
+        type=SciePlatform.for_value,
+        choices=SciePlatform.values(),
+        help=(
+            "The platform to produce the native PEX scie executable for. Can be specified multiple "
+            "times."
+        ),
+    )
+    parser.add_argument(
+        "--scie-pbs-release",
+        dest="scie_pbs_release",
+        default=None,
+        type=str,
+        help=(
+            "The Python Standalone Builds release to use. Currently releases are dates of the form "
+            "YYYYMMDD, e.g.: '20240713'. See their GitHub releases page at "
+            "https://github.com/indygreg/python-build-standalone/releases to discover available "
+            "releases. If left unspecified the latest release is used. N.B.: The latest lookup is "
+            "cached for 5 days. To force a fresh lookup you can remove the cache at "
+            "<USER CACHE DIR>/science/downloads."
+        ),
+    )
+    parser.add_argument(
+        "--scie-python-version",
+        dest="scie_python_version",
+        default=None,
+        type=Version,
+        help=(
+            "The portable CPython version to select. Can be either in `<major>.<minor>` form; "
+            "e.g.: '3.11', or else fully specified as `<major>.<minor>.<patch>`; e.g.: '3.11.3'. "
+            "If you don't specify this option, Pex will do its best to guess appropriate portable "
+            "CPython versions. N.B.: Python Standalone Builds does not provide all patch versions; "
+            "so you should check their releases at "
+            "https://github.com/indygreg/python-build-standalone/releases if you wish to pin down "
+            "to the patch level."
+        ),
+    )
+
+
+def render_options(options):
+    # type: (ScieOptions) -> str
+
+    args = ["--scie", str(options.style)]
+    for platform in options.platforms:
+        args.append("--scie-platform")
+        args.append(str(platform))
+    if options.pbs_release:
+        args.append("--scie-pbs-release")
+        args.append(options.pbs_release)
+    if options.python_version:
+        args.append("--scie-python-version")
+        args.append(".".join(map(str, options.python_version)))
+    return " ".join(args)
+
+
+def extract_options(options):
+    # type: (Namespace) -> Optional[ScieOptions]
+
+    if not options.scie_style:
+        return None
+
+    python_version = None  # type: Optional[Union[Tuple[int, int], Tuple[int, int, int]]]
+    if options.scie_python_version:
+        if (
+            not options.scie_python_version.parsed_version.release
+            or len(options.scie_python_version.parsed_version.release) < 2
+        ):
+            raise ValueError(
+                "Invalid Python version: '{python_version}'.\n"
+                "Must be in the form `<major>.<minor>` or `<major>.<minor>.<release>`".format(
+                    python_version=options.scie_python_version
+                )
+            )
+        python_version = cast(
+            "Union[Tuple[int, int], Tuple[int, int, int]]",
+            options.scie_python_version.parsed_version.release,
+        )
+        if python_version < (3, 8):
+            raise ValueError(
+                "Invalid Python version: '{python_version}'.\n"
+                "Scies are built using Python Standalone Builds which only supports Python >=3.8.\n"
+                "To find supported Python versions, you can browse the releases here:\n"
+                "  https://github.com/indygreg/python-build-standalone/releases".format(
+                    python_version=options.scie_python_version
+                )
+            )
+
+    return ScieOptions(
+        style=options.scie_style,
+        platforms=tuple(OrderedSet(options.scie_platforms)),
+        pbs_release=options.scie_pbs_release,
+        python_version=python_version,
+    )
+
+
+def build(
+    configuration,  # type: ScieConfiguration
+    pex_file,  # type: str
+    url_fetcher=None,  # type: Optional[URLFetcher]
+    env=ENV,  # type: Variables
+):
+    # type: (...) -> Iterator[ScieInfo]
+
+    return science.build(configuration, pex_file, url_fetcher=url_fetcher, env=env)
--- a/pex/scie/__init__.py
+++ b/pex/scie/__init__.py
@@ -3,8 +3,10 @@
 
 from __future__ import absolute_import
 
+import os.path
 from argparse import Namespace, _ActionsContainer
 
+from pex.compatibility import urlparse
 from pex.fetcher import URLFetcher
 from pex.orderedset import OrderedSet
 from pex.pep_440 import Version
@@ -17,6 +19,7 @@ from pex.scie.model import (
     ScieStyle,
     ScieTarget,
 )
+from pex.scie.science import SCIENCE_RELEASES_URL, SCIENCE_REQUIREMENT
 from pex.typing import TYPE_CHECKING, cast
 from pex.variables import ENV, Variables
 
@@ -39,6 +42,7 @@ def register_options(parser):
 
     parser.add_argument(
         "--scie",
+        "--par",
         dest="scie_style",
         default=None,
         type=ScieStyle.for_value,
@@ -106,6 +110,19 @@ def register_options(parser):
             "to the patch level."
         ),
     )
+    parser.add_argument(
+        "--scie-science-binary",
+        dest="scie_science_binary",
+        default=None,
+        type=str,
+        help=(
+            "The file path of a `science` binary or a URL to use to fetch the `science` binary "
+            "when there is no `science` on the PATH with a version matching {science_requirement}. "
+            "Pex uses the official `science` releases at {science_releases_url} by default.".format(
+                science_requirement=SCIENCE_REQUIREMENT, science_releases_url=SCIENCE_RELEASES_URL
+            )
+        ),
+    )
 
 
 def render_options(options):
@@ -121,6 +138,9 @@ def render_options(options):
     if options.python_version:
         args.append("--scie-python-version")
         args.append(".".join(map(str, options.python_version)))
+    if options.science_binary_url:
+        args.append("--scie-science-binary")
+        args.append(options.science_binary_url)
     return " ".join(args)
 
 
@@ -138,7 +158,7 @@ def extract_options(options):
         ):
             raise ValueError(
                 "Invalid Python version: '{python_version}'.\n"
-                "Must be in the form `<major>.<minor>` or `<major>.<minor>.<release>`".format(
+                "Must be in the form `<major>.<minor>` or `<major>.<minor>.<patch>`".format(
                     python_version=options.scie_python_version
                 )
             )
@@ -156,11 +176,18 @@ def extract_options(options):
                 )
             )
 
+    science_binary_url = options.scie_science_binary
+    if science_binary_url:
+        url_info = urlparse.urlparse(options.scie_science_binary)
+        if not url_info.scheme and url_info.path and os.path.isfile(url_info.path):
+            science_binary_url = "file://{path}".format(path=os.path.abspath(url_info.path))
+
     return ScieOptions(
         style=options.scie_style,
         platforms=tuple(OrderedSet(options.scie_platforms)),
         pbs_release=options.scie_pbs_release,
         python_version=python_version,
+        science_binary_url=science_binary_url,
     )
 
 
--- a/pex/scie/configure-binding.py
+++ b/pex/scie/configure-binding.py
@@ -1,0 +1,30 @@
+# Copyright 2024 Pex project contributors.
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+from __future__ import print_function
+
+import os
+import sys
+
+
+def write_bindings(
+    env_file,  # type: str
+    installed_pex_dir,  # type: str
+):
+    # type: (...) -> None
+    with open(env_file, "a") as fp:
+        print("PYTHON=" + sys.executable, file=fp)
+        print("PEX=" + os.path.realpath(os.path.join(installed_pex_dir, "__main__.py")), file=fp)
+
+
+if __name__ == "__main__":
+    write_bindings(
+        env_file=os.environ["SCIE_BINDING_ENV"],
+        installed_pex_dir=(
+            # The zipapp case:
+            os.environ["_PEX_SCIE_INSTALLED_PEX_DIR"]
+            # The --venv case:
+            or os.environ.get("VIRTUAL_ENV", os.path.dirname(os.path.dirname(sys.executable)))
+        ),
+    )
+    sys.exit(0)
--- a/pex/scie/model.py
+++ b/pex/scie/model.py
@@ -133,6 +133,7 @@ class ScieOptions(object):
     python_version = attr.ib(
         default=None
     )  # type: Optional[Union[Tuple[int, int], Tuple[int, int, int]]]
+    science_binary_url = attr.ib(default=None)  # type: Optional[str]
 
     def create_configuration(self, targets):
         # type: (Targets) -> ScieConfiguration
--- a/pex/scie/model.py
+++ b/pex/scie/model.py
@@ -1,0 +1,250 @@
+# Copyright 2024 Pex project contributors.
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+from __future__ import absolute_import
+
+import itertools
+import os
+import platform
+from collections import defaultdict
+
+from pex.enum import Enum
+from pex.platforms import Platform
+from pex.targets import Targets
+from pex.third_party.packaging import tags  # noqa
+from pex.typing import TYPE_CHECKING, cast
+
+if TYPE_CHECKING:
+    from typing import DefaultDict, Iterable, Optional, Set, Tuple, Union
+
+    import attr  # vendor:skip
+else:
+    from pex.third_party import attr
+
+
+class ScieStyle(Enum["ScieStyle.Value"]):
+    class Value(Enum.Value):
+        pass
+
+    LAZY = Value("lazy")
+    EAGER = Value("eager")
+
+
+class SciePlatform(Enum["SciePlatform.Value"]):
+    class Value(Enum.Value):
+        @property
+        def extension(self):
+            # type: () -> str
+            return (
+                ".exe"
+                if self in (SciePlatform.WINDOWS_AARCH64, SciePlatform.WINDOWS_X86_64)
+                else ""
+            )
+
+        def binary_name(self, binary_name):
+            # type: (str) -> str
+            return "{binary_name}{extension}".format(
+                binary_name=binary_name, extension=self.extension
+            )
+
+        def qualified_binary_name(self, binary_name):
+            # type: (str) -> str
+            return "{binary_name}-{platform}{extension}".format(
+                binary_name=binary_name, platform=self, extension=self.extension
+            )
+
+        def qualified_file_name(self, file_name):
+            # type: (str) -> str
+            stem, ext = os.path.splitext(file_name)
+            return "{stem}-{platform}{ext}".format(stem=stem, platform=self, ext=ext)
+
+    LINUX_AARCH64 = Value("linux-aarch64")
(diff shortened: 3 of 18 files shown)


jsirois:
Reviewers - yet another big one. Thanks in advance for any time you can spare. This 1st commit has no tests, those are coming in a bit, but I wanted to get this out in case you wanted to start reading. There has been pretty extensive manual testing, both for perf (see binding command that resulted to bring perf down to `--sh-boot` levels in all cases) and for feature-matrix complexity.

pex/scie/science.py
```diff
@@ -0,0 +1,315 @@
...
+        "scie_jump": {"version": SCIE_JUMP_VERSION},
+        "files": [{"name": "configure-binding.py"}, {"name": "pex"}],
+        "commands": [
+            {
+                "env": env,
+                "exe": "{scie.bindings.configure:PYTHON}",
+                "args": ["{scie.bindings.configure:PEX}"],
+            }
+        ],
+        "bindings": [
```
jsirois, line 96: This is important for perf. Its pretty bad to package up your "native Python executable" and see it take ~70ms (for cowsay) when a plain `--venv --sh-boot` gets ~20ms. The binding gets any PEX that is scie'd up to `--sh-boot` perf levels.

pex/scie/model.py
```diff
@@ -0,0 +1,250 @@
...
+
+    def create_configuration(self, targets):
+        # type: (Targets) -> ScieConfiguration
+        return ScieConfiguration.from_targets(self, targets)
+
+
+@attr.s(frozen=True)
+class ScieConfiguration(object):
+    @classmethod
+    def from_tags(
```
jsirois, line 145: This is unused, but would / will be used by `pex3 scie create pre-existing.pex`. The idea is to get the wheel tags for all the distributions in the pre-existing PEX and use those to determine the platforms to target.

sureshjoshi:
Looking to carve off some time this evening to review this, but before I start, would it be safe to say that this is a (strict?) subset of the equivalent functionality when using `science` + a lift.toml to create a naively packaged pex + interpreter (e.g. excluding busy box functionality and custom bindings).

jsirois:
Yes. You'll find a nod to this and a pointer to science docs in the `--scie` help string as a consequence (i.e.: if you need to get more fancy, go there instead).

pex/scie/__init__.py
```diff
@@ -0,0 +1,175 @@
...
+            "on first boot of the PEX scie on a given machine if needed. The URL(s) to fetch the "
+            "portable CPython interpreter from can be customized by exporting the "
+            "PEX_BOOTSTRAP_URLS environment variable pointing to a json file with the format: "
+            '`{{"ptex": {{<file name 1>: <url>, ...}}}}` where the file names should match those '
+            "found via `SCIE=inspect <the PEX scie> | jq .ptex` with appropriate replacement URLs. "
+            "Specifying `--scie {eager}` will embed the portable CPython interpreter in your PEX "
+            "scie making for a larger file, but requiring no internet access to boot. If you have "
+            "customization needs not addressed by the Pex `--scie*` options, consider using "
+            "`science` to build your scies (which is what Pex uses behind the scenes); see: "
+            "https://science.scie.app.".format(lazy=ScieStyle.LAZY, eager=ScieStyle.EAGER)
```
jsirois, line 65: @sureshjoshi I see that your Pants plugin accepts an optional custom lift manifest, parses it if present, then injects bits into it. I think to support that sort of thing in a principled way, I'd have to parse the user supplied manifest and confirm they do not set the following keys:
+ ptex
+ scie_jump
+ files: with matching names
+ interpreters: with matching ids
+ commands: with a default command (I use this to launch the PEX)
+ bindings: with a matching name (needed for the default command to work)

Additionally, I'd have to advertise that I bind ptex to "ptex" for lazy scies, and bind `configure:PYTHON` and `configure:PEX`.

Without all this I don't see how the user supplied manifest can work with Pex needs fruitfully. Can you think of any other corners? Perhaps I'm overthinking. Do you need this functionality?

jsirois: I guess for `ptex` and `scie_jump` I could allow user-specified versions (but no more) IFF those versions were compatible with a lower bound.

sureshjoshi: I was hacking around tonight, trying to envision how I'd re-build something like `pantsible` (for example). 

One idea was to manipulate the embedded manifest after `pex` generates it (add the custom bindings and whatnot by piping the file to another tool), but then I realized I don't think I'd want to be able to dynamically modify the manifest of what should be a "sealed" binary, as that would be crazy for supply chain purposes - and I don't want to be able to dynamically alter the commands the executable could call.

In the case of the plugin (which, I wouldn't really use as a reference for anything - as I made it a few years ago to solve an immediate deployment problem on a client project), I think we try to use the optional lift.toml where possible and inject the target names under certain conditions. 

For this PR, I don't see any problems with deferring all of those concerns, but I'm of two minds.
- `pex` being able to accept a custom manifest template that has to be perfectly structured, with/without certain keys feels a bit hacky
- Using a separate tool (`science`), which overlaps with a lot of what `pex` would provide, feels off too

Would it make sense/be possible for `science` to defer to `pex` in some way, for the embedded Python interpreter? I'm trying to envision some sort of cleaner composition between two tools which have similar base functionality - but `science` allows some added knobs.

```toml
[lift]
name = "pantsible"
description = "Ansible with an embedded Python interpreter."
platforms = ... inferred from pex ...

[[lift.interpreters]] -> ... inferred from pex ...

[[lift.files]]
name = "pex"

[[lift.commands]]
name = "ansible"
exe = "{scie.bindings.venv}/venv/bin/ansible"
args = []

...
```

Although, one immediate problem I see here... I think I'm conflating a `pex` file with the `pex` CLI.

pex/scie/configure-binding.py
```diff
@@ -0,0 +1,30 @@
...
+        print("PYTHON=" + sys.executable, file=fp)
+        print("PEX=" + os.path.realpath(os.path.join(installed_pex_dir, "__main__.py")), file=fp)
+
+
+if __name__ == "__main__":
+    write_bindings(
+        env_file=os.environ["SCIE_BINDING_ENV"],
+        installed_pex_dir=(
+            # The zipapp case:
+            os.environ["_PEX_SCIE_INSTALLED_PEX_DIR"]
```
sureshjoshi, line 25: Should this also be `os.environ.get`? Or is `_PEX_SCIE_INSTALLED_PEX_DIR` guaranteed to exist?

pex/scie/science.py
```diff
@@ -0,0 +1,320 @@
...
+                    continue
+            except (CalledProcessError, InvalidVersion):
+                continue
+            return binary
+    return None
+
+
+def _science_binary_url(suffix=""):
+    # type: (str) -> str
+    return "https://github.com/a-scie/science/releases/download/v{version}/{binary}{suffix}".format(
```
sureshjoshi, line 188: Didn't these used to be located at `a-scie/lift`? Has that/Will that change?

pex/scie/__init__.py
sureshjoshi: Reading through the PR, another thought that popped into my head is allowing for the `pex` CLI's generated TOML to act as an overlay or merge-manifest with a local one. 

Whether that functionality is in `science` or `pex` CLI - overlaying/overwriting the user created manifest seems reasonable.

pex/scie/__init__.py
```diff
@@ -0,0 +1,175 @@
...
+
+    python_version = None  # type: Optional[Union[Tuple[int, int], Tuple[int, int, int]]]
+    if options.scie_python_version:
+        if (
+            not options.scie_python_version.parsed_version.release
+            or len(options.scie_python_version.parsed_version.release) < 2
+        ):
+            raise ValueError(
+                "Invalid Python version: '{python_version}'.\n"
+                "Must be in the form `<major>.<minor>` or `<major>.<minor>.<release>`".format(
```
sureshjoshi, line 141:
```suggestion
                "Must be in the form `<major>.<minor>` or `<major>.<minor>.<patch>`".format(
```

For consistency with the CLI docs above

pex/scie/science.py
```diff
@@ -0,0 +1,320 @@
...
+    def binary_name(self, binary_name):
+        # type: (str) -> str
+        return self.target.platform.binary_name(binary_name)
+
+    def qualified_binary_name(self, binary_name):
+        # type: (str) -> str
+        return self.target.platform.qualified_binary_name(binary_name)
+
+
+MIN_SCIENCE_VERSION = Version("0.3.0")
```
sureshjoshi, line 51: Should this be wrapped in a `Version`? That strips trailing 0's, but we use those trailers when trying to pull down a release from GH:

Required:
`https://github.com/a-scie/lift/releases/download/v0.3.0/science-macos-aarch64`

Actual:
```bash
python3 -m pex fastapi uvicorn --scie eager -o foo.pex -- foo.py
https://github.com/a-scie/lift/releases/download/v0.3/science-macos-aarch64
HTTP Error 404: Not Found
```

pex/scie/science.py
```diff
@@ -0,0 +1,320 @@
...
+                "build",
+                "--dest-dir",
+                dest_dir,
+            ]
+        )
+        if use_platform_suffix:
+            args.append("--use-platform-suffix")
+        args.append(manifest.path)
+        with open(os.devnull, "wb") as devnull:
+            process = subprocess.Popen(args=args, stdout=devnull, stderr=subprocess.PIPE)
```
sureshjoshi, line 277: I'm getting permissions error somewhere around here, I had to manually give `science` a `+x`

pex/scie/science.py
```diff
@@ -0,0 +1,320 @@
...
+    # type: (...) -> str
+    return os.path.join(env.PEX_ROOT, "scies", "science", str(MIN_SCIENCE_VERSION), *components)
+
+
+def _qualified_science_binary_name():
+    # type: () -> str
+    return SciePlatform.current().qualified_binary_name("science")
+
+
+def _science_binary_names():
```
sureshjoshi, line 160: Last note before I crash, should the fat `science` be downloaded instead of the ptex'd one? It's being downloaded with the intent of immediate usage.

What I don't know offhandedly is whether the lazy `science` binary can be leveraged to save a few megs off future updates.

pex/scie/__init__.py
jsirois: > I was hacking around tonight, trying to envision how I'd re-build something like pantsible (for example).

Well, `pantsible` uses a feature specific to scies over and above a PEX, namely the busybox support. It makes sense to me to just directly support this with `--scie-busybox [list of entry points]`. If you specify that then Pex emits a manifest with no default command and just named commands for each listed entry point.

> Would it make sense/be possible for science to defer to pex in some way, for the embedded Python interpreter? I'm trying to envision some sort of cleaner composition between two tools which have similar base functionality - but science allows some added knobs.

Well, science is general purpose - Any language; so it doesn't really make sense for it to know about Python let alone Pex. It does have a Provider interface to supply interpreters and that has exactly 1 implementation currently, that provides PBS interpreters. A PEX provider might make sense.

That said, Pex creates PEXes - single file executables. These do not have:
1. BusyBox support: You need [conscript](https://pypi.org/project/conscript/), for example, for that.
2. Bindings support: I.E.: Pex offers you no way to do pre-launch setup. You just have to write Python code to do 1 time setup in your main if you want that or provide alternate entry points fired off with `{PEX_MODULE=foo,PEX_SCRIPT=bar} ./my.pex`
As such, I think it makes sense for Pex to offer the ability to take your PEX file and turn it into a scie that behaves _exactly_ the same, with nothing extra except maybe running faster.  Everything you'd do in a custom manifest, afaict, would add things the PEX _cannot_ already do. At that point, having to move up a layer and use science yourself with a custom lift manifest to buold your app not using Pex directly makes sense. I.E.: what scie-pants has to do. The Pants app is more complex than just what the Pants PEX does / has tight perf overhead concerns; so it makes sense to move up to the higher layer.

> Reading through the PR, another thought that popped into my head is allowing for the pex CLI's generated TOML to act as an overlay or merge-manifest with a local one.

That's exactly what I meant by all this: https://github.com/pex-tool/pex/pull/2466#discussion_r1678670539 It seems to me you can't just overlay, you must confirm the key mechanisms Pex uses in its lift are not destroyed by the merge before merging.

pex/scie/science.py
jsirois: Facepalm. Thanks! I think this explains the CI failures which are 404 and not 403 - which is I think how GH indicates a rate limit.

pex/scie/__init__.py
jsirois: Yup - fixed.

pex/scie/science.py
jsirois: Wow, my face is getting red from all the face-palm. I kept Version, but use its `raw` property when constructing the URL now.

pex/scie/science.py
jsirois: Yup - that's a bug. I already have science on my machine so this code path did not get stressed locally. Fixed.

pex/scie/science.py
jsirois: I think it does make sense to use the fat binary - but more because it provides exactly 1 download URL and since Pex tries to support firewall use cases, I only need to add 1 option `--scie-science-binary <URL>` to complement the existing `PEX_BOOTSTRAP_URLS` escape hatch for the PBS downloads. I'll fix that and add the option.

pex/scie/configure-binding.py
jsirois: It's always set right here in the lift manifest: https://github.com/pex-tool/pex/pull/2466/files/82c629dffaa2aeb2e9a0f9210fcc7a46aeb1c43f#diff-3ed834e0e145c7460bbbb64066efe81446a388f9afce7073d5d77eac77626c79R107

Originally I placed the code inline in the binding command args like so: `..., "-c", dedent("""...""")]`. That made the availability of `_PEX_SCIE_INSTALLED_PEX_DIR` obvious since it was set in `env` a few lines above.

pex/scie/science.py
jsirois: Fixed to fat and now sporting `--scie-science-binary <FILE or URL>`.

pex/scie/__init__.py
```diff
@@ -39,6 +42,7 @@ def register_options(parser):
 
     parser.add_argument(
         "--scie",
+        "--par",
```
jsirois, line 45: As the issues attached to this PR prove, people in the world know "PAR"; so it seems to make sense to add a `--par` alias to this one option for discoverability by those people. If they need more than the default `--par` treatment, they really must learn about scies and `--scie-*` advanced options anyhow.

benjyw:
Will try and make some time to review this in a few hours. But very cool feature!

jsirois:
OK, CI is now down to erroring on the Linux runners having `~/.netrc` as a directory and the Python netrc stdlib not dealing with this 🤦 . I can work around this in science - where the error is originating from - but for now I'd like to solve just Pex issues; so I'll work around in CI instead.

pex/scie/__init__.py
sureshjoshi: > It seems to me you can't just overlay, you must confirm the key mechanisms Pex uses in its lift are not destroyed by the merge before merging.

Ah okay, I thought that overlaying pex's generated output on top of a user-supplied one (wiping out whatever is already there) would work - but I guess there are a lot of edge cases that might crop up along the way.

> At that point, having to move up a layer and use science yourself with a custom lift manifest to build your app not using Pex directly makes sense. I.E.: what scie-pants has to do. The Pants app is more complex than just what the Pants PEX does / has tight perf overhead concerns; so it makes sense to move up to the higher layer.

Agreed. This is what my brain was converging to by the time I stopped reviewing last night. 

The reason why I mentioned what is ultimately a "PEX PBS provider" is to hopefully ease the problem of pex `platforms`/constraints conflicting with the `platforms` and `interpreters` inside of a `science` lift.toml. Or at least, some way to deduce at runtime the valid Python interpreters for the given pex file(s).

jsirois: > Ah okay, I thought that overlaying pex's generated output on top of a user-supplied one (wiping out whatever is already there) would work - but I guess there are a lot of edge cases that might crop up along the way.

Yeah - just blindly over-writing what a user deemed important seems like a recipe for disaster.

> The reason why I mentioned what is ultimately a "PEX PBS provider" is to hopefully ease the problem of pex platforms/constraints conflicting with the platforms and interpreters inside of a science lift.toml. Or at least, some way to deduce at runtime the valid Python interpreters for the given pex file(s).

The current PR does derive the platforms and interpreters from the Pex input by default, you commonly only need to override with --scie-platform, --scie-pbs-release, --scie-python-version if you want:
1. To create less scies than the PEX's natiral set of platform would indicate (I have a test here I'll be pushing shortly that exercises this platform restricting use).
2. To make a reproducible scie: By default a Python version is picked using `<major>.<minor>` which leaves open the possibility PBS does a new patch release and when you re-build the PEX scie later, you get that new patch release and not the current one. In some use scenarios - not cool.

I'm not sure if I'm missing your point on this one or you just didn't see that ecisting functionality in the code / tests.

sureshjoshi: I'm referring to downstream tools like `science`, not pex, in this case. As in "once you've created a pex, then ..."

Anyways, the things I have in my mind are probably out of scope of this PR, and if they're important enough, or strongly enough use-cased, I can open a new ticket later.

jsirois: Gotcha. So I think the PEX interpreter Provider would just use the `pex3 scie create ...` logic I referenced here: https://github.com/pex-tool/pex/pull/2466#discussion_r1678497485

I.E.: not create the scie, but use the `ScieConfiguration.from_tags` API to implement platform / interpreter selection via the calculated `ScieConfiguration`'s `ScieTarget` targets which include platform, pbs_release and python_version.

That said, the current science Provider interface only allows providing an interpreter and not a set of platforms; so new API work would need to be done in science anyhow it seems to plug all this in.

jsirois: I guess the current API does allow enough for a PEX interpreter Provider to error when asked to produce an interpreter distribution via `Provider.distribution(platform)` for a platform the PEX does not support. That's probably actually enough:

```toml
[lift]
name = "example"
platforms = [
    "linux-aarch64",
    "linux-x86_64",
    "macos-aarch64",
]

[[lift.files]]
name = "pex"

[[lift.interpreters]]
id = "cpython"
provider = "PEX"
pex = "{pex}"
```

Here if I ran `science lift --file pex=my-py37.pex build ...` the PEX interpreter Provider could fail since CPython 3.7 is not supported and if I ran `science lift --file pex=my-py38.pex build ...` it could fail fast if, for example, there were no linux-aarch64 distributions in the latest PBS release.

sureshjoshi: Yep, there we go - that's the kinda thing I see value in.

jsirois: The restricting use case for `--scie-platform` I mentioned now has a test in 61f55a4 as does auto platforms detection.

jsirois:
Alright reviewers, the tests are now complete. Good for a final review.

@sureshjoshi I'm happy to break off a feature request for either or both of the `--scie-manifest` and `--scie-busybox` ideas that came up in our [thread above](https://github.com/pex-tool/pex/pull/2466#discussion_r1678670539), just let me know if either makes sense / are features you will use.

sureshjoshi:
> I'm happy to break off a feature request for either or both of the `--scie-manifest` and `--scie-busybox` ideas that came up in our [thread above](https://github.com/pex-tool/pex/pull/2466#discussion_r1678670539), just let me know if either makes sense / are features you will use.

Yep, after this lands, I can play around with it a bit more and see where it leads to.

In the meantime, I want to confirm that this is the expected behaviour.

```python
# foo.py
import uvicorn
from fastapi import FastAPI

app = FastAPI()


@app.get("/")
async def root():
    return {"message": "Hello World"}

if __name__ == "__main__":
    uvicorn.run(app, host="localhost", port=8000)
```

```
% python3.12 -m pex fastapi uvicorn --scie eager --scie-python-version 3.11  -o foo.pex -- foo.py 

% SCIE=inspect ./foo
...
 "files": [
    {
      "name": "cpython-3.11.9+20240713-aarch64-apple-darwin-install_only.tar.gz",
      "key": "cpython",
...

% ./foo
Python 3.12.4 (main, Jun  6 2024, 18:26:44) [Clang 15.0.0 (clang-1500.3.9.4)] on darwin
(InteractiveConsole)
>>> import fastapi
>>>
```

In my example, as the pex was built with python3.12, the pex shebang is `/usr/bin/env python3.12` - so even though the scie is bundled with Python 3.11, we are expecting to enter a 3.12 REPL, correct?

Based on the comment in the thread above:
> As such, I think it makes sense for Pex to offer the ability to take your PEX file and turn it into a scie that behaves exactly the same, with nothing extra except maybe running faster.

```bash
% python3.11 ./foo.pex 
Python 3.12.4 (main, Jun  6 2024, 18:26:44) [Clang 15.0.0 (clang-1500.3.9.4)] on darwin
(InteractiveConsole)
>>> 
```

The current behaviour matches what would happen if I just ran the pex with python3.11, so everything seems to line up and I'm just confirming my understanding of the feature.

sureshjoshi (approved):
Looks good to me after the back and forth. Would still be good to get eyeballs from someone more familiar with `pex` itself than me. 

My familiarity was more with the `scie` side of things.

jsirois:
> Looks good to me after the back and forth. Would still be good to get eyeballs from someone more familiar with pex itself than me.

AFAICT that is currently basically no one except me.

sureshjoshi:
> > Looks good to me after the back and forth. Would still be good to get eyeballs from someone more familiar with pex itself than me.
> 
> AFAICT that is currently basically no one except me.

😆  Good point

zmanji:
I don't have time to review this but i do have a question.

If one needs to customize the binaries, they would need to use `science` to create new binaries right?

jsirois:
> The current behaviour matches what would happen if I just ran the pex with python3.11, so everything seems to line up and I'm just confirming my understanding of the feature.

Well ... you did a super weird thing to though. What do you think you meant by the trailing `-- foo.py`?! Did you mean to use `--exe foo.py`? Or were you just stressing buggy use cases? The `-- foo.py` you used just throws away those extra args, which is probably a bug - you should be warned at least. So you just get a foo.pex (an thus a foo scie) without an entrypoint.

All that weird aside, what actually happened here is this: You build a platform specific PEX for Python 3.12, but instead of letting Pex use that to configure a 3.12 PBS, you overrode that and said 3.11 is fine - which it's not. When the boot binding runs, Pex is smart enough to test the current PBS 3.11 interpreter, find it can't load the PEX, then continue on to try other Pythons on the PATH. It finds a python3.12, which works to load the PEX and then writes out these bindings on my machine:
```
cat /home/jsirois/.cache/nce/5f4d759f14822688a76e0fd21f7a93897017bba9ba2218635023781d324ee362/locks/configure-bfbf6d1d4ddde46844370bf7672b02dfc07b0e8318fb2ce7b277b8436167a67b
PYTHON=/usr/bin/python3.12
PEX=/home/jsirois/.cache/nce/5f4d759f14822688a76e0fd21f7a93897017bba9ba2218635023781d324ee362/bindings/pex_root/unzipped_pexes/c27c9d03a91f03a2286d5901502f2ab7872918e5/__main__.py
```

So, as for `--scie-platform`, the use case for `--scie-pbs-release` and `--scie-python-version` is generally _narrowing_ the values that naturally arise from the PEX in question. The PEX here only supports 3.12; but you pushed the version out of bounds.

I guess I probably should blank out `PATH` in the boot binding to keep things hermetic:
```
:; python3.12 -m pex fastapi uvicorn --scie eager --scie-python-version 3.11 -o foo.pex
:; ./foo
Failed to find compatible interpreter on path None.

1.) /home/jsirois/.cache/nce/1f91c44febc850376a35ae77e1d45f7c823994b0c80293bbbc17e647eb893853/cpython-3.11.9+20240713-x86_64-unknown-linux-gnu-install_only.tar.gz/python/bin/python3.11 CPython==3.11.9

  Failed to resolve requirements from PEX environment @ /home/jsirois/.cache/nce/263d2999f5e4edddedbbcb29b3aeaf6f49d373ee26a76de93ad97d16f9959b0d/bindings/pex_root/unzipped_pexes/c27c9d03a91f03a2286d5901502f2ab7872918e5.
  Needed cp311-cp311-manylinux_2_35_x86_64 compatible dependencies for:
   1: pydantic-core==2.20.1
      Required by:
        pydantic 2.8.2
      But this pex had no ProjectName(raw='pydantic-core', validated=False, normalized='pydantic-core') distributions.
   2: MarkupSafe>=2.0
      Required by:
        Jinja2 3.1.4
      But this pex had no ProjectName(raw='MarkupSafe', validated=False, normalized='markupsafe') distributions.
   3: httptools>=0.5.0; extra == "standard"
      Required by:
        uvicorn 0.30.1
      But this pex had no ProjectName(raw='httptools', validated=False, normalized='httptools') distributions.
   4: pyyaml>=5.1; extra == "standard"
      Required by:
        uvicorn 0.30.1
      But this pex had no ProjectName(raw='pyyaml', validated=False, normalized='pyyaml') distributions.
   5: uvloop!=0.15.0,!=0.15.1,>=0.14.0; (sys_platform != "win32" and (sys_platform != "cygwin" and platform_python_implementation != "PyPy")) and extra == "standard"
      Required by:
        uvicorn 0.30.1
      But this pex had no ProjectName(raw='uvloop', validated=False, normalized='uvloop') distributions.
   6: watchfiles>=0.13; extra == "standard"
      Required by:
        uvicorn 0.30.1
      But this pex had no ProjectName(raw='watchfiles', validated=False, normalized='watchfiles') distributions.
   7: websockets>=10.4; extra == "standard"
      Required by:
        uvicorn 0.30.1
      But this pex had no ProjectName(raw='websockets', validated=False, normalized='websockets') distributions.
Error: Failed to establish atomic directory /home/jsirois/.cache/nce/263d2999f5e4edddedbbcb29b3aeaf6f49d373ee26a76de93ad97d16f9959b0d/locks/configure-9150f882feea5a550e5936c10776cf44573934ed3669cd27f5bd99ec8ef75f90. Population of work directory failed: Boot binding command failed: exit status: 1

The ./foo scie contains no alternate boot commands.
```

What do you think @sureshjoshi? Keep it behaving just like the PEX and bouncing down the PATH to find an interpreter that works (this means we shipped the wrong Python but the target machine had the right one), or keep things hermetic and fail as my experiment above does?

---
FWIW, I debugged all this with 2 techniques:
1. Sanity check what's going on: `rm -rf ~/.cache/nce && RUST_LOG=trace PEX_VERBOSE=1 ./foo`
2. Debug the binding step: `SCIE_BINDING_ENV=SCIE=split ./foo dist && _PEX_SCIE_INSTALLED_PEX_DIR=fake SCIE_BINDING_ENV=/dev/fd/0 PEX_VERBOSE=1 python3.11 dist/pex dist/configure-binding.py`

jsirois:
> If one needs to customize the binaries, they would need to use science to create new binaries right?

@zmanji in short, probably yes.

You could use `science`, but you can also just use `cat` plus a copy of the `scie-jump` (and a copy of `ptex` if you want lazy loading). See: https://github.com/a-scie/jump/blob/main/docs/packaging.md for more, but `science` is just a high level tool that dogfoods itself and these low level tools to provide a ntive python `science` binary that make assembling scies a bit easier.

zmanji:
On being hermetic, I will just say that pex's strength is being hermetic out of the box with flags to disable that if needed. I think a pex built with this feature should strip the PATH by default.

jsirois:
> On being hermetic, I will just say that pex's strength is being hermetic out of the box with flags to disable that if needed. I think a pex built with this feature should strip the PATH by default.

I like it! Even though this breaks the "PEX scie works just like the PEX" ~guaranty, it breaks the one part about a PEX this fixes, which is sealing in the interpreter. The only reason the PEX needs to bounce around to find a compatible Python if there even is one, is because of that 1 glaring bit of non-hermiticity in traditional PEXes.

sureshjoshi:
In my case, I wasn't trying to generate an exe or script - I was just trying to make a packaged repl with fastapi, uvicorn, and my foo.py (which seemed to work, as far as I could tell). I grabbed that example from something I was doing a couple of weeks ago on one of my many weird side-tangents. I'm sure there's a better way, but it worked one time I tried it, and I just ran with it since it's just a scratchpad.

> Keep it behaving just like the PEX and bouncing down the PATH to find an interpreter that works (this means we shipped the wrong Python but the target machine had the right one), or keep things hermetic and fail as my experiment above does?

Alright, yeah, my behavioural expectation test was presuming the goal was: "PEX scie works just like the PEX" - which it does.

BUT, having said that, I think being hermetic is preferable. Building with and bundling different interpreters is an easy blunder to make, and the last place you want to find that error is after deployment.

jsirois:
> In my case, I wasn't trying to generate an exe or script - I was just trying to make a packaged repl with fastapi, uvicorn, and my foo.py (which seemed to work, as far as I could tell). 

@sureshjoshi it did not. The foo.py was not included.

sureshjoshi:
> > In my case, I wasn't trying to generate an exe or script - I was just trying to make a packaged repl with fastapi, uvicorn, and my foo.py (which seemed to work, as far as I could tell).
> 
> @sureshjoshi it did not. The foo.py was not included.

🤦🏽 

It was just loading the local foo.py all along. 

Whelp, at least my pain and suffering led to a hermetic scie.

jsirois:
> It was just loading the local foo.py all along.

@sureshjoshi yes. Thanks for that though - as you said, everything is better as a result - except perhaps your sanity. So, people seem to never `zipinfo` on their PEXes, but its really helpful. So helpful, I went through alot of effort to make it so that you can do that to your PEX scie as well.

sureshjoshi:
Yeah, I unzipped and grepped, but I had the file referenced otherwise - so it showed up in my grep, but it was just a filename, not the file itself.

As I said, very weird tangents I was messing around with 🤦🏽

jsirois:
@benjyw I'm headed to the hills for a bit; so I'm going to proceed to merge this and get out a release. I feel good about the current commitments, but I'll circle back if you spot bugs or have questions.

benjyw:
Sounds fine, I'll take another look ASAP - I'm on vacation in Europe so code reviews are backing up.

PR #2468, opened 2024-07-17
Add support for `--scie-busybox`.

jsirois:
This allows you to turn your PEX into a native BusyBox scie by
specifying which entry points to expose as BusyBox commands.


jsirois:
Reviewers, a quickie before I hit the road. I'll be back on 7/23 to add tests.

For example though:
```
# Build a BusyBox PEX scie with 3 entry-points: 1 console script and 2 module eps from the stdlib:
:; python -mpex cowsay --scie lazy --scie-busybox cowsay,venv:venv,http-server:http.server -obusy-box

# Since the exe name does not match any ep, you get BusyBox help:
:; ./busy-box
Error: Could not determine which command to run.

Please select from the following boot commands:

cowsay
http-server
venv

You can select a boot command by setting the SCIE_BOOT environment variable or else by passing it as the 1st argument.

# Now use the BusyBox:
:; ./busy-box http-server -h
usage: server.py [-h] [--cgi] [--bind ADDRESS] [--directory DIRECTORY] [port]

positional arguments:
  port                  specify alternate port (default: 8000)

options:
  -h, --help            show this help message and exit
  --cgi                 run as CGI server
  --bind ADDRESS, -b ADDRESS
                        specify alternate bind address (default: all interfaces)
  --directory DIRECTORY, -d DIRECTORY
                        specify alternate directory (default: current directory)

# Now install a link farm for more convenient use (if the install directory is on your PATH anyhow):
:; SCIE=install ./busy-box -s /tmp/bin
:; ls -l /tmp/bin/
total 0
lrwxrwxrwx 1 jsirois jsirois 39 Jul 17 15:15 cowsay -> /home/jsirois/dev/pex-tool/pex/busy-box
lrwxrwxrwx 1 jsirois jsirois 39 Jul 17 15:15 http-server -> /home/jsirois/dev/pex-tool/pex/busy-box
lrwxrwxrwx 1 jsirois jsirois 39 Jul 17 15:15 venv -> /home/jsirois/dev/pex-tool/pex/busy-box

:; /tmp/bin/venv -h
usage: venv [-h] [--system-site-packages] [--symlinks | --copies] [--clear] [--upgrade] [--without-pip] [--prompt PROMPT] [--upgrade-deps] ENV_DIR [ENV_DIR ...]

Creates virtual Python environments in one or more target directories.

positional arguments:
  ENV_DIR               A directory to create the environment in.

options:
  -h, --help            show this help message and exit
  --system-site-packages
                        Give the virtual environment access to the system site-packages dir.
  --symlinks            Try to use symlinks rather than copies, when symlinks are not the default for the platform.
  --copies              Try to use copies rather than symlinks, even when symlinks are the default for the platform.
  --clear               Delete the contents of the environment directory if it already exists, before environment creation.
  --upgrade             Upgrade the environment directory to use this version of Python, assuming Python has been upgraded in-place.
  --without-pip         Skips installing or upgrading pip in the virtual environment (pip is bootstrapped by default)
  --prompt PROMPT       Provides an alternative prompt prefix for this environment.
  --upgrade-deps        Upgrade core dependencies: pip setuptools to the latest version in PyPI

Once an environment has been created, you may wish to activate it, e.g. by sourcing an activate script in its bin directory.
```

benjyw (approved):
I'll wait for the tests, but this LGTM!

pex/scie/model.py
```diff
@@ -30,6 +33,88 @@ class Value(Enum.Value):
     EAGER = Value("eager")
 
 
+@attr.s(frozen=True)
+class ConsoleScriptsManifest(object):
```
jsirois, line 37: @sureshjoshi I have not updated the `--scie-busybox` help string yet to reflect this idea - I wanted to run it by you 1st:

You can now, for example, say:
```
python -m pex ansible --scie eager --scie-busybox *:* -o ansible
```

That nets you a binary that executes as `ansible`, but allows `SCIE=install ./ansible ...` to get the BusyBox of all ansible scripts. Now ansible is clean in terms of transitive deps not having unwanted scripts. But for cases where that is so, you can restrict to all scripts in a dist:
```
python -m pex ansible --scie eager --scie-busybox ansible:*,ansible-core:* -o ansible
```

Or use some combination of the above subtracting off individual scripts or all scripts from a distribution by prefixing with `-`.

Good, bad or ugly?

jsirois: Ok, switched from `-` to `!` so the CLI use is not a total PITA, but otherwise filled out help docs and tests. I think this makes sense.

jsirois:
Alright reviewers, I got to this earlier than expected and I think its good to go for a final review.

pex/scie/model.py
sureshjoshi: Hey sorry, just a heads up that I just made it down to the states for a vacation, so I wouldn't be able to review anything until I'm back in the cold north

benjyw: I'd defer to @sureshjoshi but this seems like reasonable syntax to me.

benjyw (approved):

pex/scie/model.py
jsirois: Yeah, the `<project-name>:*` and `*:*` are excessive, (the `*`'s are not needed at all for parsing disambiguation) but my though was just `<project-name>:` and `:` respectively would confuse people (they did enough in Pants-land). I think `*` as the wildcard is more expected and neither `<project-name>:*` nor `*:*` are likely to match in the current directory; so they are not shell-onerous.

sureshjoshi: Okay, so on my Mac, I needed to escape the asterisks - otherwise, it failed over:

Just want to confirm my usage in this case (which works for me):

```bash
% python3.12 -m pex ansible-core==2.17.1 --scie eager --scie-busybox 'ansible-core:*' -o pexsible 

% SCIE=list ./pexsible
ansible
ansible-config
ansible-connection
ansible-console
ansible-doc
ansible-galaxy
ansible-inventory
ansible-playbook
ansible-pull
ansible-vault

% SCIE_BOOT=ansible ./pexsible --version
ansible [core 2.17.1]

% SCIE_BOOT=ansible-vault ./pexsible --version
ansible-vault [core 2.17.1]
```

Also, just so it's in writing and I assume it's expected because of the pre-compilation/venv bindings in my lift file (https://github.com/sureshjoshi/pantsible/blob/39ec69754f96d045a247889e304a6ff1da587882/lift.toml#L67-L80):

Using the `scie` generated from [pantsible](https://github.com/sureshjoshi/pantsible) can run the same `--version` command about 100ms faster than the pex-generated one (130ms vs 230ms)

jsirois: > I needed to escape the asterisks - otherwise, it failed over:

... Did you mean to include failure output? I'd like to get the syntax right and if the current glob syntax is problematic it would be good to get right. I will not yank the syntax once this is released per the usual Pex guarantees.

> Using the scie generated from [pantsible](https://github.com/sureshjoshi/pantsible) can run the same --version command about 100ms faster than the pex-generated one (130ms vs 230ms)

Umm - add `--venv` to your Pex command above. You're comparing apples and oranges.

jsirois: > `SCIE_BOOT=ansible ./pexsible --version`

Perhaps muscle-memory, but kind of the whole point of the BusyBox is not to know its a scie, just that it's a BusyBox. You can:
```
:; python3.12 -m pex ansible-core==2.17.1 --scie eager --scie-busybox 'ansible-core:*' -o pexsible --venv prepend

:; ./pexsible
Error: Could not determine which command to run.

Please select from the following boot commands:

ansible
ansible-config
ansible-connection
ansible-console
ansible-doc
ansible-galaxy
ansible-inventory
ansible-playbook
ansible-pull
ansible-vault

You can select a boot command by setting the SCIE_BOOT environment variable or else by passing it as the 1st argument.

:; ./pexsible ansible --version
ansible [core 2.17.1]
  config file = None
  configured module search path = ['/home/jsirois/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/jsirois/.cache/nce/ffd012984ec0d4868f68d5c04bd135846d325383c4ea57206134508b9a995ad8/bindings/pex_root/venvs/f1a995cbdcdbc34175508ccddaf0bd2a8c9654c0/d21d9b019bc5fc18b995639990107cd0fa54b7dd/lib/python3.12/site-packages/ansible
  ansible collection location = /home/jsirois/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/jsirois/.cache/nce/ffd012984ec0d4868f68d5c04bd135846d325383c4ea57206134508b9a995ad8/bindings/pex_root/venvs/f1a995cbdcdbc34175508ccddaf0bd2a8c9654c0/d21d9b019bc5fc18b995639990107cd0fa54b7dd/bin/ansible
  python version = 3.12.4 (main, Jul 25 2024, 22:42:01) [Clang 18.1.8 ] (/home/jsirois/.cache/nce/ffd012984ec0d4868f68d5c04bd135846d325383c4ea57206134508b9a995ad8/bindings/pex_root/venvs/f1a995cbdcdbc34175508ccddaf0bd2a8c9654c0/d21d9b019bc5fc18b995639990107cd0fa54b7dd/bin/python)
  jinja version = 3.1.4
  libyaml = True
```

sureshjoshi: > Umm - add --venv to your Pex command above. You're comparing apples and oranges.

It's official: Disneyworld fried my brain... I removed the `venv` while debugging the asterisk syntax and it never came back. 

```bash
% python3.12 -m pex ansible-core==2.17.1 --scie eager --scie-busybox 'ansible-core:*' -o pexsible 
-> works

% python3.12 -m pex ansible-core==2.17.1 --scie eager --scie-busybox ansible-core:* -o pexsible
zsh: no matches found: ansible-core:*

bash-3.2$ python3.12 -m pex ansible-core==2.17.1 --scie eager --scie-busybox ansible-core:* -o pexsible 
-> works
```

sureshjoshi: > Perhaps muscle-memory, but kind of the whole point of the BusyBox is not to know its a scie

Yep, I was re-using the `SCIE=inspect` and `SCIE=list` functionality a bunch, so was just quicker to add the BOOT part 🤷🏽

jsirois: Ok, I'm not a zsh'er, but many are so that's probably not awesome. @sureshjoshi do you have opinions about what should be done with the current `ansible-core:*` / `*:*` syntax? My thought was that `ansible-core:` / `:` would confuse people since it has in Pants, but maybe not?

sureshjoshi: I think the colons will/would/do confuse everyone (I'm not exempt to that confusion from time to time). Especially wildcards, which are generally agreed to be asterisks.

So, just my entirely personal opinion, I don't mind string escaping stuff if I need to - but I guess if that's not the norm with the rest of the CLI, that could be ugly too. 

The other option is to escape it, which... Yeah, not great either.

```bash
python3.12 -m pex ansible-core==2.17.1 --venv --scie eager --scie-busybox ansible-core:\* -o pexsible 
```

This is strange though, because I'm sure there are other zsh/bash/powershell-even tools that take in an asterisk without grief (git maybe?)

@huonw @benjyw Any thoughts on this? My vacation-brain is clearly not ready for prime time.

jsirois: Afaict the zsh answer can only be in completers it has set up. If it has none (this case), it tries to file glob. Unlike bash, if a file glob misses it errors. That's my guess.

Based on that, probably the easiest escape is simply `--scie-busybox=ansible-core:*` ... presumably built in zsh globbing turns off when the start looks like a switch?

jsirois: Ok, not it. It appears this is just a PITA bit about zsh?:
```
:; git log -Sansible-core:*

:; zsh
Gill-Windows% git log -Sansible-core:*
zsh: no matches found: -Sansible-core:*
Gill-Windows% git log -S'ansible-core:*'
Gill-Windows%
```

This being said, it sounds like you're in favor of the existing glob syntax then @sureshjoshi. I'll wait for your PR approve before submitting.

jsirois: Aha, NOMATCH is the culprit:
```
:; zsh +o nomatch
Gill-Windows% git log -Sansible-core:*
Gill-Windows%
```

It appears the aggressively glob and error if no match is the default though. That seems like a crappy default, but perhaps there is good reason for it.

huonw: > @huonw ... Any thoughts on this? 

If you're asking about `*` and escaping, I'll think out loud:

- There is a risk of people writing a script that doesn't escape properly and it seeming to work until they add a file with an unlucky name. 
- If one added `foo:bar`, the immediate consequence would be expanding `*:*` to `--scie-busybox foo:bar`.
- It looks like this would be silently accepted and result in a pex with an unexpected configuration (at least, I don't see validation that the project name `foo` is understood, but maybe I missed it).
- Having validation that all input project names are known distributions would reduce that consequence; is that possible?
- Alternatively, use some other shell-safe signifier. I don't have great ideas, only `...`, e.g. `ansible:...` or `...:...` but that's pretty ugly. `*` matches by expectations more.

Summary: I can't think of a better `*` I like, but maybe we could validate the `foo:bar` values are actually known entrypoints (at least the project name) to reduce the consequence of problems here.

pex/scie/__init__.py
```diff
@@ -72,6 +74,35 @@ def register_options(parser):
             "https://science.scie.app.".format(lazy=ScieStyle.LAZY, eager=ScieStyle.EAGER)
         ),
     )
+    parser.add_argument(
+        "--scie-busybox",
+        dest="scie_busybox",
+        type=str,
+        default=[],
+        action="append",
+        help=(
```
huonw, line 83: Minor feedback: these docs look good, but they're a bit of a wall of text that'll usually be read in the context of a terminal without any formatting aids, and thus I'd expect most people (even well-intentioned diligent ones) to fail to synthesise the information out of it.

Is it possible to make it easier for users by breaking it up into paragraphs and/or lists?

huonw (commented):
Nice feature. NB. I haven't reviewed the functionality in detail.

pex/scie/__init__.py
jsirois: It is not - Argparse doesn't support formatting I know of. I've both avoided spending time on any bespoke argument system to remedy this or turning towards documentation, which need major work going on 10 years now. One or both are coming, but I'm prioritizing robust users so far in favor of righting the Pex ship going on 8 years. I do finally feel close though.

huonw: Ah I see. Thanks for the background. 👍

pex/scie/model.py
jsirois: > but maybe we could validate the foo:bar values are actually known entrypoints (at least the project name) ...

Well, in `x:y` form, the x is not, in fact, a project name - its a module name and need not bear any relation to project name (setuptools -> setuptools + pkg_resources, etc.). So, to validate those module entrypoints would require checking all dists' `top_level.txt` metadata and verifying the leading component in the module name (split by dots) had at least 1 hit and then drilling into that dist further to resolve other module components, if any. Alternatively, since there are no foreign platforms in play when reifying to a scie - the scie interpreter could be used to try an import. That then would demand equal treatment for scripts. Right now there is 0 validation that the scripts you specify directly (no wildcards) exist either. As a point of comparison, there is no checking for entrypoint validity when using `-m <entrypoint spec>` today with Pex. You have to specify:
```
  --validate-entry-point
                        Validate the entry point by importing it in separate
                        process. Warning: this could have side effects. For
                        example, entry point `a.b.c:m` will translate to `from
                        a.b.c import m` during validation. (default: False)
```

jsirois: And I lied to make things worse. Scies are, in fact, generated for foreign platforms; so only local platforms could be checked via import and there may be 0 local platforms - so the import check, if it was done, would only work some of the time. The `top_level.txt` metadata + assuming file system based imports (no magic import hooks like `__pex__`) might be more robust in more cases. No matter how you slice it, its not pretty and clean to double check this stuff for the PEX / scie builder. They really actually have to test their binary in the end.

huonw: I'm feeling fairly confused at the moment, so feel free to overrule me and move on, but I'd (somewhat arrogantly) suggest that if I'm confused that a user who's _not at all_ involved with pex development will be even more confused.

This is sprawling, so summary:

- discussion of separating the script and module entry points behaviour
- idea that `:all:` could be the globbing specifier?
- acknowledgement that validating module entry points is hard

> Well, in x:y(:z) form, the y is not, in fact, a project name

Ah, hm, so reading more closely, it seems like there's a difference between the meaning of `x` in `x:*` and `x:y`?

That is, `--scie-busybox 'x:*'` isn't equivalent to `--scie-busybox 'x:y1,x:y2,...` with an appropriately selected list of `y` values. (AIUI, the latter is it's attempting to define multiple entry points called `x` and thus is malformed in its own way.)

Assuming this is the case, I think that's... strongly unfortunate and will lead to confusion, it leaves a big rough edge to users to cut themselves on... sure, they should be testing, but pex isn't helping guide them into the "pit of success". 

Maybe a key improvement would making the different "add scripts" and "add module entrypoints" behaviours separate and more explicit?

(This could also open the door for using `:all:` for "globbing", inspired by pip.)

At the most explicit extreme would be a full set of args like:

- console scripts: `--scie-busybox-include-all-scripts`, `--scie-busybox-include-scripts-from=project`, `--scie-busybox-exclude-scripts-from=project `, `--scie-busybox-include-script=specific-script`, `--scie-busybox-exclude-script=specific-script`
- module entrypoints: `--scie-busybox-entry-point=name:module(:func)`

This'd tangentially solve the globbing issue, but it's quite a lot of args!

Brainstorming: more sensible middle grounds might be:

1. `--scie-busybox-scripts-from=...` (taking a one or more of: `*` (or `:all:`) for all projects, a project name, a `!` exclusion) and `--scie-busybox-script=...` (one or more of a script name or `!`) and `--scie-busybox-entry-point=name:module(:func)`. 
2. `--scie-busybox-script=...` (one or more of all-projects glob, single project glob<sup>1</sup>, script name, or `!` exclusion) and `--scie-busybox-entry-point=...`.

<sup>1</sup>: Obviously `project:*` could still work if `*` is the globbing specifier... but if it's `:all:`, `project:all:` seems like it can work (i.e. isn't ambiguous), but it's weird!

Examples: to add all scripts except for the script called `bar` and all scripts in project `foo`, but still add `foo`'s `foo-script`:

- extreme: `--scie-busybox-include-all-scripts --scie-busybox-exclude-scripts-from=foo --scie-busybox-include-script=foo-script --scie-busybox-exclude-script=bar`
- 1: `--scie-busybox-scripts-from=:all:,!foo --scie-busybox-scripts=foo-script,!bar`
- 2: `--scie-busybox-scripts=:all:,!foo:all:,foo-script,!bar`



> That then would demand equal treatment for scripts. Right now there is 0 validation that the scripts you specify directly (no wildcards) exist either

Yep, in the context of this comment location and discussion of wildcards, I was focusing on `ConsoleScriptsManifest` and its `iter_distribution` loop. This existing infrastructure was hinting to me that validating the project names and "individual" script names wouldn't be a massive step.

> No matter how you slice it, it's not pretty and clean to double check this stuff for the PEX / scie builder. They really actually have to test their binary in the end

Yep, definitely convinced that validation requiring (potentially impossible) dynamic imports is a dead-end, at least for default behaviour.

> You have to specify: [...] --validate-entry-point

If someone specifies `--validate-entry-point`, will these `--scie-busybox` entrypoints be validated too?

> They really actually have to test their binary in the end.

Yep, acknowledged that this is the gold standard!

jsirois: How about:
```
--scie-busybox adhoc=module.deeper:func,@,!fortune,!@ansible-core
```

Left to right:
+ Include a script named `adhoc` pointing to `module.deeper:func`.
+ Include a script named `cowsay`.
+ Include all scripts in `all` projects.
+ But exclude a script named `fortune`.
+ And also exclude all scripts in the `ansible-core` project.

That's from the rules:
1. Name scripts by their name, whether pre-existing in a project distribution or when creating a virtual ad-hoc script for a random entry point module (and optionally a `:<function name>` in that module) of your choosing (This is the verbatim syntax of setup.py console scripts).
2. As an extension of direct URL reference requirements (<project name>@<URL>) you can index into a project name to get its scripts with @, but from the left; so `<script>@<project name>(@<URL>)` where the right hand (@<URL>) is illegal here, the project distributions were already resolved! - It's just for illustration of the pre-existing syntax logical ~extension.

With the `@` operator for indexing into a project names dists you get the following permutations:
1. `cowsay@`, `!cowsay@` : Include (or exclude) the script `cowsay` in any project distribution.
2. `@ansible-core`, `!@ansible-core` : Include (or exclude) all scripts (we don't pin a name) in `ansile-core`'s distributions.
3. `@`: Include all scripts (we don't pin a script name) in all projects' distributions (we don't pin a project name).

N.B.: The `cowsay@` and `!cowsay@` are redundant with `cowsay` and `!cowsay` but keep the `@` syntax uniform. This means the bare implied "in any projects' distributions" of just `cowsay` _could_ be jettisoned, but then <script name>(=<script definition) loses a bit of its symmetry.

Simply leaving off a name in either position in the `@` project name addressing syntax is enough to wildcard. Much like a bare `:` discussed earlier, I fear this will confuse people and they might be more comfortable spelling out a wildcard like `*@*`. @huonw you proposed the Pip standard of `:all:@:all:` - That solves shell globbing but I'm not a fan on arbitrary asthetic grounds. I'm happy though to not care and take reviewer consensus on an explicit wildcard or not supporting one at all and just using the missing name on either side of the `@` to indicate all things on that side of the `@`.

@sureshjoshi I'd love your feedback too. 2 random opinions is sort of worthless. 2 out of 3 or 3 out of 3 feels a lot better.

jsirois: > If someone specifies --validate-entry-point, will these --scie-busybox entrypoints be validated too?

@huonw not right now for BusyBox entry points, but for a `-c` / `-m` single entry point, yes.

I agree that's not in the spirit of "adding --scie {eager,lazy}" nets you a thing ~just like a PEX, but native. Here the whole idea of a BusyBox is new and exclusive to scies, but...

I think I should add this since the option and appropriate warning are already there and the problem of a pure foreign platform PEX already was there - another bit of ship-righting to do. That limitation should probably be called out, right now the implementation is just YOLO and we've never had a bug report, but adding legalese to the wall of `--help` text might make sense.

sureshjoshi: Alright, turns out that this was a lot more complicated than I had thought. I guess that's why we never got [wildcard support in Docker](https://github.com/moby/moby/issues/17237).

**Globbing specifier:**

I have no problem with `*`, so long as it's called out in examples that it should be escaped (either as `\*`, or `--scie-busybox "ansible:*,ansible-core:*"`). That seems like a perfectly reasonable option, and I think using the most obvious dev-centric wildcards makes the most sense, rather than `:all:` or `@` or `::` etc...

The `@` could be reasonable too, but doing that JUST to prevent escaping a string feels a little over-the-top too. My original point was that, this might be the only CLI arg that would get stringified, but I would take that over new syntax.

**Functionality**

However, in the last responses - the scope of what is possible expanded - and gotta be honest, I'm more confused now.

To keep this grounded and a bit more concrete in my head, here are some ways I've used lifts already with `science`.

---
1. `pantsible`

Take a python library (ansible) with a suite of commands, and re-bundle it with Python, and then expose all of ansible's bin commands. Maybe there are a couple of commands we wouldn't want to export using the `!`.

Covered above in a variety of formats.

```toml
[[lift.commands]]
name = "ansible"
exe = "{scie.bindings.venv}/venv/bin/ansible"

[[lift.commands]]
name = "ansible-config"
exe = "{scie.bindings.venv}/venv/bin/ansible-config"

...

[[lift.commands]]
name = "ansible-vault"
exe = "{scie.bindings.venv}/venv/bin/ansible-vault"
```

---

2. Web server ("apigateway.pex")

A server that I bundled with some additional tooling, so by default, we run the server, but if I use the other commands, we might (e.g.) run a DB migration, or whatever support stuff I wanted.

Ignoring the pre-compile bindings, this looks like a use case for the `adhoc=module.deeper:func` style, am I correct?

```toml
[[lift.commands]]
exe = "{scie.bindings.venv}/venv/bin/uvicorn"
args = ["apigateway.main:app", "--port", "7999"]

[[lift.commands]]
name = "seed"
exe = "#{cpython:python}"
args = ["{:apigateway-pex}", "-m", "seeder.main"]
``` 

---

3. CLI or run-of-the-mill script/utility (typer.pex)

A `click`/`typer` CLI which has it's own commands/subcommands - so while I arguably COULD use a busy box here - I didn't, because the CLI tool itself handled that - and in this case, I could just run the embedded Python against my pex with the appropriate entrypoint/script setup.

This would be similar if I was just running a one-off script too. Use pex to setup the correct entrypoint/script and then get the embedded Python to run it.

```toml
[[lift.commands]]
exe = "#{cpython:python}"
args = [
    "{:typer-pex}",
]
```

---

4. Permutation of the above pexes all bundled together

---

I don't know what an intuitive way of creating that would be via the command line, as it already gets tricky even using `toml`.

jsirois: You #2 has nothing to do with how to specify an entry point, which is all the script vs module entry point syntaxes do.

You get #2 right now (minus sealing in cli.args like `--port` by creating a BusyBox and ensuring the output file name matches the command you wish to default.

I can imagine supporting sealing in args, envy bars Python args via the pre-existing PEX options for those IFF -c or -m is used in the CLI args and 1 of the scie-busybox commands uses that script (-c) or that module entry point (-m). That could be a purely additive functionality follow up PR though.

sureshjoshi: > it is absolutely not a globbing specifier. I think you misread, but that just damns the idea I guess then

Yep, I definitely mis-understood the intention of that.

>  by creating a BusyBox and ensuring the output file name matches the command you wish to default

👍🏽

jsirois: Alright - this is getting ~nowhere. Let me step back and add a `--scie` doc page. I actually like where the overloaded `:` initial proposal got pushed to by @huonw with `:`, `=` and `@` dismbiguating the various punctuation roles. It still leaves the wildcard question open, just nothing in the position to wildcard that position vs an excplicit wildcard symbol (`*` or `:all:` or ...), but it at least should provide a better format to explain the complex-ish interface to specifying BusyBox entry points.

huonw: > Instead of adding flags, this un-overloads syntax. : used to mean many things, now it means just index a function in a module, which is pre-existing Python-ecosystem syntax.

Good approach, I like it!

With that change, I'm happy where this landed. Either of `*@*` or `:all:@:all:` are fine by me, definitely agree that `:all:` doesn't look particularly nice... tolerable (to me), but not nice.

I'm unsure where we landed with validation of the `@`-specified module scripts, but definitely 100% happy to go with `*` if there is validation. And like 80% happy to go with `*` without the validation. I.e. happy with `*` either way 👍

jsirois: I actually lean towards dropping the wildcard with the switch from `:` -> (`=`, `:`, `@`). There is no need for a wildcard (just as there was none for `:`), but I think `@ansible-core`, `!@ansible-core`, and (less so) `@` all make sense on their own side-stepping the shell potentially turning `*` into invalid entry points. As I said above though, I'm writing up a documentation page to go with this PR; so we'll see how the explanation reads there.

jsirois: Hoisted on my own petard. Can't use `@` since argparse is configured with `@fromfile` support via a user contribution in 2021. I'll noodle a bit more...

jsirois: > Ignoring the pre-compile bindings, this looks like a use case for the adhoc=module.deeper:func style, am I correct?

Aha, yes - for the "seed" command, not the default command. There were 2 things going on in that second example and I didn't read closely enough. So, you'd run:
```
pex uvicorn -c uvicorn --inject-args "apigateway.main:app --port 7999" --scie eager --scie-busybox seed=seeder:main,uvicorn -o uvicorn
```

I've just finished the --inject-args / --inject-python-args / --inject-env support, which was a bug to be missing under the "behaves just like the PEX" mantra.

jsirois: Ok. @sureshjoshi, @huonw and @benjyw probably worth another look. I've switched to `@` syntax, made the needed change in the Pex argparse setup to still support @fromfiles, fixed `--inject*` propagation to the matching BusyBox command, and added a new doc page rendered here: https://jsirois.github.io/pex/scie.html.

pex/pep_427.py
```diff
@@ -282,34 +282,49 @@ def record_files(
...
-                    # -*- coding: utf-8 -*-
-                    import importlib
-                    import sys
-
-                    object_ref = "{object_ref}"
-                    modname, qualname_separator, qualname = object_ref.partition(':')
-                    entry_point = importlib.import_module(modname)
-                    if qualname_separator:
-                        for attr in qualname.split('.'):
-                            entry_point = getattr(entry_point, attr)
```
jsirois, line 303: The re-factoring of `{EntryPoint,CallableEntryPoint}` -> `NamedEntryPoint <<containing>> {ModuleEntryPoint,CallableEntryPoint}` highlighted that this generated console script was wrong for modules (i.e.: when this if branch is not taken) since a module is not a callable! It turns out that in practice, console script entrypoints are always functions (callables) and not modules and so no bug reports. That said, I switched Pex's own `pex` entrypoint from a function to the `pex` module in its own `pyproject.toml` and verified that the pex wheel still builds fine and then a Pex PEX venv built from that wheel had a `pex` console script that then failed. Thus, the fix.

docs/scie.md
```diff
@@ -0,0 +1,310 @@
+# PEX scies
```
huonw, line 1: I think this is a really impactful feature, but users are very likely to not know what "scie" refers to, and the lede is somewhat buried, both in the sidebar and the text. 

image

I think people will be able to understand and discover the feature more easily a heading like:

```suggestion
# PEX with included Python interpreter
```

Potentially the first two sentences could be flipped too, e.g.:

> PEX supports building "scie" executables that contain both a CPython interpreter and your PEX'd code, in addition to traditional [PEP-441](https://peps.python.org/pep-0441/) PEX zip file. You can create a PEX scie by adding `--scie eager` to your `pex` command line.

huonw: Other that, I like the docs a lot 👍

pex/scie/model.py
```diff
@@ -33,86 +34,166 @@ class Value(Enum.Value):
...
+                    else:
+                        yield NamedEntryPoint(
+                            name=dist_entry_point.name, entry_point=dist_entry_point.entry_point
+                        )
 
-    name = attr.ib()  # type: str
-    entry_point = attr.ib()  # type: str
+            if not_found or wrong_project:
+                raise ValueError(
+                    # TODO(John Sirois): XXX: Craft an error message.
```
huonw, line 182: Reminder about this

huonw (approved):
I like it, only minor feedback

docs/scie.md
jsirois: Yeah, agreed - scie means nothing still. Fixed.

pex/scie/model.py
jsirois: Fixed and added test coverage for missing / mis-placed scripts.

docs/scie.md
jsirois: Re-rendered: https://jsirois.github.io/pex/scie.html

jsirois:
@sureshjoshi I'd like to proceed with this by tomorrow (August 1st) evening. Let me know if you need more time to review; otherwise, I'll assume you're fine with the `@` syntax and no explicit wildcards.

sureshjoshi:
@jsirois I'm unavailable for any in-depth review until probably Tuesday/Wednesday - but I think we've converged on something and the docs make sense to me

sureshjoshi (approved):

PR #2490, opened 2024-08-02
Allow `--inject-env` overrides for `--scie-busybox`.

jsirois:
Previously, the injected environment variables could not be overriden by
setting them in the environment the PEX scie busybox was executed in.
This behavior did not match traditional PEX behavior.

### DIFF
--- a/tests/integration/scie/test_pex_scie.py
+++ b/tests/integration/scie/test_pex_scie.py
@@ -424,23 +424,31 @@ def make_project(
                 """\
                 from __future__ import print_function
 
+                import functools
+                import os
                 import sys
 
 
-                def one():
-                    print("{name}1:", *sys.argv[1:])
+                def _print(label=""):
+                    env_var = {name!r}.upper()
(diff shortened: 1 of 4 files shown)


tests/integration/scie/test_pex_scie.py
```diff
@@ -737,6 +753,151 @@ def test_scie_busybox_console_script_injections(
...
+        args=[os.path.join(bin_dir, "bar-script1"), "injected?"]
+    )
+    assert b"bar2: injected?\n" == subprocess.check_output(
+        args=[os.path.join(bin_dir, "bar-script2"), "injected?"]
+    )
+
+
+@skip_if_no_provider
+@execution_mode_args
+def test_scie_busybox_console_script_inject_python_args(
```
jsirois, line 819: The `--inject-python-args` needed no fixing, but I added the this test to make sure all 3 injection types were covered.

benjyw (approved):

PR #2496, opened 2024-08-06
Add PEX info to the PEX repl.

### LINKED ISSUE #157: First example in docs doesn't work (or I'm understanding it wrong?)
https://pex.readthedocs.org/en/stable/buildingpex.html says:

> The easiest way to build .pex files is with the pex utility, which is made available when you pip install pex. Do this within a virtualenv, then you can use pex to bootstrap itself:
> 
> ```
> $ pex pex requests -c pex -o ~/bin/pex
> ```
> 
> This command creates a pex file containing pex and requests, using the console script named “pex”, saving it in ~/bin/pex. At this point, assuming ~/bin is on your $PATH, then you can use pex in or outside of any virtualenv.

But when I try building a .pex that way, it isn't able to `import requests`:

```
$ pex pex requests -c pex -o /tmp/pex
$ /tmp/pex
Python 2.7.9 (default, Feb 10 2015, 03:29:19)
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.56)] on darwin
(InteractiveConsole)
>>> import requests
ImportError: No module named requests
```

Using pex-1.0.3:

```
$ pip show pex

---
Metadata-Version: 2.0
Name: pex
Version: 1.0.3
Summary: The PEX packaging toolchain.
Home-page: https://github.com/pantsbuild/pex
Author: UNKNOWN
Author-email: UNKNOWN
License: Apache License, Version 2.0
Location: /usr/local/lib/python2.7/site-packages
Requires: setuptools
```

jsirois:
It should now be clear when you've dropped into a PEX repl (vs. a plain
Python repl). This may help stem some of the confusion brough about by
running `pex` with no arguments. In addition, you gain some
introspection capabilities via the `pex_info` REPL command.

Fixes #157


jsirois:
For the #157 OP case of running a PEX pex with no args:
image

And the more general case of a normal PEX repl built via:
```console
:; pex --interpreter-constraint ">=3.8" --pip-version latest requests --sh-boot -orequests.pex
```
image

jsirois:
Reviewers, another big one - but broken up into its bits anyhow. Thanks in advance.

This will probably be the last of these. It's not clear to me review makes much sense anymore except when I actually have questions you might help answer. I don't really like lobbing these things at you guys and I'm sure you don't enjoy being lobbed at.

benjyw (approved):
Splitting this into multiple coherent commits made it quite straightforward to review, thanks for that!

benjyw:
> Reviewers, another big one - but broken up into its bits anyhow. Thanks in advance.
> 
> This will probably be the last of these. It's not clear to me review makes much sense anymore except when I actually have questions you might help answer. I don't really like lobbing these things at you guys and I'm sure you don't enjoy being lobbed at.

I am happy to do any code reviews that are actually useful to you! Or when you think it might improve something. But just rubber-stamping makes little sense, I agree.

Issue #2201, opened 2023-08-01
proposal: `pex3 cache` introspection/gc command


cosmicexplorer:
### Discussed in https://github.com/pantsbuild/pex/discussions/2200

<div type='discussions-op-text'>

<sup>Originally posted by **cosmicexplorer** August  1, 2023</sup>
*forked from a response to https://github.com/pantsbuild/pex/pull/2175#issuecomment-1652279145:*

> There is an expense here in ~duplicating cached zips and Pants / Pex are already both notorious amongst users for excessive cache sizes
> Without that, this feature definitely needs to be behind a flag (--i-opt-in-to-cache-doubling - clearly not spelled like that!). Now you already mentioned being behind a flag, so I think you're on board there.

# Cache GC Policies
Generalizing this a bit, I recall that pantsd used to have a flag for how often it garbage collects the rust store--if there are concerns about the bloat of pex cache directories, are there any opportunities for pex itself to help the user automate the cache management outside of just `rm -rf ~/.pex`? What is currently the easiest way to implement e.g. LRU eviction? I guess I can do something like this?
```bash
> find ~/.pex -atime '+30' -or -atime '+7' -size '+300M' -type f -exec rm -rf '+'
```
The above probably works, but I'm wondering if the dilemma about cache bloat that you describe is partially because the user isn't given enough tools to mediate it? Or am I misinterpreting you?

## Insight: evict cache entries based on *usage frequency*

In particular, one GC heuristic that pex (or pip) itself would be in the best place to record is not just how *recently* each cache entry was accessed, but how *often*. Something like this could be fun:
```bash
> pex3 cache evict -accessed '>30 days' -or \( -size '>300M' -accessed '<1 per 1 day' \)
448M    ~/.pex/stitched_dists/7a4763a35d0824ebb172b00f8d0241ff231404c4b7d97dd5ea870d5afca336a4/tensorflow_gpu-2.5.3-cp38-cp38-manylinux2010_x86_64.whl
...
2.5GB to be removed. Delete? [Y/n] y
2.5GB deleted.
```
Does that sound like a fruitful thing to investigate further? Or are there better ways to address the disk usage pressure?

# Prior Art
*Examples of this from other tools:*

## `pip` example
One useful bit of prior art is the new `pip cache` subcommand within pip (it's on the `main` branch, not sure which version it first appeared in):
```bash
[...]

PR #2512, opened 2024-08-20
Introduce `--pre-resolved-dists` resolver.

### LINKED ISSUE #1907: Pex CLI doesn't offer as much control over pex building process as the PEXBuilder
As a major PEX user twitter leverages the PEXBuilder class to build pexes instead of shelling out to the PEX cli for a few reasons.

1. The `--sources-directory` option to add sources to a pex doesn't offer fine enough control over the final import paths of source files. In order to make `--sources-directory` work we would have to prepare all the sources in merged directories. E.g Often multiple directories need to be merged into the same Chroot directory.
2. We make use of PEXBuilder.add_dist_location to add pre-installed wheels (unzipped wheels) to pexes. We have found that the overhead of using direct file URLs via `--requirement` on the CLI and even the main `PEXBuilder.add_distribution` methods to add significant overhead by invoking pip.pex for each distribution. We can improve build times significantly by relying on the existence of a mechanism for fetching 3rdparties that is outside of pex itself.
3. ... will update if I think of more ...

These may not be things that the Pex CLI wants to support but figured I would list them out based on the fact that PEXBuilder isn't under a stable API contract with users.

jsirois:
When building a PEX or creating a venv via `pex3 venv create` you can
now tell Pex to use a set of pre-resolved distributions. This is similar
to using `--no-pypi --find-links ...` except that:
1. Its roughly 3x faster since Pip is not asked to do any resolution.
2. It requires all the distributions specified form an already complete
   resolve.

One way to obtain distributions that meet criteria 2 is to use
`pip download -d ...` or `pip wheel -w ...` to pre-resolve the
distributions you need.

Closes #1907

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,15 @@
 # Release Notes
 
+## 2.19.0
+
+This release adds support for a new `--pre-resolved-dists` resolver as
+an alternative to the existing Pip resolver, `--lock` resolver and
+`--pex-repository` resolvers. Using `--pre-resolved-dists dists/dir/`
+behaves much like `--no-pypi --find-links dists/dir/` except that it is
+roughly 3x faster.
+
+* Support `--pre-resolved-dists` resolver. (#2512)
+
 ## 2.18.1
 
 This release fixes `--scie-name-style platform-parent-dir` introduced in
--- a/pex/bin/pex.py
+++ b/pex/bin/pex.py
@@ -53,6 +53,7 @@ from pex.resolve.requirement_configuration import RequirementConfiguration
 from pex.resolve.resolver_configuration import (
     LockRepositoryConfiguration,
     PexRepositoryConfiguration,
+    PreResolvedConfiguration,
 )
 from pex.resolve.resolver_options import create_pip_configuration
 from pex.resolve.resolvers import Unsatisfiable, sorted_requirements
@@ -136,7 +137,9 @@ def configure_clp_pex_resolution(parser):
         ),
     )
 
-    resolver_options.register(group, include_pex_repository=True, include_lock=True)
+    resolver_options.register(
+        group, include_pex_repository=True, include_lock=True, include_pre_resolved=True
+    )
 
     group.add_argument(
         "--pex-path",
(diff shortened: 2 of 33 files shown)


jsirois:
@hrfuller I think this solves the last discrepancy in #1907. Although you can't add dists via the Pex API, you can do so using this new `--pre-resolved-dists`. I know you're now separated from your original use case; so can't test the speed difference, but it should be ~same. I included some perf test results above at any rate. One advantage over adding dists via the API is that this supports sub-setting; i.e. point at one big consistent resolve pool of dists with `--pre-resolved-dists ...` and you can resolve the whole thing or any subset by just changing the root requirement list you hand Pex.

jsirois:
Pants folks, I added you since this may be one way to get `uv` in the loop, since its a way to have Pex step out of the resolve loop. Although neither [`uv pip download`](https://github.com/astral-sh/uv/issues/3163) nor [`uv pip wheel`](https://github.com/astral-sh/uv/issues/1681) exist today, they don't sound completely opposed. They just need someone to do the work.

jsirois:
@zmanji this might be a way for you to get offline Pex support, re: https://github.com/pex-tool/pex/issues/2201#issuecomment-1694144209

jsirois:
N.B.: The bulk of the delta lines is a test lockfile at ~1800 lines.

zmanji:
> @zmanji this might be a way for you to get offline Pex support, re: [#2201 (comment)](https://github.com/pex-tool/pex/issues/2201#issuecomment-1694144209)

Yes with this flag pex can work in a docker container without internet access or with nix or similar.

I am assuming the workflow if you have a lockfile will be:
1. Export the lock to `requirements.txt`, creating a subset if needed
2. Use pip to download all of the artifacts.
3. Use `--pre-resolved-dists` with the `requirements.txt` to build the resulting pex.

jsirois:
@zmanji yes, exactly. The tests do this here:
https://github.com/pex-tool/pex/pull/2512/files#diff-3aad399daa6b77a0c3e83cdca2225461ef07f4aebe306abdb80ed4b8bfd2fb19R39-R45

You do have to ensure that either the download or the original lock, though, is 100% wheels of course.

zmanji (approved):
Overall the code/tests and idea lgtm!

PR #2513, opened 2024-08-21
Introduce `pex3 cache {dir,info,purge}`.

### LINKED ISSUE #1176: Pex interpreter cache is not upgrade / downgrade safe.
As reported in #1175, a PEX file run on a machine that has a `~/.pex` cache populated by older versions of Pex can encounter uncaught errors reading old INTERP-INFO files. Unlike the rest of the `~/.pex` cache structure which is directory name based with hashes, the interpreter cache also includes the INTERP-INFO file format; so changes to that format should be handled gracefully. In particular, most simply, Pex should probably just nuke / re-generate any INTERP-INFO it cannot read.

### LINKED ISSUE #1655: Evaluate PEX_ROOT cache structure.
There are a few things to evaluate:
1. Versioning: Likely the cache should be versioned to allow clean incompatible changes to structure.
2. Hashing: cache keys have been built off sha1 content hashes since day1. Analyze attack vulnerability and see if sha256 could be used in (most) places. There are some places, maybe only in venvs, where path length becomes a problem due to shebang length limits.
3. Scavenging: What protions of the cache can be scavenged without affecting performance of X too badly. This is effectively about coming up with a pex3 cache tool that allows inspection and selective pruning.

### LINKED ISSUE #2201: proposal: `pex3 cache` introspection/gc command

jsirois:
Re-structure the Pex cache to both support versioning as well as adding
access tracking for shared (normal) use and for exclusive use when 
portions of the cache need to be deleted. With this new ground work, add
a new `pex3 cache {dir,info,purge}` family of commands for inspecting
and safely trimming the Pex cache.

Closes #1176
Closes #1655
Closes #2201

### DIFF
--- a/pex/cli/commands/cache/command.py
+++ b/pex/cli/commands/cache/command.py
@@ -1,0 +1,391 @@
+# Copyright 2024 Pex project contributors.
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+from __future__ import absolute_import, print_function
+
+import os
+from argparse import Action, ArgumentError, _ActionsContainer
+from datetime import datetime
+
+from pex.cache import access as cache_access
+from pex.cache.dirs import CacheDir
+from pex.cli.command import BuildTimeCommand
+from pex.cli.commands.cache.bytes import ByteAmount, ByteUnits
+from pex.cli.commands.cache.du import DiskUsage
+from pex.commands.command import OutputMixin
+from pex.common import pluralize, safe_rmtree
+from pex.exceptions import reportable_unexpected_error_msg
+from pex.jobs import iter_map_parallel, map_parallel
+from pex.orderedset import OrderedSet
+from pex.result import Error, Ok, Result
+from pex.typing import TYPE_CHECKING
+from pex.variables import ENV
+
+if TYPE_CHECKING:
+    from typing import IO, Dict, Iterable, List, Optional, Tuple, Union
+
+
+class HandleAmountAction(Action):
+    def __init__(self, *args, **kwargs):
+        kwargs["nargs"] = 0
+        super(HandleAmountAction, self).__init__(*args, **kwargs)
+
+    def __call__(self, parser, namespace, value, option_str=None):
+        if option_str in ("-H", "--human-readable"):
+            amount_func = ByteAmount.human_readable
+        elif option_str.startswith("--"):
+            amount_func = ByteAmount.for_unit(ByteUnits.for_value(option_str[2:]))
+        else:
+            raise ArgumentError(
+                argument=self,
+                message=reportable_unexpected_error_msg(
+                    "The HandleAmountAction was used in an unexpected place."
+                ),
+            )
+        setattr(namespace, self.dest, amount_func)
+
+
+class Cache(OutputMixin, BuildTimeCommand):
+    """Interact with the Pex cache."""
+
+    @staticmethod
+    def _add_amount_argument(parser):
+        # type: (_ActionsContainer) -> None
+
+        parser.add_argument(
+            "-H",
+            "--human-readable",
+            *["--{unit}".format(unit=unit.value) for unit in ByteUnits.values()],
+            dest="amount_func",
+            action=HandleAmountAction,
+            default=ByteAmount.bytes,
+            help=(
+                "How to display disk usage amounts; defaults to bytes. The -H / --human-readable "
+                "options display amounts in a human readable way and each of the other unit "
+                "options displays amounts with that specified unit."
+            )
+        )
+
+    @classmethod
+    def _add_info_arguments(cls, parser):
+        # type: (_ActionsContainer) -> None
+
+        cls._add_amount_argument(parser)
+        parser.add_argument(
+            "-S",
+            "--sort-size",
+            dest="sort_by_size",
+            action="store_true",
+            help=(
+                "Don't actually purge cache entries; instead, perform a dry run that just prints "
+                "out what actions would be taken"
+            ),
+        )
+        parser.add_argument(
+            "-r",
+            "--reverse",
+            dest="reverse",
+            action="store_true",
+            help=(
+                "Don't actually purge cache entries; instead, perform a dry run that just prints "
+                "out what actions would be taken"
+            ),
+        )
+        cls.add_output_option(parser, entity="Pex cache information")
+
+    @classmethod
+    def _add_purge_arguments(cls, parser):
+        # type: (_ActionsContainer) -> None
+
+        cls._add_amount_argument(parser)
+        parser.add_argument(
+            "--entries",
+            action="append",
+            type=CacheDir.for_value,
+            choices=CacheDir.values(),
+            default=[],
+            help=(
+                "Specific cache entries to purge. By default, all entries are purged, but by "
+                "specifying one or more particular --entries to purge, only those entries (and any "
+                "other cache entries dependent on those) will be purged."
+            ),
+        )
+        parser.add_argument(
+            "-n",
+            "--dry-run",
+            dest="dry_run",
+            action="store_true",
+            help=(
+                "Don't actually purge cache entries; instead, perform a dry run that just prints "
+                "out what actions would be taken"
+            ),
+        )
+        parser.add_argument(
+            "-R",
+            "--report-savings",
+            dest="report_savings",
+            action="store_true",
+            help="Report the disk usage savings from the purge.",
+        )
+
+        cls.add_output_option(parser, entity="Pex purge results")
+
+    @classmethod
+    def add_extra_arguments(cls, parser):
+        subcommands = cls.create_subcommands(
+            parser,
+            description="Interact with the Pex cache via the following subcommands.",
+        )
+
+        with subcommands.parser(
+            name="dir",
+            help="Print the current Pex cache directory path.",
+            func=cls._dir,
+            include_verbosity=False,
+        ) as dir_parser:
+            cls.add_output_option(dir_parser, entity="Pex cache directory")
+
+        with subcommands.parser(
+            name="info",
+            help="Present information about Pex cache status.",
+            func=cls._info,
+            include_verbosity=False,
+        ) as inspect_parser:
+            cls._add_info_arguments(inspect_parser)
+
+        with subcommands.parser(
+            name="purge",
+            help="Purge the Pex cache safely.",
+            func=cls._purge,
+            include_verbosity=False,
+        ) as purge_parser:
+            cls._add_purge_arguments(purge_parser)
+
+    def _dir(self):
+        # type: () -> Result
+
+        with self.output(self.options) as fp:
+            print(ENV.PEX_ROOT, file=fp)
+
+        return Ok()
+
+    @staticmethod
+    def _collect_info(cache_dir):
+        # type: (CacheDir.Value) -> Tuple[CacheDir.Value, DiskUsage]
+        return cache_dir, DiskUsage.collect(cache_dir.path())
+
+    def _render_usage(self, disk_usage):
+        # type: (Union[DiskUsage, Iterable[DiskUsage]]) -> str
+
+        if isinstance(disk_usage, DiskUsage):
+            du = disk_usage
+            prefix = ""
+        else:
+            du = DiskUsage.aggregate(path=ENV.PEX_ROOT, usages=disk_usage)
+            prefix = "Total: "
+
+        return "{prefix}{size} in {subdir_count} {subdirs} and {file_count} {files}.".format(
+            prefix=prefix,
+            size=self.options.amount_func(du.size),
+            subdir_count=du.subdirs,
+            subdirs=pluralize(du.subdirs, "subdirectory"),
+            file_count=du.files,
+            files=pluralize(du.files, "file"),
+        )
+
+    def _info(self):
+        # type: () -> Result
+
+        with self.output(self.options) as fp:
+            print("Path: {path}".format(path=ENV.PEX_ROOT), file=fp)
+            print(file=fp)
+
+            disk_usages = []  # type: List[DiskUsage]
+            for cache_dir, disk_usage in sorted(
+                map_parallel(
+                    inputs=CacheDir.values(), function=self._collect_info, noun="cache directory"
+                ),
+                key=lambda info: (
+                    (info[1].size, info[0].name) if self.options.sort_by_size else info[0].name
+                ),
+                reverse=self.options.reverse,
+            ):
+                disk_usages.append(disk_usage)
+                print(
+                    "{name}: {path}".format(name=cache_dir.name, path=cache_dir.rel_path), file=fp
+                )
+                print(cache_dir.description, file=fp)
+                print(self._render_usage(disk_usage), file=fp)
+                print(file=fp)
+            print(self._render_usage(disk_usages), file=fp)
+            print(file=fp)
+
+        return Ok()
+
+    def _purge_cache_dir(self, cache_dir):
+        # type: (CacheDir.Value) -> Tuple[CacheDir.Value, Optional[DiskUsage]]
+        cache_dir_path = cache_dir.path()
+        du = DiskUsage.collect(cache_dir_path) if self.options.report_savings else None
+        if not self.options.dry_run:
+            safe_rmtree(cache_dir_path)
+        return cache_dir, du
+
+    @staticmethod
+    def _log_delete_start(
+        lock_file,  # type: str
+        out,  # type: IO[str]
+    ):
+        # type: (...) -> None
+
+        try:
+            import psutil  # type: ignore[import]
+
+            # N.B.: the `version_info` tuple was introduced in psutil 0.2.0.
+            psutil_version = getattr(psutil, "version_info", None)
+            if not psutil_version or psutil_version < (5, 3, 0):
+                print(
+                    "The psutil{version} module is available, but it is too old. "
+                    "Need at least version 5.3.0.".format(
+                        version=(
+                            " {version}".format(version=".".join(map(str, psutil_version)))
+                            if psutil_version
+                            else ""
+                        )
+                    ),
+                    file=out,
+                )
+                psutil = None
+        except ImportError as e:
+            print("Failed to import psutil:", e, file=out)
+            psutil = None
+
+        if not psutil:
+            print("Will proceed with basic output.", file=out)
+            print("---", file=out)
+            print(
+                "Note: this process will block until all other running Pex processes have exited.",
+                file=out,
+            )
+            print(
+                "To get information on which processes these are, re-install Pex with the", file=out
+            )
+            print("management extra; e.g.: with requirement pex[management]", file=out)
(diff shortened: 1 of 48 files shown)


jsirois:
Pants folks - I added you since you have: https://github.com/pantsbuild/pants/issues/11167

This initial introduction of Pex cache management commands does not have any JSON output option - its just for humans currently. The linked Pants issue seems to require structured information though. Much like Pants needing to come up with a representation it prefers for dependency graphs (I assume that's still not done), this integration point also needs similar spec'ing. If you have opinions, or better, a spec, I'd be happy to follow up with support for various alternate output formats for the cache usage information.

pyproject.toml
```diff
@@ -72,6 +72,11 @@ classifiers = [
 subprocess = [
   "subprocess32>=3.2.7; python_version < '3'"
 ]
+management = [
+  # N.B.: Released on 2017-09-01 and added support for the `process_iter(attrs, ad_value)` API we
+  # use in `pex.cache.access`.
+  "psutil>=5.3"
```
jsirois, line 78: N.B.: If Pants wants to use these cache management commands and it still uses the Pex PEX releases, I'll either need to add a `pex+management` PEX to the release that embeds psutil for all supported platforms or else start releasing Pex PEX scies, which naturally handle platform specific deps. I favor the scies here since the pex+management approach doesn't scale well once more CLI-specific-deps are added, but I wanted to get Pants maintainers opinions on this.

jsirois: ... there's a longer story on why I went with psutil / access info gathering at delete-time, but suffice it to say, recording usage info in the read-write access lock path (I used sqlite3) added ~2-30ms overhead to every PEX launch and I did not deem that acceptable.

jsirois:
Some examples.

Get cache info sorted by size:
```console
:; python -mpex.cli cache info -HS
Path: /home/jsirois/.cache/pex

Pex Docs: docs/0
Artifacts used in serving Pex docs via `pex --docs` and `pex3 docs`.
0 bytes in 0 subdirectories and 0 files.

Abbreviated Platforms: platforms/0
Information calculated about abbreviated platforms specified via `--platform`.
291 kB in 13 subdirectories and 26 files.

User Code: user_code/0
User code added to PEX files using `-D` / `--sources-directory`, `-P` / `--package` and `-M` / `--module`.
531 kB in 117 subdirectories and 168 files.

Packed Bootstraps: bootstrap_zips/0
PEX runtime bootstrap code, zipped up for `--layout packed` PEXes.
1.42 MB in 2 subdirectories and 4 files.

Unzipped PEXes: unzipped_pexes/0
The unzipped PEX files executed on this machine.
3.12 MB in 483 subdirectories and 1824 files.

Packed Wheels: packed_wheels/0
The same content as 'installed_wheels/0', but zipped up for `--layout packed` PEXes.
6.32 MB in 6 subdirectories and 12 files.

Interpreters: interpreters/0
Information about interpreters found on the system.
14.4 MB in 374 subdirectories and 704 files.

Bootstraps: bootstraps/0
PEX runtime bootstrap code.
19.5 MB in 125 subdirectories and 1657 files.

Pex Tools: tools/0
Caches for the various `PEX_TOOLS=1` / `pex-tools` subcommands.
21.4 MB in 254 subdirectories and 1886 files.

Built Wheels: built_wheels/0
Wheels built by Pex from resolved sdists when creating PEX files.
29.6 MB in 156 subdirectories and 59 files.

Isolated Pex Code: isolated/0
The Pex codebase isolated for internal use in subprocesses.
45.0 MB in 367 subdirectories and 3803 files.

Scie Tools: scies/0
Tools and caches used when building PEX scies via `--scie {eager,lazy}`.
151 MB in 5 subdirectories and 40 files.

Lock Artifact Downloads: downloads/0
Distributions downloaded when resolving from a Pex lock file.
236 MB in 379 subdirectories and 1980 files.

Pip Versions: pip/0
Isolated Pip caches and Pip PEXes Pex uses to resolve distributions.
324 MB in 3349 subdirectories and 17300 files.

Virtual Environments: venvs/0
Virtual environments generated at runtime for `--venv` mode PEXes.
583 MB in 9165 subdirectories and 40569 files.

Pre-installed Wheels: installed_wheels/0
Pre-installed wheel chroots used to both build PEXes and serve as runtime `sys.path` entries.
961 MB in 8975 subdirectories and 51739 files.

Total: 2.40 GB in 23770 subdirectories and 121771 files.
```

Dry run purge of just installed_wheels cache:
```console
:; python -mpex.cli cache purge --entries installed_wheels -nRH
Would purge requested entries from /home/jsirois/.cache/pex: installed_wheels/0
Would also purge those entries transitive dependents in: unzipped_pexes/0, venvs/0

Would have purged cache Unzipped PEXes from unzipped_pexes/0
3.12 MB in 483 subdirectories and 1824 files.

Would have purged cache Pre-installed Wheels from installed_wheels/0
961 MB in 8975 subdirectories and 51739 files.

Would have purged cache Virtual Environments from venvs/0
583 MB in 9165 subdirectories and 40569 files.

Total: 1.55 GB in 18623 subdirectories and 94132 files.
```

And go for it (no psutil):
```console
:; python -mpex.cli cache purge --entries installed_wheels -RH
Purging requested entries from /home/jsirois/.cache/pex: installed_wheels/0
Also purging those entries transitive dependents in: unzipped_pexes/0, venvs/0

Failed to import psutil: No module named 'psutil'
Will proceed with basic output.
---
Note: this process will block until all other running Pex processes have exited.
To get information on which processes these are, re-install Pex with the
management extra; e.g.: with requirement pex[management]

Attempting to acquire cache write lock (press CTRL-C to abort) ...
^C
No cache entries purged.
```

With psutil:
```console
:; python -mpex.cli cache purge --entries installed_wheels -RH
Purging requested entries from /home/jsirois/.cache/pex: installed_wheels/0
Also purging those entries transitive dependents in: unzipped_pexes/0, venvs/0

Waiting on 2 in flight processes (with shared lock on /home/jsirois/.cache/pex/access.lck) to complete before deleting:
---
1. pid 281904 started by jsirois at 2024-09-02 15:45:24
   Pex env: {'PEX': '/home/jsirois/dev/pex-tool/pex/empty.pex'}
   cmdline: ['/home/jsirois/.pyenv/versions/3.11.9/bin/python3.11', '/home/jsirois/.cache/pex/unzipped_pexes/0/292f879052303680091fdcb445c2a746967b4e0f']
2. pid 282594 started by jsirois at 2024-09-02 15:45:50
   Pex env: {'PEX_TOOLS': '1', 'PEX': '/home/jsirois/dev/pex-tool/pex/empty-tools.pex'}
   cmdline: ['/home/jsirois/.pyenv/versions/3.11.9/bin/python3.11', '/home/jsirois/.cache/pex/unzipped_pexes/0/2c278c488639385dd1ca190c245fc4e8da7a0f30', 'repository', 'extract', '-f', '/tmp/find-links', '--serve']

Attempting to acquire cache write lock (press CTRL-C to abort) ...
^C
No cache entries purged.
```

And ending the processes with the shared lock:
```console
:; python -mpex.cli cache purge --entries installed_wheels -RH
Purging requested entries from /home/jsirois/.cache/pex: installed_wheels/0
Also purging those entries transitive dependents in: unzipped_pexes/0, venvs/0

Waiting on 2 in flight processes (with shared lock on /home/jsirois/.cache/pex/access.lck) to complete before deleting:
---
1. pid 281904 started by jsirois at 2024-09-02 15:45:17
   Pex env: {'PEX': '/home/jsirois/dev/pex-tool/pex/empty.pex'}
   cmdline: ['/home/jsirois/.pyenv/versions/3.11.9/bin/python3.11', '/home/jsirois/.cache/pex/unzipped_pexes/0/292f879052303680091fdcb445c2a746967b4e0f']
2. pid 282594 started by jsirois at 2024-09-02 15:45:43
   Pex env: {'PEX_TOOLS': '1', 'PEX': '/home/jsirois/dev/pex-tool/pex/empty-tools.pex'}
   cmdline: ['/home/jsirois/.pyenv/versions/3.11.9/bin/python3.11', '/home/jsirois/.cache/pex/unzipped_pexes/0/2c278c488639385dd1ca190c245fc4e8da7a0f30', 'repository', 'extract', '-f', '/tmp/find-links', '--serve']

Attempting to acquire cache write lock (press CTRL-C to abort) ...

Purged cache Unzipped PEXes from unzipped_pexes/0
3.12 MB in 483 subdirectories and 1824 files.

Purged cache Pre-installed Wheels from installed_wheels/0
961 MB in 8975 subdirectories and 51739 files.

Purged cache Virtual Environments from venvs/0
583 MB in 9165 subdirectories and 40569 files.

Total: 1.55 GB in 18623 subdirectories and 94132 files.
```

pex/cli/commands/cache/command.py
```diff
@@ -0,0 +1,391 @@
...
+        # type: (_ActionsContainer) -> None
+
+        cls._add_amount_argument(parser)
+        parser.add_argument(
+            "-S",
+            "--sort-size",
+            dest="sort_by_size",
+            action="store_true",
+            help=(
+                "Don't actually purge cache entries; instead, perform a dry run that just prints "
```
benjyw, line 80: The help strings here and for `--reverse` need updating.

pex/cli/commands/cache/command.py
```diff
@@ -0,0 +1,391 @@
...
+            "-n",
+            "--dry-run",
+            dest="dry_run",
+            action="store_true",
+            help=(
+                "Don't actually purge cache entries; instead, perform a dry run that just prints "
+                "out what actions would be taken"
+            ),
+        )
+        parser.add_argument(
```
benjyw, line 123: Why wouldn't we always emit this?

benjyw (approved):
Really nice.

pyproject.toml
jsirois: Well, two reasons:
1. It means embedding all these wheels in the Pex PEX or the Pex+management PEX - so, in part size:
    + psutil-5.9.5-cp27-cp27m-macosx_10_9_x86_64.whl
    + psutil-5.9.5-cp27-cp27m-manylinux2010_i686.whl
    + psutil-5.9.5-cp27-cp27m-manylinux2010_x86_64.whl
    + psutil-5.9.5-cp27-cp27mu-manylinux2010_i686.whl
    + psutil-5.9.5-cp27-cp27mu-manylinux2010_x86_64.whl
    + psutil-5.9.5-cp36-abi3-macosx_10_9_x86_64.whl
    + psutil-5.9.5-cp36-abi3-manylinux_2_12_i686.manylinux2010_i686.manylinux_2_17_i686.manylinux2014_i686.whl
    + psutil-5.9.5-cp36-abi3-manylinux_2_12_x86_64.manylinux2010_x86_64.manylinux_2_17_x86_64.manylinux2014_x86_64.whl
    + psutil-5.9.5-cp38-abi3-macosx_11_0_arm64.whl
2. But ut means inventing a bit of new functionality to support a `Universal` `Target` to complement the current `AbbreviatedPlatform`, `CompletePLatform` and `LocalInterpreter` `Target`s - this new `Target` type would only work when the Pex resolve was against a --lock or a --pex-repository and it would be able to grab all the wheels listed in 1 despite a machine not having all the corresponding `LocalInterpreter` targets available.

pex/cli/commands/cache/command.py
jsirois: Collecting du with a cold page cache is slow.

That said, it should be roughly as slow as the rm. Since both do the same fs walk in close succession, it seems the second, rm walk only adds ~20% instead of doubling the time. Here I test with a ~1GB Pex cache:
```console
:; hyperfine -p 'rm -rf ~/.cache/pex && cp -r ~/.cache/pex-bak ~/.cache/pex && free && sync && sudo sh -c "echo 3 > /proc/sys/vm/drop_caches" && free && sleep 2' 'python -mpex.cli cache purge' 'pytho
n -mpex.cli cache purge -RH'
Benchmark 1: python -mpex.cli cache purge
  Time (mean ± σ):     851.7 ms ±  14.0 ms    [User: 327.0 ms, System: 711.2 ms]
  Range (min … max):   826.7 ms … 873.2 ms    10 runs

Benchmark 2: python -mpex.cli cache purge -RH
  Time (mean ± σ):     998.0 ms ±   8.4 ms    [User: 460.2 ms, System: 811.0 ms]
  Range (min … max):   986.5 ms … 1017.9 ms    10 runs

Summary
  python -mpex.cli cache purge ran
    1.17 ± 0.02 times faster than python -mpex.cli cache purge -RH
```

I guess having to wait 6 minutes instead of 5 for a ~300GB Pex cache is good enough for a 1st release. I can add the inverse `--no-report` later if needed.

pex/cli/commands/cache/command.py
jsirois: Fixed.

pyproject.toml
jsirois: Hrm, I think I need to do most of 2 anyhow to support creating a lock for the Pex scies with something like `pex3 lock create --project ".[management]" --pip-version latest --style universal --target-system linux --target-system mac --interpreter-constraint "CPython==3.12.*" -o pex-scie.lock`. Alternatively, I could create a manual workflow that generated a complete platform for each of the 4 supported platforms using the scie PBS and generate a strict multi-lock using those 4 `--complete-platform` targets.

benjyw: If that's a hassle then an alternative is for Pants to install pex as a dist, with the management extra, when it needs to run management commands?

jsirois: Aha, OK. Yeah - that would work. I should have the Pex scies + release PRs up today.

PR #2514, opened 2024-08-21
Support custom setuptools & wheel versions.

### LINKED ISSUE #1895: Expose setuptools and wheel version flags for use with non-vendored Pips.

jsirois:
When using non-vendored Pip (i.e.: specifying a custom `--pip-version`),
you can now override the versions of setuptools and wheel Pex bootstraps
for Pip with the existing `--extra-pip-requirement` option and the
override will be respected. Trying to override setuptools or wheel for
vendored Pip will raise an error however since its versions are
specialized to support all Pex resolve features under Python 2.7.

Fixes #1895

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,15 @@
 # Release Notes
 
+## 2.17.0
+
+This release brings support for overriding the versions of setuptools
+and wheel Pex bootstraps for non-vendored Pip versions (the modern ones
+you select with `--pip-version`) using the existing
+`--extra-pip-requirement` option introduced in the [2.10.0 release](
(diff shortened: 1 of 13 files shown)


jsirois:
@danxmoran had requested this feature via Pants some time ago; so adding Pants reviewers for this one.

huonw (approved):
Nice one

PR #2523, opened 2024-09-10
Add `--scie-only` & `--scie-name-style`.

jsirois:
These options allow the files output when using `--scie {eager,lazy}`
scies to be better controlled. The `--scie-only` option provides a way
to ensure no PEX file is output, just the resulting scie. The
`--scie-name-style` option allows control over how the scie's target
platform is incorporated in the scie file name.


jsirois:
@martina-oefelein I'm interested in your thoughts on the new `--scie-only` and `--scie-name-style` options. The bits you are likely most interested in ordered from most relevant to least:
+ New options: https://github.com/pex-tool/pex/pull/2523/commits/d570fb9d5f32e302c82fc03bc03dcf259eb573ce#diff-bbf96d2c6fdcaa284ebb9e1fc92f6485b122a18e9ed241d96e943c5a90fbe168R86-R115
+ Tests or the new options: https://github.com/pex-tool/pex/pull/2523/commits/d570fb9d5f32e302c82fc03bc03dcf259eb573ce#diff-9d49325ca5168fedd20faae90ae081421469df69d073fd83be97c9d7ebcaad3cR1085-R1125
+ Pex scie release dogfood of the new options: https://github.com/pex-tool/pex/pull/2523/commits/225dda6e1060b878cf82382bf9b289a1cba95a85#diff-312c2b9fe952517b20c422ea9493fe918be43dc68b8e47233c2d1990e9bdcbbeR109-R117

I think this incorporates all your feedback from #2516 given the hard requirement I have of not breaking backwards compatibility; i.e: I could not change how thinks work by default and I could only add options unfortunately to get the desired results.

jsirois:
Pants reviewers: this is not as bad as the diffstat looks. The change in [da66a85](https://github.com/pex-tool/pex/pull/2523/commits/da66a859e4157c8326393eb7ada827828a099a2a) was machine-generated and accounts for the lions share of the delta lines:
```
    Run `tox -e gen-scie-platform -- --all`.

 package/complete-platforms/linux-aarch64.json |  496 ++++++++++
 package/complete-platforms/linux-x86_64.json  |  874 +++++++++++++++++
 package/complete-platforms/macos-aarch64.json |  604 ++++++++++++
 package/complete-platforms/macos-x86_64.json  | 2629 ++++++++++++++++++++++++++++++++++++++++++++++++++
 package/pex-scie.lock                         |  138 +++
 5 files changed, 4741 insertions(+)
```

scripts/create-packages.py
```diff
@@ -53,6 +60,96 @@ def build_pex_pex(
...
+            "--scie-pbs-release",
+            scie_config.pbs_release,
+            "--scie-python-version",
+            scie_config.python_version,
+            "--scie-pbs-stripped",
+            "--scie-hash-alg",
+            "sha256",
+            "--scie-busybox",
+            "@pex",
+            "--scie-busybox-pex-entrypoint-env-passthrough",
```
jsirois, line 129: N.B.: This means the released Pex scies, unlike the Pex PEX, support all 3 top-level Pex command 1st class if you leave the file name as-is. For example:
```console
:; dist/pex-linux-x86_64
Error: Could not determine which command to run.

Please select from the following boot commands:

pex
pex-tools
pex3

You can select a boot command by setting the SCIE_BOOT environment variable or else by passing it as the 1st argument.
```

So you can:
```console
# Run `pex`:
:; dist/pex-linux-x86_64 pex ...

# Run `pex3`:
:; dist/pex-linux-x86_64 pex3 ...

# Run `pex-tools`:
:; dist/pex-linux-x86_64 pex-tools ...
```

Alternatively you can do something like:
```console
# Install the binaries with top-level names (the `--symlink` is optional):
:; SCIE=install dist/pex-linux-x86_64 --symlink bin

# Run `pex`:
:; bin/pex ...

# Run `pex3`:
:; bin/pex3 ...

# Run `pex-tools`:
:; bin/pex-tools ...
```

Finally, if you re-name the downloaded Pex scie to `pex`, it behaves ~just like the Pex PEX and you must use env vars to access the alternate entry points; e.g: `PEX_SCRIPT=pex3 pex ...`.

.github/workflows/release.yml
```diff
@@ -97,14 +97,20 @@ jobs:
...
         uses: a-scie/actions/changelog@v1.6
         with:
           changelog-file: ${{ github.workspace }}/CHANGES.md
           version: ${{ needs.determine-tag.outputs.release-version }}
+      - name: Append Hashes to Changelog
+        run: |
+          changelog_tmp="$(mktemp)"
+          cat "${{ steps.prepare-changelog.outputs.changelog-file }}" <(echo '***') dist/hashes.md \
+            > "${changelog_tmp}"
+          mv "${changelog_tmp}" "${{ steps.prepare-changelog.outputs.changelog-file }}"
```
jsirois, line 113: The upshot of this is the Releases will get this style content between the change-log and the files list below:
***
|file|sha256|size|
|----|------|----|
|pex|b5fb8b2469dd8f208a7805c7ec51850e81f4a4952e4808827b8e07589bc2496f|4306002|
|pex-linux-aarch64|1861382d2c273105a4b7013054e7acbab02f9ce78de7baa7d5e744c398eb4204|24854091|
|pex-linux-x86_64|e42f3231f605147ccb93118c2bb1386ca95dfac624b19cb405e91db9886cd7ad|28404512|
|pex-macos-aarch64|b1f18587833c9af437449dddf53dc778b1705eaace750b0f02ecee3bba5d2103|22320777|
|pex-macos-x86_64|fb8e9c580a9dc7ed0aae9f5b4ecad44287d4db585980400c2c3893687e545e74|22816650|

pex/bin/pex.py
```diff
@@ -1375,6 +1375,8 @@ def do_main(
                         ),
                         V=options.verbosity,
                     )
+                if scie_configuration.options.scie_only:
+                    os.unlink(pex_file)
```
benjyw, line 1379: I assume this means that in certain cases of early failure the .pex file will be left behind as cruft. I imagine that is fine, but @martina-oefelein may have thoughts.

benjyw (approved):
Thanks for the explanations throughout, will be helpful when integrating the PEX scies.

pex/bin/pex.py
jsirois: Yes, exactly. The existing scie creation code also goes to lengths to preserve the temporary lift.toml it generates when there is a scie building failure; so this intent to leave behind the inputs to the failure for inspection is consistent. In the lift.toml case the fact that it is being plopped out is actually printed to the console.

jsirois:
Alrighty, now available here: https://github.com/pex-tool/pex/releases/tag/v2.18.0

pex/bin/pex.py
martina-oefelein: Sorry, I was offline. Yes, I think leaving the .pex file behind on failure is fine.

PR #2548, opened 2024-10-02
Support `pex3 cache prune --older-than ...`.

jsirois:


pex/cache/data.py
```diff
@@ -0,0 +1,434 @@
...
+                bootstrap_hash, code_hash = cursor.fetchone()
+                yield BootstrapDir.create(bootstrap_hash)
+                yield UserCodeDir.create(code_hash)
+
+            with closing(
+                conn.execute(
+                    """
+                    SELECT name, install_hash, wheel_hash FROM wheels
+                    JOIN zipapp_deps ON zipapp_deps.wheel_install_hash = wheels.install_hash
+                    JOIN zipapps ON zipapps.pex_hash = zipapp_deps.pex_hash
```
jsirois, line 274: There is no need for this join.

pex/cache/data.py
```diff
@@ -0,0 +1,434 @@
...
+    # type: (...) -> Iterator[InstalledWheelDir]
+
+    with _db_connection(conn=connection) as conn:
+        for placeholders, keys in _iter_key_chunks(venv_dirs, extract_key=lambda v: v.short_hash):
+            with closing(
+                conn.execute(
+                    """
+                    SELECT name, install_hash, wheel_hash FROM wheels
+                    JOIN venv_deps ON venv_deps.wheel_install_hash = wheels.install_hash
+                    JOIN venvs ON venvs.short_hash = venv_deps.venv_hash
```
jsirois, line 301: There is no need for this join and there is also no need for the venvs table at all any more.

PR #2554, opened 2024-10-08
Plumb reproducible build env vars more thoroughly.

jsirois:
This helps reduce cache sizes where sdists are built into wheels as well
as expanding the set of sdists Pex can lock reproducibly.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,20 @@
 # Release Notes
 
+## 2.20.3
+
+This release fixes both PEX building and lock creation via
+`pex3 lock {create,sync}` to be reproducible in more cases. Previously,
+if a requirement only available in source form (an sdist, a local
+project or a VCS requirement) had a build that was not reproducible due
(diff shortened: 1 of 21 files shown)


jsirois:
For the record, the Pip cache size difference where the cache is blown away, then `tox -epy311-integration` is run before and after this change:
```
:; du -sh ~/.cache/pex.it-full/pip/0/ /home/jsirois/.cache/pex/pip/1/
389M    /home/jsirois/.cache/pex.it-full/pip/0/
362M    /home/jsirois/.cache/pex/pip/1/
```

jsirois:
Pants folks: added you FYI since this fixed a problem observed by a Pants user here: https://github.com/pantsbuild/pants/discussions/21145 and so you may want this release since that user didn't actually attempt to fix the underlying problem in the wild with the "unstructured" project's build.

benjyw:
https://github.com/pantsbuild/pants/pull/21545 upgrades Pants to use Pex 2.20.3, which includes this change.

PR #2569, opened 2024-10-23
Optimize pure delete lock updates.

### LINKED ISSUE #2566: Special case lock updates that are pure delete.
If a lock is being updated - via `pex3 lock {sync,update}` - and all updates are deletes, Pip need not be consulted at all. Pex can simply walk the graph from the delete roots and remove all locked requirements no longer referenced.

jsirois:
When a lock update only asks for deletes, we now process the "resolve"
offline without consulting Pip.

Fixes #2566

### DIFF
--- a/pex/resolve/lockfile/updater.py
+++ b/pex/resolve/lockfile/updater.py
@@ -279,6 +279,7 @@ class ResolveUpdater(object):
             original_requirements=tuple(original_requirements.values()),
             update_constraints_by_project_name=update_constraints_by_project_name,
             deletes=frozenset(deletes),
+            pure_delete=bool(deletes and not replace_requirements),
             lock_configuration=lock_configuration,
             pip_configuration=pip_configuration,
             dependency_configuration=dependency_configuration,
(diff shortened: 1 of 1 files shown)


jsirois:
N.B.: Existing tests cover all the interesting delete cases, including transitive dependencies of deleted node being partially retained when something else depends on those, and they pass.

Here proj-b 1 depends on proj-a and proj-c:
https://github.com/pex-tool/pex/blob/e0aac4fb8087f1fda1d65d883381e9a98c7f282d/tests/integration/cli/commands/test_lock_update_issues_2332_2334.py#L441-L464

benjyw (approved):
Nice!

jsirois:
My claim of an existing complex delete case in https://github.com/pex-tool/pex/pull/2569#issuecomment-2433459076 was a lie, only simple deletes are tested through the new code path. That case has a delete and an update; so does not use the new code path and uses Pip instead. I've added a complex pure delete test case in #2576. It passes with no changes; so I'll process with the release before landing that change.

Issue #2528, opened 2024-09-13
Support `pex3 cache` operation at the distribution level.


jsirois:
It would be useful to be able to manipulate distributions in the Pex cache since these are the chief contributors to cache size and the cache items most directly identifiable to an end user since they ask for them in requirements.

Subject to spelling, the actions would be:
* [ ] `pex3 cache add` a distribution to prep for offline use.
* [ ] `pex3 cache remove` one or more distributions safely (paying mind to unzipped_pexes and venvs symlinks).
* [ ] `pex3 cache list` to list distributions and their sizes much like the broader pex3 cache info.

jsirois:
Arguably even better would be the ability to prune items not recently used. This requires the least understanding of Pex cache internals and is likely the most familiar metric. I'll land this capability 1st since it requires much of the same infrastructure needed to support the OP.

jsirois:
#2586 provides most of the infra needed for this.

PR #2586, opened 2024-10-30
Support `pex3 cache prune --older-than ...`.

jsirois:
Work towards #2528.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,17 @@
 # Release Notes
 
+## 2.24.0
+
+This release adds `pex3 cache prune` as a likely more useful Pex cache
+management command than the existing `pex3 cache purge`. By default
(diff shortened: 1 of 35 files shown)


jsirois:
Pants folks - this is FYI. It looks like Pants has not adopted Pex features in quite some time, but `pex3 cache prune` may be compelling. Unlike `pex3 cache purge`, it requires no knowledge of Pex internals and just does the right thing. The defaults (prune anything not used in the last 2 weeks) probably make sense to use as-is. If Pants had a cron job facility, this should probably be one of them.

N.B.: As with `pex3 cache purge`, the prune will block for in-flight PEX runs / Pex CLI invocations to complete. This blocking is not super user friendly out of the box, but is more so when psutil is installed. The released Pex PEX scies all have psutil included.

huonw:
Thanks for the heads-up.

PR #2592, opened 2024-11-13
add `--keyring-provider` flag to configure keyring-based authentication

tdyas:
Add a `--keyring-provider` flag to allow Pex callers to configure Pip  to use a keyring provider without the need to  loosen the Pip isolated mode entirely as would be done if `--use-pip-config` were passed.


pex/pip/tool.py
```diff
@@ -409,6 +414,10 @@ def _spawn_pip_isolated(
 
         command = pip_args + list(args)
 
+        # Append any pass-through pip args.
```
jsirois, line 417: The motivating option is a global option; so you can't pass it here. It would have to be above before `args` is appended (see: https://pip.pypa.io/en/stable/cli/pip/#cmdoption-keyring-provider).

For this sort of reason, which neither using Pip config files nor env vars has a similar ordering issue, I'm leery of a generic Pip flag pass through mechanism. I'd be happy to support `--keyring-provider` 1st class though if that works for you.

jsirois: Ok, looks like global options, including `--keyring-provider` can be passed in any position; so I have this wrong. I'd still be more in favor of a dedicated `--keyring-provider` pass through though since it seems likely a user would pass an option that, say `pip download` or `pip install` takes but `pip debug` does not. Since Pex does, in fact call `pip debug` to implement `--platform`, this would cause a surprising failure.

tdyas: > I'd be happy to support `--keyring-provider` 1st class though if that works for you.

The overall motivating use case is a change in Pants to pass `--keyring-provider` through. Will rework as a `--keyring-provider` option as suggested.

> Since Pex does, in fact call pip debug to implement --platform, this would cause a surprising failure.

Good to know.

tdyas: Also, this is still a WIP since I am having trouble running tests locally and am going to rely on CI to run the tests for me.

jsirois: You can try substituting `tox` with `./dtox.sh` in any tox command line fwiw.

jsirois:
The "./dtox.sh -e py38-pip22_3_1-integration --shard 2/2" shard failure is bitrot from 11/11 addressed here: https://github.com/pex-tool/pex/pull/2595

tdyas:
> The "./dtox.sh -e py38-pip22_3_1-integration --shard 2/2" shard failure is bitrot from 11/11 addressed here: #2595

With `dtox.sh`, my unit test appears to pass. I'll leave this PR in WIP state though while I work on the Pants-side of the ultimate change.

tests/test_pip.py
```diff
@@ -415,16 +415,21 @@ def test_keyring_provider(
     download_dir = os.path.join(str(tmpdir), "downloads")
     assert not os.path.exists(download_dir)
 
-    with ENV.patch(PIP_KEYCHAIN_PROVIDER="invalid") as env, environment_as(**env):
-        assert "invalid" == os.environ["PIP_KEYCHAIN_PROVIDER"]
+    with ENV.patch(PIP_KEYRING_PROVIDER="invalid") as env, environment_as(**env):
```
jsirois, line 418: I'm not sure if you've seen [/home/jsirois/dev/pex-tool/pex/tests/integration/test_keyring_support.py](https://github.com/pex-tool/pex/blob/main/tests/integration/test_keyring_support.py), but its probably what you eventually want to amend to test the "import" + `--extra-pip-requirement` and "subprocess" + PATH means of hooking the keyring provider.

tdyas: Couldn't I just update that test to use the new `--keyring-provider` option instead of the existing `--use-pip-config` option? (Or paramterize the test to try both methods?)

jsirois: ~the latter is _exactly_ what I was suggesting, yes.

jsirois: Hopefully you've noticed though how tortured the very concept of a keyring provider is. There is a nasty bootstrap problem for any locked down org.

tdyas: Added a test.

Weird failure with https://github.com/pex-tool/pex/actions/runs/11918318085/job/33215488401#step:5:5425 since the test should not have tried to execute with a Pip that does not support the option. Any thoughts?

jsirois: Absolutely. I'll hint you 1st: how do you think Pex implements `--pip-version any-not-vendored-version`?

jsirois: Following up a little more, `--keyring-provider` was added for Pip 23.1. Python 3.10.15 and older come with `ensurepip` that yield at most Pip 23.0.1; so you can only count on Python>=3.11.4 (3.11 only hit bundled Pip>=23.1 here: https://github.com/python/cpython/pull/103752) for bootstrapping any `--pip-version` >= 23.1 with `--keyring-provider` being passed as part of the bootstrap process.

tdyas: I think the intended user of my Pants change will be just fine with Python >= 3.11.4. Is that enough to ensure bootstrap of a pip supporting `--keyring-provider` though? I assume the pip bootstrap happens with the user Python chosen by Pants and not the Pants-specific PBS Python used by scie-pants?

tdyas:
@jsirois: Would you be amenable if I submitted an upgrade of the vendored pip in Pex? Is it feasible or desirable to do so? I.e., what should I know about that solution path which I probably do not currently know?

jsirois:
It would not. Hard rules are not breaking people and the current vendored Pip is the last version to support 2.7; so it has to stay. I'll think on a solution a bit more over the next few days and report if I can concoct anything.

tdyas:
> I would not. Hard rules are not breaking people and the current vendored Pip is the last version to support 2.7; so it has to stay. I'll think on a solution a bit more over the next few days and report if I can concoct anything.

What about vendoring two versions of pip? Keep the existing vendored pip for Python 2.7 compatibility and then vendor a newer pip (with for example --keyring-provider`) support. The second vendored copy will be documented as something we update.

Pex can then choose either offer the choice of the newer vendored pip (or better yet in my view) use it instead of the older vendored copy except where Python 2.7 is involved.

Thoughts?

tests/test_pip.py
jsirois: > I think the intended user of my Pants change will be just fine with Python >= 3.11.4. Is that enough to ensure bootstrap of a pip supporting --keyring-provider though?

Not with the code as-it-is, since the `python -mvenv ...` bootstrap trick is not used as aggressively as it could be. That code pointer I gave does highlight the existing mechanism though which should point the way to using it more aggressively.

In short though, knowing which version of Python you have is clearly trivial; then using that to fail fast or move forward with bootstrapping Pip via `-mvenv` and passing `--keyring-provider` after that is definitely in the realm of just putting in the work.

As I mentioned below though, I'll think on this whole situation a bit and report back. There may be some better way than either this or picking a random new Pip to vendor (then can't live without it feature with bootstrapping issue comes up in a newer Pip and surely vendoring a 3rd version is a no go, etc...).

tdyas:
I tried the macOS/x86 SCIE. Apple really does not like unsigned binaries. Opened https://github.com/pex-tool/pex/issues/2621 to discuss.

tdyas:
@jsirois: Any advice on the CI failure at https://github.com/pex-tool/pex/actions/runs/12720783453/job/35462921525?pr=2592#step:5:5585 for `pip: no such option: --keyring-provider`? The test was of Pip v23.3.2 which should support the `--keyring-provider` option.

jsirois:
Before looking deeply, is this question coming with the fact Pex generally bootstraps newer Pips using vendored Pip 20.3.4?

tdyas:
> Before looking deeply, is this question coming with the fact Pex generally bootstraps newer Pips using vendored Pip 20.3.4 understood / remembered from earlier conversations?

Potentially. Then my question becomes, how do I exclude use of the option under test (i.e., `--keyring-provider`) when bootstrapping the newer Pip in the test?

jsirois:
1st, proof:
```
Failed to spawn a job for /development/pex/.tox/py311-pip23_3_2-integration/bin/python: pid 13544 -> /tmp/pytest-of-runner/pytest-0/popen-gw1/test_import_provider_Tru-fbc100/pex-root/venvs/1/29f65b504ee0cb8e829c458f76fa5bc8a2c798b8/30cada8422d156b16f1bae989acbfba03c485755/bin/python /tmp/pytest-of-runner/pytest-0/popen-gw1/test_import_provider_Tru-fbc100/pex-root/venvs/1/29f65b504ee0cb8e829c458f76fa5bc8a2c798b8/30cada8422d156b16f1bae989acbfba03c485755/pex --disable-pip-version-check --no-python-version-warning --exists-action a --no-input --use-deprecated legacy-resolver --isolated --keyring-provider import --log /tmp/pex-pip-log.hjvi9c3z/pip.log -v --cache-dir /tmp/pytest-of-runner/pytest-0/popen-gw1/test_import_provider_Tru-fbc100/pex-root/pip/1/20.3.4-patched/pip_cache download --dest /tmp/pytest-of-runner/pytest-0/popen-gw1/test_import_provider_Tru-fbc100/pex-root/downloads/1/.tmp/resolver_download.bdwjtmj_/development.pex..tox.py311-pip23_3_2-integration.bin.python pip==23.1.2 setuptools==67.7.2 wheel==0.40.0 pex-test-backend --index-url http://localhost:59775/root/pypi/+simple/ --find-links /tmp/pytest-of-runner/pytest-0/popen-gw1/test_import_provider_Tru-fbc100/find-links --trusted-host localhost:59775 --retries 5 --timeout 15 exited with 2 and STDERR:
```

Search that line for 20.3.4.

tdyas:
```
Failed to spawn a job for /development/pex/.tox/py311-pip23_3_2-integration/bin/python: pid 13544 -> /tmp/pytest-of-runner/pytest-0/popen-gw1/test_import_provider_Tru-fbc100/pex-root/venvs/1/29f65b504ee0cb8e829c458f76fa5bc8a2c798b8/30cada8422d156b16f1bae989acbfba03c485755/bin/python /tmp/pytest-of-runner/pytest-0/popen-gw1/test_import_provider_Tru-fbc100/pex-root/venvs/1/29f65b504ee0cb8e829c458f76fa5bc8a2c798b8/30cada8422d156b16f1bae989acbfba03c485755/pex --disable-pip-version-check --no-python-version-warning --exists-action a --no-input --use-deprecated legacy-resolver --isolated --keyring-provider import --log /tmp/pex-pip-log.hjvi9c3z/pip.log -v --cache-dir /tmp/pytest-of-runner/pytest-0/popen-gw1/test_import_provider_Tru-fbc100/pex-root/pip/1/20.3.4-patched/pip_cache download --dest /tmp/pytest-of-runner/pytest-0/popen-gw1/test_import_provider_Tru-fbc100/pex-root/downloads/1/.tmp/resolver_download.bdwjtmj_/development.pex..tox.py311-pip23_3_2-integration.bin.python pip==23.1.2 setuptools==67.7.2 wheel==0.40.0 pex-test-backend --index-url http://localhost:59775/root/pypi/+simple/ --find-links /tmp/pytest-of-runner/pytest-0/popen-gw1/test_import_provider_Tru-fbc100/find-links --trusted-host localhost:59775 --retries 5 --timeout 15 exited with 2 and STDERR:
```

So yes, it was installing the newer `pip`.

jsirois:
> Then my question becomes, how do I exclude use of the option under test (i.e., --keyring-provider) when bootstrapping the newer Pip in the test?

IIUC this is not a test issue to be fixed but a production issue to be fixed. Fix prod and test just works. So, in prod, that option can only be emitted for versions of Pip that support it. Presumably you're not conditionally emitting the option. That said, the fact the option is conditional means the bootstrapping Pip will not be using a keyring provider, which is an issue. I feel like we've been around this loop before?

tdyas:
> That said, the fact the option is conditional means the bootstrapping Pip will not be using a keyring provider, which is an issue. I feel like we've been around this loop before?

Somewhat discussed before with respect to how Pants needs to ensure a recent enough Pip is available so that `--keyring-provider` is supported. It is only the this specific test variant (Python 3.11 & Pip 23.3.2) failing. If I configure the test to use the Python 3.11 `ensurepip` version (24.0) instead of the vendored Pip, maybe that would overcome the issue?

jsirois:
You absolutely need to fix the prod issue. Pex must never pass the option to a Pip that does not support it. IIRC you have everything you need in the Pip class.

tdyas:
> You absolutely need to fix the prod issue. Pex must never pass the option to a Pip that does not support it. IIRC you have everything you need in the Pip class.

My concern there is that a Pex caller using `--keyring-provider` expects Pip to have the option and Pex should then error if it ever in a position to try and invoke a version of Pip without that option. And that behavior would mean that the Pex caller won't be able to use the option unless the caller itself takes care to manage the Pip versions being used.

tdyas:
The alternative of conditionally omitting the option for Pip versions which do not support it would seem to be unexpected behavior for any Pex caller using `--keyring--provider` since the Pip invocation with the option omitted might fail because it cannot authenticate.

jsirois:
Ok - that seems reasonable. It would probably be nice to fail fast then in the appropriate place when cleaning up the PR at the end.

As far as the test goes, you're more in the weeds than I am and it sounds like you know the key variables. I can page all this in tomorrow morning if you're still stuck on test issues.

tdyas:
A potential solution was staring me in the face in the keyring tests: Manually bootstrap a `find-links` directory with a newer version of Pip which supports the `--keyring-provider` option. For example, see [this comment in `test_keyring_support.py`](https://github.com/pex-tool/pex/blob/8b9a06ed74f9066fbda04054114fca920136feea/tests/integration/test_keyring_support.py#L296):

>    N.B.: Pip keyring support presents a catch-22 unless there is a non-authenticated source to
>    resolve the keyring distributions from in the 1st place; so we use a find-links repo
>    pre-populated with the keyring dependencies we need to authenticate; just as a user would
>    have to.

The latest commit has this proposed hybrid solution: Do not set `--keyring-provider` on Pip versions which do not support it, but if Pex would have set it, then warn the user about the issue and how to solve it using `find-links` directory.

jsirois:
> A potential solution was staring me in the face in the keyring tests: Manually bootstrap a find-links directory

Yeah:
https://github.com/pex-tool/pex/blob/4c45f05e47da7501dbbd9ac6ca47df5894370d6a/tests/integration/test_keyring_support.py#L211-L230

pex/pip/tool.py
```diff
@@ -393,6 +401,36 @@ def _spawn_pip_isolated(
...
             pip_args.append("--isolated")
 
+        # Configure a keychain provider if so configured and the version of Pip supports the option.
+        # Warn the user if Pex cannot pass the `--keyring-provider` option and suggest a solution.
+        if package_index_configuration and package_index_configuration.keyring_provider:
+            if self.version.version >= PipVersion.v23_1.version:
+                pip_args.append("--keyring-provider")
+                pip_args.append(package_index_configuration.keyring_provider)
+            else:
+                logger.warning(
```
jsirois, line 411: Please use `pex.pex_warnings.warn(...)`.

pex/pip/tool.py
```diff
@@ -393,6 +401,36 @@ def _spawn_pip_isolated(
...
+                    "".join(
```
jsirois, line 412: Maybe `texwrap.dedent` instead - this seems unwieldy.

pex/pip/tool.py
```diff
@@ -393,6 +401,36 @@ def _spawn_pip_isolated(
...
+                        [
+                            "The PEX_KEYRING_PROVIDER option is set, but Pip v{} does not support ".format(
```
jsirois, line 414: Unless I missed something, `PEX_KEYRING_PROVIDER` is not a thing (nor should it be, since this is a build-time only option and not a runtime knob). You probably are thinking in `PIP_KEYRING_PROVIDER` terms. You should be presenting the option the user actually passes to the Pex CLI.

pex/pip/tool.py
```diff
@@ -393,6 +401,36 @@ def _spawn_pip_isolated(
...
+                                PipVersion.v23_1
+                            ),
+                            "Pex is ignoring PEX_KEYRING_PROVIDER for this particular Pip invocation.\n\n",
+                            "Note: If this Pex invocation fails, it may be because Pex is trying to use its vendored Pip ",
+                            "v{} to bootstrap a newer Pip version which does support `--keyring-provider`, ".format(
+                                PipVersion.VENDORED.version
+                            ),
+                            "but you configured Pex/Pip to use a Python package index which is not available ",
+                            "without additional authentication.\n\n",
+                            "In that case, please manually create a `find-links` directory with that newer version of Pip, so ",
```
jsirois, line 427: A little less strong: `s/please/you might/` - you just really have no clue what the situation is when you wade in to give advice. In that case I favor weak language to bias against super paint-by-numbers users who do exactly what you say with 0 thinking and then complain.

jsirois:
> Given [this test failure], seems like [this test](https://github.com/pex-tool/pex/blob/30c2ec8e888804b2e4e6aa87980254b5d4d7d2f1/tests/test_pip.py#L405) will either need a manually-bootstrapped Pip version or (most likely) just restrict to running with Pip v23.1 and higher?

Tom, can you fill in `[this test failure]` with a link?

tdyas:
> > Given [this test failure], seems like [this test](https://github.com/pex-tool/pex/blob/30c2ec8e888804b2e4e6aa87980254b5d4d7d2f1/tests/test_pip.py#L405) will either need a manually-bootstrapped Pip version or (most likely) just restrict to running with Pip v23.1 and higher?
> 
> Tom, can you fill in `[this test failure]` with a link?

Filled in. https://github.com/pex-tool/pex/actions/runs/12752942699/job/35543486671?pr=2592#step:5:1947

jsirois:
@tdyas I am super duper confused.
1. test_keyring_provider[20.3.4-patched] is https://github.com/pex-tool/pex/actions/runs/12752942699/job/35543486671?pr=2592#step:5:1947
2. Your "this test" link is line 405 here:
https://github.com/pex-tool/pex/blob/30c2ec8e888804b2e4e6aa87980254b5d4d7d2f1/tests/test_pip.py#L398-L417

Those two are unrelated. Is one of your two links wrong?

tdyas:
> @tdyas I am super duper confused.
> 
> 1. test_keyring_provider[20.3.4-patched] is https://github.com/pex-tool/pex/actions/runs/12752942699/job/35543486671?pr=2592#step:5:1947
> 
> 2. Your "this test" link is line 405 here:
> 
> https://github.com/pex-tool/pex/blob/30c2ec8e888804b2e4e6aa87980254b5d4d7d2f1/tests/test_pip.py#L398-L417
> 
> 
> 
> Those two are unrelated. Is one of your two links wrong?
> 
> 
> 
> Skipping past 20 questions perhaps:
> 
> The failing test is yours. It fails because it has not adapted to the fact that you now conditionally inject a `--keyring-provider` arg to the Pip arglist:
> 
> ```
> 
> >           keyring_arg = job._command.index("--keyring-provider")
> 
> E           ValueError: tuple.index(x): x not in tuple
> 
> ```
> 
> And so the corresponding question makes no sense. Presumably you want to test both branches of your new code - the branch that injects the arg and the branch that does not and warns.

Sorry will fix link after I get back from my walk to clear my head of these very code weeds.

tdyas:
> The failing test is yours. It fails because it has not adapted to the fact that you now conditionally inject a --keyring-provider arg to the Pip arglist:

Good point. And that is the failing test. Will update.  (I running both macOS and Linux laptops on this PR and pasted link from wrong machine.)

pex/pip/tool.py
```diff
@@ -393,6 +398,34 @@ def _spawn_pip_isolated(
...
+        # Configure a keychain provider if so configured and the version of Pip supports the option.
+        # Warn the user if Pex cannot pass the `--keyring-provider` option and suggest a solution.
+        if package_index_configuration and package_index_configuration.keyring_provider:
+            if self.version.version >= PipVersion.v23_1.version:
+                pip_args.append("--keyring-provider")
+                pip_args.append(package_index_configuration.keyring_provider)
+            else:
+                warn_msg = textwrap.dedent(
+                    """
+                    The PIP_KEYRING_PROVIDER option is set, but Pip v{THIS_VERSION} does not support the `--keyring-provider` option
```
tdyas, line 410: @jsirois : Is there any additional context you would want in this warning message?

pex/resolve/resolver_options.py
```diff
@@ -135,8 +135,25 @@ def register(
...
     register_use_pip_config(parser)
 
+    parser.add_argument(
+        "--keyring-provider",
+        metavar="PROVIDER",
+        dest="keyring_provider",
+        type=str,
+        default=None,
+        help=(
+            "Configure Pip to use the given keyring provider to obtain authentication for package indexes. "
```
tdyas, line 148: @jsirois: I expanded the help text here.

Should the docs site have a place discussing how to use keyring?

pex/pip/tool.py
tdyas: Instead of `PIP_KEYRING_PROVIDER`, this warning should probably mention the Pex `--keyring-provider` option instead. This PR configures keyring provider based on that `--keyring-provider` option and not the environment variable.

Although, this begs the question, should Pex be looking in the environment for `PIP_KEYRING_PROVIDER` at all?

tdyas:
CI failed due to an infrastructure error in downloading. https://github.com/pex-tool/pex/actions/runs/12790796085/job/35657416561#step:5:2870

Weirdly, I don't see an option in the GitHub Actions UI for the build to retry just the failed job.

pex/pip/tool.py
jsirois: It doesn't:
```
:; git grep PIP_KEYRING_PROVIDER
tests/integration/test_keyring_support.py:                PIP_KEYRING_PROVIDER="subprocess",
tests/integration/test_keyring_support.py:            env=make_env(PIP_KEYRING_PROVIDER="import", **devpi_clean_env),
```

And it shouldn't since `--use-pip-config` is a bit all bets off.

So I think you're talking to yourself here about the doc you added. I already agreed / pushed for `s/PEX_KEYRING_PROVIDER/--keyring-provider/` higher up in this conversation.

tdyas: Good point. Will update to `--keyring-provider`.

pex/resolve/resolver_options.py
```diff
@@ -135,8 +135,25 @@ def register(
...
+            "You must be using Pip v23.1 or higher in order since keyring support was only added in Pip v23.1. "
```
jsirois: The `or higher in order since` bit needs some sort of re-wording fix.

pex/resolve/resolver_options.py
jsirois: Probably, buts that's true of alot of things. I'm fine with you just landing this.

pex/pip/tool.py
tdyas: Done.

pex/resolve/resolver_options.py
tdyas: Reworded it to use "later versions" instead of "higher". Thoughts?

jsirois: The `or higher in order since` grammatical issue is gone so lgtm.

jsirois:
Thanks @tdyas. I'm going to get a PEP-723 locking feature banged out this afternoon which will also include the release prep for 2.29.0 which will include this change.

jsirois (approved):

jsirois:
Ok, now released in 2.29.0: https://github.com/pex-tool/pex/releases/tag/v2.29.0

tdyas:
> Ok, now released in 2.29.0: https://github.com/pex-tool/pex/releases/tag/v2.29.0

Thanks!

PR #2624, opened 2024-12-13
Add a Pex PEX scie for armv7l.

jsirois:


jsirois:
@Lauszus you said you don't need this, but FYI. The dogfooding here feels important to me.

jsirois:
@sureshjoshi this is just FYI. IIRC you do some embedded work too and maybe targeting scies for those boards is useful. On the other end of the spectrum, s390x is already making its way through a-scie which Pex will add support for and possibly also powerpc64le. More generally, I'm happy to add support for any target with good docker multi arch support / qemu support. These are now operationally simple to add at 0 cost (to me).

sureshjoshi:
@jsirois Yep, I do a decent chunk of embedded linux stuff - and there were so many times that something like this would have been handy in a pinch. I've got nothing of the sort at the moment, but I'll be messing around on an upcoming project later in 2025.

PR #2627, opened 2024-12-13
Prepare the 2.27.1 release.

jsirois:


benjyw (approved):
Thanks for the quick turnaround!

jsirois:
No problem - this is just PEX standard operating procedure. I'll be in flight when this goes green; so the actual release won't complete until about 5 hours from now.

jsirois:
The Mac failure is a GHA runner-specific flake fwict (I've seen it several times over the last few weeks). If you self serve on CI retries after the rest settle, great. If not, I'll get to that tonight.

PR #2707, opened 2025-02-27
Fix dup requirement extra merging during PEX boot.

### LINKED ISSUE #2706: pex not handling dependant packages correctly when multiple extras of the same package are installed individually
`dlt` is a package that installs `pyarrow` in its `parquet` extra: https://github.com/dlt-hub/dlt/blob/devel/pyproject.toml#L123 and has many other extras as well. We're seeing that pex brings in `pyarrow` inconsistently if you install `dlt` with multiple extras as separate dependencies, compared to if you install it with all the extras simultaneously.

To reproduce, on pex 2.33.0 on mac:

```
pex 'dlt[snowflake]' 'dlt[filesystem]' 'dlt[parquet]'  --pip-version=latest --platform=current --resolver-version=pip-2020-resolver -o bug.pex
```

outputs a pex where you cannot import pyarrow. When you run the pex file verbosely, it outputs the following to explain why:

```
pex: Skipping activation of `pyarrow<18,>=12.0.0; (python_version >= "3.9" and python_version < "3.13") and (extra == "bigquery" or extra == "parquet" or extra == "motherduck" or extra == "athena" or extra == "synapse" or extra == "clickhouse" or extra == "dremio" or extra == "lancedb" or extra == "deltalake" or extra == "pyiceberg")` due to environment marker de-selection
```

However, if you install all three extras simultaneously, it creates a PEX where you correctly can import pyarrow:
```
pex 'dlt[snowflake,filesystem,parquet]'  --pip-version=latest --platform=current --resolver-version=pip-2020-resolver -o good.pex
```

In both of the above cases if you use pip install instead of pex, the resulting venv allows you to import pyarrow, as expected.

jsirois:
Previously, using root requirements of `foo[bar] foo[baz]` would produce
the same PEX as `foo[bar,baz]`, (save for the root requirements list)
but the former PEX would not properly activate both the `bar` and `baz`
extras during boot. This is now fixed.

Fixes #2706.

### DIFF
--- a/tests/integration/test_issue_2706.py
+++ b/tests/integration/test_issue_2706.py
@@ -1,0 +1,72 @@
+# Copyright 2025 Pex project contributors.
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+from __future__ import absolute_import
+
+import shutil
+import subprocess
+
+from pex.cache.dirs import CacheDir
+from pex.common import safe_mkdir
+from pex.compatibility import commonpath
+from pex.pip.version import PipVersion
+from pex.venv.virtualenv import InstallationChoice, Virtualenv
+from testing import built_wheel, run_pex_command
+from testing.pytest_utils.tmp import Tempdir
+
+
+def test_extras_from_dup_root_reqs(tmpdir):
+    # type: (Tempdir) -> None
+
+    find_links = tmpdir.join("find-links")
+    safe_mkdir(find_links)
+
+    if PipVersion.DEFAULT is not PipVersion.VENDORED:
+        Virtualenv.create(
+            tmpdir.join("pip-resolver-venv"), install_pip=InstallationChoice.YES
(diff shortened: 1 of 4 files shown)


tests/integration/test_issue_2706.py
```diff
@@ -0,0 +1,72 @@
...
+                "pip-2020-resolver",
+                "foo[bar]",
+                "foo[baz]",
+                "-o",
+                pex,
+            ]
+        ).assert_success()
+
+        installed_wheel_dir = CacheDir.INSTALLED_WHEELS.path(pex_root=pex_root)
+        for module in "foo", "bar", "baz":
```
jsirois, line 60: N.B.: The `import baz` fails when the `yield qualified_requirement.with_extras(...)` in the fix above is reverted to just be the old `yield qualified_requirement`.

PR #2713, opened 2025-03-09
Fix PEX scie contents when split.

jsirois:
The embedded PEX now has the executable bit set and it retains the name
requested by the user in nearly all cases.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,9 +2,19 @@
 
 ## 2.33.3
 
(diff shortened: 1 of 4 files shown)


jsirois:
This was prompted by https://github.com/pantsbuild/pants/issues/21100#issuecomment-2705295818

Related to but independent of https://github.com/a-scie/jump/pull/283

PR #2714, opened 2025-03-10
Enable true Zip64 support.

jsirois:
Allow the full range of PEXes under the Zip64 limit even on old Pythons.

### DIFF
--- a/pex/common.py
+++ b/pex/common.py
@@ -183,6 +183,16 @@ class MktempTeardownRegistry(object):
 _MKDTEMP_SINGLETON = MktempTeardownRegistry()
 
 
+# When Zip64 support was added to Python long ago, the limit was set to
+# half of what it should be. Pex is always dealing with zips it creates
+# via Python and so it does not need to worry about zip implemntations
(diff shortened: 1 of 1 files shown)


jsirois:
For example:
```console
# N.B.: torch.pex build using Pex without this change and torch-zip64-mp.pex built with this change.
:; ls -1sb torch.pex torch-zip64-mp.pex
2991260 torch.pex
2990100 torch-zip64-mp.pex

# Note that the non-monkeypatched PEX is bigger.
# This is because Python thought it needed to write the extra Zip64 entries since the file
# was bigger than 2GB - the fake limit baked into Python we monkeypatched around here.

# The PEX _should_ be usable even by zip implementations without Zip64 support like the
# Python zipimporter C module, but it isn't.
:; python2.7 torch.pex -c ''
/home/jsirois/.pyenv/versions/2.7.18/bin/python2.7: can't find '__main__' module in 'torch.pex'

:; python3.13 torch.pex -c ''

# All fixed!
:; python2.7 torch-zip64-mp.pex -c ''

:; python3.13 torch-zip64-mp.pex -c ''
```

PR #2717, opened 2025-03-12
Do not include `ptex` in `--scie eager` scies.

jsirois:
This saves ~5MB for eager scies targeting 64 bit systems.


jsirois:
Facepalm.

PR #2720, opened 2025-03-18
Upgrade to `science` 0.12.2 to fix PBS rate limits.

jsirois:
C.F.: https://github.com/a-scie/lift/issues/127


jsirois:
Reviewers, this is FYI since @achimnol is a direct customer of yours and others building Pex scies via extra args may run into rate limit issues in CI.

PR #2722, opened 2025-03-19
Fix PEP-723 script metadata parsing.

jsirois:
This pre-emptively fixes the issue reported here:
  https://github.com/astral-sh/uv/issues/12303

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Release Notes
 
+## 2.33.6
+
+Fix PEP-723 script metadata parsing to skip metadata blocks found in multiline strings.
+
+* Fix PEP-723 script metadata parsing. (#2722)
+
 ## 2.33.5
 
 This release fixes rate limit issues building CPython Pex scies by bumping to science 0.12.2 which
--- a/pex/pep_723.py
+++ b/pex/pep_723.py
@@ -6,7 +6,6 @@ from __future__ import absolute_import
 import ast
 import re
 from collections import OrderedDict, defaultdict
-from typing import DefaultDict
 
 from pex import toml
 from pex.common import pluralize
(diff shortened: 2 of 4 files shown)


jsirois:
Apparently @ofek thinks this is bad form, but as a tool maintainer that generally has to work around PEPs that are too loose leading to crazy situations in the wild, I think this is just par for the course. I'd rather Pex not choke and ruin a user's day than tell them they're doing it wrong. Of course I think the underlying issue is, yet again, the PyPA. It should not be a forked entity from Python. If PEP-723 were implemented in Python by Python and shipped with Python, it could unambiguously parse this data as part of its normal bytecode parsing phase.

ofek:
> Of course I think the underlying issue is, yet again, the PyPA.

I'm not sure why you always mention that the issue at the end of the day is the PyPA; wouldn't the issue literally be the spec? I don't see how any other conclusion can be made 😅 

> If PEP-723 were implemented in Python by Python and shipped with Python, it could unambiguously parse this data as part of its normal bytecode parsing phase.

No, that's a bad idea and was rejected: https://peps.python.org/pep-0723/#why-not-use-possibly-restricted-python-syntax

Python isn't a closed ecosystem and many kinds of tools from all kinds of languages need to be able to parse the block.

jsirois:
> I'm not sure why you always mention that the issue at the end of the day is the PyPA; wouldn't the issue literally be the spec? I don't see how any other conclusion can be made

My point is there is no spec involving Python source code can avoid ambiguity without parsing the Python source code. As such, if there were instead no spec and Python shipped with its own single set of tooling, then this would all be avoided. Non-python ecosystem tooling could either shell out to python tooling - easy to do if there is 1 and only 1 tooling set - or else bit the bullet and parse the source code too.

ofek:
Hmm, based on your last comment the scope of what we both are talking about appears to differ. Just to be clear, in your ideal future of unified tooling for the Python ecosystem (a good goal indeed!) the non-Python projects I'm talking about still exist and see high usage. For example, no one is going to stop using Dependabot based on what our community does and it is written in Ruby.

jsirois:
> For example, no one is going to stop using Dependabot based on what our community does and it is written in Ruby.

We are speaking in the same terms. If dependabot wants to parse Python deps in script files, it can call `python -t deps --type 3rdparty this/file/here.py` ... assuming syntax of that rosy future where python ships with built-in tools (`-t`).

PR #2729, opened 2025-03-31
Avoid fast-path in `--sh-boot` script for `PEX_PYTHON`, `PEX_PYTHON_PATH` and `PEX_PATH`

huonw:
Similar to #2726, there's some additional variables that are need to avoid the `--sh-boot` fast-path.

### DIFF
--- a/tests/integration/test_sh_boot.py
+++ b/tests/integration/test_sh_boot.py
@@ -345,3 +345,61 @@ def test_issue_2726_pex_tools(
 
     # run the tools with seeded PEX_ROOT
     _check_pex_tools()
+
+
+@execution_mode
+def test_issue_2728_pex_python_and_pex_python_path(
+    tmpdir,  # type: Any
+    execution_mode_args,  # type: List[str]
+):
+    # type: (...) -> None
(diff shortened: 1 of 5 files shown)


tests/integration/test_sh_boot.py
```diff
@@ -345,3 +345,61 @@ def _check_pex_tools():
...
+    pex_root = os.path.join(str(tmpdir), "pex_root")
+
+    def _check_pex_with_python():
+        for env in [dict(PEX_PYTHON="FIXME"), dict(PEX_PYTHON_PATH="FIXME")]:
+            output = subprocess.check_output(
+                args=[pex, "-c", "import sys; print(sys.version)"],
+                env=make_env(PEX_ROOT=pex_root, **env),
+                stderr=subprocess.STDOUT,
+            )
+            assert False, "FIXME"
```
huonw, line 368: I am unclear on how to test these (and ran out of time today to investigate existing tests that use them): AIUI, the tests are run by tox with a specific interpreter available, and other may not be.

If there's a tip for how these should be structured, that'd be appreciated, otherwise I'll look at existing tests more, soon.

PR #2759, opened 2025-05-10
Update `--resolve-local-platforms` docs.

jsirois:
These pre-dated `--complete-platform`s and now account for those.

### DIFF
--- a/pex/resolve/target_options.py
+++ b/pex/resolve/target_options.py
@@ -170,11 +170,15 @@ def _register_platform_options(
         default=False,
         action=HandleBoolAction,
         help=(
-            "When --platforms are specified, attempt to resolve a local interpreter that matches "
-            "each platform specified. If found, use the interpreter to resolve distributions; if "
-            "not (or if this option is not specified), resolve for each platform only allowing "
-            "matching binary distributions and failing if only sdists or non-matching binary "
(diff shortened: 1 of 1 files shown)


jsirois:
As noted [here](https://github.com/pantsbuild/pants/pull/20823/files#r2083209314), these bad docs threw you off @huonw in one instance although it appears you read through these bad docs and actually used the option in combination with `--complete-platform` in Pants FAAS support in another instance. My brain does that all the time too.

benjyw (approved):
Thanks for the clarification.

Is there a way to turn off what you referred to elsewhere as "YOLO mode"? I.e., don't fall back to building on the interpreter executing pex, but require a local matching interpreter or fail fast?

jsirois:
Filling in history 1st off:
+ https://github.com/pex-tool/pex/pull/2075
+ https://github.com/pex-tool/pex/issues/2073

> Is there a way to turn off what you referred to elsewhere as "YOLO mode"? I.e., don't fall back to building on the interpreter executing pex, but require a local matching interpreter or fail fast?

Sure, use `--no-build`. I will note this contradicts the internal decision made for FAAS where the combination of `--complete-platform` and `--resolve-local-platforms` is used to try to get a good result in as many situations as possible.

PR #2804, opened 2025-07-03
Update script_metadata.py to fix parse utf8 script on zh_CN locale widows

tqjason:
windows 11 zh-CN
python 3.10
python script is utf-8
error when parse


pex/resolve/script_metadata.py
```diff
@@ -58,7 +58,7 @@ def apply_script_metadata(
     )  # type: OrderedSet[str]
     requires_python = SpecifierSet()
     for script in scripts:
-        with open(script) as fp:
+        with open(script, encoding="utf-8") as fp:
```
jsirois, line 61: Thanks for making an effort at a fix. I will note that Pex definitely does not support Windows yet. It works fine under WSL though. That said, should be able to export the `PYTHONUTF8=1` environment variable to work around these sorts of issues globally - regardless of Pex.

___
All that said - addressing your PR:

Pex supports 2.7 and >=3.5; so this won't work:
```console
:; python2.7 -c 'open("/etc/hosts", encoding="utf-8")'
TypeError: 'encoding' is an invalid keyword argument for this function
```

I think something like this would work though:
```python
with open(script) as fp:
    script_metadata = ScriptMetadata.parse(fp.read().decode("utf-8"), source=fp.name)
```

Except `ScriptMetadata.parse` is typed to take `str`, which will likely need to be changed to `Text` to be compatible with Python 2.7:
https://github.com/pex-tool/pex/blob/52e986d2407ce1a6c4b4a634ffbbc281712eeab4/pex/pep_723.py#L229-L237

That will have ripple effects.

To test all this out you'll need to have [`uv` installed](https://docs.astral.sh/uv/getting-started/installation/) and then run:
```console
uv run dev-cmd -stest
```

That will run all formatting linting including MyPy to type-check.

tqjason:
Thank you for your reply. I didn't realize the potential issues regarding the applicability of the operating environment. This was my imprudence. However, after locally modifying script_metadata.py on Windows, I successfully generated scies for three platforms: win-amd64-cp-39-cp39, macos-x86_64-cp-39-cp39, and linux_x86_64-cp-39-cp39. The win-amd64 one ran successfully, while the MacOS and Linux versions haven't been tested yet.

tqjason:
I have tried `fp.read().decode("utf-8")` . Not working. Because error is on `fp.read()`, before do `decode("utf-8")`

tqjason:
maybe `open(script, 'rb')` ?

jsirois:
> maybe open(script, 'rb') ?

Oops. Yes.

jsirois:
@tqjason have you tried setting the `PYTHONUTF8=1` environment variable when building the PEXes as an alternative workaround? That may be the way to go here.

Looking into the code fix I find:
+ PEP-263: https://peps.python.org/pep-0263/
+ PEP-686: https://peps.python.org/pep-0686/

PEP-263 is the active one and defines the default source encoding as ascii. If the Python file does not use ascii it should be including an encoding comment line. I think the current Pex code parsing script metadata should be fixed to read the file as binary, then look for an encoding comment line and use that encoding or else ascii to then decode the bytes into a string.

PEP-686 is final, but not active until the Python 3.15 release, at which point the default encoding switches from ascii to utf-8.

If you agree with this assessment of the current Python rules for source file encoding, then the proper fix is outlined. Do you want to push forward along those lines?

PR #2806, opened 2025-07-03
Fix PEP-723 script parsing file encoding handling.

jsirois:
Supercedes #2804


jsirois:
@tqjason this should do it. Thanks for bringing this issue to my attention.

PR #2807, opened 2025-07-03
Let PEX scies install PEXes in the `PEX_ROOT`.

jsirois:
Previously, the `PEX_ROOT` was synthesized to be inside the nce 
`SCIE_BASE` which led to very long `PEX_ROOT` cache paths. This foiled
PEX scies from working on Windows in many cases. The Windows case
aside, having PEXes in PEX scies behave like normal PEXes is a goal
which this supports. Although a PEX scie user should be none-the-wiser,
the creator of a PEX scie gets the runtime PEX behavior and
debuggability they expect of a non-scie PEX.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,17 @@
 # Release Notes
 
+## 2.44.0
+
+This release expands PEX scie support on Windows to more cases by changing how the `PEX_ROOT` is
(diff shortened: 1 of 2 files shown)


jsirois:
@PsyVeteran with this release (2.44.0), Pex may work for more of your Windows use cases. Still no guarantees for Windows though!

Issue #2834, opened 2025-07-21
Pylock.toml with cyclic dependencies: `maximum recursion depth exceeded`


pimdh:
Hi,

Thanks for this fantastic tool!

With the following `pylock.toml` file, which is an excerpt of a real lock file:
```
lock-version = "1.0"
created-by = "uv"
requires-python = "==3.11.*"


[[packages]]
name = "mkdocstrings"
version = "0.29.1"
dependencies = [{ name = "mkdocstrings-python", version = "1.16.12" }]
index = "https://pypi.org/simple"
sdist = { url = "https://files.pythonhosted.org/packages/41/e8/d22922664a627a0d3d7ff4a6ca95800f5dde54f411982591b4621a76225d/mkdocstrings-0.29.1.tar.gz", upload-time = 2025-03-31T08:33:11Z, size = 1212686, hashes = { sha256 = "8722f8f8c5cd75da56671e0a0c1bbed1df9946c0cef74794d6141b34011abd42" } }
wheels = [
    { url = "https://files.pythonhosted.org/packages/98/14/22533a578bf8b187e05d67e2c1721ce10e3f526610eebaf7a149d557ea7a/mkdocstrings-0.29.1-py3-none-any.whl", upload-time = 2025-03-31T08:33:09Z, size = 1631075, hashes = { sha256 = "37a9736134934eea89cbd055a513d40a020d87dfcae9e3052c2a6b8cd4af09b6" } },
]

[[packages]]
name = "mkdocstrings-python"
version = "1.16.12"
dependencies = [{ name = "mkdocstrings", version = "0.29.1" }]
index = "https://pypi.org/simple"
sdist = { url = "https://files.pythonhosted.org/packages/bf/ed/b886f8c714fd7cccc39b79646b627dbea84cd95c46be43459ef46852caf0/mkdocstrings_python-1.16.12.tar.gz", upload-time = 2025-06-03T12:52:49Z, size = 206065, hashes = { sha256 = "9b9eaa066e0024342d433e332a41095c4e429937024945fea511afe58f63175d" } }
wheels = [
[...]

PR #2835, opened 2025-07-21
Handle cyclic dependencies in pylock.toml

pimdh:
Fixes issue #2834 

I had to make `Package` not frozen, because, as far as I know, it's impossible to realize a cyclic graph with frozen objects.


jsirois:
Thanks @pimdh! I fixed up some formatting issues (`uv run dev-cmd -stest`) and restored Package immutability by making dependencies just contain what's needed by other code; namely the package index and the package name.

PR #2894, opened 2025-09-02
Support dependency replacement with `--override`.

jsirois:
Instead of just accepting PEP-508 requirement strings for `--override`
values, Pex now also accepts values of the form
`<project name>=<requirement>` to override any transitive requirement
on one project with a requirement on a completely different project.

This can be useful when you need to supply a patch to an existing
published project and would prefer to depend on wheels you pre-build
instead of using a VCS source dependency `--override`, which can be slow
to build.

Addresses #2891 and completes the suggested use cases in #2425 for which
`--override` was originally introduced.

### DIFF
--- a/pex/dependency_configuration.py
+++ b/pex/dependency_configuration.py
@@ -4,10 +4,11 @@
 from __future__ import absolute_import
 
 import itertools
+import re
 from argparse import Namespace, _ActionsContainer
 from collections import defaultdict
 
-from pex.dist_metadata import Distribution, Requirement
+from pex.dist_metadata import Distribution, Requirement, RequirementParseError
 from pex.orderedset import OrderedSet
 from pex.pep_503 import ProjectName
 from pex.pex_info import PexInfo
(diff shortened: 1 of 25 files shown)


tests/integration/test_override_replace.py
```diff
@@ -0,0 +1,379 @@
...
+        "create",
+        "--style",
+        "universal",
+        "--interpreter-constraint",
+        ">=3.9,<3.12",
+        "--indent",
+        "2",
+        "-o",
+        lock,
+    ] + multi_override_args
```
jsirois, line 323: This is a bit of involved test, but suffice it to say, it stresses this scenario:
+ universal lock covering Python 3.9, 3.10 and 3.11
+ one override for `ansicolors` with a local patch, for - say - 3.9 only
+ a second override for `ansicolors` with a different local patch, for - say - 3.11

The test helps prove that in this sort of case, where the overrides have markers that make their application only partial for the full range of a universal lock, the original dependency is used for the range of the universal lock where the override markers do not apply.

The fact that this did not used to work, even for non-replacement `--override`s, was a bug.

pex/dependency_configuration.py
```diff
@@ -152,9 +212,17 @@ def register(parser):
...
+            "modify an existing dependency on a project name by changing extras, version "
+            "constraints or markers or else they can completely swap out the dependency for a "
+            "dependency on another project altogether. For the former, simply supply the "
+            "requirement you wish. For example, specifying `--override cowsay==5.0` will override "
+            "any transitive dependency on cowsay that has any combination of extras, version "
+            "constraints or markers with the requirement `cowsay==5.0`. To completely replace "
+            "cowsay with another library altogether, you can specify an override like "
+            "`--override cowsay=my-cowsay>2`. This will replace any transitive dependency on "
+            "cowsay that has any combination of extras, version constraints or markers with the "
+            "requirement `my-cowsay>2`. This option can be used multiple times."
```
jsirois, line 225: @cburroughs hopefully its now more clear exactly what `--override` can be expected to do. I do not go into gory detail about the ability to say complex things like:
```console
--override 'a=b; os_name == "posix"' \
--override 'a=c; os_name != "posix" and python_version >= "3"' \
--override 'a=d; os_name != "posix" and python_version < "3"'
```
But as noted elsewhere, this works and is tested; so I'd like to just leave it as one of those use cases that you might expect should work and then turns out it does.

Please verify the `--help` is now a bit more clear (or not).

jsirois: @cburroughs I'm going to get a release out tonight but I'll circle back to address any feedback you have.

PR #2898, opened 2025-09-05
Respect `--interpreter-constraint` impl in locks.

jsirois:
Previously, if an interpreter implementation was specified in a
`--style universal` lock `--interpreter-constraint`, that information
was discarded; so, for example, even with an `--interpreter-constraint` 
of `CPython==3.13.*` specified, the lock resolve would consider `PyPy`
wheels in-play.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,14 @@
 # Release Notes
 
+## 2.55.1
+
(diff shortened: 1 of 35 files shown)


cburroughs:
This successfully created a lockfile with my original gnarly set of internal wheels and requirements.

PR #2903, opened 2025-09-12
Support scopes for `--index` and `--find-links`.

jsirois:
This allows scoping package repositories to only certain projects or
certain target environments or an even narrower combination of the two.

In the context of `--style universal` lock creation, this admits
automatically splitting the lock into two or more locked resolves
using different universes of `--indexes` and `--find-links`. This
new splitting support is also extended to any top-level requirements
that are split via markers as well.

This should help along these issues:
+ https://github.com/pantsbuild/pants/issues/18293
+ https://github.com/pantsbuild/pants/issues/18965
+ https://github.com/pantsbuild/pants/issues/20808

And could be used as the basis to help with an issue like:
https://github.com/pantsbuild/pants/issues/22581

### DIFF
--- a/tests/integration/cli/commands/test_split_universal_locks.py
+++ b/tests/integration/cli/commands/test_split_universal_locks.py
@@ -340,7 +340,7 @@ def pin(
 def assert_expected_split_repos_and_requirements_lock(lock_file):
     # type: (str) -> None
 
-    if sys.version_info.releaselevel == "final":
+    if sys.version_info.releaselevel != "final":
         # This test is complicated by Python development releases and is covered by several
         # production releases in CI; so we skip.
         return


tests/test_target_system.py
```diff
@@ -0,0 +1,83 @@
...
+    assert not marker_env.evaluate(Marker("python_version < '3.10'"))
+    assert marker_env.evaluate(Marker("'3.9' <= python_version"))
+    assert not marker_env.evaluate(Marker("'3.10' > python_version"))
+
+    assert marker_env.evaluate(Marker("python_full_version == '3.12.*'"))
+    assert not marker_env.evaluate(Marker("python_full_version == '3.14.*'"))
+    assert marker_env.evaluate(Marker("python_full_version == '3.11.*'"))
+    assert marker_env.evaluate(Marker("'3.12.*' == python_full_version"))
+    assert not marker_env.evaluate(Marker("'3.14.*' == python_full_version"))
+    assert marker_env.evaluate(Marker("'3.11.*' == python_full_version"))
```
jsirois, line 55: N.B.: I had to fix the universal marker evaluation system in this PR which had me looking harder at the code and marker spec. AFAICT - even though no one seems to write markers like this - it's supposed to be perfectly fine to place the marker variable on the RHS of the expression. I fixed universal marker evaluation to handle this, but `packaging` does not. I've filed https://github.com/pypa/packaging/issues/934 to track down if that's a bug in packaging or a too-loose spec or me somehow missing something obvious here.

PR #2918, opened 2025-09-26
Fix `pex3 lock create/export` for VCS URLs with userinfo

rooperuu:
Fixes issues in `pex3 lock create/export` with VCS URLs that contain userinfo such as `git+ssh://git@github.com/USER/REPO.git`.

Changes:

- Resolves `commit_id` correctly for URLs that have userinfo masked with `****`
- Takes potential userinfo into account when partitioning URLs


jsirois:
Thanks @rooperuu! I think what you did here is correct. I added a test that fails without your locker fix at any rate, and now passes.

jsirois (approved):

PR #2968, opened 2025-10-25
Add support for `--pip-version 25.3`.

jsirois:


jsirois:
N.B.: `test_redacted_requirement_handling` should fail for Pip 25.3 shards due to setuptools leaving an `.egg-info` dir in the root of VCS zip archives under the new, enforced PEP-517 builds in Pip 25.3. No other common build backends have this issue (I tested hatch, poetry and pdm and I assume uv-build is also clean). I'll have a final fix for that following.

PR #3007, opened 2025-11-13
Update buildingpex.rst

Rolfdv:
Update docs to reference Ansible to match the following example, rather than Sphinx


jsirois:
Thanks Rolf.

PR #3049, opened 2025-12-21
Issues/3040

### LINKED ISSUE #3040: Pip bootstrap install/upgrade does not use standard options e.g. --isolated
Pip bootstrapping [found here](https://github.com/pex-tool/pex/blob/b42697bac49c8527523c29924dd65b6ca19c0273/pex/pip/installation.py#L258) does not apply standard options/flags that are used elsewhere (from [_spawn_pip_isolated](https://github.com/pex-tool/pex/blob/b42697bac49c8527523c29924dd65b6ca19c0273/pex/pip/tool.py#L392)). Bootstrapping can fail due to this in cases where existing Pip user config (via env var or `pip.conf`) causes invalid or unresolvable install queries, such as insufficient authentication provided for a set private index-url or an index-url being inaccessible in general.

jsirois:
Fixes #3040

### DIFF
--- a/pex/cli/commands/lock.py
+++ b/pex/cli/commands/lock.py
@@ -45,7 +45,6 @@ from pex.pip.version import PipVersionValue
 from pex.requirements import LocalProjectRequirement
 from pex.resolve import project, requirement_options, resolver_options, target_options
 from pex.resolve.config import finalize as finalize_resolve_config
-from pex.resolve.configured_resolver import ConfiguredResolver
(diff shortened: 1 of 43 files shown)


jsirois:
Thanks @Niccolum . Unfortunately this is still red and it's just refactor commits so far that do not address the issue. This will be slower than I like, but I hope to have a fix out before the new year.

jsirois:
Closing in favor of #3054 which avoids the thorny refactor.

PR #3053, opened 2025-12-25
Add scie support for direct exec of venv scripts.

jsirois:
Previously this support was only provided for busybox scies, but now you
can pass `--no-scie-pex-entrypoint-env-passthrough` for single script
entrypoint `--venv` PEX scies as well.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,15 @@
 # Release Notes
 
+## 2.76.0
+
+This release adds support for `--no-scie-pex-entrypoint-env-passthrough` to trigger direct execution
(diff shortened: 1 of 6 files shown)


jsirois:
@cburroughs indeed. Happy New Year!

Yes, it's stable. I depend on it in dev-cmd which Pex, lift (science) and dev-cmd itself all use to make up their "build tool": https://github.com/jsirois/dev-cmd/blob/main/pyproject.toml

That said this obviates the need: https://github.com/pex-tool/pex/pull/3072

jsirois:
I do hope to breakout that code as its own psuedo-build backend distribution, but it's coupled to some PY2 / PY3 compat code and utils; so it's some work I've been lazy about getting to.

PR #3131, opened 2026-03-27
Support `--uploaded-prior-to` for Pip v26.0 and newer

### LINKED ISSUE #3128: support passing `--uploaded-prior-to` for recent pips
pip 26.0 added support for  [--uploaded-prior-to](https://github.com/pypa/pip/pull/13625). This is useful for reproducible testing, as another way to work around partial uploads to registries, and as a supply chain attack mitigation .  I'd like to be able to pass it through to `pip`.  Naively I think this could be grouped with the other "Resolver Options".

cburroughs:
Pip 26.0 adds support for --uploaded-prior-to for filtering packages by their upload time to an index. See
<https://pip.pypa.io/en/stable/user_guide/#filtering-by-upload-time> for details.

This change passes through --uploaded-prior-to on recent enough Pip versions, and emits a warning on older ones.

NOTE: When Pip is building an sdist, this setting also applies to build dependencies and it is thus easy when using far-in-the-past values to end up in a tangle of incompatible Python/setuptools versions.  This is out of Pex's control, but might be the first thing you hit if you are like me and reach for old `cowsay` versions right away.

closes #3128

### DIFF
--- a/tests/integration/cli/commands/test_lock_uploaded_prior_to.py
+++ b/tests/integration/cli/commands/test_lock_uploaded_prior_to.py
@@ -93,9 +93,3 @@ def test_uploaded_prior_to_filters_to_older_version(tmpdir):
         expected_error_re=r".*PEXWarning: .*Using Pip .* instead",
         re_flags=re.DOTALL,
     )
-
-    lock = json_codec.load(lock_file)
-    assert 1 == len(lock.locked_resolves)
-    locked_resolve = lock.locked_resolves[0]
-    assert 1 == len(locked_resolve.locked_requirements)
-    assert locked_resolve.locked_requirements[0].pin.version >= Version("6.1")


pex/resolve/resolver_options.py
```diff
@@ -867,6 +882,14 @@ def create_pip_configuration(
             )
         )
 
+    if options.uploaded_prior_to and (pip_version or PipVersion.DEFAULT) < PipVersion.v26_0:
+        raise InvalidConfigurationError(
+            "Pip {pip_version} too old to support `--uploaded-prior-to` flag added in {version_26_0}.".format(
+                pip_version=pip_version if pip_version else "%s (default)" % PipVersion.DEFAULT,
```
jsirois: Just noticed this. Totally fine, but Python 2.7 text supports the `format` method; just not the `f"..."` fstring shortcut; so this works: `"{} (default)".format(PipVersion.DEFAULT)`. It's not true in this case, but generally I go full verbose and do `"{name_in_scope} ...".format(name_in_scope=name_in_scope)` only due to a hand-wave notion that if I ever drop 2.7 support you get `s/"{name_in_scope} ...".format(name_in_scope=name_in_scope)/f"{name_in_scope} ..."/` (prefix an f and lop off the `format` call).

In short, there are uses, but I got rid of most of them and the remaining are all in very OG files from back in the Wickman / early Twitter days:
```console
:; grep -R --exclude-dir _vendored ".format(" pex/ | wc -l
1369

:; grep -R --exclude-dir _vendored " % " pex/ | wc -l
47

:; grep -R --exclude-dir _vendored " % " pex/ | cut -d: -f1 | sort -u
pex/bin/pex.py
pex/bootstrap.py
pex/common.py
pex/compatibility.py
pex/compiler.py
pex/distutils/commands/bdist_pex.py
pex/environment.py
pex/executor.py
pex/interpreter.py
pex/pex_builder.py
pex/pex_info.py
pex/pex.py
pex/tracer.py
pex/variables.py
pex/venv/virtualenv_16.7.12_py
```

PR #3159, opened 2026-04-29
Defer requirement parse-error source to failure path

cburroughs:
Constructing the `repr` for error handling requiers going through Distribution.name and eventually the `email` module for parsing.  This is -- at least to me -- surprisingly expensive.  If I profile (concretely this is with Linux `perf`) generating a lockfile the top 3 moduels on-cpu are `email.feedparser`, `email._policybase`, and `email.message`.

Comparison using the requirements and constraints from https://github.com/pypa/pip/issues/13281#issuecomment-2837189308

```
$ hyperfine --warmup 1 --runs 5 --parameter-list ref v2.93.0,csb/defer-repr --command-name '{ref}' --prepare 'git checkout {ref}'  "python3.14 -m pex.cli lock create --style universal --prefer-wheel --indent 2 -o lock.json --pip-version=26.1 -r reqs.txt  --constraint c.txt --interpreter-constraint='CPython==3.11.*'"
Benchmark 1: v2.93.0 (ref = v2.93.0)
  Time (mean ± σ):     138.271 s ±  5.397 s    [User: 119.267 s, System: 11.153 s]
  Range (min … max):   134.099 s … 147.518 s    5 runs

Benchmark 2: csb/defer-repr (ref = csb/defer-repr)
  Time (mean ± σ):     88.597 s ±  0.911 s    [User: 72.145 s, System: 10.089 s]
  Range (min … max):   87.479 s … 89.574 s    5 runs

Summary
  csb/defer-repr (ref = csb/defer-repr) ran
    1.56 ± 0.06 times faster than v2.93.0 (ref = v2.93.0)
```

NOTE: The comparison above is with python3.14 so
https://github.com/python/cpython/issues/106628 is already included.


CHANGES.md
```diff
@@ -1,10 +1,14 @@
...
+## 2.93.1
+
+ * Defer requirement parse-error source to failure path. (#3159)
+
 ## 2.93.0
 
 This release adds support for `--pip-version 26.1`.
 
-* Add support for `--pip-version 26.1`. (#3155)
+* Add support for `--pip-version 26.1`. (#3158)
```
jsirois: Bad change, but you're fighting that PR (https://github.com/pex-tool/pex/pull/3158), which will be what grabs 2.93.1. You'll be 2.93.2.

pex/pip/dependencies/requires.py
```diff
@@ -48,7 +49,10 @@ def requires(self, *args, **kwargs):
             modified_requires = []
             orig = orig_requires(self, *args, **kwargs)
             for req in orig:
-                requirement = PexRequirement.parse(str(req), source=repr(self))
+                try:
```
jsirois, line 52: There's only 1 other hit for `Requirement.parse(..., source=)` in the repo and that's in a test; so I guess this is fine, but a comment would be good - because otherwise this is easy to come back to later and undo.

jsirois (approved):
Thanks!

jsirois:
There was a bad cache image update (these: https://github.com/pex-tool/pex/actions/workflows/build-cache-image.yml); so I'll babysit re-tries on this to get it through. I've re-tagged back to a known-good cache image; so it shouldn't be too bad.

jsirois:
Thanks again @cburroughs. This is now released: https://github.com/pex-tool/pex/releases/tag/v2.93.2

PR #3180, opened 2026-05-23
Use `PEX_EXTRA_SYS_PATH.start` for Python 3.15 venvs.

jsirois:


jsirois:
@cburroughs I added you to the review only at your discretion. It seems like you're following the PyPA at some level now; so you might be interested or Pants (or its customers) might be affected by `.start` files.

cburroughs:
Thanks.  I have been trying to find a workflow to follow parts of discuss.python.org more consistently.

PR #3184, opened 2026-06-03
Upgrade to CPython 3.15.0b2.

jsirois:


jsirois:
Submitting to have cache changes available for rebuilds and then red cleanup from there...

PR #3204, opened 2026-07-12
Fix download cache coherence

mjlbach:
## Summary

This PR previously made locked `AtomicDirectory` work directories owner-specific. After digging further, that change addressed a symptom rather than the cause of the failures we were seeing, so this PR has been re-pointed at the root cause. It is now three commits, each independently reviewable:

1. **Accept legacy v2 downloaded artifact metadata.** The metadata version was bumped 2 → 3 in 826ef8e3 (#3112, first released in 2.91.0) to record local-project editability. Since then, a download cache populated by an older Pex fails every `DownloadedArtifact.load` with a `LoadError`, mass-invalidating the entire cache on first use. v2 metadata predates editable support entirely, so it is now read in place as `editable=False`. To preserve v3 semantics, `store()` also cross-checks the requested artifact's editability against the cached metadata and rebuilds on mismatch (previously, since `DownloadDir` is keyed only by file hash, an entry stored with one editability was silently returned for requests wanting the other).

2. **Remove invalid download cache entries under the artifact lock.** On `LoadError`, `DownloadManager.store` called `safe_rmtree(download_dir)` on the shared, finalized cache directory *with no lock held*. Readers are lock-free by design (`atomic_directory` only locks when its target does not exist), so under concurrent resolves one process could delete a directory another process had just repaired or was actively reading. Combined with a metadata version bump, this cascades: every process that observed the old metadata queues up a deletion of the freshly repaired directory. The invalid entry is now re-checked under the artifact lock before removal; the rebuild happens after the lock is released because Pex's in-process file lock is intentionally non-reentrant (`atomic_directory` re-checks after re-acquiring).

3. **Use BSD file locks for all `DownloadDir` writers.** On Linux, POSIX record locks (`fcntl`/`lockf`) and BSD locks (`flock`) live in independent lock tables and do not exclude one another. The fingerprint-addressed download cache had writers in both domains: `LockDownloader` uses BSD locks, while `ArtifactDownloader._fingerprint_and_move` and the `DownloadManager` default used POSIX locks on the same lockfile — so a lock creation running concurrently with a lock resolve could have two writers inside the critical section for the same artifact. Pex already requires working `flock` for cache access (`pex/cache/access.py`), so this adds no new platform requirements.

## The failure we observed

After upgrading Pants (and with it Pex ~2.81 → 2.95) with the Pex cache restored across the upgrade, CI began failing nondeterministically with downloaded wheels disappearing during concurrent resolution. The mechanism is the interaction of commits 1 and 2: the restored cache was all v2 metadata, so every concurrent Pex process hit `LoadError` on the same shared directories and issued unlocked `safe_rmtree` calls — deleting entries that sibling processes had just repaired or were reading.

## Why the work-directory approach was replaced

The earlier change here gave each locked owner a unique work directory to protect against a "cancelled owner cleaning up under a new owner". Two problems with that:

- Pants preemption (dynamic concurrency restarts) SIGKILLs the entire process group with no graceful phase, so a cancelled owner never runs late cleanup, and no descendant survives to write into an adopted work directory. In the `atomic_directory` context manager, `cleanup()` also always precedes `unlock()`, so a live owner cannot clean up after losing the lock unless lock exclusion is already broken — which commit 3 addresses at the source.
- With unique names, a SIGKILLed owner's work directory (holding a partial download) is never adopted or cleaned by the next owner, and nothing prunes it, so preemption-heavy CI would leak cache space unboundedly.

If owner-specific work directories are still wanted as insurance against exclusion failures (e.g. NFS), I'm happy to follow up separately with a stale work-directory cleanup to accompany them.

## Validation

- `tests/resolve/lockfile/test_download_manager.py` — v2 read-in-place, editable-mismatch rebuild, removal-under-lock ordering, and re-check-after-locking (a concurrently repaired entry is not deleted).
- `tests/resolve/test_downloads.py` — `_fingerprint_and_move` uses the BSD lock domain.

All 10 focused tests pass.

### DIFF
--- a/pex/resolve/downloads.py
+++ b/pex/resolve/downloads.py
@@ -11,6 +11,7 @@ from pex.artifact_url import ArtifactURL, Fingerprint
 from pex.atomic_directory import atomic_directory
 from pex.cache.dirs import CacheDir, DownloadDir
 from pex.common import safe_mkdir, safe_mkdtemp
+from pex.fs.lock import FileLockStyle
 from pex.hashing import Sha256
 from pex.jobs import Job, Raise, SpawnedJob, execute_parallel
 from pex.pip import foreign_platform
@@ -63,7 +64,9 @@ class ArtifactDownloader(object):
         hashing.file_hash(path, digest)
         fingerprint = digest.hexdigest()
         target_dir = DownloadDir.create(file_hash=fingerprint)
-        with atomic_directory(target_dir) as atomic_dir:
+        # All writers to the fingerprint-addressed DownloadDir cache must use the same lock domain.
+        # POSIX record locks and BSD flock locks do not exclude one another on Linux.
+        with atomic_directory(target_dir, lock_style=FileLockStyle.BSD) as atomic_dir:
             if not atomic_dir.is_finalized():
                 shutil.move(path, os.path.join(atomic_dir.work_dir, os.path.basename(path)))
         return Fingerprint.from_hashing_fingerprint(fingerprint)
--- a/pex/resolve/lockfile/download_manager.py
+++ b/pex/resolve/lockfile/download_manager.py
@@ -8,7 +8,7 @@ import json
 import os
 
 from pex import hashing
-from pex.atomic_directory import atomic_directory
+from pex.atomic_directory import AtomicDirectory, atomic_directory
 from pex.cache.dirs import DownloadDir
 from pex.common import safe_mkdtemp, safe_rmtree
 from pex.fs.lock import FileLockStyle
@@ -173,10 +173,30 @@ class DownloadManager(Generic["_A"]):
             if not retry:
                 raise ResultError(Error(str(e)))
 
-            TRACER.log(
-                "Found outdated downloaded artifact metadata, upgrading: {err}".format(err=e)
-            )
-            safe_rmtree(download_dir)
+            if hasattr(artifact, "fingerprint"):
+                # `atomic_directory` deliberately takes a lock only when its target does not
+                # exist. Re-check invalid metadata under the artifact lock: another process may
+                # have repaired it after our failed load but before we acquired the lock.
+                cached_download = AtomicDirectory(target_dir=download_dir, locked=True)
+                with cached_download.locked(lock_style=self._file_lock_style):
+                    try:
+                        return DownloadedArtifact.load(download_dir)
+                    except DownloadedArtifact.LoadError as locked_error:
+                        TRACER.log(
+                            "Found outdated downloaded artifact metadata, upgrading: "
+                            "{err}".format(err=locked_error)
+                        )
+                        safe_rmtree(download_dir)
+            else:
+                # Unfingerprinted artifacts use a private temporary directory, not the shared
+                # download cache, so there is no artifact lock to acquire for cleanup.
+                TRACER.log(
+                    "Found outdated downloaded artifact metadata, upgrading: {err}".format(err=e)
+                )
+                safe_rmtree(download_dir)
+
+            # Do not rebuild while holding the artifact lock: Pex's in-process file lock is
+            # intentionally non-reentrant. `atomic_directory` will re-check after acquiring it.
             return self.store(artifact, project_name, retry=False)
 
     def _download(
--- a/pex/resolve/lockfile/download_manager.py
+++ b/pex/resolve/lockfile/download_manager.py
@@ -129,7 +129,7 @@ class DownloadManager(Generic["_A"]):
     def __init__(
         self,
         pex_root=ENV,  # type: Union[str, Variables]
-        file_lock_style=FileLockStyle.POSIX,  # type: FileLockStyle.Value
+        file_lock_style=FileLockStyle.BSD,  # type: FileLockStyle.Value
     ):
         # type: (...) -> None
         self._pex_root = pex_root
--- a/tests/resolve/lockfile/test_download_manager.py
+++ b/tests/resolve/lockfile/test_download_manager.py
@@ -10,6 +10,7 @@ import pytest
 from pex import hashing
 from pex.artifact_url import ArtifactURL, Fingerprint
 from pex.cache.dirs import CacheDir
+from pex.fs.lock import FileLockStyle
 from pex.hashing import Sha1Fingerprint, Sha256Fingerprint
 from pex.pep_503 import ProjectName
 from pex.resolve.locked_resolve import FileArtifact
@@ -106,6 +107,12 @@ def test_storage_cache(
     downloaded_artifact2 = download_manager.store(artifact, project_name)
     assert downloaded_artifact1 == downloaded_artifact2
     assert 1 == len(download_manager.save_calls)
+
+
+def test_download_cache_uses_bsd_locks(download_manager):
+    # type: (FakeDownloadManager) -> None
+
+    assert FileLockStyle.BSD is download_manager._file_lock_style
 
 
 def test_storage_version_upgrade(
--- a/tests/resolve/test_downloads.py
+++ b/tests/resolve/test_downloads.py
@@ -1,0 +1,39 @@
+# Copyright 2026 Pex project contributors.
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+from contextlib import contextmanager
+
+from pex.fs.lock import FileLockStyle
+from pex.resolve import downloads
+from pex.resolve.downloads import ArtifactDownloader
+from pex.typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from typing import Any, Iterator
+
+
+def test_fingerprint_download_uses_bsd_lock(tmpdir, monkeypatch):
+    # type: (Any, Any) -> None
+
+    artifact = tmpdir.join("artifact.whl")
+    with open(artifact, "wb") as fp:
+        fp.write(b"content")
+    lock_styles = []
+
+    @contextmanager
+    def observe_atomic_directory(_target_dir, lock_style=None):
+        # type: (str, Any) -> Iterator[Any]
+        lock_styles.append(lock_style)
+
+        class Finalized(object):
+            @staticmethod
+            def is_finalized():
+                # type: () -> bool
+                return True
+
+        yield Finalized()
+
+    monkeypatch.setattr(downloads, "atomic_directory", observe_atomic_directory)
+
+    ArtifactDownloader._fingerprint_and_move(artifact)
+    assert [FileLockStyle.BSD] == lock_styles


jsirois:
@mjlbach is this theoretical or did you hit a real problem? This mechanism has been hammered by the Pants rust engine highly concurrently for years. I have no bug reports from that channel.

Keep in mind, all the critical uses are through the `atomic_directory` context manager.

mjlbach:
No, this is from a real CI failure. The failing run used Pants 2.32.1, PEX 2.95.1, CPython 3.12.7, and `[python].run_against_entire_lockfile = true` on ephemeral Linux x86_64 runners. Four independent jobs failed in the same run with the same shape but different artifacts.

Representative trace (timestamps and repository-specific paths removed):

```text
[INFO] Starting: Building pytest.pex
[INFO] Starting: Installing the user resolve lockfile
[INFO] Canceled: Building pytest.pex
[INFO] Starting: Building pytest.pex
[INFO] Completed: Building pytest.pex
[ERROR] ProcessExecutionFailure: Process 'Building pytest.pex' failed with exit code 1.

stderr:
There was 1 error downloading required artifacts:
1. mdurl 0.1.2 from https://files.pythonhosted.org/.../mdurl-0.1.2-py3-none-any.whl
    [Errno 2] No such file or directory:
    '~/.cache/pants/named_caches/pex_root/downloads/1/<hash>.lck.work/mdurl-0.1.2-py3-none-any.whl'

[INFO] Canceled: Installing the user resolve lockfile
```

The other three jobs lost `grpc_interceptor-0.15.4`, `httpx-0.28.1`, and `pyyaml-6.0.3` from their respective `downloads/1/<hash>.lck.work` directories. In some jobs `pytest.pex` was cancelled and restarted; in others the user-resolve installation was cancelled and restarted.

I have not produced a standalone PEX-only reproducer. With this patch applied from an organization-owned fork, the equivalent Pants CI path has repeatedly completed after exercising the same cancellation/restart sequence. I'm happy to provide additional PEX debug output or test a narrower fix if there is another lifecycle involved here.

jsirois:
Please send logs and confirm no network mounts are involved in the relevant paths.

mjlbach:
The representative logs are embedded in my previous comment here: https://github.com/pex-tool/pex/pull/3204#issuecomment-4952907233. That comment includes the cancellation/restart sequence, the missing `.lck.work` path, and the three corroborating failures from the same run.

No network mount is configured for the relevant paths. These are ephemeral Linux VMs, and the workflow uses `pantsbuild/actions/init-pants@v11` with its default named-cache location:

```text
/home/runner/.cache/pants/named_caches
```

The action uses `actions/cache@v5` to download a cache archive and extract it into that local path before Pants starts. In the representative failed job, the sequence was:

```text
20:46:44  begin named-cache restore
20:46:46  downloaded the 766 MB cache archive
20:46:46  /usr/bin/tar -xf .../cache.tzst -P -C $GITHUB_WORKSPACE
20:46:47  cache restore completed successfully
20:46:56  Pants started the concurrent PEX operations
20:47:29  missing downloads/1/<hash>.lck.work/... failure
```

There is no cache extraction or upload running concurrently with Pants. The post-job upload occurs after the job, and each matrix job receives its own extracted copy on a separate ephemeral VM. Within Pants sandboxes, `.cache/pex_root` is the normal append-only-cache symlink to the host-local `~/.cache/pants/named_caches/pex_root`; it is not a bind to shared storage between jobs.

We also explicitly disable Pants remote cache reads/writes, and the init action's experimental GHA remote-cache mode is disabled. The LMDB process cache is restored separately and is not the `pex_root` path.

One potentially relevant detail: the named-cache action permits prefix fallback restores, and this job restored a prior successful archive whose key matched the fallback prefix rather than the current exact lockfile hash. That can make the cache stale, but PEX should still own and repair its contents under the atomic-directory lock. I can disable named-cache restoration entirely for a comparison run if that would help isolate restored state from same-process cancellation.

jsirois:
My 1st step is confirming the failure is in a code path that uses the atomic_directory context manager.

There are a few very old ones that do not and instead just use AtomicDirectory directly.

mjlbach:
Let me know if I can help! I dug into this after we upgraded pants and started seeing flaky CI. I haven't been able to get a 100% deterministic repro but have not seen it since this was merged on our pex fork.

jsirois:
There was once a Pants (maybe 4 years ago now?) bug that `kill -9` processes, that bug may be back, and your workaround may paper over it.

jsirois:
@mjlbach what version was the Pants upgrade from? (To help bisect).

mjlbach:
2.32.0 and 2.32.1 no issue on 2.31.0

Sent via Superhuman ( ***@***.*** )

jsirois:
Thanks. Referencing the old issue for my own study: https://github.com/pantsbuild/pants/pull/18632

mjlbach:
I did some more digging and wanted to run a few things by you:

- [`DownloadedArtifact.load()` rejects all v2 metadata](https://github.com/pex-tool/pex/blob/098d3d0686a7d4c0ec2de6fd0dfddad10f30c2db/pex/resolve/lockfile/download_manager.py#L98-L114), although v3 only added `editable`. Would it be valid to accept v2 as `editable=False` when the requested artifact is non-editable, while still rebuilding it for an editable request?
- On a metadata load failure, [`download_dir` is removed outside the artifact lock](https://github.com/pex-tool/pex/blob/098d3d0686a7d4c0ec2de6fd0dfddad10f30c2db/pex/resolve/lockfile/download_manager.py#L170-L180). Should this instead acquire the artifact lock, re-check after acquiring it, remove only if it is still invalid, and release the lock before rebuilding?
- The same `DownloadDir` namespace appears to use different lock domains: [`LockDownloader` explicitly selects BSD locks](https://github.com/pex-tool/pex/blob/098d3d0686a7d4c0ec2de6fd0dfddad10f30c2db/pex/resolve/lock_downloader.py#L224-L240), while [`DownloadManager` defaults to POSIX](https://github.com/pex-tool/pex/blob/098d3d0686a7d4c0ec2de6fd0dfddad10f30c2db/pex/resolve/lockfile/download_manager.py#L129-L157) and [`ArtifactDownloader._fingerprint_and_move()` uses the default lock style](https://github.com/pex-tool/pex/blob/098d3d0686a7d4c0ec2de6fd0dfddad10f30c2db/pex/resolve/downloads.py#L59-L68). Since `lockf` and `flock` do not mutually exclude on Linux, should every writer to this cache use the same lock style, presumably BSD?

mjlbach:
Updated the PR: removed commit 1 and commit 3, and removed the added tests. The PR now contains only the minimal lock-guarded cache invalidation fix from commit 2. The existing targeted download-manager test suite passes (4 passed).

jsirois:
> Updated the PR, I don't think commit 3 is strictly necessary.

Commit 1 is pretty much irrelevant, IIUC you're just trying to lessen the frequency of what you think you solve in commit 2 - is that right? If so I'd prefer it is removed to kill the noise in the PR.

Do you mind removing commit 1 + commit 3 and don't worry about the tests. I'd like to get the the fix understood and minimal and confirmed by you in your repo if possible.

The lock style alignment can come later. I think the real answer there should be wholesale switch to flock on Unix, but previously backward compatibility concerns prevented that. I think I broke backwards compatibility in that regard long ago though; so the right way forward is to embrace that here and just use flock globally. I'm happy to take up that after your more minimal fix is in here - if indeed the `rmtree` under the lock guard is the real fix.

jsirois:
Alright - LGTM. I let CI rip @mjlbach but if you could report back in the affirmative that this scoped fix indeed fixes your CI flakes, that would be lovely.

mjlbach:
> Commit 1 is pretty much irrelevant, IIUC you're just trying to lessen the frequency of what you think you solve in commit 2 - is that right? If so I'd prefer it is removed to kill the noise in the PR.

Yes, sorry.

> Do you mind removing commit 1 + commit 3 and don't worry about the tests. I'd like to get the the fix understood and minimal and confirmed by you in your repo if possible.

Done and removed the tests

I'm testing now and not seeing the issue but will let CI hit a few more times.

cburroughs:
I use Pex by way of Pants and use ephemeral k8s pods in CI.  I've seen similar  `[Errno 2] No such file or directory` messages before (specifically May with Pants `2.31.1rc0` which used Pex `v2.81.0`)   At the time I was using a complicated "Voids the Warranty" configuration with PEX_CACHE split across multiple file systems to try to mitigate problems with two proprietary vendors.  I was unable to reproduce the issue outside of that environment and it felt wrong to risk any open source volunteer goose chases with an incomplete report going towards something that might be specific to a proprietary situation.

I wanted to provide that as context for a few -- I hope -- clarifying questions:

> Four independent jobs failed

Is the Pex cache (or full Pants named cache) shared between these jobs?  Either by running on the same node, or by a tarball of the cache from a prior job being seeded?

> CI began failing nondeterministically with downloaded wheels disappearing during concurrent resolution

And it continues to fail? Can you say something like "X% of builds fail in this way"?

> I dug into this after we upgraded pants and started seeing flaky CI

It sounds like you might be past when this was easy to do, but were you able to test either "Does rolling back Pants fix it?" or "Does running new-Pants with old-Pex have the same behavior?"

I'm currently going through old CI logs to see if I can find any more recent occurrences that may have been masked by a retry.

jsirois:
@cburroughs if you look at the change, @mjlbach found what appears to be a fairly obvious bug that would seem to explain this.

cburroughs:
I did look at the code and agree.  (It was actually working backwards from the code instead of the description that reminded me of what I had seen before.) My personal hesitation comes from being suspicious of this location in the past and being unable to prove a problem really exited, which gives me a nagging feeling that I'm missing something.

mjlbach:
@jsirois do you mind re-running? One of the tests looks like it had a network issue.

jsirois:
Ok, I have had time to fully analyze all this:

@mjlbach your reported bug cannot come from the rmtree fix. That fix is needed generally though for this reason:

Given: An AtomicDirectory is finalized and never locked iff its target directory exists.
An rmtree deletes all files under the target directory before finally removing the target directory.

As such, an unlocked rmtree on an atomic directory should lead to observible missing files in a presumably finalized atomic directory. The OP case did not have this problem. It had the problem of not seeing a work directory file.

Now the BSD / Posix mix issue does allow two processes to obtain the atomic_directory lock, meaning two processes can arrive in this code:
https://github.com/pex-tool/pex/blob/098d3d0686a7d4c0ec2de6fd0dfddad10f30c2db/pex/atomic_directory.py#L282-L318

While one of those processes is executing finalize on line 316, another can be executing safe_mkdir on line 308, and that line will yank workdir contents out from under the finalize.

@cburroughs & @mjlbach do you both buy this?
@mjlbach further, have you hit a repro yet using the cache restores of v2 metadata + this fix as it stands? You should have.

mjlbach:
The four failing jobs ran on separate ephemeral GitHub Actions runners. Each restored its own local copy of the same 766 MB Pants named-cache archive through the same fallback cache key, so there was no live filesystem shared between jobs. Each job then failed on a different artifact with the same missing `<hash>.lck.work` pattern.

The failing run used Pants 2.32.1 and CPython 3.12.7. We did not observe the issue on Pants 2.31.0, but I did not run a controlled comparison using newer Pants with the older PEX version. I would guess the failure rate was around 20%, although I did not measure it systematically. The strongest occurrence was four failures in one run.

One important qualification is that our Hedra CI workaround used the earlier owner-specific work-directory patch, not the current scoped patch in this PR.

The restored v2 metadata explanation is also an inference. The logs prove that an older fallback cache archive was restored, but we did not inspect that archive and record its metadata versions.

jsirois:
Fair warning, I'm working on a PR in parallel to fix this minimally and completely just to go through the exercise with code in my hands - which helps my brain.

mjlbach:
I'm totally ok if you take over :) My instinct is always just to file a PR with a bug report.

jsirois:
@mjlbach what I really want is your analysis of my analysis and a report back if you hit the issue in your CI with this current fix, which I contend you should, since this fix has 0 impact on the work dir, which is the OP error case.

jsirois:
> My instinct is always just to file a PR with a bug report.

I do appreciate that - good form. This one was weird though since the bug report is in Pants and the PR here.

mjlbach:
Yes, I'm re-running CI until we hit it.

tobni:
Unsolicited armchair analysis:

I think `DownloadManager.store` can create unlocked writers due to the repeated `is_finalized` check?

I dont really have the full context here but re-stating after an atomic dir is yielded that might not lock and write to it as if it were seems race-prone.

mjlbach:
I was able to repro the issue, I'm not able to reproduce the issue with my original commit 3 fixing the locks.

PR #3207, opened 2026-07-14
Fix concurrent use of artifact downloads.

jsirois:
Previously there were 2 issues handling downloaded artifact metadata:

1. When download metadata was out of date or invalid, the download
directory was non-atomically deleted and re-created via a fresh
download. Fix the non-atomicity by retaining the download directory and
just atomically upgrading the metadata files in the download directory
themselves.

2. Additionally switch to using `flock` for all unix locking. This
ensures locking a given directory works across processes and threads
and all parts of the code base. Previously the choice between `fcntl`
and `flock` led to attempting to lock the same path with different lock
types which did not properly mutually exclude.

Both of the issues here were brought to light by @mjlbach in #3204.


mjlbach:
Thank you so much for looking into this!

jsirois:
@mjlbach I'd love your review of this. I pre-emptively prepared a release, but if you and other reviewers here see problems, we can advance #3204 and I can circle back later.

jsirois:
@cburroughs and @tobni adding you both for a look since you got involved. The more eyes the better on locking-related issues like this. I really appreciate any time you can devote to review.

pex/resolve/lockfile/download_manager.py
```diff
@@ -51,27 +52,58 @@ def metadata_filename(artifact_dir):
...
+        project_name,  # type: ProjectName
+        download_dir,  # type: str
+        filename,  # type: str
+    ):
+        # type: (...) -> DownloadedArtifact
+
+        artifact_digests = ArtifactDigests(artifact)
+        archive = os.path.join(download_dir, filename)
+        hashing.file_hash(archive, artifact_digests.digest)
+        artifact_digests.check(project_name=project_name)
```
tobni: This check is comparing different fingerprints in the VCS case, I think?

pex/resolve/lockfile/download_manager.py
```diff
@@ -149,63 +244,74 @@ def store(
...
+            filename = self._download(artifact, project_name, download_dir)
 
         try:
             return DownloadedArtifact.load(download_dir)
         except DownloadedArtifact.LoadError as e:
-            if not retry:
-                raise ResultError(Error(str(e)))
+            if filename is None:
+                dists = [
+                    path for path in os.listdir(download_dir) if is_wheel(path) or is_sdist(path)
```
tobni: I think this discards local project entries.

pex/resolve/lockfile/download_manager.py
```diff
@@ -149,63 +244,74 @@ def store(
...
+                        )
+                    )
+                filename = dists[0]
 
             TRACER.log(
                 "Found outdated downloaded artifact metadata, upgrading: {err}".format(err=e)
             )
-            safe_rmtree(download_dir)
-            return self.store(artifact, project_name, retry=False)
+            return DownloadedArtifact.upgrade_existing(
```
tobni: This seems to have slightly different guarantees? Previous impl could recover from any unloadable entry, this will hard error on missing/corrupt wheels. Hard for me to trace if that was a prior guarantee.

tobni (commented):
The race looks gone but I think some corner-cases along the v2->v3 upgrade path might be left as errors in this patch.

pex/resolve/lockfile/download_manager.py
jsirois: That is right. In fact all of the Pex cache sitting behind finalized `atomic_directory` is this way. That's just accepted. If an atomic_directory block completes without raising, that is an assertion the contents of that directory are Ok. If something external to Pex corrupts the contained files, Pex generically is broken until you run `pex3 cache purge`.

The reality is the prior code was only trying to upgrade a single metadata file and was using nuking the whole atomic directory as a "cheap and easy" way to do this. That way had bugs though. This way is more direct and saves un-needed re-downloads. The downloaded artifact was never the problem - just the metadata Pex generated for it was too old.

pex/resolve/lockfile/download_manager.py
jsirois: Yes - good catch. This new scheme will need some re-work / re-thought.

pex/resolve/lockfile/download_manager.py
jsirois: Aha - it does. Fixed.

jsirois:
Alright - both remaining issues above that @tobni discovered are fixed in the latest commit by delegating finding and hashing artifacts to the `DownloadManager` subclass - which knows these details intimately for the artifact types it deals with.

tests/integration/test_pex_bootstrapper.py
```diff
@@ -308,7 +308,14 @@ def assert_boot(python=None):
...
         args = [python] if python else []
         args.extend([pex, "-c", "import psutil, sys; print(sys.executable)"])
-        output = subprocess.check_output(args=args, stderr=subprocess.PIPE)
+        pex_python_path = [sys.executable]
+        if python:
+            pex_python_path.append(python)
+        output = subprocess.check_output(
+            args=args,
+            stderr=subprocess.PIPE,
+            env=make_env(PEX_PYTHON_PATH=os.pathsep.join(pex_python_path)),
```
jsirois, line 317: Constraining with PEX_PYTHON_PATH was an unrelated fix. I have 3 versions of Python 3.14 on my machine which confounded the `assert_boot(other_interpreter)` case.

pex/resolve/lockfile/download_manager.py
jsirois: Ok, an IT now catches this.

pex/resolve/lockfile/download_manager.py
jsirois: And an IT now exercises the local project case as well as all other artifact type cases.

pex/resolve/lockfile/create.py
```diff
@@ -229,6 +229,43 @@ def store_all(
...
+                        artifacts="\n".join(
+                            "{idx}. {artifact}".format(idx=idx, artifact=artifact)
+                            for idx, artifact in enumerate(archives, start=1)
+                        ),
+                    )
+                )
+            )
+        filename = archives[0]
+        archive = os.path.join(download_dir, filename)
+        hashing.file_hash(archive, digest=digest)
```
tobni, line 266: Likely harmless but I think this hashes differently than the resolve case for VCS?

tobni (approved):
LGTM

pex/resolve/lockfile/create.py
jsirois: What do you mean by the resolve case?

tobni: When pexes are built from lockfiles. VCSArtifactDownloadManager uses digest_vcs_archive, here it is file_hash.

jsirois: Ok, yeah. So that prompts the fact that this code path, though used, has done ~nothing for quite some time. Embarrassing! The only good thing I can say is the ITs are so extensive at this point, its hard to slip much by them except locking / timing bugs or corners of the PyPA / abuse thereof in the wild Pex users have not yet hit.

Big delete coming in this file in a followup PR. Thanks for pointing this out.

---
For posterity, this code was introduced in https://github.com/pex-tool/pex/pull/1650. The point was to take artifacts downloaded by `pip download --dest <download dir> ...` - which is the "resolve" that drives lock creation, and move them to their individual `${PEX_ROOT}/downloads/2/<artifact hash>` cache dirs where they could be seen when the lock next was used to short-circuit missing artifact downloads. The `<artifact hash>` was already either read from the index metadata or else filled in by the `Locker` (which uses `digest_local_project`, `digest_vcs_repo` and `fingerprint_downloaded_vcs_archive` appropriately producing verified digests). So the DownloadManager base class was being used - very awkwardly - and the digesting bit of its API was going unused / ignored for those very pre-verified digests of local projects and VCS artifacts.

mjlbach:
Thank you!

jsirois:
@mjlbach you're welcome. Thanks you for finding the issue and digging enough to highlight the two generally pertinent issues - the mixed locks and the atomic directory delete. The latter is impossible to do atomically without a reader-writer lock and all dir readers taking a read lock. Pex does not have that infra for any case except the PEX_ROOT directory itself where `pex3 cache {prune,purge}` takes the write lock for delete and all other PEX runtime uses take the read lock very early in the boot phase.

I'll comment again when the release is out.

jsirois:
Alright, the fix is released at https://github.com/pex-tool/pex/releases/tag/v2.97.3
Thanks again @mjlbach and @tobni and sorry for the delay - this took longer than I'd like.
I'll be following up with cleanup of lock indexing (inconsistent - although unused - lock hashing issue) in a separate change.

PR #3211, opened 2026-07-15
Simplify lock creation download caching.

jsirois:
This worked previously, but it was not at all obvious. In particular,
it looked like digests calculated in the caching phase would not match
those calculated when locking. Simplification reveals download caching
only deals with artifacts that already have a fingerprint, and that
fingerprint can be re-used.

Prompted by @tobni's review of #3207.

C.F.: #1650 whose functionality this preserves.


jsirois:
@tobni I don't expect a review, but FYI. Thanks again for prompting inspection of this code.

Issue #2658, opened 2025-02-04
Windows Support


jsirois:
This tracks the second major effort at bringing Windows support to Pex. Notes will go here, issues may be broken out for actual issues.

jsirois:
Since Python 3.9 is now the current last supported release from CPython (see: https://devguide.python.org/versions/#python-release-cycle), I'll be starting there.

jsirois:
Major issues:
+ [ ] `\r\n` vs `\n`
+ [ ] `\` vs `/` in paths
+ [ ] `Lib` and `Scripts` vs `lib` and `bin` / general differences in venv layout.
+ [ ] `.exe` and `PATHEXT` for executable & script file names
+ [ ] console scripts need to be a binary - stub system
+ [ ] ~260 char path length limit: You can exceed with a registry edit (requires Admin) and an exe with a manifest enabling (Python exes have this manifest). This is available to Windows 10+ from August 2, 2016 on in the 1607 "Redstone 1" release. See:
    + https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry#enable-long-paths-in-windows-10-version-1607-and-later
    + https://en.wikipedia.org/wiki/Windows_10,_version_1607

jsirois:
Ok, the basic machinery is now solid: https://github.com/pex-tool/pex/runs/37826704273

Currently at 740 passed, 849 failed and 148 skipped.

jsirois:
745 passed, 844 failed and 148 skipped

jsirois:
757 passed, 780 failed and 201 skipped

rbebb:
I know this is still a WIP, but Windows support would be awesome!

ryanscovill:
@jsirois Appreciate this effort. Any updates?

jsirois:
@ryanscovill sure:

> Just being able to run a PEX file on Windows would be huge.

This now works via https://github.com/pex-tool/pex.rc. You can either download the appropriate `pexrc` binary for your machine from the releases page and run `pexrc inject <existing PEX>` or else you can build the pex with `--rc` on the latest versions of Pex (Pex will just download the appropriate `pexrc` binary for you and run inject on the PEX it builds for you).

What this means:
+ Building a PEX on Windows is still _not_ supported. It may or may not work (I know it works for some cases because the https://github.com/pex-tool/pex.rc Windows arm64 and amd64 CI test shards do this).
[...]

PR #3213, opened 2026-07-18
Add support for windowed PEX scies on Windows.

jsirois:
More work towards #2658.


jsirois:
Noting the fix in c09a7f3 deals with https://github.com/pypa/pip/pull/14150 as caught by the nightly Pip adhoc CI job. This is otherwise unrelated to this PR, but saves trees.

PR #3225, opened 2026-07-21
Fix PEP-691 fingerprinting against authenticated indexes.

### LINKED ISSUE #3224: PEP-691 fingerprinting cannot authenticate to basic-auth indexes (AWS CodeArtifact): lock creation downloads every artifact and builds sdists
## Summary

When a `--style universal` lock is created against an index that requires Basic authentication and serves the PEP-691 JSON Simple API, Pex cannot authenticate the follow-up requests made by its PEP-691 `FingerprintService`. Pex warns and then falls back to downloading every retained artifact for each selected locked version just to fingerprint it. Callers such as Pants suppress Pex warnings, making the fallback operationally silent.

Because `pip download --no-deps <direct-sdist-url>` can run PEP-517 metadata hooks, retained sdists may then fail the lock for reasons unrelated to dependency resolution—for example, `psycopg2-binary` without `pg_config`, or `ddtrace==3.19.1` importing `pkg_resources` after build isolation installs setuptools 82, which removed it. Even when every retained sdist can prepare metadata, lock creation needlessly downloads and hashes a potentially large artifact set.

This surfaced when AWS CodeArtifact began serving PEP 691 in July 2026. Pip >=22.2 prefers JSON through content negotiation, so locks that had regenerated successfully against PEP-503 HTML began failing without a client-side configuration change. With HTML, artifact URLs include `#sha256=` fragments and Pex obtains hashes directly from Pip's log; with JSON, hashes are separate `files[].hashes` values and are not included in Pip's logged URLs.

### 1. PEP-691 endpoints are scraped from Pip's credential-redacted log

`Locker.analyze` records endpoints from Pip lines of this form:

```text
Fetched page <url> as application/vnd.pypi.simple.v1+json
```

For an index configured as `https://user:secret@host/simple`, Pip logs the endpoint as literal `https://user:****@host/simple/<project>/`. Before the fix, Pex attempted to open that URL:

```text
pex: Fetching PEP-691 index metadata from https://aws:****@<host>/pypi/<repo>/simple/<project>/ for application/vnd.pypi.simple.v1+json
```

The stdlib opener cannot parse its authority:

```text
PEXWarning: Failed to fetch project metadata, continuing: nonnumeric port: '****@<host>'
```

Pex separately retains the real repository credential, but the redacted userinfo in the endpoint prevents that credential from being used.

### 2. Clean endpoint URLs still fail against CodeArtifact's challenge

With the redacted userinfo removed and a matching machine-scoped password entry supplied, the pre-fix `URLFetcher` still starts unauthenticated. CodeArtifact accepts Basic authentication but answers an unauthenticated request with `401` and `WWW-Authenticate: Bearer ...`. Python's stdlib Basic/Digest challenge handlers do not support that scheme and can raise:

```text
ValueError: AbstractDigestAuthHandler does not support the following scheme: 'Bearer'
```

As a result, the known Basic credential is never sent. This is in the same family as #1803, which #1811 fixed for artifact downloads by delegating them to Pip; the independently implemented PEP-691 metadata client retained the problematic stdlib authentication path.

`FingerprintService.fingerprint` intentionally treats these request errors as non-fatal, emits `Failed to fetch project metadata, continuing`, and leaves artifacts without fingerprints. Higher-level lock creation then downloads each unresolved retained artifact and hashes it locally.

## Reproduction

Against a CodeArtifact repository with a public PyPI upstream, from a lock host without `pg_config`:

```shell
TOKEN=$(aws codeartifact get-authorization-token ... --query authorizationToken --output text)
pex3 lock create \
  --pip-version 24.2 --style universal --no-pypi \
  --index "https://aws:${TOKEN}@<domain>-<acct>.d.codeartifact.<region>.amazonaws.com/pypi/<repo>/simple" \
  psycopg2-binary==2.9.12 -o lock.json
```

On v2.66.0 and `main` at bea9c2e2, this fails with `metadata-generation-failed` / `pg_config executable not found` after downloading all 12 retained artifacts for the selected release.

The challenge failure can be reproduced without AWS by giving `URLFetcher.get_body_stream` a matching machine-scoped password entry and pointing it at a server that accepts Basic auth but returns `401` with `WWW-Authenticate: Bearer` when the initial request lacks authorization.

## Expected behavior

PEP-691 metadata requests should authenticate using the known credential for the configured index, hashes should be recovered from the JSON `files[].hashes` mappings, and no retained artifacts should be downloaded solely for fingerprinting. This should produce the same lock artifact set and hashes as the equivalent PEP-503 HTML response with `#sha256=` fragments.

## Fix

#3225 strips credentials from Pip-redacted endpoint URLs, sends known machine-scoped Basic credentials preemptively, safely handles unsupported challenge schemes, and adds composed regression coverage from the redacted Pip log line through authenticated PEP-691 SHA-256 recovery. It has been validated against the CodeArtifact repository that exposed the issue: the reproducer completes with zero fallback fingerprint downloads and produces hashes identical to the HTML-index lock.

jasonwbarnett:
Fixes #3224.

Two changes, both required for `pex3 lock create` to fingerprint artifacts from an authenticated PEP-691 index (AWS CodeArtifact) instead of downloading every one of them:

1. `Locker` strips credentials from the PEP-691 endpoints it scrapes from the Pip log. Pip redacts credentials in the URLs it logs (`https://user:****@...`), so the scraped values can never authenticate and do not even parse as a netloc in the stdlib openers (`nonnumeric port: '****@...'`). The `URLFetcher` making the PEP-691 requests re-attaches real credentials for the endpoint's machine from its `PasswordDatabase`.
2. `URLFetcher.get_body_stream` sends known credentials preemptively via a Basic `Authorization` header — from the machine-scoped `PasswordDatabase` entry matching the URL, or from credentials embedded in the URL itself (which are moved out of the netloc, since the stdlib openers cannot parse a credentialed netloc). This is what makes authentication succeed: CodeArtifact accepts Basic auth but answers unauthenticated requests with `401 WWW-Authenticate: Bearer`, which the stdlib challenge handlers cannot process, so a challenge/response round trip never authenticates. Sending credentials up front avoids provoking a challenge at all.

Machine-less `default` netrc entries are deliberately never sent preemptively — they remain challenge-only, so credentials are not offered to arbitrary hosts.

Which mechanisms are load-bearing was verified by instrumenting each one during a real lock against a CodeArtifact index: the endpoint strip fires (`has_redacted_credentials=True`) and the preemptive Basic header fires. An earlier revision of this PR also guarded the stdlib auth handlers against challenge schemes they do not implement; that code never executed on the success path and only changed the error surfaced when credentials are unknown, so it has been dropped (see #3224 for the detail).

Validated against the CodeArtifact repository that surfaced #3224:

- before (`main` @ bea9c2e2): `pex3 lock create --pip-version 24.2 --style universal --no-pypi --index https://aws:<token>@<host>/pypi/<repo>/simple psycopg2-binary==2.9.12` exits 1 (`pg_config` sdist metadata build failure) after downloading the release's 11 other artifacts to fingerprint them;
- after: exits 0 with zero fingerprint downloads, and the resulting lock's artifact set and sha256s are byte-identical to those produced when the same index serves PEP 503 HTML (`#sha256=` fragments), including the sdist hash, which matches the index's advertised `hashes.sha256` exactly. Same result for other previously-failing projects (`ddtrace` et al).
- With only the endpoint strip and no preemptive auth, the `Bearer` challenge is reached, `AbstractDigestAuthHandler does not support the following scheme: 'Bearer'` is raised and swallowed to a warning, and the lock fails as before — confirming (2) is what fixes this.

`format`, `lint-check`, `typecheck` and the touched tests (`tests/test_auth.py`, `tests/test_fetcher.py`, `tests/resolve/test_locker.py`, `tests/resolve/pep_691/`) pass locally; the fetcher tests exercise a local HTTP server that mimics CodeArtifact's Bearer-challenge behavior.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_011ftLY926gkJA6yMbu813cG

### DIFF
--- a/pex/auth.py
+++ b/pex/auth.py
@@ -165,3 +165,21 @@ class PasswordDatabase(object):
     def append(self, entries):
         # type: (Iterable[PasswordEntry]) -> PasswordDatabase
         return PasswordDatabase(entries=self.entries + tuple(entries))
+
+    def lookup(self, url):
+        # type: (Text) -> Optional[PasswordEntry]
+        """Finds the password entry for the machine serving the given URL, if any.
+
+        N.B.: Only machine-scoped entries are considered; the credentials of a machine-less
+        `default` netrc entry are only safe to present in response to a challenge, not
+        preemptively to arbitrary hosts.
+        """
+        try:
+            machine = Machine.from_url(url)
+        except ValueError:
+            return None
+
+        for entry in self.entries:
+            if entry.machine == machine:
+                return entry
+        return None
--- a/pex/auth.py
+++ b/pex/auth.py
@@ -179,7 +179,9 @@ class PasswordDatabase(object):
         except ValueError:
             return None
 
-        for entry in self.entries:
+        # Later entries are appended from more explicit configuration sources and should take
+        # precedence over earlier entries (e.g. an index URL credential over a netrc credential).
+        for entry in reversed(self.entries):
             if entry.machine == machine:
                 return entry
         return None
--- a/pex/fetcher.py
+++ b/pex/fetcher.py
@@ -3,6 +3,7 @@
 
 from __future__ import absolute_import
 
+import base64
 import contextlib
 import os
 import socket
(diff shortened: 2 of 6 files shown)


jsirois:
@jasonwbarnett thanks for your efforts here. I think https://github.com/pex-tool/pex/releases/tag/v2.98.5 should allow you to close this PR, but let me know if not.

jasonwbarnett:
> @jasonwbarnett thanks for your efforts here. I think https://github.com/pex-tool/pex/releases/tag/v2.98.5 should allow you to close this PR, but let me know if not.

I'll try it out today and let you know.

jasonwbarnett:
@jsirois it works! 🎉

PR #3226, opened 2026-07-27
Fix `--venv` PEX cache handling.

### LINKED ISSUE #3223: `pex3 cache prune` deletes in-use `--scie eager --venv` venvs
A `--venv --scie eager` PEX boots by exec'ing its cached venv interpreter directly, via the build-time-baked `configure` scie binding. That fast path never calls `record_access`, so the venv's `.last-access` mtime stays frozen at first-boot time no matter how often the scie runs. `pex3 cache prune --older-than` decides staleness purely from that mtime, so it eventually deletes the venv of a scie that's in daily use. The next boot of the same, unchanged scie then fails with `ENOENT` — it execs the deleted path instead of rebuilding.

## Environment

pex 2.98.2, CPython 3.10 (PBS), `--scie eager --venv`, Linux x86-64.

## Minimal reproduction

```bash
set -eu
cd "$(mktemp -d)"
printf 'import sys\nprint(sys.executable)\n' > app.py

# Build and boot once to create the venv.
pex --venv --scie eager -D . -m app -o ./hello
./hello   # prints .../venvs/3/<pex_hash>/<contents_hash>/bin/python

# Backdate .last-access to simulate a long-lived pin.
LA=$(find ~/.cache/pex/venvs -maxdepth 4 -name .last-access -not -path '*/s/*' | head -1)
touch -d '30 days ago' "$LA"

# Boot again, successfully.
./hello

# BUG 1: a successful boot did not refresh .last-access.
stat -c '%y' "$LA"   # still 30 days ago

# BUG 2: prune deletes the in-use venv, and the scie can no longer boot.
pex3 cache prune --older-than '7 days'
./hello              # ENOENT: No such file or directory
```

## Root cause

`record_access` is the only thing that refreshes `.last-access`, and its only runtime venv caller is `ensure_venv`:

```
pex/pex_bootstrapper.py:639-640
    if record_access:
        cache_access.record_access(venv_dir)
```

The `--scie eager` fast path skips `ensure_venv` entirely — it execs `<venv>/bin/python <venv>/__main__.py` from the baked binding, and the venv's own launcher never calls `record_access`. Prune, meanwhile, scans by mtime alone with no in-use guard:

```
pex/cache/access.py:127
    last_access = os.stat(_last_access_file(pex_dir)).st_mtime
```

## Suggested fixes (any one suffices)

1. Record access on the scie fast path, matching `ensure_venv`.
2. Make prune skip venvs still referenced by a scie short-link/binding.
3. Have the scie rebuild the venv on boot when the baked path is missing.

**Disclaimer**: I used AI to investigate this, the problem occurred in production environment once a while.

jsirois:
Fix `--venv` PEX scie cache robustness & fix missing `--venv`
last-access updates.

Fixes #3223

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,12 @@
 # Release Notes
 
+## 2.98.3
+
+This release fixes `--venv` PEX scies robustness to PEX_ROOT cache deletion and  also fixes `--venv`
+PEX behavior with regard to accurate `pex3 cache prune` last access time pruning.
+
+* Fix `--venv` PEX cache handling. (#3226)
+
 ## 2.98.2
 
 This release fixes generation of foreign platform PEX scies to respect the foreign platform cache
--- a/pex/cache/dirs.py
+++ b/pex/cache/dirs.py
@@ -317,7 +317,7 @@ class UnzipDir(AtomicCacheDir):
     def iter_all(cls, pex_root=ENV):
         # type: (Union[str, Variables]) -> Iterator[UnzipDir]
         for unzip_dir in glob.glob(CacheDir.UNZIPPED_PEXES.path("*", pex_root=pex_root)):
-            if os.path.isdir(unzip_dir):
+            if os.path.isdir(unzip_dir) and not unzip_dir.endswith(".work"):
                 pex_hash = os.path.basename(unzip_dir)
                 yield UnzipDir(path=unzip_dir, pex_hash=pex_hash, pex_root=pex_root)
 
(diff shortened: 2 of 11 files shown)


jsirois:
Thanks for your report @messense, this should solve the issues you found ~optimally for both robustness and boot latency. I'll let you know when this has released in v2.98.3 tomorrow.

jsirois:
Alright @messense, this is now released in https://github.com/pex-tool/pex/releases/tag/v2.98.3
Please let me know if that release doesn't fix things for you.

PR #3249, opened 2026-08-18
Handle foreign too-long Python script shebangs.

### LINKED ISSUE #3248: `--venv-repository`: data-script interpreter is not rewritten when the source venv path is long
**Pex:** 2.100.4 · **uv:** 0.11.32 · **Python:** CPython 3.12.10 · **Platform:** Linux x86_64

Subsetting a venv from a uv venv (`--venv-repository`) leaves scripts that come from a wheel's `*.data/scripts/` directory pointing at the **source** uv venv's interpreter instead of the new venv's python, so the new venv is not self-contained.

The trigger is the shebang length limit: when `<uv-venv>/bin/python` exceeds 127 bytes, uv writes a `/bin/sh` re-exec wrapper rather than a plain shebang.

```sh
#!/bin/sh
'''exec' '/very/long/path/to/uv-venv/bin/python' "$0" "$@"
' '''
```

Pex rewrites line 1, which is already `#!/bin/sh`, so the interpreter on line 2 survives into the new venv.

## Reproducer

This script demonstrates the bug using `dill`, which ships `undill` as a data script.

### repro.sh

```bash
#!/usr/bin/env bash
# `pex venv create --venv-repository`: a wheel's data script keeps the source uv venv's
# interpreter, because that venv's path is too long for a shebang.
set -euo pipefail

WORK=/tmp/pex-uv-repro

# "$SRC/bin/python" over the 127-byte shebang limit, so uv writes its
# /bin/sh re-exec wrapper instead of a plain shebang.
SRC=$WORK/a_directory_name_long_enough_to_push_the_shebang_over_the_127_byte_limit_aaaaaaaaaaaaaaaaaaaaa/venv
OUT=$WORK/out-venv
DEPENDENCY=dill

mkdir -p $WORK
cd $WORK

# Install PEX if needed
[ -x pex ] || { curl -fsSL -o pex https://github.com/pex-tool/pex/releases/download/v2.100.4/pex; chmod +x pex; }

uv venv --clear --python 3.12 "$SRC" >/dev/null
uv pip install --python "$SRC/bin/python" --quiet "$DEPENDENCY"

# Subset venv $SRC into $OUT
PEX_SCRIPT=pex3 ./pex venv create --force --dest-dir "$OUT" --venv-repository "$SRC" --pre $DEPENDENCY

echo
echo "== the executable '$OUT/bin/undill' references the old venv path on line 2:"
sed -n '1,2p' "$OUT/bin/undill"

echo
echo "== so it breaks as soon as the source uv venv is gone:"
mv "$SRC" "$SRC.gone"
"$OUT/bin/undill" || echo "^ exit $?"
```

### Expected

`$OUT/bin/undill` should point at the new venv's own python, i.e. Pex should rewrite the `'''exec' '<python>'` wrapper
form as well and never leaves a script pointing outside the venv it created.

### Actual

`$OUT/bin/undill` fails because it tries to load the python executable from the original venv path.

Script output:
```plain
Using CPython 3.12.10 interpreter at: /workdir/.pyenv/versions/3.12.10/bin/python3.12
Creating virtual environment at: a_directory_name_long_enough_to_push_the_shebang_over_the_127_byte_limit_aaaaaaaaaaaaaaaaaaaaa/venv
Activate with: source a_directory_name_long_enough_to_push_the_shebang_over_the_127_byte_limit_aaaaaaaaaaaaaaaaaaaaa/venv/bin/activate

== the executable '/tmp/pex-uv-repro-dill/out-venv/bin/undill' references the old venv path on line 2:
#!/bin/sh
'''exec' '/tmp/pex-uv-repro-dill/a_directory_name_long_enough_to_push_the_shebang_over_the_127_byte_limit_aaaaaaaaaaaaaaaaaaaaa/venv/bin/python' "$0" "$@"

== so it breaks as soon as the source uv venv is gone:
/tmp/pex-uv-repro-dill/out-venv/bin/undill: 2: exec: /tmp/pex-uv-repro-dill/a_directory_name_long_enough_to_push_the_shebang_over_the_127_byte_limit_aaaaaaaaaaaaaaaaaaaaa/venv/bin/python: not found
^ exit 127
```

jsirois:
Detect `#!/bin/sh` too-long script shebang re-directors created by other
tools like pip and uv for data scripts when round-tripping installed
wheels from venvs.

Fixes #3248

### DIFF
--- a/tests/integration/test_issue_3248.py
+++ b/tests/integration/test_issue_3248.py
@@ -1,0 +1,85 @@
+# Copyright 2026 Pex project contributors.
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
+
+from __future__ import absolute_import
+
+import shutil
+import subprocess
+import sys
+from textwrap import dedent
+
+import pytest
+
+from pex.venv.virtualenv import Virtualenv
+from testing import IS_PYPY
+from testing.cli import run_pex3
+from testing.pytest_utils.tmp import Tempdir
+
+
+@pytest.mark.skipif(
+    IS_PYPY or sys.version_info < (3, 8),
+    reason=(
+        "Use of uv is required and uv only supports Python >= 3.8 and the `undill` script does not "
+        "work with PyPy."
+    ),
+)
+def test_uv_too_long_data_scripts_shebang(tmpdir):
+    # type: (Tempdir) -> None
+
+    uv_venv_dir = tmpdir.join("uv-venv", "too-long", "x" * max(1, 128 - len(tmpdir.path)))
+    subprocess.check_call(
+        args=[
+            "uv",
+            "venv",
+            "--python",
+            "{major}.{minor}".format(major=sys.version_info[0], minor=sys.version_info[1]),
+            uv_venv_dir,
+        ]
+    )
+
+    uv_venv = Virtualenv(uv_venv_dir)
+    subprocess.check_call(
+        args=["uv", "pip", "install", "--python", uv_venv.interpreter.binary, "dill"]
+    )
+
+    pickled = tmpdir.join("hello.pkl")
+    uv_venv.interpreter.execute(
+        args=[
+            "-c",
+            dedent(
+                """\
+                import dill
+
+
+                with open({pickled!r}, "wb") as fp:
+                    dill.dump(["hello", "world"], fp)
+                """
+            ).format(pickled=pickled),
+        ]
+    )
+
+    pex_root = tmpdir.join("pex-root")
+    pex_venv_dir = tmpdir.join("pex-venv")
+    run_pex3(
+        "venv",
+        "create",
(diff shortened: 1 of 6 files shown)


tests/integration/test_issue_3248.py
```diff
@@ -0,0 +1,85 @@
...
+    pex_venv = Virtualenv(pex_venv_dir)
+
+    def assert_undill_works():
+        assert b"['hello', 'world']\n" == subprocess.check_output(
+            args=[pex_venv.bin_path("undill"), pickled]
+        )
+
+    assert_undill_works()
+    shutil.rmtree(uv_venv_dir)
+    assert_undill_works()
```
jsirois, line 85: N.B.: This was the failing line before the fix.

PR #3269, opened 2026-09-02
Raise on Directory Listing Errors in `deterministic_walk`

apetti1920:
### Summary
This change modifies `deterministic_walk` to raise `OSError` on directory listing failures by default, rather than silently ignoring them as `os.walk` does.

Under `os.walk`'s default `onerror=None` behavior, if a directory cannot be listed (e.g., due to permission issues or concurrent deletion), it is silently dropped along with all its contents. The walk completes successfully, returning a truncated tree. When consumers (such as `pex` during build time) use this truncated tree to reproduce a directory structure, they write out a partial copy. This can lead to subtle, hard-to-debug issues where downstream consumers receive an incomplete artifact (e.g., a wheel missing critical metadata files) because the error was masked at the walking stage.

By defaulting `onerror` to a function that re-raises the error, this PR ensures that any loss of data during traversal is surfaced immediately at the point of failure, allowing the build to fail fast with a clear error code and traceback.

### Key Changes
*   **New Helper Function**: Added `_reraise_walk_error` to explicitly re-raise `OSError` exceptions caught by `os.walk`.
*   **Default Behavior Update**: `deterministic_walk` now sets `onerror=_reraise_walk_error` if the user does not explicitly provide an `onerror` argument (either positional or keyword).
*   **Documentation**: Updated docstrings to explain the difference between `os.walk`'s silent discard behavior and the new raising behavior, noting that callers can still opt into the silent behavior by explicitly passing `onerror=None`.
*   **Test Coverage**:
    *   Added `test_deterministic_walk_raises_on_listing_error`: Verifies that an `OSError` is raised when a subtree becomes inaccessible during iteration.
    *   Added `test_deterministic_walk_onerror_can_be_overridden`: Ensures callers can still explicitly pass an `onerror` callback (positionally or via keyword) to handle errors gracefully if desired.
    *   Added `test_pex_builder_add_dist_surfaces_listing_errors`: A regression test using a mock `os.walk` that simulates listing errors to ensure `PEXBuilder` fails fast rather than caching a truncated wheel.

### Notable Implementation Details
*   **Argument Detection**: The implementation uses a heuristic to detect if `onerror` was explicitly passed. It checks `len(args) <= _ONERROR_POSITION` (position 2 in `os.walk`'s signature) and `"onerror" not in kwargs`. This allows users to pass `onerror` as a positional argument without triggering the default override.
*   **Scope of Errors**: The commit message notes that this only addresses *listing* errors. `os.walk` is also silent about `stat` errors (classifying entries as files) and partial listing errors (discarding remaining entries after a failure mid-list). The comments in the docstring highlight these remaining silent failures for future awareness, though this PR focuses on the most impactful listing errors.
*   **Backwards Compatibility**: Existing code that relies on `deterministic_walk` continuing despite errors will now fail. However, this is the desired "fail-safe" behavior for build tools. Users who need the old "best-effort" behavior can explicitly pass `onerror=None`.


jsirois (approved):
Thanks @apetti1920! Looks good to me.

pex/atomic_directory.py
```diff
@@ -258,6 +254,9 @@ def _atomic_directory(
 ):
     # type: (...) -> Iterator[None]
 
+    # We use double-checked locking with the check being target_dir existence and the lock being an
```
jsirois, line 257: Oops - this wasn't "Fix fmt." (which would be `uv run dev-cmd --skip test`). It was a change I had laying around. So this is unrelated, but fine - just a comment move to a more appropriate spot after #3263.

PR #3281, opened 2026-09-16
Make SSL optional in compatibility/fetcher

### LINKED ISSUE #3282: pex.compatibility / pex.fetcher fail to import on interpreters without SSL support
## Problem

Interpreters built without SSL support (e.g. embedded / AOSP Python builds, or CPython built with
`--without-ssl`) don't define `urllib.request.HTTPSHandler` (Python 3) or `urllib2.HTTPSHandler`
(Python 2) at all.

Because `pex/compatibility.py` does an unconditional `from urllib.request import HTTPSHandler as
HTTPSHandler` (and the Python 2 equivalent from `urllib2`), and `pex/fetcher.py` eagerly calls
`initialize_ssl_context()` and constructs an `HTTPSHandler(...)` at module scope, merely *importing*
`pex.compatibility` or `pex.fetcher` raises an `ImportError` on such interpreters — even for code
paths that never actually need to make an HTTPS request.

## Expected behavior

Importing `pex.compatibility` / `pex.fetcher` should succeed regardless of whether the interpreter
has SSL support. Only code paths that actually attempt to establish an HTTPS connection should fail,
and they should fail with a clear, actionable error rather than a bare `ImportError` at import time.

## Additional context

Fixed in #3281, which:
- Wraps the `HTTPSHandler` imports in `pex/compatibility.py` in `try`/`except ImportError`, exposing
  a `HAS_SSL` flag.
- Makes `pex/fetcher.py` skip eager SSL context initialization and skip registering an
  `HTTPSHandler` on `URLFetcher` when `HAS_SSL` is `False`.
- Raises a clear `IOError` (instead of an import-time `ImportError`) if something actually tries to
  establish an HTTPS connection on an interpreter without SSL support.

DonnaIsDoughnut:
Fixes #3282

Interpreters built without SSL support (e.g. embedded / AOSP Python builds) don't define `urllib.request.HTTPSHandler` / `urllib2.HTTPSHandler` at all. Previously, importing `pex.compatibility` or `pex.fetcher` on such an interpreter would fail immediately.

This wraps the SSL-specific imports in try/except and adds a `HAS_SSL` flag. `URLFetcher` only registers an `HTTPSHandler` when SSL is available, and eagerly initializing the default SSL context is skipped in that case. Attempting to actually establish an HTTPS connection without SSL support now raises a clear `IOError` instead of failing at import time.


jsirois:
@DonnaIsDoughnut Pex does not use fetcher at runtime (i.e.: PEX files do not use fetcher). The Pex runtime does use compatibility though. As such, it seems to me compatibility ought to just be broken up and fetcher can own its imports. Since the Pex buildtime always needs a fetcher with https support and the Pex runtime never needs a fetcher, this should be a cleaner split.

This is relevant: https://github.com/pex-tool/pex/issues/88. Towards that end, what chip architectures do you target for embedded? I want to make sure pex.rc supports those if possible. Current target coverage can be seen here: https://github.com/pex-tool/pex.rc/blob/main/rust-toolchain

I'll take a look at splitting things up as I suggest later today.

jsirois:
Ok @DonnaIsDoughnut, I lied. The URLFetcher can be used at PEX runtime in 1 case:
```console
:; PEX_TOOLS=1 ./my.pex venv --pip /install/pex/in/this/venv
```
If the Python interpreter used does not have the `ensurepip` module (which some Debian setups, for example, can be missing).
So this PR both re-organizes fetcher to own its imports and uses lazy imports of URLFetcher where needed: https://github.com/pex-tool/pex/pull/3283

jsirois:
@DonnaIsDoughnut please try out https://github.com/pex-tool/pex/releases/tag/v2.103.1 and close this PR if that fix works for your case.

jsirois:
@DonnaIsDoughnut I'm going to close this next Friday, September 25th if I haven't head back from you about your experience with https://github.com/pex-tool/pex/releases/tag/v2.103.1.

DonnaIsDoughnut:
Working perfectly. Thank you very much!

PR #3283, opened 2026-09-16
Isolate `URLFetcher` use in the PEX runtime.

### LINKED ISSUE #3282: pex.compatibility / pex.fetcher fail to import on interpreters without SSL support

jsirois:
Fixes #3282


pex/windows/__init__.py
```diff
@@ -67,6 +66,8 @@ def _fetch_stub(stub_name):
     stub_dir = os.path.join(_CACHE_DIR, stub_name)
     with atomic_directory(stub_dir) as atomic_dir:
         if not atomic_dir.is_finalized():
+            from pex.fetcher import URLFetcher
```
jsirois, line 69: N.B.: Besides the new lazy import in `Virtualenv` above, this is the only other lazy import introduced in PEX runtime code paths. This code path cannot trigger however for use of Pex from PyPI or the released Pex PEXes since those include the windows stubs pre-embedded. This case only occurs when using Pex to build itself in a fresh checkout that has no stubs bootstrapped yet.

PR #3279, opened 2026-09-14
Support creating a venv for a foreign platform.

jasonwbarnett:
Closes #3278.

A venv's layout is fixed by the Python implementation, version and OS; the
machine architecture plays no part. So a local interpreter can lay a venv out
for a foreign platform it differs from in architecture alone, leaving just the
venv Python to point at wherever the foreign interpreter will live.

The new `pex3 venv create --link-python` does that pointing. The path need not
exist when the venv is created, which is what makes the foreign case work: the
venv, its scripts and the foreign wheels all land locally and the Python link
only resolves once the venv reaches its final resting place. That is also
useful on its own for a venv built somewhere other than where it will run.

When the target is foreign, `--link-python` is required and a stand-in
interpreter is searched for on `--python-path` (`$PATH` by default), matching
the target's interpreter tag, ABI and `platform_system`. Nothing matching means
a fail fast naming what was needed. `--compile` now uses the venv's own Python
rather than the one running Pex, so a foreign venv gets bytecode for its own
version.

Built and verified on `linux-aarch64` for `linux_x86_64-cp-311-cp311`:

```
bin/python -> /usr/local/bin/python3.11
home = /usr/local/bin
lib/python3.11/site-packages/psutil/_psutil_linux.abi3.so: ELF 64-bit LSB shared object, x86-64
```

N.B.: `tests/integration/cli/commands/test_venv_create.py` grew a cross-arch
case that leans on `ensure_python_interpreter(PY310)`; it could not run where I
wrote it, since `pyenv install 3.10.7` will not build on Ubuntu 24.04. Its
assertions were checked against a real foreign venv built with 3.11 instead.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01JXoAkfaDfspdEqDzgHoTzG

PR #3285, opened 2026-09-17
Fix `pex3 venv create --link-python` for `--pex-repository`.

jasonwbarnett:
`pex3 venv create --link-python` for a foreign platform fails when the venv is installed from a `--pex-repository`:

```
A distribution for psutil could not be resolved for <local python>.
Found 1 distribution for psutil that does not apply:
1.) The wheel tags for psutil 5.9.5 are cp36-abi3-manylinux_2_12_x86_64, ... which do not match
    the supported tags of <local python>: cp313-cp313-manylinux_2_39_aarch64 ...
```

The distributions themselves install fine — it is populating the venv's *sources* that fails. `_populate_first_party` re-resolves the repository PEX to build the `pex-repl` script's `activated_dists`, and that resolve runs against a local interpreter, which a PEX built for the foreign platform alone holds nothing for. `_install_from_pex` has already resolved the distributions for the target, so this passes those down instead; the `PEX_TOOLS=1 ... venv` path keeps re-resolving, where the PEX and the interpreter always agree.

This is the combination Pants uses — it always invokes `pex3 venv create` with `--pex-repository` — so #3279 is unreachable from Pants without it.

Repro and verification, on a linux aarch64 host:

```
pex psutil==5.9.5 --platform linux_x86_64-cp-313-cp313 -o psutil.pex
pex3 venv create -d venv --pex-repository psutil.pex --platform linux_x86_64-cp-313-cp313 \
    --python-path $(command -v python3.13) --link-python /opt/python/bin/bob
```

Before: the error above. After: `bin/python`, `bin/python3` and `bin/python3.13` all link to `/opt/python/bin/bob`, and `lib/python3.13/site-packages/psutil/_psutil_linux.abi3.so` is `ELF 64-bit LSB shared object, x86-64`.

`test_foreign_target_link_python_from_pex_repository` covers it. I could not run it locally — the `py310` fixture's CPython 3.10.7 build fails in my environment (`ensurepip` exits non-zero) — so CI is its first run; the commands above are the same shape by hand. `uv run dev-cmd format lint typecheck` is green.

Written with Claude Code, as was #3279.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_012imZdyz1fhVXRNKPjbhSDw

PR #3286, opened 2026-09-18
Fix `pex3 venv create --link-python` for `--pex-repository`.

jsirois:
`pex3 venv create --link-python` for a foreign platform fails when the venv is installed from a `--pex-repository`:

```
A distribution for psutil could not be resolved for <local python>.
Found 1 distribution for psutil that does not apply:
1.) The wheel tags for psutil 5.9.5 are cp36-abi3-manylinux_2_12_x86_64, ... which do not match
    the supported tags of <local python>: cp313-cp313-manylinux_2_39_aarch64 ...
```

The distributions themselves install fine — it is populating the venv's *sources* that fails. `_populate_first_party` re-resolves the repository PEX to build the `pex-repl` script's `activated_dists`, and that resolve runs against a local interpreter, which a PEX built for the foreign platform alone holds nothing for. `_install_from_pex` has already resolved the distributions for the target, so this passes those down instead; the `PEX_TOOLS=1 ... venv` path keeps re-resolving, where the PEX and the interpreter always agree.

This is the combination Pants uses — it always invokes `pex3 venv create` with `--pex-repository` — so #3279 is unreachable from Pants without it.

Repro and verification, on a linux aarch64 host:

```
pex psutil==5.9.5 --platform linux_x86_64-cp-313-cp313 -o psutil.pex
pex3 venv create -d venv --pex-repository psutil.pex --platform linux_x86_64-cp-313-cp313 \
    --python-path $(command -v python3.13) --link-python /opt/python/bin/bob
```

Before: the error above. After: `bin/python`, `bin/python3` and `bin/python3.13` all link to `/opt/python/bin/bob`, and `lib/python3.13/site-packages/psutil/_psutil_linux.abi3.so` is `ELF 64-bit LSB shared object, x86-64`.

`test_foreign_target_link_python_from_pex_repository` covers it. I could not run it locally — the `py310` fixture's CPython 3.10.7 build fails in my environment (`ensurepip` exits non-zero) — so CI is its first run; the commands above are the same shape by hand. `uv run dev-cmd format lint typecheck` is green.

Written with Claude Code, as was #3279.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_012imZdyz1fhVXRNKPjbhSDw


jsirois:
Supercedes #3285.

